PR #17, opened 2020-07-06
Display json and edit link

alban-auzeill:


frontend/src/RulePage.js
```diff
@@ -49,10 +50,15 @@ export function RulePage(props) {
 
   let description = <div>Loading...</div>;
   if (!descIsLoading && !descError) {
-    description = <div dangerouslySetInnerHTML={{__html: descHTML}}/>;
+    description = <div>
+      <div dangerouslySetInnerHTML={{__html: descHTML}}/>
+      <hr />
+      <a href={editOnGithubUrl}>Edit on Github</a><br/>
+      <hr />
+      <pre>{JSON.stringify(metadataJSON, null, 2)}</pre>
```
nicolas-harraudeau-sonarsource, line 58: metadataJSON might not be loaded yet. I suggest to create a variable and set it in the `if (!metadataIsLoading && !metadataError) {` as we do for the title.

alban-auzeill: done

PR #32, opened 2021-02-05
Check links embeded in adocs files

tomverin:


checklinks.py
```diff
@@ -0,0 +1,36 @@
...
+from urllib.request import urlopen
+from urllib.error import URLError
+
+error=0
+
+def live_url(url):
+  try:
+    code = urlopen(url).code  
+    if (code / 100 >= 4):
+      print(f"Nothing there: {url}")
```
nicolas-harraudeau-sonarsource, line 14: Could you print the actual status? 400, 401, 403 and 404 status codes have completely different meanings. Last time I worked with URLs, the answer I was getting was different if I did it from curl or a browser. Knowing what the exact status code is will help us understand if the problem comes from the script or if the URL is really broken.

checklinks.py
```diff
@@ -0,0 +1,36 @@
...
+      error=1
+    else:
+      print(f"Ok: {url}")
```
nicolas-harraudeau-sonarsource, line 17: Just a side-note: this should be only printed in "debug" or "info" mode. We don't have to do it right now but we should start using loggers pretty soon or we will have a hard time finding errors in the logs.

checklinks.py
tomverin: sure will do

checklinks.py
tomverin: good point, I'll integrate a logger next

nicolas-harraudeau-sonarsource (changes requested):
It works. I just suggested a few improvements to avoid testing multiple times the same links and not using a global variable.

checklinks.py
```diff
@@ -0,0 +1,36 @@
...
+    for link in soup.findAll('a'):
+      live_url(link.get('href'))
+
+
+for directory, dirnames, filenames in os.walk("."):
+  for filename in filenames:
+    if filename.endswith('.html'):
+      findurl_in_html(os.path.join(directory,filename))
+if error == 1:
+  exit(1)
```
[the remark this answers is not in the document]
tomverin: Can't find all html files since they don't already exist, they are created on the fly by the asciidoc validation process, when a file is valid, I can then test html urls against it

rspec-tools/rspec_tools/cli.py
```diff
@@ -9,10 +10,16 @@ def cli(debug):
 @cli.command()
 @click.option('--rule', help='Validate only the rule matching the provided ID.')
 def validate(rule):
-    '''Validate rules.'''
-    # TODO
-    if rule == '42':
-        raise RuleNotFoundError(rule)
+  '''Validate rules.'''
```
nicolas-harraudeau-sonarsource, line 13: I think we can remove this command. It was only there as an example. I can do it later if you prefer.

rspec-tools/tests/links/test_links.py
```diff
@@ -0,0 +1,35 @@
+from click.testing import CliRunner
+from rspec_tools import cli
+from rspec_tools import checklinks
+
+def test_find_urls():
+  urls={}
+  checklinks.findurl_in_html("tests/links/404/404.html",urls)
+  assert len(urls) == 1
```
nicolas-harraudeau-sonarsource, line 8: This wouldn't detect a bug where the function adds the wrong key or value to the dict. You can simply do:
```
assert urls == {'https://www.google.com/404': 'tests/links/404/404.html'}

File: rspec-tools/tests/links/test_links.py [hunk 6]
```diff
@@ -0,0 +1,35 @@
...
+
+def test_live():
+  assert checklinks.live_url("https://www.google.com")
```
nicolas-harraudeau-sonarsource, line 11: This test could break if google is down. Yes it happens rarely but we should avoid depending on external services ;)

You can mock `urlopen` and make it return whatever code you want, or even raise exceptions.
In our case it would look something like:

```python
from unittest.mock import patch

def test_url():
  url = 'https://www.mytestwebsite42.com';
  with patch('urllib.request.urlopen') as mock_urlopen:
    mock_urlopen.return_value.code = 200
    assert checklinks.live_url(url)
    mock_urlopen.assert_called_once()
    assert mock_urlopen.call_args[0][0].full_url == url
```
Same for every other test.

nicolas-harraudeau-sonarsource (changes requested):

rspec-tools/rspec_tools/checklinks.py
```diff
@@ -0,0 +1,82 @@
...
+    data=None, 
+    headers={
+        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/35.0.1916.47 Safari/537.36'
+    }    
+  )
+  try:
+    code = urlopen(req,timeout=5).code  
+    if (code / 100 >= 4):
+      print(f"ERROR: {code} Nothing there")
+      return url
```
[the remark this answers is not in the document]
tomverin: iteration of the code lead me there, and I did not see any better, looks much more clear with returning False indeed

rspec-tools/rspec_tools/checklinks.py
```diff
@@ -0,0 +1,82 @@
...
+  for filepath in pathlib.Path(dir).glob('**/*.html'):
+    filename=str(filepath.absolute())
+    findurl_in_html(filename,urls)
+  print("All html files crawled")
+  print("Testing links")
+  for key in urls:
+    print(f"{key} in {len(urls[key])} files")
+    result=live_url(key)
+    if result != True:
+      errors.append(result)
```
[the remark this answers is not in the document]
tomverin: Indeed

rspec-tools/tests/links/test_links.py
tomverin: done

rspec-tools/rspec_tools/cli.py
tomverin: Done

rspec-tools/rspec_tools/checklinks.py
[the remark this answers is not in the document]
tomverin: Done

PR #44, opened 2021-02-22
Improve rule metadata validation

nicolas-harraudeau-sonarsource:
This adds the command `rspect-tools validate-rules`.
There are quite a few `rule.adoc` and `metadata.json` files because I copied a few rules in tests to avoid having tests rely on production rules.

The "validate_asciidoc" check does not pass because we have 500+ rules with invalid metadata :s

We could later restrict the validation to modified rules only, but for now I suggest we keep it as is and fix the rules in Jira.


rspec-tools/rspec_tools/cli.py
```diff
@@ -48,5 +50,27 @@ def create_rule(languages: str, user: Optional[str]):
     pull_request = rule_creator.create_new_rule_pull_request(token, rule_number, lang_list, user=user)
 
 
+@cli.command()
+@click.argument('rules', nargs=-1)
+def validate_rules(rules):
+  '''Create a new rule.'''
```
arseniy-sonar, line 56: Looks like a C&P artifact

validate_asciidoc_and_metadata.sh
```diff
@@ -1,15 +1,16 @@
+# Validate metadata
```
arseniy-sonar, line 1: This looks like an independent part of the script that could be put directly into the .cirrus.yml (perhaps even into a separate `script_` section) for better CI UX

rspec-tools/rspec_tools/rules.py
```diff
@@ -0,0 +1,77 @@
+
+import json
+from pathlib import Path
+from typing import Final, Generator, Iterable, Optional
+
+
+METADATA_FILE_NAME: Final[str] = 'metadata.json'
+
+class RuleLanguage:
```
arseniy-sonar, line 9: I find this name confusing w.r.t. to what this class represents, especially when used as return type for `Rule.languages`. For a `Language` and `RuleLanguage` I'd expect only details about the language (that might be specific to a rule), and not a rule specific to the language.
How about `LanguageSpecificRule`, or just `SpecificRule` (with the symmetric renaming of `Rule` to `GenericRule`)?

rspec-tools/rspec_tools/rules.py
```diff
@@ -0,0 +1,77 @@
...
+
+  def __init__(self, rule_path: Path):
+    self.rule_path = rule_path
+
+  @property
+  def id(self) -> str:
+    return self.rule_path.name
+
+  @property
+  def languages(self) -> Generator[RuleLanguage, None, None]:
```
arseniy-sonar, line 48: It is confusing when this function returns something other than just languages. I think a name like `languageSpecificRules` or `specificRules` or `specializations` would be more obvious (regardless of the comment on the `RuleLanguage` class)

rspec-tools/rspec_tools/cli.py
```diff
@@ -48,5 +50,27 @@ def create_rule(languages: str, user: Optional[str]):
     pull_request = rule_creator.create_new_rule_pull_request(token, rule_number, lang_list, user=user)
 
 
+@cli.command()
+@click.argument('rules', nargs=-1)
+def validate_rules(rules):
```
arseniy-sonar, line 55: As far as I can tell, current version only validates the metadata.json and ignores the asciidoc description.
Furthermore, validation of the asciidoc description is planned in the future.
I think it is more appropriate to speak only about metadata validation for now, and use names like `validate_rules_metadata` to clearly delineate the scope of the tool.
Once asciidoc validation is in place (e.g., `validate_rules_description`), we could merge the two into `validate_rules` or keep them separate.

arseniy-sonar (changes requested):
Impressive
I would mention in the PR title that it concerns only the metadata validation.

rspec-tools/rspec_tools/cli.py
nicolas-harraudeau-sonarsource: Indeed

validate_asciidoc_and_metadata.sh
nicolas-harraudeau-sonarsource: Works for me.

rspec-tools/rspec_tools/rules.py
nicolas-harraudeau-sonarsource: Good idea, ok for `LanguageSpecificRule` and `GenericRule`.

PR #66, opened 2021-05-04
RULEAPI-609 [after transition] Support superseded rules

amelie-renard-sonarsource:


arseniy-sonar:
Did you make the changes only in this repository?
I need to know, because if so, once I merge they will be overwritten by the export script the next night. In that case I will leave the PR up, like the RULEAPI-555 about manually translating rspecs, until "after the transition" when the export script is shut down.

amelie-renard-sonarsource:
Yes I only did the changes in this repository, I don't see how I could add a comment in the description of the Jira RSPECs...
I think this PR can remain open until the shift is done, there is no blocker

PR #74, opened 2021-05-20
RULEAPI-612 Enable full-text search over the non-implemented RSPECS...

amelie-renard-sonarsource:
... in the PRs along with the rspecs on the master branch

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1913,6 +1913,131 @@
         }
       }
     },
+    "@octokit/auth-token": {
+      "version": "2.4.5",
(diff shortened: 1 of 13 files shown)


amelie-renard-sonarsource:
Everything seemed okay to me (as far as my knowledge of this would allow me to judge :))
I just modified the style of the "non implemented" label as it seemed a bit aggressive to me
And about the test, you already did one to check the creation of the PR link. Going further would mean to mock a PR and check that it's correctly handled in `process_incomplete_rspecs` but I'm afraid that's a bit too complex for the usefulness...

arseniy-sonar (approved):
This functionality calls for some tests, but that would require mocking Octokit, as well as Git which seems harder than implementing the functionality itself. Let's say RULEAPI-571 should cover this.

PR #77, opened 2021-05-26
Create rule XXXX

arseniy-sonar:


arseniy-sonar:
reopening to test rspec frontend resilience

PR #78, opened 2021-05-31
RULEAPI-624 Fix the RSPEC index in the presense of PRs with noncompliant file structure.

arseniy-sonar:
1. Force the deployment to fail for any unexpected failure in the indexing of the un-merged open PRs (for the non implemented rules).
2. Ignore the PRs that promise to create a certain rule (i.e. "Create rule S1234") and fail to deliver on that promise: have no folder `rules/S1234`.

### DIFF
--- a/frontend/src/deployment/index.ts
+++ b/frontend/src/deployment/index.ts
@@ -12,6 +12,8 @@ import { process_incomplete_rspecs, PullRequest } from './pullRequestIndexing';
 
 import { PR_DIRECTORY, RULE_SRC_DIRECTORY, RULE_DST_DIRECTORY } from './paths';
 
+process.on('unhandledRejection', up => { throw up })
+
 // eslint-disable-next-line @typescript-eslint/no-unused-expressions
 yargs(process.argv.slice(2))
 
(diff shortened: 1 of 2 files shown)


amelie-renard-sonarsource:
- If I understand correctly, GH page was empty because only the PR title was renamed to "Create SXXXX". So the script looked for a folder named "SXXXX" that did not exist... Is that correct?
- The script also looks at PRs that are closed. Is that something we want? (I genuinely don't know the answer to this question...)

arseniy-sonar:
* I believe the GH page deployment failed earlier, when I opened a PR named "Create rule S7018" using an old `coverage` branch that had nothing to do with S7018. The script was looking for `rspec/s7018` in that PR and couldn't find it, so it failed and produced no index at all.
* I believe the script works only with open PRs (see `status: open` in the [PR indexing file](https://github.com/SonarSource/rspec/blob/master/frontend/src/deployment/pullRequestIndexing.ts#L24)).

The closed PRs are of two sorts: merged (these we do not care about), and discarded (i.e., a rule idea was never implemented). The latter might be of some interest. I think they should be marked differently from the implemented rules and the not-yet-implemented rules. At the moment we have no meaningful PRs in this category, so I propose to postpone implementing it for when we have some to be able to evaluate the implementation. Here is a ticket to track this idea: https://jira.sonarsource.com/browse/RULEAPI-628

amelie-renard-sonarsource (approved):

PR #88, opened 2021-06-03
RULEAPI-616: Stabilize the link validation CI check by preserving the probing results between runs

arseniy-sonar:
two fluentcpp links (in S5402 and S5419) that are down now are injected into the url-probing history, so they will not fail the link validation for the following 7 days. If fluentcpp.com is still down by then, I suppose, we will have to remove or replace them.

### DIFF
--- a/rspec-tools/rspec_tools/checklinks.py
+++ b/rspec-tools/rspec_tools/checklinks.py
@@ -63,8 +63,6 @@ def url_was_reached_recently(url: str):
 def live_url(url: str, timeout=5):
   if url.startswith('#'):
     return True
-  if url in EXCEPTIONS:
-    return True
   try:
     req = requests.Request('GET', url, headers = {'sec-ch-ua': '" Not A;Brand";v="99", "Chromium";v="90"',
                                                   'sec-ch-ua-mobile': '?0',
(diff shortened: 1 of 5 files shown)


rspec-tools/rspec_tools/checklinks.py
```diff
@@ -2,23 +2,68 @@
...
+
+def url_was_reached_recently(url: str):
+  global link_probes_history
+  if url not in link_probes_history:
+    return False
+  last_time_up = link_probes_history[url]
+  print(f"{url} was reached most recently on {last_time_up}")
+  spread = random.randrange(PROBING_SPREAD)
+  probing_cooldown = PROBING_COOLDOWN + datetime.timedelta(minutes=spread)
+  diff = (datetime.datetime.now() - last_time_up)
```
amelie-renard-sonarsource, line 61: useless `diff`

rspec-tools/rspec_tools/checklinks.py
```diff
@@ -2,23 +2,68 @@
...
+def rejuvenate_url(url: str):
+  global link_probes_history
+  link_probes_history[url] = datetime.datetime.now()
+
+def url_is_long_dead(url: str):
+  global link_probes_history
+  if url not in link_probes_history:
+    return True
+  last_time_up = link_probes_history[url]
+  print(f"{url} was reached most recently on {last_time_up}")
```
amelie-renard-sonarsource, line 50: If a link is long dead, this sentence will be printed twice (first time with `url_was_reached_recently`)
It makes the script a bit too verbose, no?
I think we can drop the message of `url_was_reached_recently`, because I don't think it is very useful... (there's already the message ""skip probing because it was reached recently"). But I let you judge on that

amelie-renard-sonarsource (changes requested):
Small comments

rspec-tools/rspec_tools/checklinks.py
arseniy-sonar: Indeed, that is redundant

amelie-renard-sonarsource (approved):

PR #90, opened 2021-06-03
RULEAPI-608 Rename unconventional headers in RSPECs and update the validation script in GitHub rspec repository

amelie-renard-sonarsource:


rspec-tools/rspec_tools/validation/description.py
```diff
@@ -11,24 +11,11 @@
...
-                                            'Noncompliant Code Examples',
-                                            'NonCompliant Code Example',
                                             'Compliant Solution',
-                                            'Compliant Solutions',
-                                            'Compliant solution',
-                                            'Compliant Solution:',
-                                            'Compliant Example',
-                                            'Compliant Code Example',
-                                            'Compliant',
                                             'See',
```
arseniy-sonar, line 15: Should we merge "See" with "See Also"?

arseniy-sonar (commented):
Looks good to me. Not sure why you left both "See" and "See Also" as separate sections

rspec-tools/rspec_tools/validation/description.py
amelie-renard-sonarsource: I some rules there are two distinct sections, "See" and "See Also".
"See" regroups links that are related to the rule
whereas
"See Also" lists other RSPECs that are related to the current one
So I'd rather not merge these two, because making this distinction makes sense to me

PR #93, opened 2021-06-05
String offset-based methods should be preferred for finding substrings from offsets (java:S4635)

nbauma109:
"Using these methods gives the same result" => no, it doesn't, it gives the position from the beginning of the String. 
The returned value has to be adjusted by subtracting the substring index parameter to obtain the same result.
Example: 
```
		String s = "java/util/Map$Entry";
		int beginIndex = s.lastIndexOf('/');
		char char1 = '$';
		System.out.println(s.substring(beginIndex).indexOf(char1)); // 4
		System.out.println(s.indexOf(char1, beginIndex)); // 13
		System.out.println(s.indexOf(char1, beginIndex) - beginIndex); // 4
```


quentin-jaquier-sonarsource (approved):

quentin-jaquier-sonarsource:
Hello @nbauma109 

It makes sense, thanks for taking the time to fix the description of the rule.

PR #96, opened 2021-06-07
RULEAPI-570 Add required filters in Github Page search

arseniy-sonar:
The issue of covered vs. targeted languages needs further work, I scoped it into RULEAPI-645.

### DIFF
--- a/frontend/src/SearchPage.tsx
+++ b/frontend/src/SearchPage.tsx
@@ -10,29 +10,30 @@ import Box from '@material-ui/core/Box';
 
 import useStyles from './SearchPage.style';
 import { useSearch } from './utils/useSearch';
+import { useFetch } from './utils/useFetch';
 import {
   SearchParamSetter,
   useLocationSearch,
   useLocationSearchState
 } from './utils/routing';
 import { SearchHit } from './SearchHit';
+import { IndexAggregates } from './types/IndexStore'
 
 export const SearchPage = () => {
   const classes = useStyles();
-  
+
   const pageSize = 20;
   const [query, setQuery] = useLocationSearchState('query', '');
 
   const [ruleType, setRuleType] = useLocationSearchState('types', 'ALL');
   const allRuleTypes: Record<string,string> = {
     'BUG': 'Bug',
-    'CODE_SMELL': 'Code Smell', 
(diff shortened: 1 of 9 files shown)


frontend/src/SearchPage.tsx
```diff
@@ -148,6 +152,26 @@ export const SearchPage = () => {
           ))}
         </TextField>
       </Grid>
+      <Grid item xs={4}>
+      <TextField
+          select
+          fullWidth
+          margin="normal"
+          variant="outlined"
+          label="Language"
```
amelie-renard-sonarsource, line 171: "Languages" as it can have more than one value?

amelie-renard-sonarsource (changes requested):
Small comments, and as discussed, it would be nicer if languages and tags were alphabetically ordered.

frontend/src/SearchPage.tsx
arseniy-sonar: You can select only one language, hence the singular of "Language". I applied the same rule to "Rule type" that can only have a single value.

frontend/src/SearchPage.tsx
```diff
@@ -10,40 +10,57 @@ import Box from '@material-ui/core/Box';
...
   const {results, numberOfHits, error, loading} = useSearch(query,
-    ruleType === "ALL" ? null : ruleType,
+    ruleType === 'ANY' ? null : ruleType,
+    ruleLang === 'ANY' ? null : ruleLang,
     ruleTags,
+    qualityProfiles,
     pageSize, pageNumber);
   const totalPages = numberOfHits ? Math.ceil(numberOfHits/pageSize) : 0;
 
+  let allRuleTags = ['confusing', 'pitfall', 'clumsy', 'junit', 'tests'];
```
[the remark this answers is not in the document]
arseniy-sonar: Indeed, having it empty until it loads is more intuitive. Emptied.

frontend/src/SearchPage.tsx
amelie-renard-sonarsource: Indeed, I don't know why I had the impression that I could select multiple languages... Perfect then

amelie-renard-sonarsource (approved):

PR #107, opened 2021-06-08
Modify rule S6291: Generalize rule for multiple databases

hendrik-buchwald-sonarsource:
Implementation ticket: [SONARKT-61](https://jira.sonarsource.com/browse/SONARKT-61)


hendrik-buchwald-sonarsource:
> Rule description looks very good to me.
> Please add references to MASVS and OWASP Mobile Top 10.
> I also think that OWASP Top 10 A3 Sensitive data exposure would fit,

Thanks! MASVS and OWASP Mobile Top 10 were provided by Eric. I have rebased the branch on master, so they also show up here now. A3 was added.

> Not a blocker but, as mention in [SonarSource/security-expected-issues#480](https://github.com/SonarSource/security-expected-issues/pull/480), I am open to discuss if we should maybe provide code example that show how to properly store the database encryption key.

As discussed today I have added it as text instead of in the code samples to avoid having the same block of code for every code sample.

PR #114, opened 2021-06-09
RULEAPI-641 Document how comments can be added to rules

amelie-renard-sonarsource:


README.adoc
```diff
@@ -178,6 +178,18 @@ you have to run `update` immediately after,
...
 when generating the deprecation notes for superseded rules.
 
+== Comment a rule
+
+Comments and links that were created on Jira have been gathered in a `comments-and-links.adoc` file for each concerned rule. +
+You can add a comment anywhere you need in a rule by adding the following lines in the `*.adoc` file:
+[source]
+----
+\ifdef::env-github,rspecator-view[]
+John Doe (9 Jun 2021): my comment on the rule
```
arseniy-sonar, line 188:
```suggestion
=== on 9 Jun 2021, 15:49:00 John Doe wrote:
My comment on the rule.
```
I think the hour of the comment can be useful. While the seconds is an overkill, I would keep consistency in the format with the generated comments (you can see an example [here](https://github.com/SonarSource/rspec/blob/master/rules/S2070/comments-and-links.adoc))

README.adoc
```diff
@@ -178,6 +178,18 @@ you have to run `update` immediately after,
 because rule-api relies on the files in the directory to determine the covered set of rules
 when generating the deprecation notes for superseded rules.
 
+== Comment a rule
+
+Comments and links that were created on Jira have been gathered in a `comments-and-links.adoc` file for each concerned rule. +
+You can add a comment anywhere you need in a rule by adding the following lines in the `*.adoc` file:
```
arseniy-sonar, line 184:
```suggestion
You can add a comment anywhere in a rule by adding the following lines in the `*.adoc` file:
```
An obvious dependant clause that hiders the clarity of the message (in my opinion).

README.adoc
```diff
@@ -178,6 +178,18 @@ you have to run `update` immediately after,
...
+[source]
+----
+\ifdef::env-github,rspecator-view[]
+John Doe (9 Jun 2021): my comment on the rule
+\endif::env-github,rspecator-view[]
+----
+This way, your comment will only be visible on the Search Page (and won't be visible for the user for example).
```
arseniy-sonar, line 191:
```suggestion
This way, your comment will only be visible in GitHub preview and on the Search Page (and will not be visible for the user).
```
IMO, "for example" is redundant here. Also, I see no need to contract "will not" into "won't" here.

arseniy-sonar (approved):
Looks good. Left some opinionated remarks, feel free to disregard.

README.adoc
amelie-renard-sonarsource: I agree with you about the hour of the comment.
But I'd rather not put the same format as what is done in `comments-and-links.adoc` because the comment could be anywhere in the rule (and not necessarily in the dedicated "Comments And Links" section)... So put it in bold like that could make the rule less readable.
I rely on people common sense to format it as it needs to be

PR #119, opened 2021-06-10
RULEAPI-640: Add language(s) label(s) to automatically created PR

arseniy-sonar:


rspec-tools/rspec_tools/utils.py
```diff
@@ -1,9 +1,68 @@
...
+    supported_langs = supported_langs.replace(' ', '')
+    supported_langs = supported_langs.replace('\n', '')
+    return supported_langs.split(',')
+
+def get_mapped_languages():
+  return LANG_TO_LABEL.keys();
+
+def parse_and_validate_language_list(languages):
+  lang_list = [lang.strip() for lang in languages.split(',')]
+  valid_langs = load_valid_languages()
```
amelie-renard-sonarsource, line 57: This action could be done after the following `if`

rspec-tools/rspec_tools/utils.py
```diff
@@ -1,9 +1,68 @@
...
+}
+
 def copy_directory_content(src:Path, dest:Path):
   for item in src.iterdir():
     if (item.is_dir()):
       shutil.copytree(item, dest)
     else:
       shutil.copy2(item, dest)
+
+def load_valid_languages():
```
amelie-renard-sonarsource, line 43: Nice improvement!

rspec-tools/rspec_tools/utils.py
```diff
@@ -1,9 +1,68 @@
...
+  with open(SUPPORTED_LANGUAGES_FILENAME, 'r') as supported_langs_file:
+    supported_langs = supported_langs_file.read()
+    supported_langs = supported_langs.replace(' or', '')
+    supported_langs = supported_langs.replace('`', '')
+    supported_langs = supported_langs.replace(' ', '')
+    supported_langs = supported_langs.replace('\n', '')
+    return supported_langs.split(',')
+
+def get_mapped_languages():
+  return LANG_TO_LABEL.keys();
```
amelie-renard-sonarsource, line 53: This function does not seem useful

rspec-tools/tests/test_create_rule.py
```diff
@@ -134,3 +137,23 @@ def test_create_new_single_lang_rule_branch(rule_creator: RuleCreator, mock_rspe
         relative_path = lang_item.relative_to(lang_root)
         actual_content = rule_dir.joinpath(lang, relative_path).read_text()
         assert actual_content == expected_content
+
```
amelie-renard-sonarsource, line 140: What about adding a test where a PR is created with an unsupported language?

amelie-renard-sonarsource (changes requested):

rspec-tools/rspec_tools/utils.py
arseniy-sonar: This function is necessary for the test that makes sure all the supported languages are mapped to some label. I'll add a comment about that.

rspec-tools/tests/test_create_rule.py
arseniy-sonar: There's no protection in `rule_creator.create_new_rule_pull_request` against unsupported languages and labels: it assumes the arguments are valid.
It is the cli.py/`create_rule` that parses and validates the language list. The parsing and validation are tested in test_utils.py.
The only part that is not tested is the cli.py, because it requires a lot of mocking for the environment (os, tmpfile, RuleCreator). I am not sure how to go about testing it, to be honest.
What do you say, we leave it for the RULEAPI-579 hardening ticket?

rspec-tools/rspec_tools/utils.py
arseniy-sonar: Moved.

rspec-tools/tests/test_create_rule.py
amelie-renard-sonarsource: Ok, thanks for the explanation.
I'm fine with your suggestion

arseniy-sonar: In the end I added a test, a little refactoring, wasn't that difficult

PR #137, opened 2021-06-18
Update S2755 description to be more consise

eric-therond-sonarsource:


rules/S2755/description.adoc
```diff
@@ -1,32 +1,27 @@
...
- <xsl:template match="/">
-  &content;
- </xsl:template>
-</xsl:stylesheet> 
-----
 
-It is recommended to disable access to external entities and network access in general.
+It's recommended to limit resolution of external entities by using one of these solutions: 
+
+* If DOCTYPE is not necessary, disable completely its use.
```
hendrik-buchwald-sonarsource, line 22: The sentence `disable completely its use` sounds a bit strange to me. For one, I think the order should rather be `disable its use completely`. But also, *disabling* *use* sounds not right. Maybe `disable this feature completely`?

The same applies for the sentence below and the same sentences in the comments of the code samples.

hendrik-buchwald-sonarsource (approved):
In general it looks good to me but I left some comments.

PR #139, opened 2021-06-18
S1451 Remove the <br> tags from inside <pre>

arseniy-sonar:
See https://github.com/asciidoctor/asciidoctor/issues/724 for the details on the empty-line stripping in the end of code blocks,
and the [long-standing issue](https://github.com/asciidoctor/asciidoctor/issues/754) about disabling this behavior.

### DIFF
--- a/rules/S1451/csharp/rule.adoc
+++ b/rules/S1451/csharp/rule.adoc
@@ -19,7 +19,6 @@ then the ``++headerFormat++`` parameter should end with an empty line like this
 
 ----	
 // Copyright (c) SonarSource. All Rights Reserved. Licensed under the LGPL License.  See License.txt in the project root for license information.   
-​
 ---- 
 
 include::../compliant.adoc[]
--- a/rules/S1451/csharp/rule.adoc
+++ b/rules/S1451/csharp/rule.adoc
@@ -17,10 +17,9 @@ namespace Foo
 ----
 then the ``++headerFormat++`` parameter should end with an empty line like this
 
-[subs=normal]
 ----	
 // Copyright (c) SonarSource. All Rights Reserved. Licensed under the LGPL License.  See License.txt in the project root for license information.   
-{empty} +
+​
 ---- 
 
 include::../compliant.adoc[]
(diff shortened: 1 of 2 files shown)


rules/S1451/csharp/rule.adoc
```diff
@@ -17,10 +17,9 @@ namespace Foo
 ----
 then the ``++headerFormat++`` parameter should end with an empty line like this
 
-[subs=normal]
 ----	
 // Copyright (c) SonarSource. All Rights Reserved. Licensed under the LGPL License.  See License.txt in the project root for license information.   
-{empty} +
+​
```
pavel-mikula-sonarsource, line 22: Let's remove the empty line all together. I don't see a reason for having horizontal whitespace (EOL) followed by horizontal whitespace (top padding/margin of the follow up heading)
```suggestion
```

rules/S1451/vbnet/rule.adoc
```diff
@@ -17,10 +17,9 @@ namespace Foo
 ----
 then the ``++headerFormat++`` parameter should end with an empty line like this
 
-[subs=normal]
 ----	
 ' Copyright (c) SonarSource. All Rights Reserved. Licensed under the LGPL License.  See License.txt in the project root for license information.   
-{empty} +
+​
```
pavel-mikula-sonarsource, line 22: Let's just remove them
```suggestion
​```

pavel-mikula-sonarsource (changes requested):

arseniy-sonar:
@pavel-mikula-sonarsource, are you sure we can remove it? It seems to me that the last empty line is the point of the code snippet.

> then the `headerFormat` parameter should end with an empty line like this

arseniy-sonar:
I did remove the zero-width spaces, but if you think they are important for the rule semantics, I'll revert the last commit.

pavel-mikula-sonarsource (approved):
LGTM. The point of this rule is to have a correct license header text, as configured in SC/SQ. The EOL is minor stuff.

PR #140, opened 2021-06-21
Modify rule S5693: Move web.config links to see part of the document

dmitry-golovinov-sonarsource:


andrei-epure-sonarsource (approved):
LGTM

PR #142, opened 2021-06-22
Comment why we do not implement S6220

alban-auzeill:


arseniy-sonar:
@alban-auzeill You might want to add this to the end of `java/rule.adoc`, so that the comment shows up in the github and sonarsource.github.io/rspec/# 

```
'''
== Comments And Links
(visible only on this page)

include::../comments-and-links.adoc[]
```

alban-auzeill:
@arseniy-sonar Thanks for your help about updating `rule.adoc`, I will add it, I thought it was done automatically by the GitHub pages generation.

PR #146, opened 2021-06-25
Modify rule S3551: Fix mismatch in CERT references between text documentation and metadata

dorian-burihabwa-sonarsource:
The reference to MET04-J has been removed from the metadata and replaced with TSM00-J which is more specific and thus relevant to the issue.


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #152, opened 2021-06-29
Modify rule S997: CPP-2396 Add exception for function definitions

amelie-renard-sonarsource:
Associated implementation PR: https://github.com/SonarSource/sonar-cpp/pull/1434

### DIFF
--- a/rules/S997/description.adoc
+++ b/rules/S997/description.adoc
@@ -7,4 +7,4 @@ Some names have to be declared in the global namespace, and are excluded from th
 * ``++main++`` (or its variants)
 * Overloads of the global ``++new++`` & ``++delete++`` operators
 
-In addition, no issue is raised for function definitions as they can not be accessed from different translation units anyway, because of the https://en.cppreference.com/w/cpp/language/definition[One-Definition-Rule].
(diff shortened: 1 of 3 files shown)


rules/S997/description.adoc
```diff
@@ -1,10 +1,10 @@
...
-
 This rule raises an issue when a name in the global namespace has external linkage and therefore can be accessed from outside a particular translation unit.
 
-
 Some names have to be declared in the global namespace, and are excluded from this rule:
 
 * ``++main++`` (or its variants)
 * Overloads of the global ``++new++`` & ``++delete++`` operators
+
+In addition, no issue is raised for function definitions as they can not be accessed from different translation units anyway, because of the https://en.cppreference.com/w/cpp/language/definition[One-Definition-Rule].
```
arseniy-sonar, line 10:
```suggestion
In addition, no issue is raised for function definitions because they can only be accessed from different translation units through a forward declaration that will be flagged.
```
I don't think ODR is relevant here. The rule should prevent polluting the global namespace regardless of where and how many times any of the names are defined.

arseniy-sonar (changes requested):
I would also move the `void  m1() { } // Noncompliant` line from noncompliant to a the compliant section.

arseniy-sonar (approved):

PR #157, opened 2021-06-30
Fix the template for multi-lang rule metadata

arseniy-sonar:


eric-therond-sonarsource (approved):
LGTM

eric-therond-sonarsource:
@arseniy-sonar I am not sure but I think we should change in the template:
```
  "qualityProfiles": ["Sonar way"]
```
to
```
  "defaultQualityProfiles": ["Sonar way"]
```

See the error here:
https://github.com/SonarSource/rspec/pull/153/checks?check_run_id=2950702961

arseniy-sonar:
Indeed, sorry for such a crude template :/
Thank you for dogfooding it @eric-therond-sonarsource!
I pushed the change

PR #162, opened 2021-07-05
Modify rule S2057: Add an exception for Records

dorian-burihabwa-sonarsource:
Add an exception to the rule as the constraint of validation on the serialUID is waived for Records.

SONARJAVA-3755


m-g-sonar (commented):
LGTM!

PR #192, opened 2021-07-09
Modify rule S5854: Fix noncompliant example

sebastian-hungerecker-sonarsource:
The comment showing the output used the wrong accent.


dorian-burihabwa-sonarsource (approved):
LGTM

PR #213, opened 2021-07-14
Modify rule S6210: Rephrase requirement on private element access

dorian-burihabwa-sonarsource:


margarita-nedzelska-sonarsource (approved):
LGTM

PR #214, opened 2021-07-15
Update rule S3247 description with the information about redundant cast

dmitry-golovinov-sonarsource:


dmitry-golovinov-sonarsource:
@andrei-epure-sonarsource  feel free to use this link to inspect the resulting file https://github.com/SonarSource/rspec/blob/csharp/S3247/rules/S3247/csharp/rule.adoc

rules/S3247/csharp/rule.adoc
```diff
@@ -22,6 +22,28 @@ if (f != null)
 }
 ----
 
+The casting of any typed variable of the ``++is++`` type pattern expression also gets marked as redundant.
```
andrei-epure-sonarsource, line 25: AFAIK we always have just one `Noncompliant Code Example` (with multiple examples inside if necessary) and only one `Compliant code example` where the issues get fixed. 

Also, a rule description should be educational rather than describing what gets marked as redundant. IMO this line doesn't bring anything more to the reader, so you can remove it and leave the code example speak for itself.

Type patterns have been introduced in C# 7.0 (March 2017) so we should be safe to directly provide the C# 7 example as compliant for both noncompliant cases. Or eventually to provide 2 examples:

```
// C# 6
var f = x as Fruit;
if (f != null)
{
  // ...
}
// C# 7
if (x is Fruit fruit)
{
 // ...
}

```

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

PR #215, opened 2021-07-19
Update S5542 description: ECB mode for RSA

eric-therond-sonarsource:


rules/S5542/kotlin/rule.adoc
```diff
@@ -16,8 +16,9 @@ val c3 = Cipher.getInstance("RSA/NONE/NoPadding") // Noncompliant: RSA without O
 val c1 = Cipher.getInstance("AES/GCM/NoPadding"); // Compliant
 
 // Recommended for RSA
-val c2= Cipher.getInstance("RSA/None/OAEPWithSHA-1AndMGF1Padding") // Compliant
 val c3 = Cipher.getInstance("RSA/None/OAEPWITHSHA-256ANDMGF1PADDING") // Compliant
+// or the ECB mode can be used for RSA (will be treated as None)
```
agigleux, line 20: I would change this to:

// or the ECB mode can be used for RSA when "None" is not available, for example in Bouncy Castle or Android context - in that case, ECB will be treated as "None" for RSA.

rules/S5542/java/rule.adoc
```diff
@@ -16,8 +16,9 @@ Cipher c3 = Cipher.getInstance("RSA/NONE/NoPadding"); // Noncompliant: RSA witho
 Cipher c1 = Cipher.getInstance("AES/GCM/NoPadding"); // Compliant
 
 // Recommended for RSA
-Cipher c2= Cipher.getInstance("RSA/None/OAEPWithSHA-1AndMGF1Padding"); // Compliant
 Cipher c3 = Cipher.getInstance("RSA/None/OAEPWITHSHA-256ANDMGF1PADDING"); // Compliant
+// or the ECB mode can be used for RSA (will be treated as None)
```
agigleux, line 20: // or the ECB mode can be used for RSA when "None" is not available, for example in Bouncy Castle or Android context - in that case, ECB will be treated as "None" for RSA.

agigleux (changes requested):

agigleux (approved):

PR #234, opened 2021-07-29
Modify rule S5860: Describe missing non-compliant pattern

yassin-kammoun-sonarsource:


rules/S5860/description.adoc
```diff
@@ -1,6 +1,7 @@
 Why use named groups only to never use any of them later on in the code?
 
-This rule raises issues every time named groups are:
+This rule raises issues every time named groups:
 
-* defined but never called anywhere in the code through their name;
-* defined but called elsewhere in the code by their number instead.
+* are defined but never called anywhere in the code through their name;
+* are defined but called elsewhere in the code by their number instead;
+* don't exist.
```
vilchik-elena, line 7: IMHO it's not clear :) I would say `are referenced while not defined`

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #250, opened 2021-08-05
Modify rule S1764 csharp and vbnet: Change scope to Main-only

pavel-mikula-sonarsource:
Related to https://github.com/SonarSource/sonar-dotnet/issues/4659


rules/S1764/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+  "scope": "Main"
```
andrei-epure-sonarsource, line 2: if you change it, you should change it for other OOP languages, too, not only C#. I don't see why at least Java shouldn't have the same behavior, and possibly other languages

I'd say this is worth a dogfood post to discuss it

andrei-epure-sonarsource (changes requested):

rules/S1764/csharp/metadata.json
pavel-mikula-sonarsource: I can raise awareness about this change, but other languages would need to implement this on their analyzer side and implement it as their own PR in this repo anyway.

andrei-epure-sonarsource: yes please do

andrei-epure-sonarsource (approved):
LGTM

PR #259, opened 2021-08-11
Create rule S3431[VBNET]: ExpectedException should not be used

csaba-sagi-sonarsource:


rules/S3431/metadata.json
```diff
@@ -1,2 +1,28 @@
...
+  ],
+  "extra": {
+    "coveredLanguages": [
+      "C#",
+      "VB.Net"
+    ],
+    "replacementRules": [
+      
+    ]
+  },
```
pavel-mikula-sonarsource, line 20: Do we need to maintain this manually? 
Do we actually still need these fields?

rules/S3431/vbnet/rule.adoc
```diff
@@ -0,0 +1,55 @@
+It should be clear to a casual reader what code a test is testing and what results are expected. Unfortunately, that's not usually the case with the ``++<ExpectedException>++`` attribute since an exception could be thrown from almost any line in the method.
```
pavel-mikula-sonarsource, line 1: This looks that can be [simplified](https://xtranet-sonarsource.atlassian.net/wiki/spaces/LANG/pages/665809/LT+Weekly+Meeting+Notes#July-30)
```suggestion
It should be clear to a casual reader what code a test is testing and what results are expected. Unfortunately, that's not usually the case with the `<ExpectedException>` attribute since an exception could be thrown from almost any line in the method.
```

pavel-mikula-sonarsource: Same in C# rule

rules/S3431/vbnet/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+
+This rule detects MSTest and NUnit ``++ExpectedException++`` attribute.
+
+
+== Noncompliant Code Example
+
+----
+<TestMethod>
+<ExpectedException(typeof(ArgumentNullException))> ' Noncompliant
```
pavel-mikula-sonarsource, line 11:
```suggestion
<ExpectedException(GetType(ArgumentNullException))> ' Noncompliant
```

rules/S3431/vbnet/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+End Sub
+----
+
+or
+
+
+----
+<TestMethod>
+Public Sub TestNullArg()
+    Assert.ThrowsException(Of ArgumentNullException)(Sub()  )
```
pavel-mikula-sonarsource, line 39: Just to express that something is supposed to happen there, because it will not compile as is
```suggestion
    Assert.ThrowsException(Of ArgumentNullException)(Sub() ... )
```

rules/S3431/vbnet/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+End Sub
+----
+
+
+== Exceptions
```
pavel-mikula-sonarsource, line 44: This is .NET-specific rule. Let's finetune the structure here. I suggest to:
- move the description to `../rule.adoc` and this exception footer to `../exceptions.adoc` and include them from C# and VB.NET variant so we don't have to maintain it twice.

rules/S3431/metadata.json
csaba-sagi-sonarsource: Let`s merge with these changes, I would say it can do no harm as other language specific PRs had them too. And I`ll check if it is really needed because I checked a few rules and saw that the coveredLanguages section is out of date so there is a chance it is not used.

pavel-mikula-sonarsource (approved):
LGTM

PR #262, opened 2021-08-12
Modify rule S3723: Consider all multiline constructs

yassin-kammoun-sonarsource:


rules/S3723/rule.adoc
```diff
@@ -1,5 +1,4 @@
-Modern browsers ignore unneeded, trailing commas, so there are no negatives to having them unless you're supporting an IE 8 application. Since they make adding new properties simpler, their use is preferred. This rule raises an issue when the last item in an object declaration or array declaration does not end with a trailing comma and does not lie on the same line as the closing curly brace or bracket.
-
+Modern browsers ignore unneeded, trailing commas, so there are no negatives to having them unless you're supporting an IE 8 application. Since they make adding new properties simpler, their use is preferred. This rule raises an issue when the last item in a multiline construct (arrays, objects, imports, exports, and functions) does not end with a trailing comma and does not lie on the same line as the closing curly brace or bracket.
```
vilchik-elena, line 1:
```suggestion
Modern browsers ignore unneeded, trailing commas, so there are no negatives to having them unless you're supporting an IE 8 application. Since they make adding new properties simpler, their use is preferred. This rule raises an issue when the last item in a multiline construct (array or object literal, import or export statement, function declaration or call) does not end with a trailing comma and does not lie on the same line as the closing curly brace, bracket or parenthesis.
```

vilchik-elena (approved):

PR #264, opened 2021-08-13
update s2076 PHP description + common metadata

eric-therond-sonarsource:


rules/S2076/php/rule.adoc
```diff
@@ -2,27 +2,48 @@ include::../description.adoc[]
...
 
-// If the value "/sbin/shutdown" is passed as binary and the web server is running as root,
+// If the command "/sbin/shutdown" is executed and the web server is running as root,
 // then the machine running the web server will be shut down and become unavailable for future requests
 
-exec( $binary ); // Noncompliant
+exec($cmd); // Noncompliant
+----
+
+PHP OS commands allow the use of shell meta-characters and thus arbitrary commands can be executed:
```
hendrik-buchwald-sonarsource, line 16: Do you mean the PHP functions to execute commands with `PHP OS commands`? At the moment it sounds like it is talking about `php-cli`.

hendrik-buchwald-sonarsource (approved):
As with the argument injection, it might make sense to briefly explain that also controlling a single argument is not necessarily secure.

rules/S2076/php/rule.adoc
eric-therond-sonarsource: Yes it's not clear maybe:
```PHP functions that execute commands authorize shell meta-characters to be used and thus additional and arbitrary commands can be executed:```

hendrik-buchwald-sonarsource: Maybe ```Shell meta-characters can be used to execute additional commands in PHP functions that execute commands:```?

eric-therond-sonarsource: looks good

PR #265, opened 2021-08-13
Create rule S5883[php]: OS commands should not be vulnerable to argument injection attacks

eric-therond-sonarsource:
ticket: https://jira.sonarsource.com/browse/SONARSEC-2622


rules/S5883/php/rule.adoc
```diff
@@ -1 +1,43 @@
...
+`escapeshellcmd` doesn't prevent additional arguments to be injected:
+
+----
+exec(escapeshellcmd("/usr/bin/cat ".$_GET["arg"])); // Noncompliant
+----
+
+`escapeshellcmd` cancels any prior sanitization with `escapeshellarg`:
+
+----
+exec(escapeshellcmd("/usr/bin/cat ".escapeshellarg($_GET["arg"])); // Noncompliant
```
hendrik-buchwald-sonarsource, line 14:
```suggestion
exec(escapeshellcmd("/usr/bin/find ".escapeshellarg($_GET["arg"])); // Noncompliant
```

I like `find` over `cat` as an example command since `cat` likely does not have any dangerous options while `find` has.

rules/S5883/php/rule.adoc
```diff
@@ -1 +1,43 @@
-include::../rule.adoc[]
+include::../description.adoc[]
+
+== Noncompliant Code Example
+
+`escapeshellcmd` doesn't prevent additional arguments to be injected:
```
hendrik-buchwald-sonarsource, line 5: I am not completely sure myself how this has to be formulated but the current formulation does not feel right to me. Maybe like this?

```suggestion
`escapeshellcmd` doesn't prevent additional arguments from being injected:
```

rules/S5883/php/rule.adoc
```diff
@@ -1 +1,43 @@
...
+exec(escapeshellcmd("/usr/bin/cat ".$_GET["arg"])); // Noncompliant
+----
+
+`escapeshellcmd` cancels any prior sanitization with `escapeshellarg`:
+
+----
+exec(escapeshellcmd("/usr/bin/cat ".escapeshellarg($_GET["arg"])); // Noncompliant
+----
+
+fifth argument of the `mail` function accepts arguments that are added to the command line:
```
hendrik-buchwald-sonarsource, line 17: Since it is a new sentence and does not start with a function name I would capitalize it.

```suggestion
Fifth argument of the `mail` function accepts arguments that are added to the command line:
```

rules/S5883/php/rule.adoc
```diff
@@ -1 +1,43 @@
...
+
+== Compliant Solution
+
+`escapeshellarg` should be used to sanitize a specific argument:
+
+----
+exec("/usr/bin/cat ".escapeshellarg($_GET["arg"]));
+----
+
+fifth argument of the `mail` function can be secured with the use of an allowlist:
```
hendrik-buchwald-sonarsource, line 34: Same.

```suggestion
Fifth argument of the `mail` function can be secured with the use of an allowlist:
```

hendrik-buchwald-sonarsource (approved):
I left some suggestions.

It might make sense to briefly mention that user-controlled arguments still might be a security problem. The compliant solution will raise a hotspot, so then the developers would know why.

rules/S5883/php/rule.adoc
eric-therond-sonarsource: I can explain that in the generic description of the rule for sure, something like that:
`Command arguments constructed from user-controlled data should be sanitized in order to escape argument delimiters (eg: ', space, - ) and preventing injection of additional arguments. A single user-controlled argument may still lead to vulnerabilities if it corresponds to a dangerous option supported by the command, like -exec with the find command, in that case signal the end of options processing in the command line by the use of -- (double dash) or restrict the options to only trusted values.`

About the command itself I think `cat` could be still used but in a way to indicate that authorizing additional arguments can lead to some problems (it's not visible in the current code examples I agree):
```
// the attacker can cat files outside of the intended /public directory by passing additional arguments /public/file.txt /private/data.csv
exec("/usr/bin/cat /public/".$_GET["arg"]);
```

This rule is mainly about **additional** argument thus I am not super comfortable by using a command in code examples that allows a single dangerous argument like `-exec` with `find`.

what do you think?

hendrik-buchwald-sonarsource: Yea, makes sense!

PR #270, opened 2021-08-18
Create rule S5855[javascript]: Regex alternatives should not be redundant

yassin-kammoun-sonarsource:


vilchik-elena (approved):

vilchik-elena:
We decided to not implement this rule as it requires automata

PR #271, opened 2021-08-18
Create rule S5996[javascript]: Regex boundaries should not be used in a way that can never be matched

yassin-kammoun-sonarsource:


vilchik-elena (approved):

vilchik-elena:
We decided to not implement this rule as it requires automata

PR #273, opened 2021-08-19
Create rule S5843[javascript]: Regular expressions should not be too complicated

yassin-kammoun-sonarsource:


rules/S5843/javascript/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+
+The complexity of a regular expression is determined as follows:
+
+
+Each of the following operators increases the complexity by an amount equal to the current nesting level and also increases the current nesting level by one for its arguments:
+
+
+* ``++|++`` - when multiple ``++|++`` operators are used together, the subsequent ones only increase the complexity by 1
+* ``++&&++`` (inside character classes) - when multiple ``++&&++`` operators are used together, the subsequent ones only increase the complexity by 1
```
vilchik-elena, line 11: I don't think such syntax exists in JS

rules/S5843/javascript/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+* Quantifiers (``++*++``, ``+``, ``++?++``, ``++{n,m}++``, ``++{n,}++`` or ``++{n}++``)
+* Non-capturing groups that set flags (such as ``++(?i:some_pattern)++`` or ``++(?i)some_pattern++``)
```
vilchik-elena, line 13: and this I also didn't find

rules/S5843/javascript/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+* Lookahead and lookbehind assertions
+
+Additionally, each use of the following features increase the complexity by 1 regardless of nesting:
+
+
+* character classes
+* back references
+
+If a regular expression is split among multiple variables, the complexity is calculated for each variable individually, not for the whole regular expression. If a regular expression is split over multiple lines, each line is treated individually if it is accompanied by a comment, otherwise the regular expression is analyzed as a whole.
```
vilchik-elena, line 22: I would remove `if it is accompanied by a comment, otherwise the regular expression is analyzed as a whole.`. Seems like a lot for trouble for little value

rules/S5843/javascript/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+    const dateParts[] = dateString.split("[-/.]");
+    int day = parseInt(dateParts[0]);
+    int month = parseInt(dateParts[1]);
+    int year = parseInt(dateParts[2]);
+    // Put logic to validate and process the date based on its integer parts here
+}
+----
+
+
+== Exceptions
```
vilchik-elena, line 49: I would drop this section. I think it duplicates what was said already

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #280, opened 2021-08-23
Create rule S5867[javascript]: Regular expressions using Unicode character classes or property escapes should enable the unicode flag

yassin-kammoun-sonarsource:


rules/S5867/javascript/rule.adoc
```diff
@@ -0,0 +1,21 @@
+When using Unicode classes like `+\p{Alpha}+` without the `+u+` flag or when using hard-coded character classes like `+[a-zA-Z]+`, letters outside of the ASCII range, such as umlauts, accented letters or letter from non-Latin languages, won't be matched. This may cause code to incorrectly handle input containing such letters.
+
+
+To correctly handle non-ASCII input, it is recommended to use Unicode classes like `+\p{Alpha}+` and enable Unicode support by using the `+u+` flag with the regex.
+
+
+== Noncompliant Code Example
+
+----
+/[a-zA-Z]/
```
vilchik-elena, line 10: tbh my idea for this rule was to not report here. I think it's noisy, not everyone needs to match non-latin letters. 
I think we should report only when `\p{}`/`\P{}` syntax is used while `u` flag is not enabled

rules/S5867/metadata.json
```diff
@@ -1,2 +1,29 @@
 {
+  "title": "Unicode-aware versions of character classes should be preferred",
```
vilchik-elena, line 2: then title should be JS-specific

vilchik-elena (changes requested):

rules/S5867/javascript/rule.adoc
```diff
@@ -0,0 +1,21 @@
...
+/\p{Alpha}/
```
vilchik-elena, line 11: Example should show the wrong behaviour

rules/S5867/metadata.json
```diff
@@ -1,2 +1,29 @@
...
+  "type": "CODE_SMELL",
```
vilchik-elena, line 3: I know it's not yet done, but I think we could make it `BUG` for JS

vilchik-elena (approved):

PR #286, opened 2021-08-30
Modify rule S6216[java]: change "in run-time" by "at runtime"

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM

PR #287, opened 2021-08-31
Modify rule S5361: Add PHP as covered language

nils-werner-sonarsource:


rules/S5361/php/metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "title": "Rule S5361: `str_replace` should be preferred to `preg_replace`"
```
pynicolas, line 2:
```suggestion
  "title": "`str_replace` should be preferred to `preg_replace`"
```

PR #300, opened 2021-09-02
Modify rule S3655[java]: mention "isEmpty" and indirect access in code examples

quentin-jaquier-sonarsource:
Inspired from https://github.com/SonarSource/sonar-java/pull/3724.

### DIFF
--- a/rules/S3655/java/rule.adoc
+++ b/rules/S3655/java/rule.adoc
@@ -15,7 +15,22 @@ Optional<String> value = this.getOptionalValue();
 String stringValue = value.get(); // Noncompliant
 ----
 
+----
+if (methodThatReturnsOptional().isEmpty()) {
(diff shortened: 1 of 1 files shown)


rules/S3655/java/rule.adoc
```diff
@@ -25,6 +29,11 @@ Optional<String> value = this.getOptionalValue();
 if (value.isPresent()) {
   String stringValue = value.get();
 }
+
```
alban-auzeill, line 32: We should show as a first compliant example the usage of:
```
optional.ifPresent(value -> doSomethingWith(value, "name"));
```

quentin-jaquier-sonarsource: I updated the examples, let me know what you think!

alban-auzeill (approved):

PR #303, opened 2021-09-02
Update rule S2699 [csharp]: Mention new supported frameworks

csaba-sagi-sonarsource:


rules/S2699/csharp/rule.adoc
```diff
@@ -3,11 +3,13 @@ A test case without assertions ensures only that no exceptions are thrown. Beyon
 
 This rule raises an exception when no assertions from any of the following frameworks are found in a test:
 
-* ``++MSTest++``
-*  ``++NUnit++``
-* ``++XUnit++``
-* ``++FluentAssertions++`` (4.x and 5.x)
-* ``++NSubstitute++``
+* `FluentAssertions` (4.x and 5.x)
```
andrei-epure-sonarsource, line 6: I would group unit test frameworks (MSTest, NUnit, xUnit) together , and the others after - this would be a more logical way

rules/S2699/csharp/rule.adoc
```diff
@@ -3,11 +3,13 @@ A test case without assertions ensures only that no exceptions are thrown. Beyon
...
+* `MSTest`
+* `NFluent`
+* `NSubstitute`
+* `NUnit`
+* `Shoudly`
+* `XUnit`
```
andrei-epure-sonarsource, line 12:
```suggestion
* `xUnit`
```

https://xunit.net/

andrei-epure-sonarsource (approved):
LGTM, please consider the comments

rules/S2699/csharp/rule.adoc
andrei-epure-sonarsource: `NUnit` is fine https://nunit.org/
and MSTest is fine

PR #308, opened 2021-09-03
Modify rule S874 (CPP-3133) Handle combined operation & C++20

loic-joly-sonarsource:
- False positive when combining operations due to integer promotion
- The problem mostly disappears in C++20


rules/S874/cfamily/rule.adoc
```diff
@@ -32,6 +43,13 @@ If the right-side operand to a shift operator is known at compile time, it is ac
 foo = 15u >> SHIFT;
 ----
 
+When combining several bitwise operations, even if all leaf operands are unsigned, if they are smaller than an `int`, some intermediate results will be of type `signed int`, due to integral promotion. However, this situation is usually not an issue, and is an exception for this rule:
```
tomasz-kaminski-sonarsource, line 46: I checked the C++17 standard, and that seems to be correct, as according to the [basic.fundamental] p3 [final-draft:](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf)
>> he range of non-negative values of a signed integer type is a subrange of
the corresponding unsigned integer type, the representation of the same value in each of the two types is
the same, and the value representation of each corresponding signed/unsigned type shall be the same

tomasz-kaminski-sonarsource (approved):

PR #311, opened 2021-09-06
Create rule S6218[kotlin]: Equals method should be overridden in data…

margarita-nedzelska-sonarsource:
… classes containing array fields


rules/S6218/metadata.json
```diff
@@ -1,2 +1,27 @@
...
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [
+  
+  ],
+  "extra": {
+    "coveredLanguages": [
+      "java",
+      "kotlin"
```
johann-beleites-sonarsource, line 14:
```suggestion
      "Java",
      "Kotlin"
```

rules/S6218/kotlin/rule.adoc
```diff
@@ -0,0 +1,40 @@
+In data classes, the default behavior of the ``++equals()++`` method is to check the equality by field values. This works well for primitive fields or fields, whose type overrides ``++equals()++``, but this behavior doesn't work as expected for array fields.
```
johann-beleites-sonarsource, line 1:
```suggestion
In data classes, the default behavior of the `equals()` method is to check the equality by field values. This works well for primitive fields or fields, whose type overrides `equals()`, but this behavior doesn't work as expected for array fields.
```

rules/S6218/kotlin/rule.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+
+By default, array fields are compared by their reference, and overriding ``++equals()++`` is highly appreciated to achieve the deep equality check. The same strategy applies to the ``++hashcode()++`` method.
```
johann-beleites-sonarsource, line 4:
```suggestion
By default, array fields are compared by their reference, so overriding `equals()` is highly recommended to ensure a deep equality check. The same applies to the `hashcode()` method.
```

rules/S6218/kotlin/rule.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+
+This rule reports an issue if a record class has an array field and is not overriding ``++equals()++`` or ``++hashcode()++`` methods.
```
johann-beleites-sonarsource, line 7:
```suggestion
This rule reports an issue if a record class has an array field and is not overriding `equals()` or `hashcode()` methods.
```

johann-beleites-sonarsource (approved):

PR #313, opened 2021-09-06
Modify rule S1705: FP when the returned value of postfix inc is actually used (CPP-3246)

loic-joly-sonarsource:


tomasz-kaminski-sonarsource (approved):
The change itself looks good. However, I am not convinced on their justification.

PR #317, opened 2021-09-07
Modify rule S5855: Add PHP as covered language

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
LGTM! Do we want to add delimiters even when we are not in a preg function?

PR #319, opened 2021-09-07
Create rule S1143[kotlin]: Jump statements should not occur in "final…

margarita-nedzelska-sonarsource:
…ly" blocks


rules/S1143/kotlin/rule.adoc
```diff
@@ -0,0 +1,71 @@
+Using `return`, `break`, `throw`, and so on from a `finally` block suppresses the propagation of any unhandled `Throwable` which was thrown in the `try` or `catch` block.
+
+
+This rule raises an issue when a jump statement (`break`, `continue`, `return`, `throw`) would force control flow to leave a `finally` block.
```
margarita-nedzelska-sonarsource, line 4: The original description mentions `goto` as well, I would rather write a dedicated description here.

rules/S1143/kotlin/rule.adoc
```diff
@@ -0,0 +1,71 @@
+Using `return`, `break`, `throw`, and so on from a `finally` block suppresses the propagation of any unhandled `Throwable` which was thrown in the `try` or `catch` block.
```
johann-beleites-sonarsource, line 1:
```suggestion
Using `break`, `continue`, `return` and `throw` inside of a `finally` block suppresses the propagation of any unhandled `Throwable` thrown in the `try` or `catch` block.
```

rules/S1143/kotlin/rule.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+== Compliant Solution
+
+----
+fun main() {
+    try {
+        doSomethingWhichThrowsException()
+        println("OK") // incorrect "OK" message is printed
+    } catch (e: RuntimeException) {
+        println("ERROR") // this message is not shown
```
johann-beleites-sonarsource, line 44: I suppose we can modify these two comments to reflect the opposite in the compliant variant?

rules/S1143/kotlin/rule.adoc
```diff
@@ -0,0 +1,71 @@
...
+----
+
+
+ifdef::env-github,rspecator-view[]
+'''
+== Comments And Links
+(visible only on this page)
+
+include::../comments-and-links.adoc[]
+endif::env-github,rspecator-view[]
```
johann-beleites-sonarsource, line 71: Never sure if we want this, to me it seems like a relict from JIRA. No strong opinion, though.

rules/S1143/kotlin/rule.adoc
margarita-nedzelska-sonarsource: Copy/Paste issue :)

rules/S1143/kotlin/rule.adoc
margarita-nedzelska-sonarsource: I think this to show comments from Jira on the RSPEC. No strong opinion here to, but try to be consistent with other languages

johann-beleites-sonarsource (approved):
LGTM!

PR #325, opened 2021-09-08
Create rule S2151[Kotlin]: "runFinalizersOnExit" should not be called

johann-beleites-sonarsource:


rules/S2151/java/rule.adoc
```diff
@@ -39,9 +28,7 @@ public static void main(String [] args) {
 ----
 
 
-== See
-
-* https://wiki.sei.cmu.edu/confluence/x/4jZGBQ[CERT, MET12-J.] - Do not use finalizers
+include::../description-tail.adoc[]
```
margarita-nedzelska-sonarsource, line 31: Not sure if Cert or MET12-J are relevant for Kotlin

johann-beleites-sonarsource: In principal I agree, it would be nicer to have Kotlin-specific links here. The reason as to why `runFinalizersOnExit` shouldn't be called is JVM specific and I couldn't find a Kotlin-specific article, even if the same holds for any JVM-based language. In the linked text they obviously assume we're talking Java but the explanation is based on the workings of the JVM. For this sort of rule I think it is nice to have some further reading, for Kotlin it doesn't seem to exist yet. So the next-best thing, IMO, is to link a Java text. WDYT?

johann-beleites-sonarsource: How about adding a note to the link informing the reader that they are about to land on a Java-centric page, that the inner workings of the JVM also apply to Kotlin here, though?

margarita-nedzelska-sonarsource: Makes sense to me

PR #326, opened 2021-09-08
Create rule S2122[Kotlin]: "ScheduledThreadPoolExecutor" should not have 0 core threads

johann-beleites-sonarsource:


rules/S2122/description.adoc
```diff
@@ -0,0 +1,4 @@
+``++java.util.concurrent.ScheduledThreadPoolExecutor++``'s pool is sized with ``++corePoolSize++``, so setting ``++corePoolSize++`` to zero means the executor will have no threads and run nothing.
+
+
+This rule detects instances where ``++corePoolSize++`` is set to zero, via either its setter or the object constructor.
```
margarita-nedzelska-sonarsource, line 4:
```suggestion
`java.util.concurrent.ScheduledThreadPoolExecutor`'s pool is sized with `corePoolSize`, so setting `corePoolSize` to zero means the executor will have no threads and run nothing.


This rule detects instances where `corePoolSize` is set to zero, via either its setter or the object constructor.
```

PR #327, opened 2021-09-08
Create rule S2189[kotlin]: Loops should not be infinite

margarita-nedzelska-sonarsource:


rules/S2189/kotlin/rule.adoc
```diff
@@ -0,0 +1,44 @@
+An infinite loop is one that will never end while the program is running, i.e., you have to kill the program to get out of the loop. Whether it is by meeting the loop's end condition or via a `break`, every loop should have an end condition.
```
johann-beleites-sonarsource, line 1: There seems to be a generic rule template used by multiple languages for this rule (see Java version, for instance). I think it would make sense to use that or at least import the shared text.

rules/S2189/metadata.json
```diff
@@ -12,7 +12,8 @@
   "extra": {
     "coveredLanguages": [
       "Java",
-      "JavaScript"
+      "JavaScript",
+      "Kotlin"
```
johann-beleites-sonarsource, line 16: Interesting, this isn't something we should necessarily fix with this PR, I wonder why all the other languages that support this rule aren't listed here.

margarita-nedzelska-sonarsource: I am not sure which languages support have exactly this rule implemented and which just have a folder here. I am not sure what we can do with it

johann-beleites-sonarsource (approved):

PR #329, opened 2021-09-08
Update rule S2695: "PreparedStatement" and "ResultSet" method s should be called with valid indices

margarita-nedzelska-sonarsource:
…s should be called with valid indices


rules/S2695/kotlin/rule.adoc
```diff
@@ -0,0 +1,33 @@
+The parameters in a `PreparedStatement` are numbered from 1, not 0, so using any "set" method of a `PreparedStatement` with a number less than 1 is a bug, as is using an index higher than the number of parameters. Similarly, `ResultSet` indices also start at 1, rather than 0
```
johann-beleites-sonarsource, line 1: You can put the parts of the rule that are common for both Kotlin & Java in a shared file (e.g. `description.adoc`) and include them with `include::../description.adoc[]`. Might help in avoiding accidental inconsistencies in the future.

rules/S2695/description.adoc
```diff
@@ -0,0 +1 @@
+The parameters in a `PreparedStatement` are numbered from 1, not 0, so using any "set" method of a `PreparedStatement` with a number less than 1 is a bug, as is using an index higher than the number of parameters. Similarly, `ResultSet` indices also start at 1, rather than 0
```
johann-beleites-sonarsource, line 1:
```suggestion
The parameters in a `PreparedStatement` are indexed beginning at 1, not 0, so using any "set" method of a `PreparedStatement` with a number less than 1 is a bug, as is using an index higher than the number of parameters. The same indexing style also applies to `ResultSet`.
```

johann-beleites-sonarsource (approved):

PR #331, opened 2021-09-10
Delete rule S1258 for C++

loic-joly-sonarsource:
This rule was never implemented for C++, and does not really make sense...
 Removing C++ from the targeted languages (keeping the Java version).

Related to CPP-206


loic-joly-sonarsource:
@arseniy-sonar, I assigned to you not really because of the content, but to be sure I did the right thing to remove a language from a rule.

arseniy-sonar (approved):

PR #332, opened 2021-09-13
Rule S3059: update description and code examples

costin-zaharia-sonarsource:


rules/S3059/csharp/rule.adoc
```diff
@@ -1,4 +1,56 @@
-include::../rule.adoc[]
+There's no point in having a ``++public++`` member in a non-``++public++`` type because objects that can't access the type will never have the chance to access the member.
+
+
+This rule raises an issue when a type has methods, fields, or inner typs with higher visibility than the type itself has.
```
mary-georgiou-sonarsource, line 4: small typo here.

mary-georgiou-sonarsource (changes requested):
There's only a small typo in line 4 (tps -> types).

rules/S3059/csharp/rule.adoc
```diff
@@ -1,4 +1,56 @@
...
+    public static decimal PI = 3.14m;  // Noncompliant
+
+    public int GetOne() // Noncompliant
+    {
+        return 1;
+    }
+
+    protected record NestedType // Noncompliant: outer class is internal
+    {
+        public bool FlipCoin() // Noncompliant: parent class is protected
```
andrei-epure-sonarsource, line 21:
```suggestion
        public bool FlipCoin() // Noncompliant: parent type is protected
```

rules/S3059/csharp/rule.adoc
```diff
@@ -1,4 +1,56 @@
...
+        // ...
+    }
+}
+----
+
+
+== Compliant Solution
+
+----
+internal class MyClass // Class visibility upgrade makes members compliant
```
andrei-epure-sonarsource, line 34:
```suggestion
public class MyClass // Class visibility upgrade makes members compliant
```

?

rules/S3059/csharp/rule.adoc
```diff
@@ -1,4 +1,56 @@
...
+    public static decimal PI = 3.14m;
+
+    public int GetOne()
+    {
+        return 1;
+    }
+
+    protected record NestedType
+    {
+        protected bool FlipCoin() // visibility changed to match class
```
andrei-epure-sonarsource, line 45:
```suggestion
        protected bool FlipCoin() // visibility changed to match type
```

andrei-epure-sonarsource (changes requested):
Is this rule really SonarWay? 

I'm pretty sure that it's debatable, it should be evaluated on how noisy it is. AFAIK we don't respect this in our own codebase, so I can imagine other teams may not respect this.

I can imagine a philosophical debate between programmers on this topic.

costin-zaharia-sonarsource:
> Is this rule really SonarWay?
> I can imagine a philosophical debate between programmers on this topic.

The rule was specified as being part of SonarWay. I agree that this can cause some frustration for some developers and that we can discuss and implement it outside of the default profile.

I still find it useful since having a public method on an internal class can cause some confusion.

There are though some exceptions we have to keep in mind:
- implicit and explicit operators always have to be public (even if they are in a type with different visibility)
- private nested types need higher visibility for their members to make them available to the parent type.

andrei-epure-sonarsource (approved):
LGTM

mary-georgiou-sonarsource (approved):
LGTM

rules/S3059/csharp/rule.adoc
```diff
@@ -52,6 +52,33 @@ public class MyClass // Class visibility upgrade makes members compliant
...
+----
+
+Nested types, even if private, can be used and inherited in the parent type. In this case, the visibility of the outer type is considered.
+
+----
+internal class MyClass
+{
+    private class NestedClass
+    {
+        public int PublicProperty { get; } // Noncompliant: should be internal
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Not sure I understand why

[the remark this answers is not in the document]
andrei-epure-sonarsource: because if I want the parent class to have access to the fields, I just make them public.

Either way a private inner class is private. I don't think that changing the visibility of the outer class should be reflected on the members of each private inner type.

Does it make sense? I don't know how to phrase it in words...

And actually I realize as writing that this is an argument for the rule not being in SonarWay.

If I want my type members to have "maximum visibility of the type" - I just make them public. If I change the parent type from "internal" to public, I may not want to change the visibility of each individual type member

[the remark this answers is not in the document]
costin-zaharia-sonarsource: > Either way a private inner class is private. I don't think that changing the visibility of the outer class should be reflected on the members of each private inner type.
> If I want my type members to have "maximum visibility of the type" - I just make them public. If I change the parent type from "internal" to public, I may not want to change the visibility of each individual type member

I think that this argument can be generalized also for the members of the outer class and this will make the rule redundant.

On the other side, the visibility of the nested class members is even harder to get right (need to consider also the outer type) so raising an issue could be more helpful.

> And actually I realize as writing that this is an argument for the rule not being in SonarWay.
I agree with this rule not being in SonarWay. I will ask the team for feedback.

PR #333, opened 2021-09-13
RULEAPI-674 Searching for RSPEC-XXXX should return the same as searching for SXXXX

vilchik-elena:


frontend/src/utils/useSearch.ts
```diff
@@ -95,3 +94,8 @@ export function useSearch(query: string, ruleType: string|null, ruleLang: string
 
   return {results, numberOfHits, error, loading};
 }
+
+// allows to search by a rule key in the following formats: SXXX, RSPEC-XXX and XXX
+function amendQuery(query: string) {
+  return query.replace('RSPEC-', 'S').replace(/(?<!S)(?<=\s|^)(\d{3,})/i, 'S$1');
```
arseniy-sonar, line 100: I feel the second part is too aggressive. For example, will I be able to find a rule about HTML `404`?

arseniy-sonar (changes requested):

frontend/src/utils/useSearch.ts
vilchik-elena: indeed, I didn't think about this case

vilchik-elena: eventually I simplified and kept only first part. It appeared that adding `S###` would not work as our search tries to find rules having all words from the query so `100 S100` will find nothing.

arseniy-sonar (approved):

PR #335, opened 2021-09-13
Modify rule S6001: Add PHP as covered language

pynicolas:


karim-ouerghemmi-sonarsource (approved):
The Noncompliant/Compliant examples are missing the regex delimiters. Otherwise, LGTM.

PR #343, opened 2021-09-14
Update rule S899: Return values should not be ignored when they conta…

margarita-nedzelska-sonarsource:
…in the operation status code


rules/S899/kotlin/rule.adoc
```diff
@@ -0,0 +1,41 @@
...
+    }
+    if (!file.delete()) {
+        // file delete failed; take appropriate action
+    }
+}
+----
+
+== See
+
+Although these resources talks about Java, the underlying information concerning the JVM are just as relevant for Kotlin.
```
johann-beleites-sonarsource, line 29:
```suggestion
Although these resources talk about Java, the underlying information concerning the JVM is just as relevant for Kotlin.
```

johann-beleites-sonarsource (approved):

PR #344, opened 2021-09-14
Update rule S2114: Collections should not be passed as arguments to their own methods

johann-beleites-sonarsource:


rules/S2114/kotlin/compliant.adoc
```diff
@@ -0,0 +1,13 @@
...
+----
+fun compliant(newList: MutableList<Any>) {
+    val objs = mutableListOf<Any>()
+    objs.add("Hello")
+
+    objs.containsAll(newList) // Compliant
+    objs.addAll(newList) // Compliant
+    objs.removeAll(newList) // Compliant
+    objs.retainAll(newList) // Compliant
+}
```
johann-beleites-sonarsource, line 12: I think it would be nice to look more consistent in how we add the compliant/noncompliant examples. I know that we use the parameter `newList` here, so I think we should either wrap the noncompliant example in a `fun` as well, or we declare `newList` as another `val`.

johann-beleites-sonarsource (commented):
We need to rebase on master before we can merge this. There have been some changes to the metadata format since the PR was opened, for instance. Keys like `coveredLanguages` don't exist anymore and can be removed.

PR #347, opened 2021-09-15
Update rule S2272: "Iterator.next()" methods should throw "NoSuchElementException"

johann-beleites-sonarsource:


vilchik-elena:
ping @johann-beleites-sonarsource

PR #354, opened 2021-09-16
RULEAPI-658 Do not diplay 'default' in languages on search page

vilchik-elena:


frontend/src/SearchHit.tsx
```diff
@@ -38,7 +38,8 @@ type SearchHitProps = {
 
 export function SearchHit(props: SearchHitProps) {
   const classes = useStyles();
-  const languages = props.data.languages.map(lang => (
+  const specifiedLanguages = props.data.languages.filter(language => language !== 'default');
```
arseniy-sonar, line 41: "specified" implies some one has specified (selected) them, which sounds misguiding to me.
How about:
```suggestion
  const actualLanguages = props.data.languages.filter(language => language !== 'default');
```
or
```suggestion
  const realLanguages = props.data.languages.filter(language => language !== 'default');
```
?

arseniy-sonar (approved):

PR #364, opened 2021-09-17
Modify rule S6226: Disable rule on Windows

marco-antognini-sonarsource:


rules/S6226/cfamily/rule.adoc
```diff
@@ -10,6 +10,8 @@ If the type is not empty, this attribute is still valid, and has no effect. This
 This rule raises an issue on each member of a class that has an empty or potentially empty (in case of templates) type and does not have ``\[[no_unique_address]]`` attribute.
 
 
+NOTE: This rule is disabled on Windows because ``\[[no_unique_address]]`` isn't supported by MSVC and Clang on this platform. See https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/#c20-no_unique_address[MSVC C{pp}20 and the /std:c{pp}20 Switch].
```
loic-joly-sonarsource, line 13: Even thjough I like it, I don't think the NOTE tag has ever been tested in our RSPECs, I'm not sure how it could be rendered on all the platforms that display rspecs. So:
- Either you replace it be something less fancy.
- Or you make a note for future you to check it on SC, SQ, SL (different variants) and rules.sonarsource.org


```suggestion
NOTE: This rule is disabled on Windows because ``\[[no_unique_address]]`` isn't supported by MSVC and Clang on this platform. See https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/#c20-no_unique_address[MSVC {cpp}20 and the `/std:c++20` Switch].
```
Or maybe less visually overwhelming:
```suggestion
NOTE: This rule is disabled on Windows because ``\[[no_unique_address]]`` https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/#c20-no_unique_address[isn't well supported by MSVC and Clang on this platform].
```

loic-joly-sonarsource (changes requested):

rules/S6226/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks. I went for your second solution, it is indeed better. As for `NOTE`, I think I will keep the scope of my ticket relatively tight for now. I'll keep it mind for another time, as it could be useful generally speaking.

loic-joly-sonarsource (approved):

PR #368, opened 2021-09-20
Modify rule S6226: Add exception section

marco-antognini-sonarsource:


rules/S6226/cfamily/rule.adoc
```diff
@@ -48,6 +48,9 @@ class HashMap {
 };
 ----
 
+== Exceptions
+
+This rule does not apply to field whose class has a non-default alignment.
```
arseniy-sonar, line 53:
```suggestion
This rule does not apply to fields whose class has a non-default alignment.
```

tomasz-kaminski-sonarsource (approved):

PR #372, opened 2021-09-20
RULEAPI-683 Reference the default language and each language by the tag

vilchik-elena:


csaba-sagi-sonarsource (approved):
LGTM. Please take a look at RULEAPI-600. I would say this PR fixes that one as well. It is now possible to copy the rule ID without any bigger trouble.

PR #373, opened 2021-09-20
RULEAPI-675: When searching for a rule-id the results should start with the corresponding RSPEC

csaba-sagi-sonarsource:


frontend/public/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html prefix="og: https://ogp.me/ns#" lang="en">
```
vilchik-elena, line 2: @csaba-sagi-sonarsource should it be in a different PR?

csaba-sagi-sonarsource: Yep, my bad, I`ll move it to a separate PR.

vilchik-elena (approved):
@csaba-sagi-sonarsource I refactored your implementation, let me know what you think

PR #378, opened 2021-09-20
Modify rule S3260 

mary-georgiou-sonarsource:
Modify rule S3260: Include also record types, since the rule is also valid for them.


costin-zaharia-sonarsource (approved):
LGTM!

PR #363, opened 2021-09-17
Modify rule S6229: Improve description

marco-antognini-sonarsource:
Fix a few typos and add the "noncompliant solution" section.

PR #379, opened 2021-09-20
RULEAPI-684: Restore the section skipped during import from Jira

arseniy-sonar:
The issue that caused the section skipping: non-canonical section names. These were mostly fixed before the export point. A few freshly created rules, days before the transition had non-canonical sections. They are mentioned in the log during export (but were missed or ignored during export). The affected rules are (all c-family): S6223, S6227, S6229, S6231

S6223, and S6231 were fixed earlier
S6229 is fixed by https://github.com/SonarSource/rspec/pull/363
This PR fixes the remaining S6227


csaba-sagi-sonarsource (approved):
LGTM

PR #380, opened 2021-09-20
Modify rule S6004 : Update after changes provided by CPP-2843

loic-joly-sonarsource:
- Small typos introduced when adding range-based for loops
- The rule no longer triggers if there is no code for the variable to leak into

### DIFF
--- a/rules/S6004/cfamily/rule.adoc
+++ b/rules/S6004/cfamily/rule.adoc
@@ -1,4 +1,4 @@
-{cpp}17 introduced a construct to create and initialize a variable within the condition of ``++if++`` and ``++switch++`` statements and {cpp}20 added this construct to range-based for loops. Using this new feature simplifies common code patterns and helps in giving variables the right scope.
+{cpp}17 introduced a construct to create and initialize a variable within the condition of `if` and `switch` statements and {cpp}20 added this construct to range-based `for` loops. Using this new feature simplifies common code patterns and helps in giving variables the right scope.
 
 
 Previously, variables were either declared before the statement, hence leaked into the ambient scope, or an explicit scope was used to keep the scope tight, especially when using RAII objects. This was inconvenient as it would lead to error-prone patterns. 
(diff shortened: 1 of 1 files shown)


rules/S6004/cfamily/rule.adoc
```diff
@@ -7,32 +7,33 @@ Previously, variables were either declared before the statement, hence leaked in
...
+void error_prone_init() {
   { // explicit scope
     std::unique_lock<std::mutex> lock(mtx, std::try_to_lock);
-    if (lock.owns_lock())
-    {
+    if (lock.owns_lock()) {
        //...
      }
   } //mutex unlock
   // ... code
```
abbas-sabra-sonarsource, line 17: should we update this example too to make it trigger?

phil-nash-sonarsource: +1

phil-nash-sonarsource (commented):
Looks good - lots of welcome fixes. But there are also a few cases of "Statment" instead of "Statement" in function names. Probably worth fixing those, too, while you are there :-)

rules/S6004/cfamily/rule.adoc
loic-joly-sonarsource: I believed `// ... code` would be enough for the explanation..
Any idea of which code to write?

tomasz-kaminski-sonarsource: I would suggest changing return type to `bool` and having `return true` in `owns_lock` branch and `return false` in the other branch.  Something like:
```
bool error_prone_init() {
{
    std::unique_lock<std::mutex> lock(mtx, std::try_to_lock);
    if (lock.owns_lock()) {
    {
       //...
       return true;
     }
  // ... code
 return false;
}
```
Note that explicit scope needs to be removed, because otherwise it is always compilant.

loic-joly-sonarsource: Thanks for the suggestion. I selected not to return in the true branch, because otherwise the code could also have been written with an else, and the variable initialized outside.
I also don't think it's an issue if the code is compliant: We don't claim this in non compliant code, just that it's code that can be rewritten too using the same technique. Maybe we could someday extend this rule to make this code non compliant (I created CPP-3273 for that).

PR #382, opened 2021-09-21
Create rule S2053[kotlin]: Hashes should include an unpredictable salt

johann-beleites-sonarsource:


rules/S2053/kotlin/rule.adoc
```diff
@@ -0,0 +1,40 @@
+include::../description.adoc[]
+
+include::../recommended.adoc[]
+
+== Noncompliant Code Example
```
margarita-nedzelska-sonarsource, line 5: Do you think it is worth mentioning that in examples that simple Random is not secure too?

margarita-nedzelska-sonarsource (approved):
Looks good!

PR #386, opened 2021-09-21
RULEAPI-673: Stop displaying "Not implemented" for languages that do implement the rule

csaba-sagi-sonarsource:
This PR also changes the behaviour of the Edit on Github link. Before if there was a PR in progress for any language it always opened the folder in the branch from which the PR was created. Now the behaviour remains the same for not implemented languages, however it targets the master branch for languages that already have an implementation.

### DIFF
--- a/frontend/src/RulePage.tsx
+++ b/frontend/src/RulePage.tsx
@@ -189,6 +189,11 @@ export function RulePage(props: any) {
     }
   }
 
+  if (coverage != "Not Covered") {
+    prUrl = undefined;
+    branch = 'master'; 
+  }
+
   let editOnGithubUrl = 'https://github.com/SonarSource/rspec/blob/' +
                         branch + '/rules/' + ruleid + (language ? '/' + language : '');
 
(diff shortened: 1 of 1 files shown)


arseniy-sonar (commented):
This is definitely an improvement. The PR link still makes little sense for languages unrelated to the PR, but it is difficult to determine the intent of the PR author "do they plan to cover yet another language?"

One thing still looks inconsistent, though: the "non implemented" badge on the rule card on the search page. right now, it is displayed for the entire rule giving an impression that the rule is not implemented at all, even if for some languages it is. Could you, maybe, list what languages the rule is not implemented for?
Feel free to disregard the wish and merge if you deem the change to require too much effort.

csaba-sagi-sonarsource:
> This is definitely an improvement. The PR link still makes little sense for languages unrelated to the PR, but it is difficult to determine the intent of the PR author "do they plan to cover yet another language?"
> 
> One thing still looks inconsistent, though: the "non implemented" badge on the rule card on the search page. right now, it is displayed for the entire rule giving an impression that the rule is not implemented at all, even if for some languages it is. Could you, maybe, list what languages the rule is not implemented for?
> Feel free to disregard the wish and merge if you deem the change to require too much effort.

I`ll take a look to see how much effort is needed to implement it. Thanks for the suggestion.

PR #387, opened 2021-09-21
RULEAPI-670 Print the metadata.json path for the files that break a json parser

vilchik-elena:
For validation see https://cirrus-ci.com/task/4900259752050688?logs=metadata_tests#L16


andrei-epure-sonarsource (approved):
LGTM

PR #390, opened 2021-09-22
Modify rule S6045: Clarify required step for custom comparators

tomasz-kaminski-sonarsource:


rules/S6045/cfamily/rule.adoc
```diff
@@ -18,8 +18,9 @@ struct StringHash {
...
+Prefer using a transparent comparator with associative "std::string" containers to avoid creating the temporary. Note that transparent comparators are strongly discouraged if used with types that are not directly comparable as it will lead to the creation of ``++O(log(container.size())))++`` temporaries with lookup functions such as ``++find++``, ``++count++``, and ``++lower_bound++``.
 
+In case when custom non-transparent functor (comparator, equality or hasher) is used, it may have different semantics than corresponding operators on `std:::string`. In such case, the heterogenous lookup can still be enabled, by declaring the ``++is_transparent++`` nested type in the fuctor, and adjusting the implementation to accept either ``++std::string_view++`` or any type (i.e. turning it into template). The later change is required, to avoid creation of `std::string` temporaries for each invocation and thus degradation of perfomance.
```
marco-antognini-sonarsource, line 23:
```suggestion
Custom non-transparent functor (comparator, equality or hasher) may have different semantics than corresponding operators on `std:::string`. In such case, the heterogeneous lookup can still be enabled, by declaring the ``++is_transparent++`` nested type in the functor, and adjusting the implementation to accept either ``++std::string_view++`` or any type (i.e. turning it into a template). The later change is required to avoid the creation of `std::string` temporaries for each invocation and thus degradation of performance.
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -74,6 +89,22 @@ void g() {
   std::string_view n{"Nemo"};
   m.find(n); // No need to create a std::string
 }
+
+struct UpToTenLess {
+  using is_transparent = void;
+
+  bool operator()(const std::string_view lhs, const std::string_view rhs) const {
```
marco-antognini-sonarsource, line 96:
```suggestion
  bool operator()(std::string_view lhs, std::string_view rhs) const {
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -18,8 +18,9 @@ struct StringHash {
 };
 ----
 
-Always prefer using a transparent comparator with associative "std::string" containers to avoid creating the temporary. Note that transparent comparators are strongly discouraged if used with types that are not directly comparable as it will lead to the creation of ``++O(log(container.size())))++`` temporaries with lookup functions such as ``++find++``, ``++count++``, and ``++lower_bound++``.
+Prefer using a transparent comparator with associative "std::string" containers to avoid creating the temporary. Note that transparent comparators are strongly discouraged if used with types that are not directly comparable as it will lead to the creation of ``++O(log(container.size())))++`` temporaries with lookup functions such as ``++find++``, ``++count++``, and ``++lower_bound++``.
```
marco-antognini-sonarsource, line 21:
```suggestion
Prefer using a transparent comparator with associative ``++std::string++`` containers to avoid creating the temporary. Note that transparent comparators are strongly discouraged if used with types that are not directly comparable as it will lead to the creation of ``++O(log(container.size())))++`` temporaries with lookup functions such as ``++find++``, ``++count++``, and ``++lower_bound++``.
```
for consistency.

marco-antognini-sonarsource (changes requested):
Clarification looks good overall.

marco-antognini-sonarsource (approved):
LGTM

PR #391, opened 2021-09-22
RULEAPI-678: Escape """ in search strings embedded into the rule pages rendered in frontend

csaba-sagi-sonarsource:
Implementation based on 

> Special characters aren’t stored in the index, which means you can’t search for them. The index only keeps text and numbers, so searching for "\\[Jira Software\\]" and "Jira Software" will have the same effect — escaped special characters ([]) won’t be included in the search. `

from https://confluence.atlassian.com/jirasoftwareserver0813/search-syntax-for-text-fields-1027135330.html


vilchik-elena (approved):
manually tested

PR #392, opened 2021-09-22
RULEAPI-687: Migrate legacy keys from Jira RSPEC

arseniy-sonar:
I exported legacy keys from Jira with a modified rule-api jar (see [these three commits](https://github.com/SonarSource/sonar-rule-api/compare/1a309ce1...az/RULEAPI-687-export-legacy-keys)) and then used [this script](https://github.com/SonarSource/languages-experimental-tooling/blob/master/personal/arseniy-zaostrovnykh/rspec/integrate-legacy-keys.js) to add the `legacyKeys` field into every metadata.json

For the record: if rule is defined for 1 language to keep the top level JSON empty, the legacy key is placed in the language-specific JSON. Otherwise (if >1 language), all legacy keys from JIRA are aggregated (across subtasks) and placed into top level JSON. That way we might loose some precision, but it's ok.


vilchik-elena (approved):
Please revmov

vilchik-elena:
hey @arseniy-sonar, I again can't comment on the changes so just commenting here. 

Please remove `Web:`, it's repo prefix, we don't need it here. Otherwise LGTM

arseniy-sonar:
@vilchik-elena, as discussed, I moved `legacyKeys` inside `meta`, because it is irrelevant for analyzers and should be skipped during metadata generation by rule-api. Used [this script](https://github.com/SonarSource/languages-experimental-tooling/blob/master/personal/arseniy-zaostrovnykh/rspec/move-legacykeys-to-extra.js) to achieve that.

vilchik-elena (approved):

PR #394, opened 2021-09-23
RULEAPI-668: Take the sonar-security analyzer into account when collecting rule coverage in the coverage script

arseniy-sonar:
I hope you don't mind the first two commits that do just some refactoring without changing the functionality of the script (it does sort the fields in json now: for the ease of comparison later).
I validated that the script output did not change using `diff <(jq --sort-keys . ../covered_rules.json) <(jq --sort-keys . covered_rules.json)`.

The functional changes are in the last commit:

- adding `sonar-security` into the list of scanned repositories; 
- prefixing versions displayed in the coverage information with the repository name, since now the language rule can be implemented in multiple repositories.

### DIFF
--- a/rule_coverage/coverage.py
+++ b/rule_coverage/coverage.py
@@ -7,11 +7,10 @@ from git import Git
 from pathlib import Path
 
 repos=['sonar-abap','sonar-cpp','sonar-cobol','sonar-dotnet','sonar-css','sonar-flex','slang-enterprise','sonar-java','SonarJS','sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml','sonar-kotlin', 'sonar-secrets']
-#repos=['sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml']
 
 def load_json(file):
   with open(file) as json_file:
-    return json.load(json_file) 
+    return json.load(json_file)
 
 def get_rules_json(path,languages,version):
   print(f"Getting rules from {os.getcwd()} {path}")
@@ -30,7 +29,7 @@ def dump_rule(name,rule,languages,version):
   else:
     for language in languages:
       store_rule(name,rule,language,version)
-  
+
 def store_rule(name,rule,language,version):
   if language not in rules:
     print(f"create entry for {language}")
@@ -39,7 +38,7 @@ def store_rule(name,rule,language,version):
       name=name[:name.find('_')]
   if name not in rules[language]:
     rules[language][name]=version
-  
+
 def dump_rules(repo,version):
   for sp_file in Path('.').rglob('sonarpedia.json'):
     print(sp_file)
@@ -50,7 +49,7 @@ def dump_rules(repo,version):
     get_rules_json(path,languages,version)
     with open(f"../{rules_filename}", 'w') as outfile:
       json.dump(rules, outfile, indent=2)
-  
+
 def checkout(repo,version,batch_mode):
   token=os.getenv('GITHUB_TOKEN')
   if not token:
@@ -64,7 +63,7 @@ def checkout(repo,version,batch_mode):
   else:
     git_repo=Repo(repo)
   if batch_mode:
-    os.chdir(repo)  
+    os.chdir(repo)
     for tag in git_repo.tags:
       if not '-' in tag.name:
         print(f"{repo} {tag.name}")
@@ -80,10 +79,10 @@ def checkout(repo,version,batch_mode):
     g=Git(repo)
     g.checkout(version)
     os.chdir(repo)
-  
+
 def main():
   parser = argparse.ArgumentParser(description='rules coverage')
-  parser.add_argument('command', nargs='+', help='see code for help')  
+  parser.add_argument('command', nargs='+', help='see code for help')
   args = parser.parse_args()
 
   global rules
@@ -93,9 +92,7 @@ def main():
     rules=load_json(rules_filename)
   else:
     rules={}
-  
-  
-    
+
   if args.command[0] == "batchall":
     print(f"batch mode for {repos}")
     for repo in repos:
@@ -104,7 +101,7 @@ def main():
     repo=args.command[1]
     print(f"batch mode for {repo}")
     checkout(repo,None,True)
-  else: 
+  else:
     repo=args.command[0]
     version=args.command[1]
     print(f"checking {repo} version {version}")
--- a/rule_coverage/coverage.py
+++ b/rule_coverage/coverage.py
@@ -12,41 +12,46 @@ def load_json(file):
   with open(file) as json_file:
     return json.load(json_file)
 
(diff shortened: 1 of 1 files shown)


rule_coverage/coverage.py
```diff
@@ -59,26 +62,33 @@ def checkout(repo,version,batch_mode):
...
-    g=Git(repo)
-    g.checkout(version)
-    os.chdir(repo)
+    return Repo(repo)
+
+def scan_all_versions(repo):
+  git_repo = checkout_repo(repo)
+  os.chdir(repo)
+  g=Git(repo)
+  for tag in git_repo.tags:
```
andrei-epure-sonarsource, line 70: [education] why is scanning all versions necessary?

rule_coverage/coverage.py
```diff
@@ -59,26 +62,33 @@ def checkout(repo,version,batch_mode):
...
+        print("checkout failed, resetting and cleaning")
+        g.reset('--hard',tag)
+        g.clean('-xfd')
+      dump_rules(repo,tag.name)
+  os.chdir('..')
+
+def scan_version(repo,version):
+  git_repo = checkout_repo(repo)
+  g=Git(repo)
+  g.checkout(version)
```
andrei-epure-sonarsource, line 85: [education] why we don't use `try...except` for this checkout?

rule_coverage/coverage.py
```diff
@@ -59,26 +62,33 @@ def checkout(repo,version,batch_mode):
...
+  for tag in git_repo.tags:
+    if not '-' in tag.name:
+      print(f"{repo} {tag.name}")
+      try:
+        g.checkout(tag.name)
+      except Exception:
+        print("checkout failed, resetting and cleaning")
+        g.reset('--hard',tag)
+        g.clean('-xfd')
+      dump_rules(repo,tag.name)
```
andrei-epure-sonarsource, line 79: [education] shouldn't `dump_rules` be together with `checkout`?

rule_coverage/coverage.py
```diff
@@ -49,7 +46,7 @@ def dump_rules(repo,version):
     sonarpedia = load_json(sp_file)
     path=str(sonarpedia_path)+'/'+sonarpedia['rules-metadata-path'].replace('\\','/')
     languages=sonarpedia['languages']
-    get_rules_json(path,languages,version)
+    get_rules_json(path,languages, repo + ' ' + version)
```
andrei-epure-sonarsource, line 49: [education] why was this change needed?

andrei-epure-sonarsource (commented):
I have some questions

rule_coverage/coverage.py
arseniy-sonar: In order to find the first version that implements the rule.
Also to find out if the rule was implemented in the past even if it is removed now. The latter point is the current behavior, which is arguably suboptimal. In the future, I think it is best to display the version first implemented the rule and the version last implemented it.

rule_coverage/coverage.py
arseniy-sonar: With the addition of sonar-security, they can come from at least two repos: the sonar-language and sonar-security (at least for some languages).

rule_coverage/coverage.py
```diff
@@ -6,12 +6,11 @@
...
-repos=['sonar-abap','sonar-cpp','sonar-cobol','sonar-dotnet','sonar-css','sonar-flex','slang-enterprise','sonar-java','SonarJS','sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml','sonar-kotlin', 'sonar-secrets']
-#repos=['sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml']
+repos=['sonar-abap','sonar-cpp','sonar-cobol','sonar-dotnet','sonar-css','sonar-flex','slang-enterprise','sonar-java','SonarJS','sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml','sonar-kotlin', 'sonar-secrets', 'sonar-security']
 
 def load_json(file):
   with open(file) as json_file:
-    return json.load(json_file) 
+    return json.load(json_file)
 
 def get_rules_json(path,languages,version):
```
andrei-epure-sonarsource, line 15: maybe use `repoAndVersion` rather than `version`? and the same for `dump_rule`, `store_rule`.

Because it may be confusing where it's the version or repoAndVersion. See `dump_rules` version and `get_rules_json` version

And maybe add here (or somewhere better) a comment to mention that some languages have rules in multiple repositories (sonar-security + sonar-lang)?

rule_coverage/coverage.py
```diff
@@ -6,12 +6,11 @@
...
   print(f"Getting rules from {os.getcwd()} {path}")
```
andrei-epure-sonarsource, line 16: maybe log also `version` ?

rule_coverage/coverage.py
andrei-epure-sonarsource: do we have a ticket for this?

rule_coverage/coverage.py
andrei-epure-sonarsource: (given that we wrap the checkout in a `try` in `scan_all_versions`)

andrei-epure-sonarsource (changes requested):
(minor comments)

rule_coverage/coverage.py
arseniy-sonar: Sounds good. done.

rule_coverage/coverage.py
arseniy-sonar: done.

rule_coverage/coverage.py
```diff
@@ -6,52 +6,56 @@
...
+repos=['sonar-abap','sonar-cpp','sonar-cobol','sonar-dotnet','sonar-css','sonar-flex','slang-enterprise','sonar-java','SonarJS','sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml','sonar-kotlin', 'sonar-secrets', 'sonar-security']
 
 def load_json(file):
   with open(file) as json_file:
-    return json.load(json_file) 
+    return json.load(json_file)
 
-def get_rules_json(path,languages,version):
-  print(f"Getting rules from {os.getcwd()} {path}")
+# repoAndVersion uniquely identifies the analyzer and version implemeting
```
andrei-epure-sonarsource, line 15:
```suggestion
# repoAndVersion uniquely identifies the analyzer and version implementing
```

rule_coverage/coverage.py
arseniy-sonar: I created one: [RULEAPI-691](https://jira.sonarsource.com/browse/RULEAPI-691)

rule_coverage/coverage.py
arseniy-sonar: Yep, I've put them together now.

PR #398, opened 2021-09-24
RULEAPI-603 Fail ci metadata check if rule with replacement has 'ready' status

vilchik-elena:


rspec-tools/rspec_tools/validation/metadata.py
```diff
@@ -16,12 +16,25 @@ def get_json_schema():
...
     validate(instance=rule_language.metadata, schema=schema)
   except ValidationError as e:
     path = f'in {e.path[-1]}' if e.path else ''
     raise RuleValidationError(f'Rule {rule_language.id} has invalid metadata {path}: {e.message}') from e
 
+def validate_status(rule_language: LanguageSpecificRule):
+  status = rule_language.metadata.get('status')
+  if has_replacement_rules(rule_language) and status in ["beta", "ready"]:
+    raise RuleValidationError(f'Rule {rule_language.id} has invalid metadata:'
+      f' status can\'t be "{status}" for a rule with the replacement rule(s)')
```
andrei-epure-sonarsource, line 34:
```suggestion
      f' status can\'t be "{status}" for a rule with replacement rule(s).')
```

(either remove `the`, or add the replacement rules in the error message `with the remplacement rule(s): SXXX, SYYY`)

rspec-tools/tests/validation/test_metadata_validation.py
```diff
@@ -43,3 +43,12 @@ def test_adding_properties_fails_validation(rule_language: LanguageSpecificRule)
     with patch.object(LanguageSpecificRule, 'metadata', new_callable=PropertyMock) as mock:
       mock.return_value = metadata
       validate_metadata(rule_language)
+
+
+def test_ready_rule_with_replacement_fails_validation(rule_language: LanguageSpecificRule):
```
andrei-epure-sonarsource, line 48: could you add two more tests:
- when the rule status is not `beta` or `ready` and has a replacement rule? - should not fail
- when the rule has 2 replacement rules (to make sure it fails in this case, too)

andrei-epure-sonarsource (changes requested):
I left a few comments

rspec-tools/tests/validation/test_metadata_validation.py
andrei-epure-sonarsource: now you changed the test with only one replacement to test with two replacements - either way you're not testing one of the cases

I still think it would be good to test both cases

andrei-epure-sonarsource (approved):
LGTM

please consider my comment

rspec-tools/tests/validation/test_metadata_validation.py
vilchik-elena: Indeed, I did it consciously, as I don't see much value in this test and I don't want to have too many tests

PR #401, opened 2021-09-27
Modify rule S6021: Expand to cover unordered containers 

tomasz-kaminski-sonarsource:


rules/S6021/cfamily/rule.adoc
```diff
@@ -1,4 +1,5 @@
-Heterogeneous containers were introduced in {cpp}14 to increase performances when working with ``++std::map++``, ``++std::set++``, ``++std::multimap++`` and ``++std::multiset++``. Before their introduction, when working with a key in such a container, it was required to use an object of the exact key type. This could lead to costly object creations when we would like to work with objects of compatible but different types:
+Heterogeneous containers were introduced in {cpp}14 to increase performances when working with ``++std::map++``, ``++std::set++``, ``++std::multimap++``, ``++std::multiset++``, and since {cpp}20, their ``++unordered_++` counterparts.
```
marco-antognini-sonarsource, line 1:
```suggestion
Heterogeneous containers were introduced in {cpp}14 to increase performance when working with ``++std::map++``, ``++std::set++``, ``++std::multimap++``, ``++std::multiset++``, and since {cpp}20, their ``++unordered_++`` counterparts.
```

rules/S6021/cfamily/rule.adoc
```diff
@@ -24,8 +25,6 @@ private:
   std::string myData;
```
marco-antognini-sonarsource, line 25: There's a typo two lines up but I cannot create a suggestion for it...
```
// Other functions to make the class behaves correctly
```
should be
```
// Other functions to make the class behave correctly
```
(no *s* at the end of *behave*)

rules/S6021/cfamily/rule.adoc
```diff
@@ -35,7 +34,21 @@ void f() {
...
-This rule raises an issue when a transparent lookup function of a heterogeneous container is used with a type that cannot be directly compared with the container key type. Only standard associative containers with expensive to create key and straightforward comparison functions are considered.
+This problem appears also for the case of an unordered container for which heterogeneous lookup is enabled (available since {cpp}20), for which equality is used to compare elements in the same bucket (having same or close hashes):
+----
+void g() {
+  std::unordered_map<std::string, int, StringHash, std::equal_to<>> m;
+  MyString str{"abc"};
+  m.find(str);
+}
+----
+Again in this case ``++str++`` will not be converted to a ``++std::string++`` when calling the function ``++find++``, and for each element of the bucket that corresponds to the hash of ``++str++`` will be compared using homogeneous ``++operator==++``,
```
marco-antognini-sonarsource, line 45:
```suggestion
In this case ``++str++`` will not be converted to a ``++std::string++`` when calling the function ``++find++``, and each element of the bucket that corresponds to the hash of ``++str++`` will be compared using homogeneous ``++operator==++``,
```

rules/S6021/cfamily/rule.adoc
```diff
@@ -35,7 +34,21 @@ void f() {
...
+and for each such comparison, a conversion will now happen. The number of elements varies depending on the hash distribution from ``++O(1)++` (in average case), to  ``++O(N)++`` in the worst case.
```
marco-antognini-sonarsource, line 46:
```suggestion
and for each such comparison, a conversion will now happen. The number of compared elements varies depending on the hash distribution from ``++O(1)++` (on average) to ``++O(N)++`` (in the worst case).
```

rules/S6021/cfamily/rule.adoc
```diff
@@ -35,7 +34,21 @@ void f() {
...
+As consequence, the performance of slow runs (where multiple hash collisions happen due the input data), is made even worse.
```
marco-antognini-sonarsource, line 47:
```suggestion
As consequence, the performance of slow runs (when multiple hash collisions happen due to the data distribution) is made even worse.
```

rules/S6021/cfamily/rule.adoc
```diff
@@ -59,16 +78,30 @@ void f() {
...
+  std::unordered_map<std::string, int> m;
+  MyString str{"abc"}; // See previous definition of MyString
+  m.find(str); // Compliant, one conversion at the start
+}
+
 ----
 Option 2: Provide heterogeneous comparisons
 
 ----
+bool operator==(MyString const &s1, std::string const &s2) {/*...*/} // automatically reversed from C++20
```
marco-antognini-sonarsource, line 92: What do you mean by "automatically reversed from C++20"?

NB: east-const is used on these lines. Maybe this should be updated too?

marco-antognini-sonarsource (changes requested):

rules/S6021/cfamily/rule.adoc
tomasz-kaminski-sonarsource: That you do need the `operator==(std::string const&, MyString const&)`, as the language will reverse order of arguments automatically

marco-antognini-sonarsource (approved):

PR #402, opened 2021-09-27
RULEAPI-694: Scan the master branch for each analyzer and add covered rules

arseniy-sonar:


rule_coverage/coverage.py
```diff
@@ -70,17 +70,19 @@ def checkout_repo(repo):
...
-  for tag in git_repo.tags:
-    if not '-' in tag.name:
-      print(f"{repo} {tag.name}")
-      scan_version(repo, tag.name)
+  versions = [tag.name for tag in git_repo.tags]
+  for version in versions:
+    if not '-' in version:
+      print(f"{repo} {version}")
+      scan_version(repo, version, version)
+  scan_version(repo, 'master', 'nightly')
```
andrei-epure-sonarsource, line 78: why is the version name `nightly`?

arseniy-sonar: I have an impression it is a common name for a build from the master branch (often performed every night). Plus this coverage script runs every night.
Do you think "master" is more intuitive for the version name?
Or another name?

andrei-epure-sonarsource: I think it's deceiving... I mean whether it's done nightly or not is a setting on each repository

Also, this script isn't aware when the last `master` build happened... I'd keep it simple and just name it `master` (IMO)

arseniy-sonar: Makes sense. I renamed it to "master"

rule_coverage/coverage.py
```diff
@@ -113,7 +115,7 @@ def main():
     version=args.command[1]
     print(f"checking {repo} version {version}")
     checkout_repo(repo)
-    scan_version(repo,version)
+    scan_version(repo, version, version)
```
andrei-epure-sonarsource, line 118:
```suggestion
    scan_version(repo, version)
```

andrei-epure-sonarsource (approved):
LGTM (one small refactoring artefact left)

PR #403, opened 2021-09-27
Improve stability of the checkout for analyzer versions(tags)

arseniy-sonar:
`g.checkout` used to often fail for some versions especially of `sonar-dotnet` due to fantom unstaged changes, which lead to incorrect version being recorded for coverage of certain rules, for example: https://github.com/SonarSource/rspec/commit/8736e561f9fadf43d53380b9bb37d9c6a01932ec

the direct `reset` called on a `Repo` object seems to work more reliably.


rule_coverage/coverage.py
```diff
@@ -76,15 +76,14 @@ def scan_all_versions(repo):
...
+  r=Repo(repo)
   os.chdir(repo)
   try:
-    g.checkout(version)
+    r.head.reference = r.commit(version)
+    r.head.reset(index=True, working_tree=True)
     dump_rules(repo, version)
   except Exception:
     print(f"{repo} {version} checkout failed, resetting and cleaning")
-    g.reset('--hard', version)
```
andrei-epure-sonarsource, line 86: on `Repo` this is not necessary anymore?

rule_coverage/coverage.py
```diff
@@ -76,15 +76,14 @@ def scan_all_versions(repo):
       scan_version(repo, tag.name)
 
 def scan_version(repo,version):
-  g=Git(repo)
+  r=Repo(repo)
```
andrei-epure-sonarsource, line 79: could you add a comment on why `Repo` is used here and not in `checkout_repo`? because it seems both functions are checking out repositories...

andrei-epure-sonarsource (changes requested):

rule_coverage/coverage.py
arseniy-sonar: yes, as I understand, `r.head.reset(index=True, working_tree=True)` is equivalent, and it is done on each call to `scan_version` unconditionally

rule_coverage/coverage.py
arseniy-sonar: Good remark. In fact, `Repo` is used here because `checkout_repo` is expected to be called before you run `scan_version`. But that makes for a poor contract, I replaced it with `checkout_repo` (which doesn't add much overhead in case it was already called before).

andrei-epure-sonarsource (approved):
LGTM

PR #404, opened 2021-09-27
Update rule S5322: Add XML and update description

hendrik-buchwald-sonarsource:
This is very similar to the last RSPEC changes that you already reviewed in PR #388. Though, instead of creating a new rule it now adds it to an existing one. I have updated the description of the existing rule since I think my description was a bit more detailed.


rules/S5322/metadata.json
```diff
@@ -25,7 +25,9 @@
   "scope": "Main",
   "securityStandards": {
     "CWE": [
-      925
+      265,
```
pierre-loup-tristant-sonarsource, line 28: Minor: why adding CWE-265 which is very generic when we already have CWE-925+CWE-926 ?

pierre-loup-tristant-sonarsource (approved):
LGTM.
I have one minor question regarding one of the CWE you chose.

pierre-loup-tristant-sonarsource:
@hendrik-buchwald-sonarsource I just noticed that there is missing links between this PR and the implementation ticket

rules/S5322/metadata.json
hendrik-buchwald-sonarsource: True, not necessary. I have removed it.

hendrik-buchwald-sonarsource:
@pierre-loup-tristant-sonarsource I have added links.

PR #407, opened 2021-09-28
RULEAPI-691: Rule coverage: Track the last version of an analyzer implementing a rule along with the first version

arseniy-sonar:
The big number of changes in the covered_rules.json are due to the incidental correction for the version traversal order: it used to be arbitrary (or lexicographical), now it is chronological, so the implementation versions for many rules are corrected.


rule_coverage/coverage.py
```diff
@@ -8,6 +8,12 @@
 
 repos=['sonar-abap','sonar-cpp','sonar-cobol','sonar-dotnet','sonar-css','sonar-flex','slang-enterprise','sonar-java','SonarJS','sonar-php','sonar-pli','sonar-plsql','sonar-python','sonar-rpg','sonar-swift','sonar-tsql','sonar-vb','sonar-html','sonar-xml','sonar-kotlin', 'sonar-secrets', 'sonar-security']
 
+canonical_names={
```
andrei-epure-sonarsource, line 11: [education] what is this for?

arseniy-sonar: This coalesces the language names of the same language used in different versions or different analyzers.
Without canonicalizing the language names, they are considered different languages, which leads to apparent end-of-support for all rules when an analyzer switches from one language name to another (as is the case for "Web" -> "HTML"), or confusion if the language is implemented by multiple analyzers using different names (which is the case for "JAVASCRIPT" in SonarJS vs. "JS" in sonar-security).

rule_coverage/coverage.py
```diff
@@ -26,23 +32,39 @@ def get_rules_json(path, languages, repoAndVersion):
...
       name=name[:name.find('_')]
   if name not in rules[language]:
-    rules[language][name] = repoAndVersion
+    rules[language][name] = {'since': repoAndVersion, 'until': repoAndVersion}
+  elif type(rules[language][name]) == dict:
+    rules[language][name]['until'] = repoAndVersion
+  else:
+    rules[language][name] = {'since': rules[language][name], 'until': repoAndVersion}
+
+def simplify_spec_for_rules_that_are_still_supported():
```
andrei-epure-sonarsource, line 62: please add a comment mentioning that this implies that `master` will be the last in the array, so the rules needs to be sorted for this function to work.

andrei-epure-sonarsource (changes requested):

rule_coverage/coverage.py
```diff
@@ -26,23 +32,39 @@ def get_rules_json(path, languages, repoAndVersion):
     else:
         continue
 
+def canonicalize(language):
+  if language in canonical_names:
+    return canonical_names[language]
+  return language
 
 def dump_rule(name, rule, languages, repoAndVersion):
   if "compatibleLanguages" in rule:
```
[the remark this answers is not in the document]
arseniy-sonar: Some analyzers implement rules for multiple languages: C-family and JS/TS. However, not all rules implemented are compatible with all the languages the analyzer supports. This field specifies what languages out of the set supported by the analyzer are compatible.

rule_coverage/coverage.py
```diff
@@ -26,23 +32,39 @@ def get_rules_json(path, languages, repoAndVersion):
...
     for language in rule['compatibleLanguages']:
       store_rule(name, rule, language, repoAndVersion)
```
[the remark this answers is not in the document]
arseniy-sonar: Oups. I missed that one. Thanks!
Better to call it inside `store_rule` to avoid such omissions.

rule_coverage/coverage.py
arseniy-sonar: Done.

rule_coverage/coverage.py
```diff
@@ -26,23 +32,39 @@ def get_rules_json(path, languages, repoAndVersion):
...
   else:
     for language in languages:
-      store_rule(name, rule, language, repoAndVersion)
+      store_rule(name, rule, canonicalize(language), repoAndVersion)
 
 def store_rule(name, rule, language, repoAndVersion):
```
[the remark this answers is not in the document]
arseniy-sonar: I agree, the code is confusing at best. I reengineered the whole script.

rule_coverage/coverage.py
```diff
@@ -26,23 +32,39 @@ def get_rules_json(path, languages, repoAndVersion):
...
+  global rules
   if language not in rules:
     print(f"create entry for {language}")
     rules[language] = {}
   if '_' in name:
       name=name[:name.find('_')]
   if name not in rules[language]:
-    rules[language][name] = repoAndVersion
+    rules[language][name] = {'since': repoAndVersion, 'until': repoAndVersion}
```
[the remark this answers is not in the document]
arseniy-sonar: Is it simpler now, after the refactoring?

rule_coverage/coverage.py
```diff
@@ -6,116 +6,174 @@
...
+  implemented_rules = {}
+  for lang in languages:
+    implemented_rules[lang] = []
   for filename in os.listdir(path):
     if filename.endswith(".json") and not filename.startswith("Sonar_way"):
-        rule=load_json(os.path.join(path, filename))
-        dump_rule(filename[:-5], rule, languages, repoAndVersion)
+        rule = load_json(os.path.join(path, filename))
+        ruleId = get_rule_id(filename)
+        for language in rule_languages(rule, languages):
```
andrei-epure-sonarsource, line 47: This logic is cumbersome.

I suggest to rename `rule_languages` to `swap_with_compatible_if_present` and make it a different variable to give the result enough sense

```
ruleLanguages = swap_with_compatible_if_present(rule, languages)
for ruleLanguage in ruleLanguages :
  if ruleLanguage not in implemented_rules:
    implemented_rules[ruleLanguage ] = []
  implemented_rules[ruleLanguage ].append(ruleId)

```

_(I apologize for the long review, but I think clean-as-you-code is the right thing to do with this file)_

rule_coverage/coverage.py
```diff
@@ -6,116 +6,174 @@
...
+  Some analyzers, like SonarJS and sonar-cpp handle multiple languages
+  by the same rule implementation. They add a special field "compatibleLanguages"
+  to the rule metadata to specify which languges each implementation is applicable to.
+  '''
+  if "compatibleLanguages" in rule:
+    return rule["compatibleLanguages"]
+  else:
+    return languages
+
+def get_implemented_rules(path, languages):
```
andrei-epure-sonarsource, line 39:
```suggestion
def get_implemented_rules(path, languagesFromSonarPedia):
```

andrei-epure-sonarsource: (clean as you code)

And please add a function comment stating that:
- normally the rules should apply for the languages in SonarPedia
- there are some repositories where some rules are implemented for a subset of the languages in SonarPedia, which is contained in  `compatibleLanguages`

It took me a while to understand this logic and a comment would have saved a lot of time.

rule_coverage/coverage.py
andrei-epure-sonarsource: Also, I am highly confused - what is the case where the rule language might not be in SonarPedia, but be present in compatibleRules?

rule_coverage/coverage.py
```diff
@@ -6,116 +6,174 @@
...
+      self.rules[language] = {}
+
+    if version == 'master':
+      self._rule_implemented_for_last_version(ruleId, language, repo_and_version)
+    else:
+      self._rule_implemented_for_intermediate_version(ruleId, language, repo_and_version)
+
+  # analyzer+version uniquely identifies the analyzer and version implementing
+  # the rule for the given languages.
+  # Rule implementations for some langauges are spread across multiple repositories
```
andrei-epure-sonarsource, line 100:
```suggestion
  # Rule implementations for some languages are spread across multiple repositories
```

rule_coverage/coverage.py
```diff
@@ -6,116 +6,174 @@
...
+  # for example sonar-java and sonar-security for Java.
+  # We use analyzer+version to avoid confusion between versions of different analyzers.
+  def add_analyzer_version(self, analyzer, version, implemented_rules):
+    for language in implemented_rules:
```
andrei-epure-sonarsource, line 104: this is a bit confusing for a non-python developer (because in Java/C# you'd have `for language in implemented_rules.keys` or smth similar), IMO I have two possible suggestions:

- rename `implemented_rules` to `implemented_rules_per_language` would make it clearer it's a dictionary
OR
- add a comment saying that the parameter is a dictionary where `language` is the key and `list of rules` is the value.

frontend/src/RulePage.tsx
```diff
@@ -178,10 +178,17 @@ export function RulePage(props: any) {
...
-      )
+    const coverageMapper = (key: any, range: any) => {
+      if (typeof range === "string") {
+        return (
+          <li >{key}: {range}</li>
+        );
+      } else {
+        return (
+          <li><span style={{textDecorationLine: 'line-through'}}>{key}: {range['since']}</span>
+              <span>(until {range['until']}) </span></li>
```
andrei-epure-sonarsource, line 189: the `until` is a bit small here... I still see `sonar-abap 3.7` and my brain intuitively ignores what's with strikethrough because... well, it's cut text so it's not necessary.

So my brain sees "Covered Since until sonar-abap 7.3..."

Is it possible to make this w/o strikethrough and something like

"Not covered in ABAP anymore. Was covered from version X to version Y."

image

```suggestion
              <span>(was covered until {range['until']}, not covered anymore) </span></li>
```

andrei-epure-sonarsource (changes requested):
Thank you for doing the refactoring of the coverage.py script - it was very hard to understand and it's a true blessing to see object oriented programming! 💯 🥇 🥂 👍 🚀 

I still have some comments, however I'm pretty sure they're the last (🤞 ).

rule_coverage/coverage.py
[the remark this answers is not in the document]
andrei-epure-sonarsource: yes, it is much, much better

frontend/src/RulePage.tsx
arseniy-sonar: done.

rule_coverage/coverage.py
arseniy-sonar: I have this comment on the `rule_languages` (`compatible_languages` now). Do you think it is enough (after renaming the function to `compatible_languages` and the argument to `languagesFromSonarPedia`)?

rule_coverage/coverage.py
arseniy-sonar: How about just `compatible_languages`?
I don't think there can is a rule with `compatibleLanguages` containing a language that is not listed in sonarpedia. I think at the moment it is a subset.

I am not sure whether we should make sure to always use a subset of the languages in sonarpedia, or if a developer specified a language not listed in sonarpedia go ahead and add it the rule as implemented for that language. This is what the coverage script does right now.

rule_coverage/coverage.py
arseniy-sonar: renamed.

rule_coverage/coverage.py
andrei-epure-sonarsource: I think it's enough

rule_coverage/coverage.py
```diff
@@ -6,116 +6,175 @@
...
+
+  # analyzer+version uniquely identifies the analyzer and version implementing
+  # the rule for the given languages.
+  # Rule implementations for some languages are spread across multiple repositories
+  # for example sonar-java and sonar-security for Java.
+  # We use analyzer+version to avoid confusion between versions of different analyzers.
+  def add_analyzer_version(self, analyzer, version, implemented_rules_per_language):
+    for language in implemented_rules_per_language:
+      for ruleId in implemented_rules[language]:
+        self.rule_implemented(ruleId, language, analyzer, version)
```
andrei-epure-sonarsource, line 107:
```suggestion
      for rule_id in implemented_rules_per_language[language]:
        self.rule_implemented(rule_id , language, analyzer, version)
```

I wonder why tests didn't fail... this isn't covered?

arseniy-sonar: There are no automated tests for this script (I am adding them as part of RULEAPI-573), and I admit I haven't try to run the script after this last change.

rule_coverage/coverage.py
```diff
@@ -6,116 +6,175 @@
...
 
-# repoAndVersion uniquely identifies the analyzer and version implementing
-# the rule for the given languages.
-# Rule implementations for some langauges are spread across multiple repositories
-# for example sonar-java and sonar-security for Java.
-# We use repoAndVersion to avoid confusion between version of different analyzers.
-def get_rules_json(path, languages, repoAndVersion):
-  print(f"Getting rules from {os.getcwd()} {path} for {repoAndVersion}")
+def get_rule_id(filename):
+  ruleId = filename[:-5]
```
andrei-epure-sonarsource, line 22: nitpick: sonarlint complains about python naming convention:
- ruleId -> rule_id
- languagesFromSonarPedia -> languages_from_sonarpedia

andrei-epure-sonarsource (approved):
LGTM! great to see this!

rule_coverage/coverage.py
arseniy-sonar: Done.

PR #408, opened 2021-09-28
RULEAPI-696: adapt the rspec/readme.adoc narrative to not mention "coveredLanguages" and to menation `-branch` in `rule-api`

arseniy-sonar:
The big number of changes in the covered_rules.json are due to the incidental correction for the version traversal order: it used to be arbitrary (or lexicographical), now it is chronological, so the implementation versions for many rules are corrected.


csaba-sagi-sonarsource (approved):
LGTM! If the coveredLanguages keeps coming back to some metadata.json files then maybe it would be nice to add a validation for it that it does not exist. However I think it will be the case only in the upcoming few days until everybody realizes that it is not needed anymore.

PR #414, opened 2021-09-29
Rule S5840: Regex patterns and their sub-patterns should not always fail

johann-beleites-sonarsource:


vilchik-elena:
ping @johann-beleites-sonarsource

PR #384, opened 2021-09-21
Modify rule S6045: Cover heterogenous lookup for unordered containers in C++20

tomasz-kaminski-sonarsource:
…

PR #417, opened 2021-09-29
Modify rule S6195: Fix grammar and typo in code

marco-antognini-sonarsource:


tomasz-kaminski-sonarsource:
I would suggest mentoning the if-constexpr rule in see also, while we are editing the rspec.

tomasz-kaminski-sonarsource (changes requested):
I would suggest mentoning the if-constexpr rule in see also, while we are editing the rspec, simliary how it is done here:
https://github.com/SonarSource/rspec/pull/384/files

marco-antognini-sonarsource:
Let me know if the format is as expected.

rules/S6195/cfamily/rule.adoc
```diff
@@ -42,3 +42,6 @@ auto negate(std::signed_integral auto n) { return -n; }
 
 http://open-std.org/JTC1/SC22/WG21/docs/papers/2016/p0225r0.html[Why I want Concepts, and why I want them sooner rather than later]
 
+=== See Also
+
+* S6017
```
tomasz-kaminski-sonarsource, line 47: The foirmat should be:
`` S6021 for when it might be a bad idea to use transparent comparators.``
I.e. something like:
`` S6017 to see when enable_if could be replaced with if constexpr`.

tomasz-kaminski-sonarsource (changes requested):

tomasz-kaminski-sonarsource (approved):

rules/S6195/cfamily/rule.adoc
marco-antognini-sonarsource: Oh, okay. It seems we are not very consistent then -- `git grep` shows that most rules are referred without additional explanations.

PR #426, opened 2021-09-30
RULEAPI-688 Display covered languages on search page

vilchik-elena:


arseniy-sonar (changes requested):
Looks very good!
I noticed just one regression: "Covered Since" does not list coverage for all language on a generic rule page (when no language is chosen) anymore, it shows "Not covered" instead, even for rules that cover all the targeted languages.
Screenshot from 2021-09-30 10-51-56

arseniy-sonar:
@vilchik-elena, note that covered_rules.json is more complex now, and mentions "until" version for rules that were removed, e.g. https://github.com/SonarSource/rspec/blob/master/frontend/public/covered_rules.json#L81
I think these rules should be marked as "not covered", and perhaps be colored into a third color, red, for example.

PR #430, opened 2021-09-30
Use GITHUB_TOKEN for PR crawler to escape GitHub API throttling

arseniy-sonar:
With the substantial increase of the number of new-rule PR, I couldn't `predeploy` the github page to be able to review #426 because github was throttling my anonymous API requests.

### DIFF
--- a/frontend/src/deployment/pullRequestIndexing.ts
+++ b/frontend/src/deployment/pullRequestIndexing.ts
@@ -18,13 +18,15 @@ export interface PullRequest {
  * Call the "process" callback for each such rule.
  * @param tmpRepoDir Path to the temporary directory with a throw-away clone of the rspec repository.
  *                   if the repository does not exist, it will clone it first.
- * @param process A callback to process each rule. Takes two arguments:
+ * @param callback A callback to process each rule. Takes two arguments:
  *                srcDir - the path to the incomplete rule found in the corresponding PR.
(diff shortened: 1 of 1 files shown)


frontend/src/deployment/pullRequestIndexing.ts
```diff
@@ -38,7 +40,11 @@ export async function process_incomplete_rspecs(tmpRepoDir: string,
   }
   const repo = await (() => {
     if (!fs.existsSync(path.join(tmpRepoDir, '.git'))) {
-      return Git.Clone.clone('https://github.com/SonarSource/rspec/', tmpRepoDir);
+      if (process.env.GITHUB_TOKEN) {
+        return Git.Clone.clone('https://' + process.env.GITHUB_TOKEN + '@github.com/SonarSource/rspec/', tmpRepoDir);
```
vilchik-elena, line 44: I didn't find any doc about such clone

vilchik-elena (approved):
LGTM, but I didn't manage to verify manually, works ok on my side even before the change

frontend/src/deployment/pullRequestIndexing.ts
arseniy-sonar: Here is the doc: https://github.blog/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/#using-oauth-with-git

PR #436, opened 2021-09-30
Improve the failure experience of "add a language" GH Action

arseniy-sonar:
Fix the redirecting URL and raise an exception earlier to leave no unstaged changes when unwinding the stack during exception processing.


csaba-sagi-sonarsource (approved):
LGTM!

PR #444, opened 2021-10-01
Modify rule S1124: update docs to include default modifier

saberduck:
This is already implemented, just docs were not updated.

Feedback from https://twitter.com/heinzkabutz/status/1443903763098521601?s=20


quentin-jaquier-sonarsource (approved):
LGTM

PR #447, opened 2021-10-04
Fix coverage update: the phantom of the old rule_coverage directory

arseniy-sonar:
This is a rebase artefact: forgotten to use the `rspec-tools` directory where the coverage script was moved to.


andrei-epure-sonarsource (approved):
LGTM

PR #448, opened 2021-10-04
Fix the wrongly removed See section in S5753 C#

arseniy-sonar:


arseniy-sonar:
closed in favor of the complete fix in #449

PR #478, opened 2021-10-12
Modify rule S6287: Remove tainted key from samples

hendrik-buchwald-sonarsource:
The specification of the rule changed after the RSPEC was created and I forgot to update the code samples in RSPEC.

For now an issue is only raised if the header is explicitly "Set-Cookie" but not if it is tainted. In the future this should change but for now we will not detect an issue if the key is tainted. Thus, it does not make sense to show it in the code samples.


gyula-sallai-sonarsource:
Looks good to me, thanks!

gyula-sallai-sonarsource (approved):

PR #481, opened 2021-10-13
Create rule S4423[terraform]: Add AWS Elasticsearch Domain

hendrik-buchwald-sonarsource:
Implementation ticket: https://jira.sonarsource.com/browse/SONARIAC-201


rules/S4423/terraform/highlighting.adoc
```diff
@@ -0,0 +1,7 @@
+=== Highlighting
+
+For `aws_elasticsearch_domain`:
+
+* Highlight `tls_security_policy` if it is specified but has the wrong value
+* Highlight `domain_endpoint_options` if it is specified but does not contain `tls_security_policy`
+* Highlight resource if `domain_endpoint_options` is not specified at all
```
loris-s-sonarsource, line 7: I checked, and by default, traffic within the VPC is unencrypted, so ✔️ 
By default, the policy is  Policy-Min-TLS-1-0-2019-07, so ✔️

loris-s-sonarsource: Should we also add this one? :
```suggestion
* Highlight resource if `domain_endpoint_options` is not specified at all
* Highlight resource if `enforce_https ` is set to false
```

loris-s-sonarsource (approved):
Only one question about whether or not we should add another highlight. Else, LGTM.

PR #482, opened 2021-10-13
Create rule S4423[cloudformation]: Add AWS Elasticsearch Domain

hendrik-buchwald-sonarsource:
Implementation ticket: https://jira.sonarsource.com/browse/SONARIAC-202


rules/S4423/cloudformation/highlighting.adoc
```diff
@@ -0,0 +1,7 @@
+=== Highlighting
+
+For `AWS::Elasticsearch::Domain` and `AWS::OpenSearchService::Domain`:
+
+* Highlight `TLSSecurityPolicy` if it is specified but has the wrong value
+* Highlight `DomainEndpointOptions` if it is specified but does not contain `TLSSecurityPolicy`
+* Highlight resource if `DomainEndpointOptions` is not specified at all
```
loris-s-sonarsource, line 7: Same as the terraform one, maybe we should add 
```suggestion
* Highlight resource if `DomainEndpointOptions` is not specified at all
* Highlight resource if `EnforceHTTPS` is set to false
```

hendrik-buchwald-sonarsource: Also a copy from my response :)

It is an issue if `enforce_https` is not enabled but for a different rule. I think it will likely trigger an issue for rule S5332.

I do set it in the code sample here. It is optional, I just think the code sample looks slightly nicer with `domain_name` and `enforce_https`. I choose them because if people use the `aws_elasticsearch_domain` resource they will likely have to define this 2 parameters.

We could also go for a minimal version:

```
resource "aws_elasticsearch_domain" "example" {
  domain_endpoint_options {
    tls_security_policy = "Policy-Min-TLS-1-0-2019-07" # Noncompliant
  }
}
```

loris-s-sonarsource: Ok, no problem :) LGTM then !

PR #557, opened 2021-10-29
Modify rule S5857: Add Python as covered language

nils-werner-sonarsource:


rules/S5857/python/rule.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+r'<.+?>'
+r'".*?"'
+----
+
+
+== Compliant Solution
+
+----
+r'<[^>]++>'
```
guillaume-dequenne-sonarsource, line 16: Python's `re` module doesn't have possessive quantifiers. I believe we should adapt the rule a bit to not recommend using them. The `description.adoc` file also mention possessive quantifiers, which we shouldn't do for Python either.

guillaume-dequenne-sonarsource (changes requested):
I believe we should adapt this rule a bit to avoid recommending possessive quantifiers, which don't exist in Python's `re` module.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #561, opened 2021-11-01
Fix "pipenv install" invocation to avoid update of Pipfile.lock

pierre-loup-tristant-sonarsource:
When python scripts executing on Cirrus Ci run`pipenv install -e .`, it triggers an update of `Pipfile.lock` which then fails to resolve dependency.
As `pipenv install` doesn't trigger an update of `Pipfile.lock` I removed the `-e` argument from install commands.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,7 +23,7 @@ tooling_tests_task:
     PYTHONPATH: .
   install_dependencies_script:
     - cd rspec-tools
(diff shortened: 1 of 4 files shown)


arseniy-sonar (approved):
Thanks!

arseniy-sonar:
Use `pipenv install` all thourghout

PR #570, opened 2021-11-05
Modify rule S1121: Describe unexpected creation of global variables

yassin-kammoun-sonarsource:
This change is motivated by a feedback from the community (see [thread](https://community.sonarsource.com/t/typescript-s1121-should-ignore-chained-assignments/27392/3?u=yassin_kammoun)).

### DIFF
--- a/rules/S1121/javascript/rule.adoc
+++ b/rules/S1121/javascript/rule.adoc
@@ -1,5 +1,7 @@
 include::../description.adoc[]
 
+Moreover, using chained assignment in declarations is also dangerous because one may accidentally create global variables, e.g. in ``++let x = y = 1;++``, if ``++y++`` is not declared, it will be hoisted as global.
+
 == Noncompliant Code Example
 
 ----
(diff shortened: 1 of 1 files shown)


rules/S1121/javascript/rule.adoc
```diff
@@ -1,5 +1,7 @@
 include::../description.adoc[]
 
+Moreover, using chained assignment in declarations is also dangerous because one may accidentally create global variables, e.g. in ``++let x = y = 1;++``, if ``++y++`` is not declared, it will be hoisted as global.
```
vilchik-elena, line 3: Should we remove this case from `Exceptions` section?

vilchik-elena (changes requested):

rules/S1121/javascript/rule.adoc
yassin-kammoun-sonarsource: Good point!

vilchik-elena (approved):
@yassin-kammoun-sonarsource before merging could you also double-check that all other exceptions are still valid?

PR #572, opened 2021-11-05
Modify rule S4784 [VBNET]: Remove unnecessary replacementRules from json

csaba-sagi-sonarsource:


andrei-epure-sonarsource (approved):
LGTM

PR #575, opened 2021-11-08
Modify rule S3229: Complete the description with bases; add compliant solution

arseniy-sonar:


rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
-Class members are initialized in the order in which they are declared in the class, not the order in which they appear in the class initializer list. To avoid errors caused by order-dependent initialization, the order of members in the initialization list should match the order in which members are declared in a class.
+Class members are initialized in the order in which they are declared in the class,
+not the order in which they appear in the class initializer list.
+To avoid errors caused by order-dependent initialization,
+the order of members in the initialization list should match the order in which members are declared in a class.
 
+Base parts of the object are initialized before direct members, in the order they are listed in the inheritance list,
+and virtual bases are initialized first.
+
+The complete oprder, as described https://en.cppreference.com/w/cpp/language/constructor#Initialization_order[here]:
```
marco-antognini-sonarsource, line 9:
```suggestion
The initialisation order, as described https://en.cppreference.com/w/cpp/language/constructor#Initialization_order[here], is:
```

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
+. If the constructor is for the most-derived class, virtual bases are initialized in the order in which they appear in depth-first left-to-right traversal of the base class declarations (left-to-right refers to the appearance in base-specifier lists)
+. Then, direct bases are initialized in left-to-right order as they appear in this class's base-specifier list
+. Then, non-static data members are initialized in order of declaration in the class definition.
```
marco-antognini-sonarsource, line 12: Github preview doesn't display this as a list. It may be better to use `*` to create a list here.

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
-Class members are initialized in the order in which they are declared in the class, not the order in which they appear in the class initializer list. To avoid errors caused by order-dependent initialization, the order of members in the initialization list should match the order in which members are declared in a class.
+Class members are initialized in the order in which they are declared in the class,
+not the order in which they appear in the class initializer list.
+To avoid errors caused by order-dependent initialization,
+the order of members in the initialization list should match the order in which members are declared in a class.
 
+Base parts of the object are initialized before direct members, in the order they are listed in the inheritance list,
+and virtual bases are initialized first.
```
marco-antognini-sonarsource, line 7: Given the next paragraph, I'm not sure this one brings any value. It could be removed I think.

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
 
 == Noncompliant Code Example
 
 ----
 #include <iostream>
 
-class C {
+struct A {
+  A(int num) {
+    std::cout <<"A(num = " << num << <<")" <<std::endl;
```
marco-antognini-sonarsource, line 21:
```suggestion
    std::cout << "A(num = " << num << ")" << std::endl;
```

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
+  }
+};
+
+struct B {
+  int b;
+};
+
+class C :public A, B {
```
marco-antognini-sonarsource, line 29:
```suggestion
class C : public A, B {
```

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
 ----
 
+== Compliant Solution
+
+----
+#include <iostream>
+
+struct A {
+  A(int num) {
+    std::cout <<"A(num = " << num << <<")" <<std::endl;
```
marco-antognini-sonarsource, line 50:
```suggestion
    std::cout << "A(num = " << num << ")" << std::endl;
```

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
+  }
+};
+
+struct B {
+  int b;
+};
+
+class C :public A, B {
```
marco-antognini-sonarsource, line 58:
```suggestion
class C : public A, B {
```

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
+public:
+  int x;
+  int y;
+
+  C(int i) : A{i}, B{i}, x(i + 1), y(i) { }  // Noncompliant
```
marco-antognini-sonarsource, line 63:
```suggestion
  C(int i) : A{i}, B{i}, x(i + 1), y(i) { }
```

rules/S3229/cfamily/rule.adoc
```diff
@@ -1,25 +1,73 @@
...
   int y;
 
-  C(int i) : y(i), x(y + 1) { }  // Noncompliant
+  C(int i) : B{i}, A{b}, y(i), x(y + 1) { }  // Noncompliant
 };
 
 int main() {
-  C c(1);
+  C c(1); // prints A(num = 0)
   std::cout << c.x << " " << c.y << std::endl;  // prints 1 1
```
marco-antognini-sonarsource, line 39: I actually wonder, is it well defined and always prints those values? I have the feeling that this (accessing members before they are initialised) is actually undefined behaviour... https://godbolt.org/z/41Gcxe7Ee seems to "confirm" this, although it would be more convincing to have the wording from the spec instead of an example but that's not the point of this rule.

It may be better to rephrase the comments to say it's UB and may print unexpected values instead of "1, 2, 1".

rules/S3229/cfamily/rule.adoc
marco-antognini-sonarsource: or _initialization_ if American English is prefered, I don't know.

marco-antognini-sonarsource (changes requested):
Thanks for the update. Reviewing this prompted a question about the existing example around undefined behaviour. I have a few suggestions; most of them are trivial.

rules/S3229/cfamily/rule.adoc
arseniy-sonar: That's because of the missing empty line before the list. I use `.` for a list to make it ordered (`*` creates an unordered list).

rules/S3229/cfamily/rule.adoc
arseniy-sonar: Indeed. Removed.

rules/S3229/cfamily/rule.adoc
arseniy-sonar: Aggreed, rephrased.

rules/S3229/cfamily/rule.adoc
marco-antognini-sonarsource: Ah, thanks for the explanation! :-)

marco-antognini-sonarsource (approved):
Thanks for the update. LGTM.

PR #590, opened 2021-11-12
Modify rule S4581: Add more non compliant examples.

mary-georgiou-sonarsource:


rules/S4581/message.adoc
```diff
@@ -1,4 +1,5 @@
 === Message
 
-Please review this use of "new Guid()".
+Please review the way you are instantiating a new Guid.
```
costin-zaharia-sonarsource, line 3: It looks like the message was out of date. 

I think it should be `Use 'Guid.NewGuid()' or 'Guid.Empty' or use Guid instantiation with arguments.` since this in the latest version of the code that we want to merge to master.

costin-zaharia-sonarsource (changes requested):
The message seems to be different that the one we want to merge on master,

rules/S4581/message.adoc
```diff
@@ -1,5 +1,3 @@
 === Message
 
-Please review the way you are instantiating a new Guid.
-
-
+Use 'Guid.NewGuid()' or 'Guid.Empty' or use Guid instantiation with arguments.
```
costin-zaharia-sonarsource, line 3: Sorry, this is really my mistake. I forgot that we changed the message to `Use 'Guid.NewGuid()', 'Guid.Empty' or the constructor with arguments.`.

PR #605, opened 2021-11-18
Modify rule S1763[cfamily]: cover co_return and [[noreturn]]

marco-antognini-sonarsource:
This effectively takes the content of `../description.adoc` and adds mentions to `co_return` and `[[noreturn]]` functions (`abort`, `std::terminate`).

---

No implementation changes are required.

### DIFF
--- a/rules/S1763/cfamily/rule.adoc
+++ b/rules/S1763/cfamily/rule.adoc
@@ -1,4 +1,6 @@
-include::../description.adoc[]
+Some statements (`return`, `break`, `continue`, `goto`, `switch`, `co_return`) and `throw` expressions move control flow out of the current code block. Furthermore, some function do not return control flow (e.g. `abort()`, `std::terminate()`, functions with the ``++[[noreturn]]++`` attribute).
+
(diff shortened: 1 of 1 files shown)


rules/S1763/cfamily/rule.adoc
```diff
@@ -1,4 +1,6 @@
-include::../description.adoc[]
+Some statements (`return`, `break`, `continue`, `goto`, `switch`, `co_return`) and `throw` expressions move control flow out of the current code block. Furthermore, some function do not return control flow (e.g. `abort()`, `std::terminate()`, functions with the ``++[[noreturn]]++`` attribute).
```
arseniy-sonar, line 1:
```suggestion
Some statements (`return`, `break`, `continue`, `goto`, `co_return`) and `throw` expressions move control flow out of the current code block. Furthermore, some function do not return control flow (e.g. `abort()`, `std::terminate()`, functions with the ``++[[noreturn]]++`` attribute).
```
either `switch` should not be in the list, or it should include other control statements such as `if`, `for`, `while`.

arseniy-sonar (changes requested):

PR #615, opened 2021-11-23
Modify rule S5332: Add message for missing option

hendrik-buchwald-sonarsource:
Implementation ticket: https://jira.sonarsource.com/browse/SONARIAC-277

SONARIAC-277 contains some cases where an issue should be raised if a specific option is not set. The new message (line 6) should cover this.


pierre-loup-tristant-sonarsource (approved):
LGTM.

In the ticket it would be nice to have an RSPEC session that point to this PR.
It should also says the highlighting and message are defined in the RSPEC.

PR #616, opened 2021-11-24
Modify rule S1238[cfamily]: exception for coroutines (CPP-3317)

arseniy-sonar:
Implementation ticket: CPP-3317


marco-antognini-sonarsource (approved):
LGTM

PR #629, opened 2021-11-29
Modify rule S5843: Fix duplicate description parts

quentin-jaquier-sonarsource:
FYI @johann-beleites-sonarsource and @nils-werner-sonarsource:
Funny/confusing problem: you both modified the rule almost at the same time (#416 and #537), and git did not manage to catch that the description is now broken for PHP, Java and Python.


nils-werner-sonarsource (approved):
LGTM!

PR #635, opened 2021-12-01
Modify rule S2222[CSharp]: improve description and code example

csaba-sagi-sonarsource:


rules/S2222/compliant.adoc
```diff
@@ -7,12 +7,29 @@ class MyClass
 
   public void DoSomething() 
   {
+    Monitor.Enter(obj); // Noncompliant
```
andrei-epure-sonarsource, line 10: this is the compliant solution ... it should contain lock

rules/S2222/csharp/description.adoc
```diff
@@ -0,0 +1,5 @@
+include::../description.adoc[]
+
+The types tracked by the rule are: Monitor, Mutex, ReaderWriterLock, ReaderWriterLockSlim and SpinLock.
```
andrei-epure-sonarsource, line 3:
```suggestion
The types tracked by the rule are: Monitor, Mutex, ReaderWriterLock, ReaderWriterLockSlim and SpinLock from the `System.Threading` namespace. 
```

rules/S2222/compliant.adoc
andrei-epure-sonarsource: maybe give a simple example with `ReaderWriterLockSlim` , just to have it (both Noncompliant and Compliant) - because in this example the fix is not to use Monitor at all

andrei-epure-sonarsource (changes requested):

rules/S2222/noncompliant.adoc
```diff
@@ -16,3 +16,22 @@ class MyClass
   }
 }
 ----
+
+or
```
andrei-epure-sonarsource, line 20: IMO you can avoid the `or` part and add the two samples in the same `MyClass` because they are using different APIs for different purposes (they're not alternatives)
```
----
class MyClass 
{
  private object obj = new object();
  public void DoSomethingWithMonitor() 
  {
    Monitor.Enter(obj); // Noncompliant
    if (IsInitialized())
    {
      // ...
      Monitor.Exit(obj);
    }
  }
  
  private ReaderWriterLockSlim lockObj = new ReaderWriterLockSlim();
  public void DoSomethingWithReaderWriteLock() 
  {
    lockObj.EnterReadLock(); // Noncompliant
    // do something without releasing in the same method
  }
}
----
```

andrei-epure-sonarsource (changes requested):
Also, now that you are on the rspec, could you also add the VB .NET version of the rule?

this should help https://converter.telerik.com/

We'll want to implement both if time allows (as this will validate the fact that the engine is language-independent). So better just to get done with the RSPEC at this point.

rules/S2222/noncompliant.adoc
andrei-epure-sonarsource: alternatively you can do the VB .NET in a different PR

andrei-epure-sonarsource: just to finish with this one

csaba-sagi-sonarsource: I am working on the VBNET part but It will be part of a new PR.

andrei-epure-sonarsource (approved):
LGTM

PR #640, opened 2021-12-06
Correct the meaning of a PR link for a rule (under specification)

arseniy-sonar:
Existence of a PR link means the rule specification is not merged to master yet,
and might still be undergoing specification.
It should not be hidden if the rule is already covered by an analyzer.

If the rule is still in the PR it means it's specification is not complete,
and might change before the PR is merged.

The original meaning of "Not implemented" stems from the deprecated policy of
merging the rspec to the master before merging the implementation of the rule
in an analyzer (which itself was caused by the technical limitation of rule-api,
since removed).

### DIFF
--- a/frontend/src/RulePage.tsx
+++ b/frontend/src/RulePage.tsx
@@ -48,7 +48,7 @@ const useStyles = makeStyles((theme) => ({
   tabScroller: {
     flexGrow: 0
   },
-  unimplemented: {
+  underspecified: {
     color: 'red'
   },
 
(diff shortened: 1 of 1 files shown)


arseniy-sonar:
Closed. As per RULEAPI-673, not taking into account that the rule is implemented for some language even when it is being specified for another language might create confusion on its own. The situation this PR is supposed to remediate should happen less often, and after implementing RULEAPI-716 - almost never.

PR #647, opened 2021-12-08
Modify rule S2222[csharp, vbnet]: Improve noncompliant example

csaba-sagi-sonarsource:


rules/S2222/csharp/description.adoc
```diff
@@ -1,3 +1,5 @@
 include::../description.adoc[]
 
-The types tracked by the rule are: Monitor, Mutex, ReaderWriterLock, ReaderWriterLockSlim and SpinLock from the `System.Threading` namespace.
\ No newline at end of file
+The types tracked by the rule are: Monitor, Mutex, ReaderWriterLock, ReaderWriterLockSlim and SpinLock from the `System.Threading` namespace.
+
+Note that the rule is currently in an experimental phase so it might have more false positives or false negatives than usual.
```
pavel-mikula-sonarsource, line 5:
```suggestion
Note that the rule is currently in an experimental phase.
```

rules/S2222/vbnet/description.adoc
```diff
@@ -1,3 +1,5 @@
 include::../description.adoc[]
 
-The types tracked by the rule are: Monitor, Mutex, ReaderWriterLock, ReaderWriterLockSlim and SpinLock from the `System.Threading` namespace.
\ No newline at end of file
+The types tracked by the rule are: Monitor, Mutex, ReaderWriterLock, ReaderWriterLockSlim and SpinLock from the `System.Threading` namespace.
+
+Note that the rule is currently in an experimental phase so it might have more false positives or false negatives than usual.
```
pavel-mikula-sonarsource, line 5: Same as C#
```suggestion
Note that the rule is currently in an experimental phase.
```

rules/S2222/vbnet/noncompliant.adoc
```diff
@@ -17,7 +17,10 @@ Class Example
 
     Public Sub DoSomethingWithReaderWriteLockSlim()
         lockObj.EnterReadLock() ' Noncompliant
-        ' do something without releasing in the same method
+        If IsInitialized() Then
```
pavel-mikula-sonarsource, line 20: We'll need to update the C# version as well by adding `noncompliant.adoc` and changing `rule.adoc` to use it.

rules/S2222/vbnet/noncompliant.adoc
```diff
@@ -17,7 +17,10 @@ Class Example
...
+            ' ..
+            lockObj.ExitReadLock()
+        End If
```
pavel-mikula-sonarsource, line 23: Noncompliant for VB doesn't make sense as it asks not to release at all

rules/S2222/vbnet/noncompliant.adoc
```diff
@@ -17,7 +17,10 @@ Class Example
 
     Public Sub DoSomethingWithReaderWriteLockSlim()
         lockObj.EnterReadLock() ' Noncompliant
-        ' do something without releasing in the same method
+        If IsInitialized() Then
+            ' ..
+            lockObj.ExitReadLock()
```
pavel-mikula-sonarsource, line 22: _Line 12 above:_ Square brackets around `Exit` are not needed. Same for noncompliant.

pavel-mikula-sonarsource (changes requested):

rules/S2222/vbnet/noncompliant.adoc
csaba-sagi-sonarsource: Why add a new file if the base example is given in C#?

rules/S2222/vbnet/noncompliant.adoc
csaba-sagi-sonarsource: What do you mean?

pavel-mikula-sonarsource: My bad, I opened a wrong file

rules/S2222/noncompliant.adoc
```diff
@@ -20,7 +20,11 @@ class MyClass
   public void DoSomethingWithReaderWriteLockSlim() 
   {
     lockObj.EnterReadLock(); // Noncompliant
-    // do something without releasing in the same method
+    if (IsInitialized())
```
pavel-mikula-sonarsource, line 23: This whole method should not be shared between Java and CSharp

rules/S2222/vbnet/noncompliant.adoc
pavel-mikula-sonarsource: Because the base example is shared with Java.

csaba-sagi-sonarsource: Yes and a looked at a couple of other rules randomly and all of them had a base example. I can remove the examples from the base but than it will not have any. WDYT?

pavel-mikula-sonarsource: We do share examples IF the code is common for Java & C# (and C, JS and other C-like languages).
We do NOT share examples if the code is not valid for at least one languages.
It is common to have Java (and C, JS and other C-like languages) as the common template. And custom one for C#.
It is also common to have C# template as shared for C# + VB.NET-only rules.

pavel-mikula-sonarsource (approved):
LGTM

PR #653, opened 2021-12-09
Modify rule S6353: Add PHP as covered language

karim-ouerghemmi-sonarsource:
- Implementation ticket: https://jira.sonarsource.com/browse/SONARPHP-1276


rules/S6353/php/rule.adoc
```diff
@@ -0,0 +1,42 @@
+include::../description.adoc[]
+
+== Noncompliant Code Example
+
+----
+preg_match("/[0-9]/", $s); // Noncompliant - same as "/\d/"
```
nils-werner-sonarsource, line 6: Do we need to wrap the expression with `preg_match` to make out point?

nils-werner-sonarsource (approved):
LGTM! See comment

quentin-jaquier-sonarsource (approved):
LGTM.

The rule is now implemented in the PHP analyzer, and we are in the process of specifying this rule for Java, it makes sense to merge this PR now.

PR #664, opened 2021-12-16
Modify rule S4275[dotnet]: Fix VB.NET examples

pavel-mikula-sonarsource:
Old examples did not compile, because field `x` collided with property `X`.

Also only one property is needed to demonstrate the issue, the other one is a nosise.


costin-zaharia-sonarsource (approved):
LGTM!

PR #672, opened 2021-12-22
S1768 Revert changes to the RSPEC that are not in sync with the code

loic-joly-sonarsource:
* One change was wrong: Enums can have negative values
* The other one is of little value, especially for a deprecated rule.


loic-joly-sonarsource:
@abbas-sabra-sonarsource : With that change, the only difference in the plugin when running RuleAPI compared to the previous version is that the rule is now deprecated (and this is a good difference that should be part of the plugin)

PR #673, opened 2021-12-22
S2665: CPP-1222 Align RSPEC with implementation

loic-joly-sonarsource:
CPP-1222


loic-joly-sonarsource:
@abbas-sabra-sonarsource : When using this version, the ruleApi will generate results semantically identical to what is currently in the plugin

PR #674, opened 2021-12-22
Modify Rule S1001: Revert changes introduced by CPP-1307

loic-joly-sonarsource:
The changes in the RSPEC were never implemented. They will be put in another PR.

Also rewrite the examples.


loic-joly-sonarsource:
@abbas-sabra-sonarsource : This description matched the current implementation, but is different from what is currently in the plugin. I have another pending PR to update the plugin.

PR #676, opened 2021-12-29
Modify rule S3442 [csharp]: Improve rule description

csaba-sagi-sonarsource:
Related to issue: https://github.com/SonarSource/sonar-dotnet/issues/5006


andrei-epure-sonarsource (approved):
LGTM

PR #686, opened 2022-01-10
Create rule S1155: VB.NET

pavel-mikula-sonarsource:


rules/S1155/vbnet/rule.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+Private Function IsEmpty(Strings As IEnumerable(Of String)) As Boolean
+    Return Strings.Count() = 0      ' Noncompliant
+End Function
+----
+
+== Compliant Solution
+
+----
+Private Function HasContent(Strings As IEnumerable(Of String)) As Boolean
```
csaba-sagi-sonarsource, line 25: Why isn't there a compliant solution for `HasContent2`?

csaba-sagi-sonarsource (changes requested):
Two minor comments left. Both are applicable for the C# part too.

rules/S1155/vbnet/rule.adoc
pavel-mikula-sonarsource: Because it's the same as `HasContent`. It was like this in the C# version of the rule as well. I can add both.

csaba-sagi-sonarsource: I don't have a strong opinion but it might be confusing for the users. If you agree please add it, otherwise ignore the comment.

csaba-sagi-sonarsource (commented):
LGTM!

csaba-sagi-sonarsource (approved):
OK, this time I'll approve too :)

PR #688, opened 2022-01-11
Modify Rule S5131: Message clarity

loris-s-sonarsource:
This RSpec message has been misunderstood by multiple Sonar* users claiming that "reflecting user-controlled data" was exactly their goal. This message aims to clarify our point when we detect XSS in code and advise users to make a fix.

### DIFF
--- a/rules/S5131/message.adoc
+++ b/rules/S5131/message.adoc
@@ -1,4 +1,3 @@
 === Message
 
-Change this code to not reflect user-controlled data.
-
+Sanitize this user-controlled data before reflecting it.
--- a/rules/S5131/message.adoc
+++ b/rules/S5131/message.adoc
@@ -1,3 +1,3 @@
 === Message
 
-Sanitize this user-controlled data before reflecting it.
+Change this code to not reflect unsanitized user-controlled data.


loris-s-sonarsource:
```
Change this code to not reflect unsanitized user-controlled data.
```

eric-therond-sonarsource:
More generally, I think it's a problem that is broader than just XSS, I just collected some current messages:
- Change this code to not construct SQL queries directly from user-controlled data
- Change this code to not perform redirects based on user-controlled data.
- Change this code to not construct the path from user-controlled data.
- Change this code to not log user-controlled data.
- Change this code to not construct the URL from user-controlled data.
- Change this code to not construct the regular expression from user-controlled data.
- Change this code to not construct the OS command from user-controlled data.
- Change this code to not construct this LDAP name or filter from user-controlled data.
- Change this code to not construct this XPath expression from user-controlled data.
- Change this code to not reflect user-controlled data.

In this list only the message about sql injection is fine to me because it uses the word _directly_.

I think what you are suggesting in your last comment is good to me as a quick fix, other possibilities could be:
- Change this code to not reflect untrused user-controlled data.
- Change this code to not directly reflect user-controlled data.

I let you decide, I don't have any preferences.

I think this problem is very close to the educational on going subject (to provide meaningful messages and not generic ones), and could be handled [in this circle](https://discuss.sonarsource.com/t/provide-more-focused-security-rules-descriptions-for-our-users/9581/7) I believe, I will also send to you an email related to that.

eric-therond-sonarsource (approved):
LGTM

loris-s-sonarsource:
> More generally, I think it's a problem that is broader than just XSS, I just collected some current messages:
> 
>     * Change this code to not construct SQL queries directly from user-controlled data
> 
>     * Change this code to not perform redirects based on user-controlled data.
> 
>     * Change this code to not construct the path from user-controlled data.
> 
>     * Change this code to not log user-controlled data.
> 
>     * Change this code to not construct the URL from user-controlled data.
> 
>     * Change this code to not construct the regular expression from user-controlled data.
> 
>     * Change this code to not construct the OS command from user-controlled data.
> 
>     * Change this code to not construct this LDAP name or filter from user-controlled data.
> 
>     * Change this code to not construct this XPath expression from user-controlled data.
> 
>     * Change this code to not reflect user-controlled data.
> 
> 
> In this list only the message about sql injection is fine to me because it uses the word _directly_.
> 
> I think what you are suggesting in your last comment is good to me as a quick fix, other possibilities could be:
> 
>     * Change this code to not reflect untrused user-controlled data.
> 
>     * Change this code to not directly reflect user-controlled data.
> 
> 
> I let you decide, I don't have any preferences.
> 
> I think this problem is very close to the educational on going subject (to provide meaningful messages and not generic ones), and could be handled [in this circle](https://discuss.sonarsource.com/t/provide-more-focused-security-rules-descriptions-for-our-users/9581/7) I believe, I will also send to you an email related to that.

Hmm... I like "untrusted"... I'm going to think about contexts where we have the most chances of being understood

loris-s-sonarsource:
> ```
> Change this code not to reflect unsanitized user-controlled data.
> ```

```
Sanitize this user-controlled data so that it contains only allowed characters.
```

```
Clean up this user-controlled data so that it contains only allowed characters.
```

rules/S5131/message.adoc
```diff
@@ -1,4 +1,3 @@
 === Message
 
-Change this code to not reflect user-controlled data.
-
+Sanitize this user-controlled data before reflecting it.
```
loris-s-sonarsource, line 3:
```suggestion
Change this code to not reflect unsanitized user-controlled data.
```

-> First baby step

loris-s-sonarsource: I choose not to go with "untrusted" because I can smell the people saying "I trust this data since it's coming from a service that I wrote myself"

PR #693, opened 2022-01-12
 RULEAPI-579: Enable project analysis in SonarQube Next

tomasz-kaminski-sonarsource:


.cirrus.yml
```diff
@@ -13,21 +20,33 @@ container_definition: &CONTAINER_DEFINITION
   namespace: default
   use_in_memory_disk: true
 
+setup_sonar_scanner: &SETUP_SONAR_SCANNER
+  setup_sonar_scanner_script:
+    - apt update -y && apt upgrade -y && apt update -y && apt install -y unzip       
+    - curl --create-dirs -sSLo $HOME/.sonar/sonar-scanner.zip https://binaries.sonarsource.com/Distribution/sonar-scanner-cli/sonar-scanner-cli-$SONAR_SCANNER_VERSION-linux.zip
+    - unzip -o $HOME/.sonar/sonar-scanner.zip -d $HOME/.sonar/
+    - export PATH=$SONAR_SCANNER_HOME/bin:$PATH
```
arseniy-sonar, line 28: are both this line and line 9 (PATH: ${SONAR_SCANNER_HOME}/bin:$PATH) necessary?

.cirrus.yml
```diff
@@ -13,21 +20,33 @@ container_definition: &CONTAINER_DEFINITION
   namespace: default
   use_in_memory_disk: true
 
+setup_sonar_scanner: &SETUP_SONAR_SCANNER
+  setup_sonar_scanner_script:
+    - apt update -y && apt upgrade -y && apt update -y && apt install -y unzip       
+    - curl --create-dirs -sSLo $HOME/.sonar/sonar-scanner.zip https://binaries.sonarsource.com/Distribution/sonar-scanner-cli/sonar-scanner-cli-$SONAR_SCANNER_VERSION-linux.zip
+    - unzip -o $HOME/.sonar/sonar-scanner.zip -d $HOME/.sonar/
```
arseniy-sonar, line 27: You are relying on `unzip` to create the `sonar-scanner-${SONAR_SCANNER_VERSION}-linux` directory. It is a timebomb, if at one point it unpacks the contents into a different directory (e.g., because the zip structure changes), it will fail to find the scanner in the PATH.
It is also not obvious that this step is the one that creates the directory added to PATH in the next step.

Can you specify explicitly that you want to unpack into the SONAR_SCANNER_HOME directory?

.cirrus.yml
```diff
@@ -13,21 +20,33 @@ container_definition: &CONTAINER_DEFINITION
...
     cpu: 1
-    memory: 1G
+    memory: 2G
   env:
     PYTHONPATH: .
   install_dependencies_script:
     - cd rspec-tools
     - pipenv install
-    - pipenv run pip install pytest
+    - pipenv run pip install pytest-cov
```
arseniy-sonar, line 41:
```suggestion
    - pipenv run pip install pytest pytest-cov
```
Just want to make it explicit that the script needs both packages.

tomasz-kaminski-sonarsource:
@Godin Just FYI I created to PR to remove the misleading reports from SonarCloud.

arseniy-sonar (changes requested):
How can I see the scan results on SonarQube?

.cirrus.yml
tomasz-kaminski-sonarsource: This code is taken verbatim for the SonarCloud page instruction for manual analysis of the project, so I think we should also dogfood them.

tomasz-kaminski-sonarsource:
To get the link of the scan result, they need to be enabled in SonarQube project settings, that I do not have access to. For this moment, the only way is to get them from analysis log in CirrusCI.

.cirrus.yml
tomasz-kaminski-sonarsource: This line is not need and have no effect.

.cirrus.yml
tomasz-kaminski-sonarsource: And looking at the unzip doc, I can explicitly directories that are in the zip to specific dir (see {{-d $HOME/.sonar/}}), but the directory with version name is part of the archieve.

arseniy-sonar (approved):

PR #694, opened 2022-01-12
Modify rule S109 for C#: new exceptions

andrei-epure-sonarsource:


andrei-epure-sonarsource:
to be merged after https://github.com/SonarSource/sonar-dotnet/pull/5247

PR #699, opened 2022-01-12
Fix python issues raised by SQ for the old code

arseniy-sonar:


arseniy-sonar:
I won't touch the code smells for the rspec-front end here because they are not related to this PR, and are attributed to it erroneously.

tomasz-kaminski-sonarsource (approved):
Looks good. Also, gives me good vibe about our tools, as the fixes requested makes sense.

PR #702, opened 2022-01-12
RULEAPI-703: Fix warnings emitted during `npm run predeploy`

marco-antognini-sonarsource:
Namely, fix a few occurrences of:

    section title out of sequence: expected level 1, got level 2

Visually, this has little or no impact on:

* rules.sonarsource.com (h2 and h3 have the same style)
* SL documentation in VSCode (the two levels have very similar styles)

at the benefit of fixing those warnings.


arseniy-sonar:
the level-1 (h2) headers are treated differently for rule specifications: they delineate the sections, and must belong to [our predefined set of sections](https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_tools/validation/description.py#L13), so this change will not pass asciidoc validation.

The root cause of these warnings, I believe, is the assumed "Description" header for the first section that is added out-of-band and is not visible for asciidoctor.

marco-antognini-sonarsource:
Closing in favour of another solution through a fork for preview.

PR #706, opened 2022-01-13
RULEAPI-722: Always cleanup temprary branch for coverage

tomasz-kaminski-sonarsource:
Unless `always()` or `failure()` is configured `step.if` behave as if `success() &&` was included.

### DIFF
--- a/.github/workflows/update_coverage.yml
+++ b/.github/workflows/update_coverage.yml
@@ -46,6 +46,7 @@ jobs:
       uses: andymckay/cancel-action@0.2
 
     - name: 'Push the updated coverage file to a new branch'
+      id: create-temp-branch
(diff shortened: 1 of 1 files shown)


.github/workflows/update_coverage.yml
```diff
@@ -78,7 +79,7 @@ jobs:
         git push origin master
 
     - name: 'Delete the temporary branch'
-      if: steps.gen-coverage.outputs.new_coverage == 'true'
+      if: always() && steps.create-temp-branch.conclusion == 'success'
```
frederic-tingaud-sonarsource, line 82: how is that different from just `if: steps.create-temp-branch.conclusion == 'success'`?

frederic-tingaud-sonarsource (approved):

PR #707, opened 2022-01-13
Modify Rule S4423[cfamily]: add missing nullptr check for libcurl examples

arseniy-sonar:
Do not merge before the FP is fixed


pierre-loup-tristant-sonarsource (approved):
LGTM

arseniy-sonar:
Hi, @abitmore 
So far no progress made on this change, but it is in our backlog and will be tackled as soon as we have available developers.

necto:
Hey @abitmore, I am glad to let you know that we have fixed CPP-3476 in our analyzer.
Unfortunately, we forgot to update the rule description, and in the meantime, the description has evolved.
I've updated the change to be compatible with the current version of the rule. @pierre-loup-tristant-sonarsource, could you please check if it is still correct and merge it?

PR #708, opened 2022-01-13
Fix UI warnings: <div> nested in <p> and list items lacking keys

arseniy-sonar:
Fix the two warnings observed when testing the SearchPage component:

- Each child in an array should have a unique "key" prop.
  -> add a key prop to list items
- `<div>` cannot be nested in `<p>` (or sth to that effect)
  -> change the relevant Typography components from "p" to "div"


frederic-tingaud-sonarsource (approved):
Applying the same fixes to the fields I added in my PR

PR #709, opened 2022-01-13
RULEAPI-571: test the UI components

arseniy-sonar:
Mostly snapshot testing, with quite big snapshots. I think they will need to be broken down in the future if they break too often.
Also no interaction testing for the moment.

Not sure what to do with these warnings caused by the `react-highlight` (it uses `highlightBlock` inside. our code doesn't use `highlightBlock` directly). Maybe they will go away if we upgrade it?
```
    console.log
      Deprecated as of 10.7.0. highlightBlock will be removed entirely in v12.0

      at deprecated (node_modules/highlight.js/lib/core.js:1490:11)

    console.log
      Deprecated as of 10.7.0. Please use highlightElement now.

      at deprecated (node_modules/highlight.js/lib/core.js:1490:11)
```

### DIFF
--- a/frontend/src/__tests__/RulePage.test.tsx
+++ b/frontend/src/__tests__/RulePage.test.tsx
@@ -5,43 +5,33 @@ import { render } from '@testing-library/react';
 import { RulePage } from '../RulePage';
 import { Router } from 'react-router-dom';
 import { createMemoryHistory } from 'history';
+import { fetchMock } from '../testutils'
+
+const rulesPath = path.join(__dirname, '..', 'deployment', '__tests__', 'resources', 'plugin_rules');
+
+function readRuleFile(ruleId, filename) {
+    return fs.readFileSync(path.join(rulesPath, ruleId, filename)).toString();
+}
 
 beforeEach(() => {
-    const rulesPath = path.join(__dirname, '..', 'deployment', '__tests__', 'resources', 'plugin_rules');
-    const specS1000 = fs.readFileSync(path.join(rulesPath, 'S1000', 'cfamily-description.html')).toString();
-    const metadataS1000 = fs.readFileSync(path.join(rulesPath, 'S1000', 'cfamily-metadata.json')).toString();
-    const specS3457 = fs.readFileSync(path.join(rulesPath, 'S3457', 'csharp-description.html')).toString();
-    const metadataS3457 = fs.readFileSync(path.join(rulesPath, 'S3457', 'csharp-metadata.json')).toString();
-    function fetchMock(url, opts) {
-        const coveredRulesUrl = `${process.env.PUBLIC_URL}/covered_rules.json`;
-        const specUrlS1000 = `${process.env.PUBLIC_URL}/rules/S1000/cfamily-description.html`;
-        const metadataUrlS1000 = `${process.env.PUBLIC_URL}/rules/S1000/cfamily-metadata.json`;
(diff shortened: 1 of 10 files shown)


frontend/src/__tests__/RulePage.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { createMemoryHistory } from 'history';
+
+beforeEach(() => {
+    const rulesPath = path.join(__dirname, '..', 'deployment', '__tests__', 'resources', 'plugin_rules');
+    const specS1000 = fs.readFileSync(path.join(rulesPath, 'S1000', 'cfamily-description.html')).toString();
+    const metadataS1000 = fs.readFileSync(path.join(rulesPath, 'S1000', 'cfamily-metadata.json')).toString();
+    const specS3457 = fs.readFileSync(path.join(rulesPath, 'S3457', 'csharp-description.html')).toString();
+    const metadataS3457 = fs.readFileSync(path.join(rulesPath, 'S3457', 'csharp-metadata.json')).toString();
+    function fetchMock(url, opts) {
+        const coveredRulesUrl = `${process.env.PUBLIC_URL}/covered_rules.json`;
```
frederic-tingaud-sonarsource, line 16: Perhaps we could remove `${process.env.PUBLIC_URL}/` from `url` and have a dictionary of {relativeUrl: answer}?

frontend/src/__tests__/SearchPage.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const searchIndex = buildSearchIndex(indexStore);
+    function fetchMock(url, opts) {
+        const indexUrl = `${process.env.PUBLIC_URL}/rules/rule-index.json`;
+        const storeDataUrl = `${process.env.PUBLIC_URL}/rules/rule-index-store.json`;
+        const aggregatesUrl = `${process.env.PUBLIC_URL}/rules/rule-index-aggregates.json`;
+        const coveredRulesUrl = `${process.env.PUBLIC_URL}/covered_rules.json`;
+        return Promise.resolve({
+            json: () => {
+                if (url === indexUrl) {
+                    return Promise.resolve(JSON.parse(JSON.stringify(searchIndex)));
```
frederic-tingaud-sonarsource, line 21: Do we really need to stringify + parse the result? Is it because we have a risk of the values being changed before the promise resolves?

frederic-tingaud-sonarsource (approved):
Two questions but they shouldn't be blocking.

frederic-tingaud-sonarsource:
> Mostly snapshot testing, with quite big snapshots. I think they will need to be broken down in the future if they break too often.

It's possible we'll want to split them in the future, yes. On the other hand, we don't have too many of them to break yet and they made the test update very simple, which is why we might find it sustainable to just keep it like that in the future.

> Not sure what to do with these warnings caused by the react-highlight (it uses highlightBlock inside. our code doesn't use highlightBlock directly). Maybe they will go away if we upgrade it?

I think it will fix itself when we'll upgrade.

frontend/src/__tests__/RulePage.test.tsx
arseniy-sonar: Done.

frontend/src/__tests__/SearchPage.test.tsx
arseniy-sonar: Yes, Lunr has some custom serialization logic so the stringify + parse produces a different object, and that is what Lunr expects when it loads a search index. I put that in a comment.

PR #714, opened 2022-01-14
Rename 'Common' to 'Text'

nils-werner-sonarsource:
Due to a PM decision, we will rename the unreleased language/domain from "Common" to "Text" [see discuss post](https://discuss.sonarsource.com/t/mmf-2574-sonarqube-and-sonarcloud-products-detect-unicode-bidi-characters-to-prevent-trojan-source-attacks/9360/14).


arseniy-sonar (changes requested):
https://github.com/SonarSource/rspec/blob/master/frontend/src/utils/useRuleCoverage.ts#L35
https://github.com/SonarSource/rspec/blob/master/frontend/src/RulePage.tsx#L112
https://github.com/SonarSource/rspec/blob/master/frontend/src/RulePage.tsx#L144

arseniy-sonar:
Also, do not forget to update rule-api: https://github.com/SonarSource/sonar-rule-api/blob/master/src/main/java/com/sonarsource/ruleapi/utilities/Language.java#L54

rules/S6389/text/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "Using bidirectional characters is security-sensitive",
   "type": "SECURITY_HOTSPOT",
-  "status": "ready",
+  "status": "beta",
```
arseniy-sonar, line 4: I was surprised that this value 'beta' passed validation, but indeed it is one of the permitted values in the json schema, and it's been there since the beginning.
However, I am concerned that it might break some RSPEC tooling, because there it was never used so far.
Do you think the rule can have the status "ready"?
If so, I'll create a ticket to make sure it doesn't break RSPEC.
If not, I'll remove "beta" from the [set](https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_tools/validation/rule-metadata-schema.json#L17) of valid values for rule status

nils-werner-sonarsource: If you like. I can leave it as `ready` for now and manually name it `beta` in the plugin. WDYT? However. I would prefer to see tickets that all types of [status](https://github.com/SonarSource/sonarqube/blob/900af07a23832dc8357eb49c3dd997976686ff29/sonar-plugin-api/src/main/java/org/sonar/api/rule/RuleStatus.java) are supported.

arseniy-sonar: Ok. Since RSPEC hardening is in progress this week, let's merge the PR and see if anything breaks. We should have the capacity to fix it quickly

PR #715, opened 2022-01-14
RULEAPI-703: Fix warnings emitted during `npm run predeploy`

marco-antognini-sonarsource:
Follow up from #702.

Parse custom string instead of file and explicitly introduce level-1 section so that asciidoctor sees it.

Preview at: https://marco-antognini-sonarsource.github.io/rspec/#/rspec/S100

Does it look reasonable?

### DIFF
--- a/frontend/src/RulePage.tsx
+++ b/frontend/src/RulePage.tsx
@@ -316,7 +316,7 @@ export function RulePage(props: any) {
       </Box>
 
       <Box>
-        <Typography variant="h4">Description</Typography>
+        <Typography variant="h4">Rule Documentation</Typography>
         <Typography component={'span'} className={classes.description}>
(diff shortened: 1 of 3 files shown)


frontend/src/RulePage.tsx
```diff
@@ -316,7 +316,7 @@ export function RulePage(props: any) {
       </Box>
 
       <Box>
-        <Typography variant="h4">Description</Typography>
+        <Typography variant="h4">Rule Documentation</Typography>
```
arseniy-sonar, line 319: Why keeping this "redundant" header at all?

arseniy-sonar (approved):
Whether you prefer removing the "Rule Documentation" header or keeping it, I think it is good to go.

PR #719, opened 2022-01-17
Modify S5144: Improved code samples

loris-s-sonarsource:
Changed snippets of the samples for consistency between implementations and to avoid potential fix pitfalls.

Also, some readability changes for chunks of code (columns with the equal sign)

### DIFF
--- a/rules/S5144/csharp/rule.adoc
+++ b/rules/S5144/csharp/rule.adoc
@@ -65,7 +65,7 @@ namespace WebApplicationDotNetCore.Controllers
                 return BadRequest();
             }
 
-            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url); // Compliant
+            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url);
 
             HttpWebResponse response  = (HttpWebResponse)request.GetResponse();
             Stream dataStream         = response.GetResponseStream();
--- a/rules/S5144/csharp/rule.adoc
+++ b/rules/S5144/csharp/rule.adoc
@@ -51,21 +51,24 @@ namespace WebApplicationDotNetCore.Controllers
             return View();
         }
 
-        private readonly string[] whiteList = { "https://www.sonarsource.com" };
+        private readonly string[] whiteList = { "www.example.com", "example.com" };
 
         public IActionResult ReadContentOfURL(string url)
         {
+            // Extract the hostname from the URL
+            URI remoteHost = new Uri(url)
+
             // Match the incoming URL against a whitelist
-            if (!whiteList.Contains(url))
+            if (!whiteList.Contains(remoteHost.Host))
             {
                 return BadRequest();
             }
 
-            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url); // Noncompliant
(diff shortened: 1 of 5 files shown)


rules/S5144/csharp/rule.adoc
```diff
@@ -51,21 +51,25 @@ namespace WebApplicationDotNetCore.Controllers
...
+
             // Match the incoming URL against a whitelist
-            if (!whiteList.Contains(url))
+            if (!whiteList.Contains(remoteHost))
             {
                 return BadRequest();
             }
 
-            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url); // Noncompliant
+            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url); // Compliant
```
hendrik-buchwald-sonarsource, line 68: The "noncompliant" was of course wrong but we do not write "compliant" comments into the "Compliant Solution" section. We only do it in security-expected-issues for ourselves. So you can just completely remove the comment.

Sometimes you will find the comment already existing, like in this rule for the other languages. In that case, you can just remove it. In the beginning, we were not aware that it should not be done but Ann then explicitly told us.

rules/S5144/java/rule.adoc
```diff
@@ -13,13 +13,18 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
...
+  URI uri           = new URI(inputUrl);
+  String remoteHost = uri.getHost();
+
+  if (!urlWhiteList.contains(remoteHost))
     throw new IOException();
 
-  URL url2 = new URL(str);
-  HttpURLConnection conn2 = (HttpURLConnection) url2.openConnection(); // compliant
+  URL url = uri.toURL();
+  HttpURLConnection conn = (HttpURLConnection) url.openConnection(); // Compliant
```
hendrik-buchwald-sonarsource, line 27: Same comment here as above.

rules/S5144/javascript/rule.adoc
```diff
@@ -20,9 +20,12 @@ Validate the url with an allowlist:
...
 function ssrf(req, res) {
-  const url = req.query.url; 
+  const white_list = [ "www.example.com", "example.com" ]
+
+  const url             = (new URL(req.query.url));
+  const remote_hostname = url.hostname;
 
-  if(url.startsWith("https://www.trustedwebsite.com/route/?query=")) { 
+  if (white_list.includes(remote_hostname)) {
     request(url, callback); // Compliant
```
hendrik-buchwald-sonarsource, line 29: Same comment here as above.

rules/S5144/php/rule.adoc
```diff
@@ -12,11 +12,15 @@ $resp = file_get_contents($url); // Noncompliant
...
 );
-$url = $_GET["url"];
-if (in_array($url, $whitelist)) {
-  $resp = file_get_contents($url);
+
+$url         = $_GET["url"];
+$parsed_url  = parse_url($url);
+
+if (in_array($parsed_url['host'], $whitelist)) {
+  $resp = file_get_contents($url); // Compliant
```
hendrik-buchwald-sonarsource, line 23: Same comment here as above.

hendrik-buchwald-sonarsource (approved):
LGTM, just a small comment about a comment.

PR #722, opened 2022-01-18
Modify rule S3735: Ignore JavaScript promises

yassin-kammoun-sonarsource:


rules/S3735/javascript/rule.adoc
```diff
@@ -31,6 +31,12 @@ void (function() {
 }());
 ----
 
+No issue is also raised when ``++void++``'s argument is a promise.
```
vilchik-elena, line 34: I think it's weird to say "also" two times. I would just drop all "also"s

rules/S3735/javascript/rule.adoc
```diff
@@ -31,6 +31,12 @@ void (function() {
...
+
+----
+const f = () => { return new Promise(() => {}); };
```
vilchik-elena, line 37: Could you put here a more real-life promise?

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #728, opened 2022-01-19
RULEAPI-682: Index multiple types and rules with no languages

marco-antognini-sonarsource:
~WIP~ -- This lacks some end-to-end testing, such as search. Will be done separately.

---

preview for S5867:
 * https://marco-antognini-sonarsource.github.io/rspec/#/rspec/?query=S5867&types=BUG
 * https://marco-antognini-sonarsource.github.io/rspec/#/rspec/?query=S5867&types=CODE_SMELL
vs
 * https://sonarsource.github.io/rspec/#/rspec/?query=S5867&types=BUG
 * https://sonarsource.github.io/rspec/#/rspec/?query=S5867&types=CODE_SMELL

preview for rules with no languages (e.g. S1076):
 * https://marco-antognini-sonarsource.github.io/rspec/#/rspec/?query=S1076&types=ANY
vs
 * https://sonarsource.github.io/rspec/#/rspec/?query=S1076&types=ANY


marco-antognini-sonarsource:
(This ain't ready for review -- just an attempt at triggering the CI jobs.)

marco-antognini-sonarsource:
@arseniy-sonar this is now ready. FWIW There are two code smells I'm not sure how to address.

frontend/src/deployment/metadata.ts
```diff
@@ -68,12 +101,20 @@ export function generateRulesMetadata(srcPath: string, dstPath: string, rules?:
...
+    if (fs.existsSync(parentFile)) {
+      return JSON.parse(fs.readFileSync(parentFile, 'utf8'));
+    }
+    return {};
+  })();
   const childFile = path.join(srcDir, 'metadata.json');
   const childJson = fs.existsSync(childFile) ? JSON.parse(fs.readFileSync(childFile, 'utf8')) : {};
-  return {...childJson, ...parentJson};
+  return { ...childJson, ...parentJson };
 }
```
arseniy-sonar, line 120: Frankly, I think "parent" and "child" names a swapped here. The semantics is correct, but it is mindboggling to think of the root rule directory as a "child" and the language subdirectory as a "parent". If you agree do you mind swapping the names (as you are rewriting this function anyway)?

frontend/src/deployment/metadata.ts
```diff
@@ -1,63 +1,96 @@
...
 
+import { LanguageSupport } from '../types/RuleMetadata';
 import { getRulesDirectories, listSupportedLanguages } from './utils';
 
+/**
+ * Save the given metadata to disk.
+ */
+function writeRuleMetadata(dstDir: string, filename: string, metadata: any) {
+  const file = path.join(dstDir, filename);
+  fs.writeFileSync(file, JSON.stringify(metadata), { encoding: 'utf8' });
```
arseniy-sonar, line 12:
```suggestion
  fs.writeFileSync(file, JSON.stringify(metadata, null, 2), { encoding: 'utf8' });
```
I agree with keeping the `rule-index-store.json` short - with no line breaks or indentation. However, I think we should keep the metadata for each rule readable, in case some want's to inspect the `language-metadata.json` file, or download it somewhere else.
These files are relatively short, so I don't think this adds much overhead, but substantially improves readability.

frontend/src/deployment/metadata.ts
```diff
@@ -1,63 +1,96 @@
...
+}
+
+/**
+ * Merge all sqKeys in an array to check rule coverage.
+ */
+function getAllKeys(allMetadata: { metadata: any }[]) {
```
arseniy-sonar, line 18: Regarding the codesmells on SQ, I think you can specify that you expect the `metadata` to be of a type that has an `sqKey` field and an (optional) `extra` field that is an object that has an (optional) `legacyKeys` field, just like you did for the `allMetadata` type.

frontend/src/deployment/metadata.ts
```diff
@@ -1,63 +1,96 @@
...
+    metadata.extra?.legacyKeys?.forEach((key: string) => set.add(key));
+    return set;
+  }, new Set<string>());
+  return Array.from(keys);
+}
+
+/**
+ * Generate the default metadata for a rule without any language-specific data.
+ */
+function generateGenericMetadata(srcDir: string, dstDir: string, branch: string, prUrl?: string) {
```
arseniy-sonar, line 30: If it is not forbidden yet, it should be forbidden to open a new-rule PR with no language, so you don't need to handle the `prUrl` parameter for rules with no language.

By "forbidden" I mean checked by CI validation. the deployment probably shouldn't crash, just emit a warning, to avoid having one broken PR to prevent the entire frontend from deployment.

frontend/src/deployment/description.ts
```diff
@@ -59,32 +82,41 @@ export function generateRulesDescription(srcPath: string, dstPath: string, rules
...
-        attributes: {
-          'rspecator-view': '',
-          docfile: ruleSrcFile,
-        },
-        safe: 'unsafe',
-        base_dir: baseDir,
-        backend: 'xhtml5',
-        to_file: false
-    };
+function getRuleAdoc(srcDir: string, language?: string) {
```
arseniy-sonar, line 89: This looks like a false abstraction. AFAIK, you call this `getRuleAdoc` from two places, and in one place you provide the `language` argument in another - not. Inside the function you branch on whether the argument is provided. So looks like you are trying to smash together two different functions `getSpecificRuleAdoc` and `getGenericRuleAdoc`. They might share some logic, but their differences unnecessarily multiply the complexity of the function body.

frontend/src/deployment/__tests__/searchIndex.test.ts
```diff
@@ -79,113 +168,131 @@ describe('index store generation', () => {
...
+    ruleIndexStore = indexStore;
+  }
 
-  function search(index: lunr.Index, query: string, field: string): string[] {
-    const hits = index.query(q => {
-      q.term(query, {
+  // Hack to avoid warnings when 'selectivePipeline' is already registered
+  if ('selectivePipeline' in (lunr.Pipeline as any).registeredFunctions) {
+    delete (lunr.Pipeline as any).registeredFunctions['selectivePipeline']
+  }
```
arseniy-sonar, line 260: should this be a part of an "afterEach" fixture?

frontend/src/deployment/__tests__/searchIndex.test.ts
```diff
@@ -79,113 +168,131 @@ describe('index store generation', () => {
...
+  return buildSearchIndex(ruleIndexStore);
+}
+
+// These search functions mimic the behaviors of useSearch.ts.
```
arseniy-sonar, line 264: How painful is it to factor out the "behaviors of useSearch.ts" and call them from this test? (To make sure they faithfully mimic these behaviors)

arseniy-sonar (changes requested):
Nice job testing the deployment scripts!
Mostly good, I just left a few nitpicks and refactoring suggestions.

frontend/src/deployment/metadata.ts
marco-antognini-sonarsource: Currently, AFAICT, those JSON files are not easily downloadable. At least I couldn't find links to them (although if you know where they are you can handcraft the URLs). So I wouldn't expect people to download them. If someone really wants them, they can easily reformat them too (using `python -m json.tool` for example).

In numbers, I just measured a reduction of 30%: before we had 10.14MB of `*-metadata.json` files, now about 7.03MB.

The readability aspect isn't modified by this change: they are pretty printed anyway in `RulePage.tsx` anyway.

frontend/src/deployment/metadata.ts
marco-antognini-sonarsource: Thanks, I knew I was forgetting something! I'll check and improve this.

frontend/src/deployment/metadata.ts
marco-antognini-sonarsource: Sure, sounds good. 👍

marco-antognini-sonarsource (commented):
Thanks for the feedback. :-)

frontend/src/deployment/metadata.ts
marco-antognini-sonarsource: Should be better now. I've used other names to remove any ambiguity. Hopefully, that helps.

frontend/src/deployment/metadata.ts
marco-antognini-sonarsource: I've improved this a bit thanks to your advice.

frontend/src/deployment/__tests__/searchIndex.test.ts
marco-antognini-sonarsource: Yup, that was a good idea -- I realized some of the testing I wrote wasn't meaningful. If you wonder, I had to replace "be should" with "should be used" to avoid triggering a match on the third rule. The new testing is a bit more relaxed but it fits more closely what the search page is doing.

frontend/src/deployment/metadata.ts
marco-antognini-sonarsource: (If you're reading the update I just pushed, wonder not: this wasn't addressed yet.)

marco-antognini-sonarsource: This wasn't trivial as I had to learn a bit about pytest and Cirrus along the way, but it should be better now: the CI job will run additional checks only on modified rules. As written, the additional checks are simple: there must be one language. Later we could add more specific checks if needed. In other words, this new check is not run on rules already on master.

.cirrus.yml
```diff
@@ -78,6 +78,8 @@ validate_metadata_task:
     dockerfile: ci/Dockerfile
     cpu: 1
     memory: 1G
+  env:
```
marco-antognini-sonarsource, line 81: This is required otherwise `git merge-base` fails.

frontend/src/deployment/metadata.ts
```diff
@@ -1,63 +1,103 @@
 import fs from 'fs';
 import path from 'path';
-import { LanguageSupport } from '../types/RuleMetadata';
 
+import { LanguageSupport } from '../types/RuleMetadata';
 import { getRulesDirectories, listSupportedLanguages } from './utils';
 
+type Metadata = any;
```
marco-antognini-sonarsource, line 7: This is still using `any`. Hopefully, this is okay. Let me know if you have a better idea.

frontend/src/utils/useSearch.ts
```diff
@@ -1,15 +1,60 @@
 import React, { useState } from 'react';
 
-import * as lunr from 'lunr'
+import lunr from 'lunr';
```
marco-antognini-sonarsource, line 3: I took the opportunity to some of the clean existing code smells. :-)

rspec-tools/rspec_tools/cli.py
```diff
@@ -68,9 +68,22 @@ def validate_rules_metadata(rules):
...
         error_counter += 1
   if error_counter > 0:
-    message = f"Validation failed due to {error_counter} errors"
-    click.echo(message, err=True)
-    raise click.Abort(message)
+    fatal_error(f"Validation failed due to {error_counter} errors")
+
+@cli.command()
+@click.argument('rule_key', required=True)
+def validate_modified_rule_metadata(rule_key: str):
```
marco-antognini-sonarsource, line 75: SQ coverage fails because of that function. I'm not sure how it should be tested... Maybe we should have a dedicated ticket to improve this coverage?

rspec-tools/rspec_tools/cli.py
```diff
@@ -68,9 +68,22 @@ def validate_rules_metadata(rules):
...
+  '''Validate the given rule, assuming it was modified and therefore requires additional checks
+     on top of validate_rules_metadata().
+  '''
```
arseniy-sonar, line 78: Something is bothering me with this differentiation: "validation for all metadata" vs. "validation for metadata of changed rules":
- if a rule did not change, why validate its metadata again (as in "validation for all metadata")?
- why run some validation for the changed rules (which does not look at what changed specifically) - will it fail for a rule once it is no longer changing?

I understand that you implemented this to make sure some validation is run for PRs introducing new rules, but if I didn't discuss it with you, I would look at this (as it happened when I looked at this before remembering our discussion) and be perplexed.

IMO, running metadata validation only on the subset of rules that changed makes sense. It wasn't done before only because metadata validation is cheap and there was no need to optimize it (unlike adoc validation). However, it would be simpler to understand if running validation for a rule would always give the same result: valid/invalid regardless of whether this rule was just created, or is an old one; modified or not; on the `master` or on a fresh branch.

I don't see a solution that would reconcile the wishes above and the distinction we wanted to make between the old rules with no language and introducing a new rule/removing the last language from an existing rule. Maybe hardcode the list of rules for which it is OK to have no language? Or allow all the `closed` rules to have no language?

arseniy-sonar (commented):
I did not have time to finish the review before the interview. I left some small notes and one that calls for some thinking. Feel free to address them now. I'll have another look after lunch.

ci/validate_metadata.sh
```diff
@@ -1,7 +1,23 @@
-# Validate metadata
+#!/bin/bash
+set -ueo pipefail
+
+invalid=0
+
+git fetch --quiet "${CIRRUS_DEFAULT_ORIGIN:-origin}" "${CIRRUS_DEFAULT_BRANCH:-master}"
+base="$(git merge-base FETCH_HEAD HEAD)"
+echo "Comparing against the merge-base: ${base}"
+modifiedRules="$(git diff --name-only "${base}" -- rules | sed -Ee 's#rules/(S[0-9]+)/.*#\1#' | sort -u)"
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: > Why does one require the CIRRUS_CLONE_DEPTH setting and the other one not?

Both actually do, and that's how I figured out how to fix the CI failures.

> How is this different from the similar snippet in validate_asciidoc.sh ?

In terms of main differences: here we only look at `rules/` and no other directories, which allows writing a simpler command.

Aside from that, here I followed shellcheck's recommendations for quotes and error handling.

> Also, like in validate_asciidoc.sh I'd extend the modifiedRules set to all rules if any relevant script is modified.

The purpose of my change was to run an extra set of checks on modified rules only, in addition to running the existing checks for all rules. So following the same logic as `validate_asciidoc.sh` was not necessary. Let's discuss whether/how we want to modify this in your other comment.

rspec-tools/rspec_tools/cli.py
marco-antognini-sonarsource: > I understand that you implemented this to make sure some validation is run for PRs introducing new rules, but if I didn't discuss it with you, I would look at this (as it happened when I looked at this before remembering our discussion) and be perplexed.
> [...]
> However, it would be simpler to understand if running validation for a rule would always give the same result: valid/invalid regardless of whether this rule was just created, or is an old one; modified or not; on the master or on a fresh branch.

That's a very fair point indeed.

> IMO, running metadata validation only on the subset of rules that changed makes sense. It wasn't done before only because metadata validation is cheap and there was no need to optimize it (unlike adoc validation).

That's simple enough to do it, so why not indeed. 👍 

> I don't see a solution that would reconcile the wishes above and the distinction we wanted to make between the old rules with no language and introducing a new rule/removing the last language from an existing rule. Maybe hardcode the list of rules for which it is OK to have no language? Or allow all the closed rules to have no language?

To keep things as simple as possible, and therefore easily modifiable in the future, a fixed list of known exception might be better. At least, when a new rule is introduced and the test fails, someone will have to acknowledge what they are doing is probably not adequate. Also, it seems to follow more closely the historical evolution of this project.

With a generic exception for closed rules, someone may remove (by mistake) language specification.

ci/validate_metadata.sh
```diff
@@ -1,7 +1,23 @@
-# Validate metadata
+#!/bin/bash
+set -ueo pipefail
+
+invalid=0
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: done in 99ee0cc

frontend/src/deployment/metadata.ts
```diff
@@ -4,18 +4,25 @@ import path from 'path';
...
  */
-function writeRuleMetadata(dstDir: string, filename: string, metadata: any) {
+function writeRuleMetadata(dstDir: string, filename: string, metadata: Metadata) {
   const file = path.join(dstDir, filename);
   fs.writeFileSync(file, JSON.stringify(metadata), { encoding: 'utf8' });
 }
 
+type SQKeyMetadata = {
+  sqKey: string,
+  extra: undefined | { legacyKeys: undefined | string[] },
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: done in 3c90f6e

rspec-tools/rspec_tools/cli.py
marco-antognini-sonarsource: > Maybe we should have a dedicated ticket to improve this coverage?

I should rephrase that: we _should_ do it. I'll open a ticket for this. Reason being: `validate_rules_metadata` was incorrect already: `rule.key` should have been `rule.id`, for example.

ci/validate_metadata.sh
```diff
@@ -1,23 +1,25 @@
...
-
-# Validate all metadata
-pipenv run rspec-tools validate-rules-metadata || exit_code=1
+if ! git diff --name-only --exit-code "${base}" -- rspec-tools/
+then
+  # Revalidate all rules
+  affected_rules="$(basename --multiple rules/*)"
+else
+  affected_rules="$(git diff --name-only "${base}" -- rules/ | sed -Ee 's#rules/(S[0-9]+)/.*#\1#' | sort -u)"
+fi
```
marco-antognini-sonarsource, line 13: I'll leave this as-is for now to avoid more scope creep, but these commands could be extracted to a shared file and reused in `validate_asciidoc.sh`.

ci/validate_metadata.sh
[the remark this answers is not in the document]
marco-antognini-sonarsource: BTW, there seems to be an existing check already: https://github.com/SonarSource/rspec/blob/1ca56eaf428fbd35c3eced18fb1b98f3a906049a/ci/validate_asciidoc.sh#L50-L56
However, I was not expecting to find metadata checks in asciidoc validation code, and I'm not sure how robust it is. For these reasons, I went ahead and updated the python scripts to properly validate metadata.

frontend/src/utils/useSearch.ts
```diff
@@ -1,15 +1,60 @@
...
 
 import { useFetch } from './useFetch';
 import { IndexedRule, IndexStore } from '../types/IndexStore';
 
+export function filterTypes(q: lunr.Query, type: string) {
```
arseniy-sonar, line 8: This family of functions, modify its first argument, which might surprise someone like me reading the code.
You might want to rename them using a more "impure" verb. for example: `addFilterForTypes`.
Even better, IMO, would be to subclass `lunr.Query` and define these functions as member-functions.

So for a function that modifies `q`, for me in order from the least surprise to the most would be the call sites:
- `q.addFilterForTypes(type)`
- `q.filterTypes(type)`
- `addFilterTypes(q, type)`
- `filterTypes(q, type)`

arseniy-sonar: btw, the function name mentions `types`, but the argument is called `type` - something does not match here.

rspec-tools/rspec_tools/validation/metadata.py
```diff
@@ -20,6 +92,37 @@ def validate_metadata(rule_language: LanguageSpecificRule):
   validate_status(rule_language)
   validate_security_standards(rule_language)
 
+
+def validate_rule_metadata(rule: GenericRule):
```
arseniy-sonar, line 96: The name is confusing here: `validate_metadata` vs `validate_rule_metadata`, although the argument types help.
How about renaming into: `validate_rule_metadata` and `validate_metadata_of_language_specific_rule`, or along those lines? To make it clear when looking at the two which one you should use in each case.

rspec-tools/rspec_tools/validation/metadata.py
```diff
@@ -20,6 +92,37 @@ def validate_metadata(rule_language: LanguageSpecificRule):
...
+  '''In addition to the test carried out by validate_metadata for each language specification,
+     a rule must have at least one language (unless it is part of the list of exception).
+  '''
+  specializations = list(rule.specializations)
+  if rule.id in RULES_WITH_NO_LANGUAGES:
+    if specializations:
+      # When this triggers, ask yourself whether the rule should be removed from RULES_WITH_NO_LANGUAGES
+      raise RuleValidationError(f'Rule {rule.id} should have no specializations')
```
arseniy-sonar, line 104: Perhaps add the (shorter version of) the comment to the error message?
e.g.
```suggestion
      # When this triggers, ask yourself whether the rule should be removed from RULES_WITH_NO_LANGUAGES
      raise RuleValidationError(f'Rule {rule.id} should have no specializations. Forgot to remove it from the list?')
```

arseniy-sonar (approved):
Great job! I left a few comments and suggestions. I probably won't be able to do another review before leaving today, but that's not needed. Whether or not you take action on my comments, it is good to go.

rspec-tools/rspec_tools/cli.py
arseniy-sonar: Agreed.

frontend/src/utils/useSearch.ts
marco-antognini-sonarsource: Good idea to rename the function. Will do.

Regarding `types` vs `type`, it's because a rule can have multiple types but the user filters using one type only. It's a bit awkward indeed.

rspec-tools/rspec_tools/validation/metadata.py
marco-antognini-sonarsource: What about `validate_rule_metadata` vs `validate_rule_specialization_metadata` to reuse existing terminology?

rspec-tools/rspec_tools/validation/metadata.py
marco-antognini-sonarsource: Done in d8ab57a.

frontend/src/utils/useSearch.ts
marco-antognini-sonarsource: Done in c44ce47.

rspec-tools/rspec_tools/validation/metadata.py
marco-antognini-sonarsource: done in c44ce47

rspec-tools/rspec_tools/cli.py
marco-antognini-sonarsource: https://jira.sonarsource.com/browse/RULEAPI-735

PR #734, opened 2022-01-21
Update S2815: it is not deprecated for C++ and not relevant for java and csharp

abbas-sabra-sonarsource:


abbas-sabra-sonarsource:
@arseniy-sonar It is not clear to me if the top-level metadata JSON is needed if the rule is one language. Kept for consistency with other rules

frederic-tingaud-sonarsource:
My understanding is that we want to keep the deprecated and closed rules for other languages, to keep the history of its evolution.

abbas-sabra-sonarsource:
> My understanding is that we want to keep the deprecated and closed rules for other languages, to keep the history of its evolution.

Yes, but looking at Jira this rule was never implemented for other languages. Probably we have this subfolder because of the targeted field from JIRA

arseniy-sonar (approved):

PR #735, opened 2022-01-21
RULEAPI-726: Add end-to-end testing

marco-antognini-sonarsource:
Works on top of #728, hence the first commits are "duplicated" between this PR and #728. See the "WIP" commits for now.

The idea is to have a self-contained end-to-end test that generates the index from metadata + adoc and performs a few search queries.


frontend/src/__tests__/RulePage.test.tsx
```diff
@@ -6,6 +6,7 @@ import { RulePage } from '../RulePage';
 import { Router } from 'react-router-dom';
 import { createMemoryHistory } from 'history';
 import { fetchMock } from '../testutils'
+import { removeSelectivePipeline } from './SearchPage.test';
```
marco-antognini-sonarsource, line 9: this is obviously unused at the moment...

frontend/src/__tests__/SearchPage.test.tsx
```diff
@@ -16,7 +17,15 @@ function normalize(obj) {
     return JSON.parse(JSON.stringify(obj));
 }
 
+export function removeSelectivePipeline() {
```
marco-antognini-sonarsource, line 20: This function and the calls to it silence some warnings in the log output.

frontend/src/__tests__/EndToEndTesting.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+      //   history.push('/rspec/#/rspec/');
+      //   const { findByText } = render(<Router history={history} > <SearchPage /></Router >);
+      //   await findByText(/Search Rule Specifications/);
+      //   await findByText(/Number of rules found: 3\b/);
+      //   await findByText(/Java Rule S100 Title/);
+      //   await findByText(/Python Rule S200 Title/);
+      //   await findByText(/Java Rule S200 Title/);
+      //   await findByText(/Rule S501 Rule is closed with no language-specific specification/);
+      // }
+      // FIXME how to reset state? If both are uncommented, the 'Search Rule Specifications' is found multiple times in the second test...
```
marco-antognini-sonarsource, line 102: FIXME here.

> how to reset state? If both are uncommented, the 'Search Rule Specifications' is found multiple times in the second test...

frontend/src/__tests__/EndToEndTesting.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+      // Test search by title.
+      {
+        const history = createMemoryHistory();
+        history.push('/rspec/#/rspec/?query=Java+Rule+S100+Title');
+        const { findByText, asFragment } = render(<Router history={history}> <SearchPage /></Router >);
+        // expect(asFragment()).toMatchInlineSnapshot();
+        await findByText(/Search Rule Specifications/);
+        // await findByText(/Number of rules found: 1/); // FIXME how to properly pass query?
```
marco-antognini-sonarsource, line 111: FIXME here

> how to properly pass query?

frontend/src/__tests__/EndToEndTesting.test.tsx
marco-antognini-sonarsource: I don't have an explanation of what is going on in the background but breaking tests into `test('bla', () => { ... });` fixed this somehow!

frontend/src/__tests__/EndToEndTesting.test.tsx
marco-antognini-sonarsource: Mocking the `useLocation` wasn't easy. I tried simpler solutions (e.g. using `jest.fn`) but they didn't work so I basically wrote a poor man mock function for this. Seems to work now.

frontend/src/__tests__/SearchPage.test.tsx
```diff
@@ -1,22 +1,14 @@
...
-
-function normalize(obj) {
-    // Lunr (the search engine) expects its objects to have been
-    // serialized and deserialized when it is queried.
-    // This is not a no-op, because, for example, it translates function references to
-    // simple labels on the serialization step, and then uses these labels to
-    // restore the function references when loading.
-    return JSON.parse(JSON.stringify(obj));
-}
+import { fetchMock, normalize, removeSelectivePipeline } from '../testutils';
```
marco-antognini-sonarsource, line 8: I had to move them out of this file because importing a `*.test.tsx` file seems to run the tests in the file that has the `import` statement.

frontend/src/__tests__/EndToEndTesting.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const Paths = {
+  src: '',
+  dst: '',
+  index: '',
+  store: '',
+  aggregates: '',
+};
+
+beforeAll(() => {
```
marco-antognini-sonarsource, line 24: In order to have multiple `test()` I had to write my own `withTestDir` with `beforeAll` and `afterAll`.

marco-antognini-sonarsource:
FWIW, This is still very much WIP (hence the commit message, this will be cleaned later on)

frontend/src/deployment/__tests__/searchIndex.test.ts
```diff
@@ -79,113 +168,131 @@ describe('index store generation', () => {
...
+    const rulesPath = path.join(__dirname, 'resources', 'metadata');
+    const [indexStore, _] = buildIndexStore(rulesPath);
+    ruleIndexStore = indexStore;
+  }
 
-  function search(index: lunr.Index, query: string, field: string): string[] {
-    const hits = index.query(q => {
-      q.term(query, {
+  // Hack to avoid warnings when 'selectivePipeline' is already registered
+  if ('selectivePipeline' in (lunr.Pipeline as any).registeredFunctions) {
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: done.

frontend/src/__tests__/EndToEndTesting.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    await findByText(/Number of rules found: 1/);
+    await findByText(/Python Rule S200 Title/);
+  });
+
+  // This test emits many warnings:
+  //   Material-UI: You have provided an out-of-range value `ANY` for the select component.
+  //   Consider providing a value that matches one of the available options or ''.
+  //   The available values are `ANY`.
+  // This is because the languages are not yet available when the page renders the first time.
+  test('Query by language: python', async () => {
```
marco-antognini-sonarsource, line 219: If this is too noisy, I can remove the test.

arseniy-sonar: It is noisy indeed, but I think it is worth fixing, for instance:
```diff
diff --git a/frontend/src/SearchPage.tsx b/frontend/src/SearchPage.tsx
index 68e78ab85..013ad019e 100644
--- a/frontend/src/SearchPage.tsx
+++ b/frontend/src/SearchPage.tsx
@@ -208,7 +208,7 @@ export const SearchPage = () => {
                   size="small"
                   variant="outlined"
                   label="Language"
-                  value={ruleLang}
+                  value={allLangs.length === 0 ? "" : ruleLang}
                   onChange={handleUpdate('lang')}
                   data-testid="rule-language"
                 >
```

frontend/src/__tests__/EndToEndTesting.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    state: null,
+    hash: 'somehash',
+    key: undefined,
+  };
+}
+
+const history = createMemoryHistory();
+beforeAll(() => {
+  history.push('/rspec/#/rspec/');
+});
```
arseniy-sonar, line 148: This part seems to be redundant: tests pass if I remove it. So maybe it should be removed?

frontend/src/__tests__/EndToEndTesting.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    await findByText(/Number of rules found: 1/);
+    await findByText(/Rule S501 Rule is closed with no language-specific specification/);
+  });
+
+  test('Query by type: BUG', async () => {
+    fakeLocation = generateFakeLocation('types=BUG');
+    const { findByText, findAllByText } = render(<Router history={history}> <SearchPage /></Router >);
+    expect(await findAllByText(/Rule Title and Description/)).toHaveLength(1 + 1); // <label> + <span>
+    await findByText(/Number of rules found: 1/);
+    await findByText(/Java Rule S200 Title/);
```
arseniy-sonar, line 186: In some tests you mention both titles "Java Rule S200 Title" and "Python Rule S200 Title", and in some only one of them, while they are actually both either present or absent. One might assume that the search also filters titles within a rule. I think it is more clear assert the presence of both, or perhaps assert the presence of both in the first test, comment that they have shared destiny, and then assert the presence of only one in all subsequent tests.

arseniy-sonar (approved):
I like these tests

frontend/src/__tests__/EndToEndTesting.test.tsx
marco-antognini-sonarsource: Thanks, I was going crazy with that one... Glad the fix is so simple!

frontend/src/__tests__/EndToEndTesting.test.tsx
marco-antognini-sonarsource: Yes, indeed. Thanks for catching this. Done.

frontend/src/__tests__/EndToEndTesting.test.tsx
marco-antognini-sonarsource: That's a very good point. I've added the check to all tests to make them as independent as possible, even if this means they overlap a bit.

PR #738, opened 2022-01-23
Modify rule S105: add comment about smart tabs

jycr:
Fixing the visual size of the indentations for everyone makes it difficult to read the code for people who need a size adapted to their needs or their context.

Indeed, the advantage of tabulation is precisely to make it possible to adapt the readability of the code according to the constraints of the user.

This is an fundamental accessibility criterion.

In addition, this allows a group of developers, some of whom prefer large indentations, and others, more restrained indentations, to work on the same code base (each configuring according to his needs, on his IDE, the visual size of the desired intentation).

Here is an article motivating the deprecation of this rule: https://dev.to/alexandersandberg/why-we-should-default-to-tabs-instead-of-spaces-for-an-accessible-first-environment-101f

### DIFF
--- a/rules/S105/comments-and-links.adoc
+++ b/rules/S105/comments-and-links.adoc
@@ -15,3 +15,16 @@ Low value rule for python as:
 * tabs are accepted if the rest of [the code base is already using it](\https://www.python.org/dev/peps/pep-0008/#id18).
 * pycodestyle already supports all PEP8 rules and there is no point for us to reimplement them.
 
+=== on 23 Jan 2022, 22:20 JY Cr wrote:
+
+Fixing the visual size of the indentations for everyone makes it difficult to read the code for people who need a size adapted to their needs or their context.
+
+Indeed, the advantage of tabulation is precisely to make it possible to adapt the readability of the code according to the constraints of the user.
+
+This is an fundamental accessibility criterion.
+
+In addition, this allows a group of developers, some of whom prefer large indentations, and others, more restrained indentations, to work on the same code base (each configuring according to his needs, on his IDE, the visual size of the desired intentation).
+
+Here is an article motivating the deprecation of this rule: https://dev.to/alexandersandberg/why-we-should-default-to-tabs-instead-of-spaces-for-an-accessible-first-environment-101f
+
+In order not to go against readability and accessibility aspects, while avoiding problems related to alignment (different concept from indentation), it is preferable to favor "[smart tabs](https://www.emacswiki.org/emacs/SmartTabs) indentation".


arseniy-sonar:
Hi, JY Cr!
Thank you for contributing to this rules repository, it is much appreciated.

I see no reason to deprecate the rule based on your argument (which I find nonetheless valid and strong).

In our products, we do our best to recommend only the rules that are objectively correct. We do this by maintaining a default quality profile called "Sonar Way". As you can see, this rule is not included in Sonar Way, partly because of the argument you cited.

Rule S105 is controversial. One argument against using tabs is the frequent situation of people using tabs for alignment (rather than indentation; due to lack of experience or tooling issues). Tab-based indentation breaks once you view the code with a different tab width hindering the code legibility. 

We do need to support our customers and users, and that means sometimes implementing controversial rules, that are neither objectively right nor wrong. This is one of such rules. This rule might make sense in a project that has inexperienced contributors and has no contributors that have an issue with the common indentation depth. Ultimately it is the call of the maintainer of the codebase whether this rule should be activated or not. (This is also indicated by the tag "convention").

We deprecate a rule when we find out that the rule can no longer be justified, or when we implement a new rule that provides more value for all cases in which the old one was relevant. This is not the case for S105.

I do find your comment to be very relevant.

I will be happy to merge this PR not as a deprecation PR, but as a comment - if you revert the changes to S105/metadata.json.

arseniy-sonar (changes requested):
Hi, JY Cr!
Thank you for contributing to this rules repository, it is much appreciated.

I see no reason to deprecate the rule based on your argument (which I find nonetheless valid and strong).

In our products, we do our best to recommend only the rules that are objectively correct. We do this by maintaining a default quality profile called "Sonar Way". As you can see, this rule is not included in Sonar Way, partly because of the argument you cited.

Rule S105 is controversial. One argument against using tabs is the frequent situation of people using tabs for alignment (rather than indentation; due to lack of experience or tooling issues). Tab-based indentation breaks once you view the code with a different tab width hindering the code legibility.

We do need to support our customers and users, and that means sometimes implementing controversial rules, that are neither objectively right nor wrong. This is one of such rules. This rule might make sense in a project that has inexperienced contributors and has no contributors that have an issue with the common indentation depth. Ultimately it is the call of the maintainer of the codebase whether this rule should be activated or not. (This is also indicated by the tag "convention").

We deprecate a rule when we find out that the rule can no longer be justified, or when we implement a new rule that provides more value for all cases in which the old one was relevant. This is not the case for S105.

I do find your comment to be very relevant.

I will be happy to merge this PR not as a deprecation PR, but as a comment - if you revert the changes to S105/metadata.json.

jycr:
You are right.
I will modify my PR to remove the changes in the metadata.

Do you think it is relevant if I propose another rule to favor "[smart tabs](https://www.emacswiki.org/emacs/SmartTabs)" so code base maintainers wishing to favor the use of tabs (for the reasons mentioned above) can activate this rule to encourage this practice if necessary?

jycr:
You are right.
I have modified my PR to remove the changes in the metadata and changed the title, according: https://github.com/SonarSource/rspec#to-modify-an-existing-rule

Do you think it is relevant if I propose another rule to favor "[smart tabs](https://www.emacswiki.org/emacs/SmartTabs)" so code base maintainers wishing to favor the use of tabs (for the reasons mentioned above) can activate this rule to encourage this practice if necessary?

arseniy-sonar:
I've merged your change in #740 (a duplicate PR; for security reasons CI does not work for external contributors). So I'll close this PR.
Thank you!

To suggest a new rule I encourage you to use our community forum: https://community.sonarsource.com/c/suggestions/rules/13

PR #739, opened 2022-01-24
Modify rule S1102: add exception for simple tables

roberto-orlandi-sonarsource:


rules/S1102/html/exceptions.adoc
```diff
@@ -0,0 +1,32 @@
+== Exceptions
+
+This rule is not applied in case of simple tables.
+
+Tables are considered as such when the headers are all in the first row OR all in the first column. The two conditions must not apply together.
```
christophe-zurn-sonarsource, line 5: I'd suggest the following change, as I find it strange as styling to have a single word as uppercase (`OR`):
```suggestion
Tables are considered as such when the headers are either all in the first row, or all in the first column. The two conditions must not apply together.
```

christophe-zurn-sonarsource (approved):
LGTM, although I'd suggest a small improvement regarding the styling of the second sentence

PR #743, opened 2022-01-25
Modify rule S5813: Fix typos

marco-antognini-sonarsource:


rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,28 @@
 In C, a string is just a buffer of characters, usually using the ``++null++`` character as a sentinel for the end of the string. Which mean that the developer has to be aware of low level details such as buffer sizes, having an extra character to store the final null character... Doing that correctly and consistently is notoriously difficult, and any error can lead to a security vulnerability, for instance giving access to sensitive data, or allowing arbitrary code execution.
 
 
-The function ``++char *strlen(char * restrict s);++`` measures the length of the string ``++s++`` (excluding the final null character). The ``++wcslen++`` does the same for wide characters, and should be used with the same guidelines.
+The function `size_t strlen(const char *s)` measures the length of the string `s` (excluding the final null character). The `wcslen` does the same for wide characters, and should be used with the same guidelines.
```
loic-joly-sonarsource, line 4: Not sure about removing `restrict`. Since this rule mostly targets a C audience, it may make sense to display it as C developers are accustomed to see it?

marco-antognini-sonarsource: I took the spelling from http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf. That's also how it's displayed on https://en.cppreference.com/w/c/string/byte/strlen. It seems `strlen` is simply different in that matter from `strcpy` and others (well, it could be `restrict`'d too in `strlen` but it would have no value given there's only one parameter).

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,28 @@
 In C, a string is just a buffer of characters, usually using the ``++null++`` character as a sentinel for the end of the string. Which mean that the developer has to be aware of low level details such as buffer sizes, having an extra character to store the final null character... Doing that correctly and consistently is notoriously difficult, and any error can lead to a security vulnerability, for instance giving access to sensitive data, or allowing arbitrary code execution.
```
loris-s-sonarsource, line 1: Hey, I saw the Discourse topic, sorry for popping out from out of nowhere. I couldn't help but notice minor mistakes in the first paragraph and think to myself that the paragraph in itself may lack clarity.

Also, the rule doesn't even mention the term "off-by-one error", which is the concrete mistake that this security hotspot wants to put in the spotlight.

Here is a suggestion for cutting off surplus details and improving clarity:

```suggestion
In C, a string is a buffer of characters that usually uses the character ``++null++`` as the end of the string. This means that the developer must be careful with operations involving the size of the buffer to avoid ''off-by-one'' errors. +
Doing this correctly and consistently is notoriously difficult, and any error can lead to a security vulnerability that allows, for example, access to sensitive data or execution of arbitrary code.
```

loris-s-sonarsource: It seems that I misunderstood the rule and that it's not really targeting off-by-one errors, but actually saying "**Ensure that the string finishes with a null byte**".

loris-s-sonarsource: I'll make a broader suggestion

loris-s-sonarsource:
```suggestion
Several functions in the standard C libraries, such as "size_t strlen(const char *s)" and "size_t wcslen( const wchar_t *str )", process strings under the assumption that they are null-terminated byte strings. +

For example, to process the string "example", this string must be stored in memory as "example\0". If this is not the case and the string lacks a null byte at the end, the function iterates in memory until it finds the null byte of an adjacent string or a segmentation error occurs.

Therefore, string parameters must end with a proper null byte. The absence of this particular byte can lead to security vulnerabilities that allow, for example, access to sensitive data or the execution of arbitrary code.
```

loris-s-sonarsource: I can't make a suggestion for the message because of Github, but I also suggest that we clarify the message and change

```
Make sure use of "strlen" is safe here.
```

To 

```
Make sure that the parameter of the strlen function is a null-terminated string here.
```

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,28 @@
...
-* There is a possibility that the string is not correctly null-terminated
+* There is a possibility that the pointer is null.
+* There is a possibility that the string is not correctly null-terminated.
 
 There is a risk if you answered yes to any of those questions.
 
 
 == Recommended Secure Coding Practices
 
 * C11 provides, in its annex K, the ``++strlen_s++`` and the ``++wcslen_s++`` that were designed as safer alternative to ``++strlen++`` and ``++wcslen++``. It's not recommended to use them in all circumstances, because they introduce a runtime overhead and require to write more code for error handling, but they perform checks that will limit the consequences of calling the function with bad arguments.
```
loris-s-sonarsource, line 17: Here is a trimmed-down version of this paragraph:

```suggestion
* Use safer functions such as C11 ``++strlen_s++`` and ``++wcslen_s++``, as they perform checks that handle typical programming errors.
+
Note, however, that they cause runtime overhead and require writing more code for error handling.
```

loris-s-sonarsource: Note: The "+" is a list continuation ([source](https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#list-continuation))

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,28 @@
...
-* Even if your compiler does not exactly support annex K, you probably have access to similar functions
-* If you are writing {cpp} code, using ``++std::string++`` to manipulate strings is much simpler and less error-prone
+* Even if your compiler does not exactly support annex K, you probably have access to similar functions.
```
loris-s-sonarsource, line 18: I think that this suggestion is unclear, we should be clear about which similar functions should be called. Or at least give guidelines on how to chose such a function

loris-s-sonarsource: Do you know what the writer had in mind at the time of writing?

rules/S5813/cfamily/rule.adoc
marco-antognini-sonarsource: There's indeed room for improvements, and thanks for the suggestions. Now, putting my "nitpicking" hat on... 😈 

"Byte" is only loosely defined in C, e.g. C11 §3.6 reads

> **byte**
> addressable unit of data storage large enough to hold any member of the basic character set of the execution environment

so I would refrain from using it. Here, "character" seems more appropriate as it's more focused/tailored to the domain of this rule. This also avoids talking about the difference between `char` and `wchar_t`.

> to process the string "example", this string must be stored in memory as "example\0".

This is only true if it is processed with `strlen`. The way the 1st sentence is phrased doesn't necessarily imply we focus on that. A small tweak is required -- but I'd need to think more to suggest a specific fix.

> the function iterates in memory until it finds the null byte of an adjacent string or a segmentation error occurs.

or nasal demons fly out of your nose. This is definitely nitpicking but anything could happen at this stage, so it may be worth modifying the sentence to say, those are only examples. Or the sentence can be changed even more to keep closer to your original goal: strip out unnecessary noise in the rule description.

This makes me think -- there are several rules related to C-strings. It may be worth having a good, beginner-level explanation of C-strings, how they work and common pitfalls in one of these rules and refer to that from the other rules. This of course requires a larger cleaner.

The other thing that would be interesting to know to move forward is: what exactly was puzzling the customer in the first place. Knowing this would ensure we actually address it without making the situation worst. 👼

rules/S5813/cfamily/rule.adoc
marco-antognini-sonarsource: > as they perform checks 

When reading this, the first question that arises in my mind is: what are they? True, this was already the case beforehand. I wonder how we could clarify this, or avoid the issue.

rules/S5813/cfamily/rule.adoc
marco-antognini-sonarsource: This is only a guess, but maybe functions such as `strnlen` that are defined in POSIX or Windows CRT?

* https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strnlen-strnlen-s?view=msvc-170
* https://man7.org/linux/man-pages/man3/strnlen.3p.html

rules/S5813/cfamily/rule.adoc
loris-s-sonarsource:
```suggestion
* Use safer functions such as C11 ``++strlen_s++`` and ``++wcslen_s++``: Both perform checks that handle typical programming errors.
+
Note, however, that they cause runtime overhead and require writing more code for error handling.
```

rules/S5813/cfamily/rule.adoc
loris-s-sonarsource: Yeah, I think that the customer issue in the first place was maybe first of all with the type typo in the text, and the rest is not really clear.

Here is a new suggestion to avoid nitpicking customer feedback haha: 

```suggestion
Several functions in the standard C libraries, such as ``++size_t strlen(const char *s)++`` and ``++size_t wcslen(const wchar_t *str)++``, process strings under the assumption that they are null-terminated strings. +

For example, to process the string "example", this string must be stored in memory as "example\0". If this is not the case and the string is missing a null character at the end, the function will continue to iterate in memory and the behavior of the code will become unstable.

Therefore, string parameters must end with a proper null character. The absence of this particular character can lead to security vulnerabilities that allow, for example, access to sensitive data or the execution of arbitrary code.
```

loris-s-sonarsource: > This is only true if it is processed with strlen. The way the 1st sentence is phrased doesn't necessarily imply we focus on that. A small tweak is required -- but I'd need to think more to suggest a specific fix.

I don't think I understand what you mean, since we specifically write "null-terminated string" in the first paragraph

rules/S5813/cfamily/rule.adoc
loris-s-sonarsource: Or : 

```suggestion
* Use safer functions that handle typical programming errors, such as C11 ``++strlen_s++`` and ``++wcslen_s++``.
+
Note, however, that they cause runtime overhead and require writing more code for error handling.
```

loris-s-sonarsource: Here is a variant, but may lack clarity on "safer":

```suggestion
* Use safer functions. The functions ''++strlen_s++'' and ''++wcslen_s++'' from the C11 revision (Appendix K) handle typical programming errors, but note that they cause run-time overhead and require more code for error handling.
```

rules/S5813/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks for the suggestion. I took part of them as is but overall I modified the description a bit more -- with fresh eyes, it seems the flow in the description was a bit strange.

Let me know what you think.

marco-antognini-sonarsource: > For the first paragraph, do you know if it is expected from the reader to know what a wide-character is?

I would say, if they are using `wcslen` they should know what a wide-character (`wchar_t`) is. If they don't, then it shouldn't prevent them from understanding the rest of the description, hopefully.

loris-s-sonarsource: noted! thanks :)

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,37 @@
-In C, a string is just a buffer of characters, usually using the ``++null++`` character as a sentinel for the end of the string. Which mean that the developer has to be aware of low level details such as buffer sizes, having an extra character to store the final null character... Doing that correctly and consistently is notoriously difficult, and any error can lead to a security vulnerability, for instance giving access to sensitive data, or allowing arbitrary code execution.
+The function `size_t strlen(const char *s)` measures the length of the string `s` (excluding the final null character). +
+The function `size_t wcslen(const wchar_t *s)` does the same for wide characters, and should be used with the same guidelines.
 
+Similarly to many other functions in the standard C libraries,
+`strlen` and `wcslen` process strings under the assumption that they are not null pointers.
```
loic-joly-sonarsource, line 5: Since a null pointer is not a string, I was disturbed by the wording

```suggestion
`strlen` and `wcslen` assume that their argument is not a null pointer.
```

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,37 @@
...
 
-The function ``++char *strlen(char * restrict s);++`` measures the length of the string ``++s++`` (excluding the final null character). The ``++wcslen++`` does the same for wide characters, and should be used with the same guidelines.
+Additionally, they expect the strings to be null-terminated.
+For example, the 5-letter string "abcde" must be stored in memory as "abcde\0" (i.e. using 6 characters) to be processed correctly.
+When a string is missing the null character at the end, these functions may continue to iterate in memory and the behavior of the program is undefined.
```
loic-joly-sonarsource, line 9: Not sure about this one. The initial wording felt strange, but I don"t know why...
```suggestion
When a string is missing the null character at the end, these functions will iterate past the end of the buffer, which is undefined behavior.
```

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,37 @@
...
 
 There is a risk if you answered yes to any of those questions.
 
 
 == Recommended Secure Coding Practices
 
-* C11 provides, in its annex K, the ``++strlen_s++`` and the ``++wcslen_s++`` that were designed as safer alternative to ``++strlen++`` and ``++wcslen++``. It's not recommended to use them in all circumstances, because they introduce a runtime overhead and require to write more code for error handling, but they perform checks that will limit the consequences of calling the function with bad arguments.
-* Even if your compiler does not exactly support annex K, you probably have access to similar functions
-* If you are writing {cpp} code, using ``++std::string++`` to manipulate strings is much simpler and less error-prone
+* Use safer functions. The C11 functions `strlen_s` and `wcslen_s` from annex K handle typical programming errors.
```
loic-joly-sonarsource, line 24: I think it's better if the note is on a different line.
```suggestion
* Use safer functions. The C11 functions `strlen_s` and `wcslen_s` from annex K handle typical programming errors. +
```

rules/S5813/cfamily/rule.adoc
```diff
@@ -1,28 +1,37 @@
...
+  Note, however, that they have a runtime overhead and require more code for error handling and therefore may not be suited to every case.
```
loic-joly-sonarsource, line 25:
```suggestion
  Note, however, that they have a runtime overhead and require more code for error handling and therefore are not suited to every case.
```
A significant part of a C community feels strongly against using those functions (which is why the annex is optional and is not consistently implemented), so I guess we can use stronger wording there.

marco-antognini-sonarsource (commented):
Thanks for the suggestions, @loic-joly-sonarsource. They look good.

loris-s-sonarsource:
Hey @marco-antognini-sonarsource, I suggest we drop the rest of my reviews because I think that they didn't target the initial idea of the PR and are hijacking it. This way, we can wrap this up and close this PR for good :)

PR #746, opened 2022-01-25
RULEAPI-727: Add interactions tests

arseniy-sonar:


frontend/src/__tests__/SearchPage.test.tsx
```diff
@@ -30,18 +31,166 @@ beforeEach(() => {
...
+    const { findByText, asFragment } = renderDefaultSearchPage();
     await findByText(/rules found/i);
     expect(asFragment()).toMatchSnapshot();
 });
 
+test('narrows search by title', async () => {
+    const { queryByText, queryByTestId, getByRole } = renderDefaultSearchPage();
+
+    // Finish rendering after fetching all the data
+    await waitFor(() => fetchMocker.finished());
```
frederic-tingaud-sonarsource, line 64: I think we could gain in readability if this waitFor and the two expects were wrapped in renderDefaultSearchPage

frederic-tingaud-sonarsource (approved):
It looks a lot more clean and robust than what I expected. In the end, I think those will be a lot more stable than the snapshots!
We might have the opportunity for a factorization of the waiting step before each search page display, but otherwise it looks very good.

PR #751, opened 2022-01-26
RULEAPI-732: Don't use chips for the status titles in search page

frederic-tingaud-sonarsource:
I'm not sure how much I love the borders. Don't hesitate to tell me if you think it's too much. 

Before/After

Busy Rule
before1
after1

Multiple results
before2
after2

Targeted tab title
before3
after3


marco-antognini-sonarsource (commented):
(Haven't looked at the impl yet.)

> I'm not sure how much I love the borders. Don't hesitate to tell me if you think it's too much.

Yeah, I think this adds quite some visual/cognitive load on the page.

I would personally remove both horizontal and vertical borders.
image





Possibly also remove the horizontal paddings on the outside to preserve the alignment on the left.
image

WDYT?

frederic-tingaud-sonarsource:
If that works with you, I'd like to go with this version:
afterthin

I know you prefer the one without lines at all, but I think it brings some readability when there are multiple lines of languages without being super aggressive like the first version.

marco-antognini-sonarsource (approved):
Sure, let's not bikeshed too much on this -- in the end I'm sure many will like it too! :-)

PR #756, opened 2022-01-27
RULEAPI-729: Reduce title fonts to make the page more readable

frederic-tingaud-sonarsource:
Before / After
beforerule
afterrule


frontend/src/RulePage.tsx
```diff
@@ -15,15 +15,34 @@ import { RuleMetadata } from './types';
...
 const useStyles = makeStyles((theme) => ({
+  '@global': {
+    h1: {
+      fontSize: '1.6rem',
+      fontWeight: 500,
+      marginTop: theme.spacing(3),
+      marginBottom: theme.spacing(3)
+    },
+    h2: {
+      color: '#720D0C',
```
marco-antognini-sonarsource, line 27: I think red-ish color might confuse people a bit (at least myself): it may be interpreted as an error. Maybe keeping it black would be more conservative.

frontend/src/RulePage.tsx
```diff
@@ -15,15 +15,34 @@ import { RuleMetadata } from './types';
...
+      fontSize: '1.2rem'
+    },
+    h3: {
+      fontSize: '1rem',
+      color: '#65666A'
```
marco-antognini-sonarsource, line 32: Same here, I would avoid "fancy" colors and keep text black.

frontend/src/RulePage.tsx
```diff
@@ -15,15 +15,34 @@ import { RuleMetadata } from './types';
 
 import './hljs-humanoid-light.css';
 
-
 const useStyles = makeStyles((theme) => ({
+  '@global': {
+    h1: {
+      fontSize: '1.6rem',
```
marco-antognini-sonarsource, line 21: Given this font size reduction, I wonder if the rule id at the top of the page should also be reduced. WDYT?

marco-antognini-sonarsource (commented):
These comments are mainly about personal preferences, so not to be taken as hard requests for changes.

frontend/src/RulePage.tsx
frederic-tingaud-sonarsource: Because it is very short (5 characters max), I am not too bothered by the current rule name size.
Here is how it would look with a smaller rule name. I couldn't explain why, but I'm not a huge fan: 
image

frontend/src/RulePage.tsx
frederic-tingaud-sonarsource: Changed to blueish

frontend/src/RulePage.tsx
frederic-tingaud-sonarsource: Changed to dark blue

marco-antognini-sonarsource: Looks great in blue!

frontend/src/RulePage.tsx
marco-antognini-sonarsource: After seeing the screenshot, I agree. Let's keep it as-is.

PR #757, opened 2022-01-28
RULEAPI-682: Fix CI failure when no rules have changed 

marco-antognini-sonarsource:
Ensure array is correctly read.
Ensure errors in subprocesses are detected and abort the script.

`shopt -s lastpipe` is used to avoid the following complex pattern:

    mapfile -t array < <(command-that-may-fail)
    pid=$!
    wait $pid || exit 1


marco-antognini-sonarsource:
I've locally reproduced the previous failure https://cirrus-ci.com/task/6396593430593536?logs=metadata_tests#L5 and tested four scenarios:
 - no changes since master
 - one modified rule
 - two modified rules
 - modified rspec-tools
All succeeded.

ci/validate_metadata.sh
```diff
@@ -1,21 +1,18 @@
 #!/bin/bash
 set -ueo pipefail
+shopt -s lastpipe # To pipe command result into mapfile and have the array variable available in the main shell process.
 
 git fetch --quiet "${CIRRUS_DEFAULT_ORIGIN:-origin}" "${CIRRUS_DEFAULT_BRANCH:-master}"
 base="$(git merge-base FETCH_HEAD HEAD)"
 echo "Comparing against the merge-base: ${base}"
 if ! git diff --name-only --exit-code "${base}" -- rspec-tools/
```
frederic-tingaud-sonarsource, line 8: We might need to check the `ci` folder too.

frederic-tingaud-sonarsource (approved):
I don't understand the lastpipe trick and what it solves, but otherwise, it looks good to me.

ci/validate_metadata.sh
marco-antognini-sonarsource: If we do that, it would be harder to test changes of this script because it would always take the same branch. So I'm not entirely convinced but it's worth thinking more about this. I've updated https://jira.sonarsource.com/browse/RULEAPI-736 to cover this point.

marco-antognini-sonarsource:
> I don't understand the lastpipe trick and what it solves, but otherwise, it looks good to me.

Thanks for the review. For future readers and reference (including myself 😅), here is the full story about `lastpipe`.

Firstly, we have this code to list into an array the rules we care about
```
mapfile -t affected_rules < <(some-command)
```
Even with `set -e`, if `some-command` fails it won't stop the script. To catch the error, we have to write more code, namely:

```
mapfile -t affected_rules < <(some-command)
pid=$!
wait $pid
```
Here, `set -e` will ensure the script stops at `wait $pid` if the process substitution `<(some-command)` failed. But that's ugly to write IMO.

So instead of using process substitution, we can (try to) pipe the result into `mapfile`:
```
some-command | mapfile -t affected_rules
```

With `set -o pipefail`, the script will stop if `some-command` fails.

But this won't work. Shellcheck actually warns about this:
> Modification of affected_rules is local (to subshell caused by pipeline).
https://github.com/koalaman/shellcheck/wiki/SC2030

It's like `bash -c 'true | x=1; echo \"$x\"'`, which prints `""`.

With `shopt -s lastpipe`, it works as expected, finally!

PR #761, opened 2022-01-31
Modify rule S4423[terraform]:Add SSL policy for Google Cloud load balancers

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-345](https://jira.sonarsource.com/browse/SONARIAC-345)
Preview: [S4423](https://sonarsource.github.io/rspec/#/rspec/S4423/terraform)

### DIFF
--- a/rules/S4423/terraform/highlighting.adoc
+++ b/rules/S4423/terraform/highlighting.adoc
@@ -1,5 +1,7 @@
 === Highlighting
 
+==== AWS
+
 For `aws_elasticsearch_domain`:
 
 * Highlight `tls_security_policy` if it is specified but has the wrong value
@@ -16,3 +18,10 @@ For `aws_apigatewayv2_domain_name`:
 * Highlight `security_policy` if it is specified but has the wrong value
 * Highlight `domain_name_configuration` if it is specified but does not contain `security_policy`
 * Highlight resource if `domain_name_configuration` is not specified at all
+
+==== GCP
+
+For `google_compute_ssl_policy`:
+
+* Highlight `min_tls_version` if it is specified but has the wrong value
+* Highlight resource if `min_tls_version` is not specified at all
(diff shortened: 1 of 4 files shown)


rules/S4423/terraform/highlighting.adoc
```diff
@@ -16,3 +18,10 @@ For `aws_apigatewayv2_domain_name`:
 * Highlight `security_policy` if it is specified but has the wrong value
 * Highlight `domain_name_configuration` if it is specified but does not contain `security_policy`
 * Highlight resource if `domain_name_configuration` is not specified at all
+
+==== GCP
+
+For `google_compute_ssl_policy`:
+
+* Highlight `min_tls_version` if it is specified but has the wrong value
+* Highlight resource if `min_tls_version` is not specified at all
```
eric-therond-sonarsource, line 27: It depends on what you will target for `profile` argument:
for now I would say that when `min_tls_version` is missing but `profile` equals to `RESTRICTED` it should be compliant.

eric-therond-sonarsource (approved):
LGTM
I just let a comment regarding the `profile` argument that you will target as you said this morning, to not forget that the good minimum tls version seems to be applied when the `RESTRICTED` profile is used.

rules/S4423/terraform/highlighting.adoc
pierre-loup-tristant-sonarsource: I did not think about this case. I tested and I confim that we should not raise on this case. Old TLS versions will still be offered but in practice the handshake will always fail as none of the TLSv1.1 or TLSv1 cipher suites can be used.

pierre-loup-tristant-sonarsource:
Thanks for your review. I will adjust the ticket and add an example in security-expected-issues.
I added custom issue messages after your review. Cloud you have a look and let me know what you think?
Thank you.

eric-therond-sonarsource:
LGTM

PR #770, opened 2022-02-01
RULEAPI-633: Autolink RSPECs

marco-antognini-sonarsource:
Demo at https://marco-antognini-sonarsource.github.io/rspec/#/rspec/S100 (S100 is a good candidate, like many other rules).


frontend/src/RulePage.tsx
```diff
@@ -323,13 +342,21 @@ function useDescription(metadata: PageMetadata, ruleid: string, language: string
...
-    return <div>
+    const description = <div>
       <div dangerouslySetInnerHTML={{ __html: descHTML }} />
       <hr />
       <a href={editOnGithubUrl}>Edit on Github</a><br />
       <hr />
       <Highlight className='json'>{metadata.jsonString}</Highlight>
     </div>;
+
+    // This should be moved somewhere else... here it is not executed when the page is loaded.
```
marco-antognini-sonarsource, line 353: I wrote something like this before 500a488b041c9d8e198c229af4375ffa6d2602d8 and it "worked" -- i.e. the links href were populated as expected. Now it doesn't work anymore. This is way outside my zone of comfort, so probably a completely different solution is required anyway.

@frederic-tingaud-sonarsource do you have any advice on how I can modify the description when it's loaded?

frederic-tingaud-sonarsource: I think when you run this, the page is not updated yet. So document.getElementByClassName gets nothing.
I was looking at this library: https://www.npmjs.com/package/html-react-parser
If we use that instead of dangerouslySetInnerHTML, you could have the DOM of descHTML before we render it. It would also help for code highlighting.

marco-antognini-sonarsource: Thanks. Looks promising indeed. I'll try to give it a shot.

frontend/src/RulePage.tsx
```diff
@@ -314,7 +315,25 @@ function usePageMetadata(ruleid: string, language: string, classes: UsedStyles):
   };
 }
 
-function useDescription(metadata: PageMetadata, ruleid: string, language: string) {
+function doesRuleExistForLanguage(rspecId: string, language: string) {
+  // FIXME implement me.
```
marco-antognini-sonarsource, line 319: I need to create a follow-up ticket for that -- I'd rather have a basic feature now than extend the scope of investigation further -- and update the comment.

frontend/src/RulePage.tsx
marco-antognini-sonarsource: Your work in #771 was exactly what I needed, thanks!

As it stands, this PR will conflict with yours. I don't mind in which order they are handled. I suppose, given the size of yours, it's best to go ahead with #771 first.

frontend/src/RulePage.tsx
frederic-tingaud-sonarsource: My issue with this FIXME, is that it hints in a direction that might not be the right one. I think the check will need to be asynchronous, so I don't expect `doesRuleExistForLanguage` to be used as-is.
Perhaps it would be simpler to put the FIXME in `getRspecPath` and not use `language` at all.

marco-antognini-sonarsource: That's a very good point. I've opened the ticked and removed unnecessary code.

frontend/src/RulePage.tsx
arseniy-sonar: Beware that #771 is contingent on releasing rule-api that ignores the changes from #771, so it might take a while to arrive.

marco-antognini-sonarsource: I don't mind waiting a bit, given the size of #771 it's probably the best thing to do in terms of collective work.

marco-antognini-sonarsource:
(requesting review to be ready once #771 is merged and speed up the process a tiny bit.)

frontend/src/deployment/description.ts
```diff
@@ -7,6 +7,21 @@ import { logger } from './deploymentLogger';
...
+function generateAutoRspecLinks(html: string) {
+  // Insert placeholder links for SXXX or RSPEC-XXX to the appropriate description page.
+  //
+  // The web application is responsible for providing the target link.
+  // The link will depend on whether the default page is viewed, and in that case it will
+  // point to the default page for the target rules, or whether a language-specific page
+  // is viewed.
+  // The distinction cannot be made when generating the HTML description because the description
+  // for the first language is also used as the default description.
+  return html.replace(
```
frederic-tingaud-sonarsource, line 19: I don't think we can do it blindly like that, because there are rules mentioning RSPEC ids in their code sections. See for example https://sonarsource.github.io/rspec/#/rspec/S3470

frederic-tingaud-sonarsource (changes requested):
I see an issue with the code sections.

frontend/src/deployment/description.ts
marco-antognini-sonarsource: Seems to work fine: https://marco-antognini-sonarsource.github.io/rspec/#/rspec/S3470/cfamily (aside from the fact this implemented rule refers to a rule that isn't implemented yet...)

What issue could we expect?

marco-antognini-sonarsource: I've included #771 and fixed the merge conflicts. But I'll need to investigate more. It seems to work fine on GitHub (cf my fork), but when running locally I get some strange issues.

1. go to http://localhost:3000/rspec#/rspec/S3470/cfamily
2. click on a link to S5963 (the one in the code snippet is not clickable, but the one under Comments And Links is)
3. in the console I get errors such as:
```
react-dom.development.js:10301 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```
but if I reload or copy/paste the link, it works...

marco-antognini-sonarsource: I still don't know why locally I get errors but I don't on my fork's github pages...

However, I've made the following observation using the same testing protocol as above:
* removing the `data-rspec-id` attributes from links in `frontend/public/rules/S5963/default-description.html` doesn't change a thing (so I restored them back)
* removing the `data-lang` attributes from `<code>` removes the errors.

I've also checked, there are no nested links in `code` on S5963.

In other words, the issue seems to happen when triggering the links on S3470 but because of `<Highlight>`.

Looking at the stacktraces, I'm not sure how to debug this... :/

frontend/src/deployment/__tests__/resources/metadata/S1000/cfamily-description.html
```diff
@@ -49,6 +49,7 @@ <h2 id="_noncompliant_code_example">Noncompliant Code Example</h2>
   fn_a();                  // Is expected to initialize "x" to 42
   if (x == 42)             // But does not, as there are 2 distinct "x" variables
   {
+     // I am an auto link in some source code: <a data-rspec-id="S987" class="rspec-auto-link">S987</a>.
```
frederic-tingaud-sonarsource, line 52: We want the opposite. Inside a <pre> section, this link should not be generated as it will not work.

frontend/src/deployment/description.ts
frederic-tingaud-sonarsource: The problems I see: Inside a `<pre>` section, we shouldn't do the change. Another problem I see when clicking on your link is that you should also disable it inside other urls like Evgeny's comment on S3470.

marco-antognini-sonarsource: Okay, so I followed your suggestion and it does indeed solve the problem. I believe it was a bug in the highlighting component because `<a>` are allowed in `<pre>` according to the [spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element) (as any other "Phrasing content"), but solving that would be more tedious than not having the links. (Let's keep that pandora box closed for now.)

The new implementation also doesn't make replacements within existing URLs.

frontend/src/deployment/__tests__/resources/metadata/S1000/cfamily-description.html
marco-antognini-sonarsource: done.

marco-antognini-sonarsource:
The fork preview was updated at https://marco-antognini-sonarsource.github.io/rspec/#/rspec/S3470

frederic-tingaud-sonarsource (approved):
That looks pretty cool and robust to me!

PR #771, opened 2022-02-01
RULEAPI-661: syntax coloring

frederic-tingaud-sonarsource:
We need to wrap code with highlight.js somehow, because asciidoctor will not do it for us.


arseniy-sonar:
The current rule-api version generates additional `<code>` tags for the rules with source language set. To make sure products do not break, let's merge this PR only once we release rule-api version that strips the `<code>` tags (so the generated HTML remains identical to the one that is already in the analyzers).

marco-antognini-sonarsource:
Due to GitHub limitations, it's complicated to comment on this large change. Could you rewrite the history of your branch to have two commits: the first that modifies `rules/` and the second for any other changes? That way I should be able to comment inline on the non-content changes.

Could you also share the script you used to modify `rules/`? That would be easier to review than the large change itself.

Having run this locally, I think it's a great step forward! (I'm not used to that color scheme but that's a detail not worth bikeshedding over.)

I've noticed however that some code snippets were not highlighted: an example is S2223 for csharp. I guess it's because there are two snippets in those sections.

frederic-tingaud-sonarsource:
Very good points from both of you!
The script I used is https://github.com/SonarSource/languages-experimental-tooling/blob/master/personal/fred-tingaud/rspec/source-language.py and it cannot handle more than one code block at the moment, which is definitely an issue.
I rebase-squashed as asked, @marco-antognini-sonarsource

frontend/src/deployment/__tests__/description.test.ts
```diff
@@ -53,8 +53,8 @@ describe('description generation', () => {
   });
 
   expect.extend({
-    toBeValid(received, validator, expectedPath) {
-      if (validator === received) {
+    toBeSameAsFile(received, validator, expectedPath) {
+      if (validator.replaceAll('\r\n', '\n') === received.replaceAll('\r\n', '\n')) {
```
marco-antognini-sonarsource, line 57: There are some test failures due to this test. https://cirrus-ci.com/task/4726331920351232?logs=tests#L39

I guess it doesn't matter if this is removed. 👼

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -47,3 +47,32 @@ def validate_parameters(rule_language: LanguageSpecificRule):
...
+                         'python': 'python',
+                         'ruby': 'ruby',
+                         'rust': 'rust',
+                         'scala': 'scala',
+                         'swift': 'swift',
+                         'tsql': 'sql',
+                         'xml': 'xml'
+                         }
+
+def validate_source_language(rule_language: LanguageSpecificRule):
```
marco-antognini-sonarsource, line 69: This check works well, very well in fact!

https://api.cirrus-ci.com/v1/task/5289281873772544/logs/asciidoc_tests.log

I only looked at the first issue. It seems due to the fact the code snippet was not handled correctly by your script: https://github.com/SonarSource/rspec/blob/75b253f2d975b36c7dfb7951500016f8d09fff57/rules/S1005/noncompliant.adoc

frederic-tingaud-sonarsource: Sadly, it is a way bigger problem behind:
We don't want to have syntax coloring at the root of the rule (they are here because they are shared between languages). But our current python test runs after solving the includes, so we can't know whether the tag is missing because it's referencing something at the root or because it was forgotten.

arseniy-sonar: Why not set the "unknown" language for the code snippets in the rule root folders?

frontend/src/deployment/__tests__/description.test.ts
```diff
@@ -51,4 +51,42 @@ describe('description generation', () => {
...
+        const receivedPath = path.join(path.dirname(expectedPath), 'received-' + path.basename(expectedPath));
+        fs.writeFileSync(receivedPath, received);
+        return {
+          message: () => `Identity check failed on ${expectedPath}.\nReceived file saved in ${receivedPath}`,
+          pass: false
+        };
+      }
+    }
+  });
+  test('generates metadata for active rules', () => {
```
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: The test is badly named, which I'll fix and is not exactly necessary for this PR, but I realized the PR was not triggering enough tests and that was because the folder "metadata" was not in line with our adocs. Thus a test to make sure it isn't so.
I also added https://jira.sonarsource.com/browse/RULEAPI-740 to go farther in that direction.

rspec-tools/rspec_tools/validation/description.py
arseniy-sonar: As clarified in slack, I meant the "text" language, i.e. our go-to language if we don't know which one it is - just to make sure every code snippet has some source language specified and no snippet is forgotten.

frederic-tingaud-sonarsource: fixed in accordance!

frontend/src/deployment/__tests__/description.test.ts
frederic-tingaud-sonarsource: our node version on the CI was old. Fixed

marco-antognini-sonarsource (commented):
It seems some sections were not updated: https://api.cirrus-ci.com/v1/task/5844615575109632/logs/asciidoc_tests.log

I see also `-- Exceptions` sections were not considered. I'm happy if this is done as a separate task to avoid exploding the size of this one.

frontend/src/deployment/__tests__/description.test.ts
```diff
@@ -53,10 +53,10 @@ describe('description generation', () => {
   });
 
   expect.extend({
-    toBeSameAsFile(received, validator, expectedPath) {
-      if (validator.replaceAll('\r\n', '\n') === received.replaceAll('\r\n', '\n')) {
+    toBeSameAsFile(received, expected, expectedPath) {
+      if (expected.replaceAll('\r\n', '\n') === received.replaceAll('\r\n', '\n')) {
         return {
-          message: () => `Identity check failed on ${expectedPath}.\nExpected:\n${validator}\n\nReceived:\n${received}`,
+          message: () => `Identity check failed on ${expectedPath}.\nExpected:\n${expected}\n\nReceived:\n${received}`,
```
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: I was confused first too. It's the failure when using not: `expect(a).not.toBeSameAsFile(b)`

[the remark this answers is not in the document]
marco-antognini-sonarsource: Ho, I see! Thanks for the explanation. 👍

marco-antognini-sonarsource (dismissed):
I think it should be good now. 🤞

frontend/src/__tests__/RulePage.test.tsx
```diff
@@ -35,7 +35,7 @@ beforeEach(() => {
         'CSH' : {'S3457': 'c#1'},
         'C': {'S100': 'c1', 'S234': {'since': 'c2', 'until': 'c3'}}}
     };
-    jest.spyOn(global, 'fetch').mockImplementation(fetchMock(mockUrls));
+    jest.spyOn(global, 'fetch').mockImplementation(fetchMock(mockUrls) as jest.Mocked<typeof fetch>);
```
arseniy-sonar, line 38: Would it make sense to declare `fetchMock` with this type (`jest.Mocked<typeof fetch>`) - if that's syntactically possible?

frontend/src/__tests__/SearchPage.test.tsx
```diff
@@ -37,7 +37,7 @@ function genMockUrls() {
 let fetchMocker = fetchMockObject(genMockUrls());
 
 beforeEach(() => {
-    jest.spyOn(global, 'fetch').mockImplementation(fetchMocker.mock);
+    jest.spyOn(global, 'fetch').mockImplementation(fetchMocker.mock as jest.Mocked<typeof fetch>);
```
arseniy-sonar, line 33: Would it make sense to declare `fetchMocker.mock` with this type (jest.Mocked<typeof fetch>) - if that's syntactically possible?

frontend/src/__tests__/RulePage.test.tsx
frederic-tingaud-sonarsource: I tried, but I get errors in utils.ts when doing it.

frontend/src/__tests__/SearchPage.test.tsx
frederic-tingaud-sonarsource: I tried, but I get errors in utils.ts when doing it.

frontend/src/deployment/__tests__/description.test.ts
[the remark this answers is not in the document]
arseniy-sonar: I am also confused. I think at least an explanation comment is in order.

frontend/src/deployment/__tests__/description.test.ts
```diff
@@ -51,4 +51,42 @@ describe('description generation', () => {
...
+      expect(rules.length).toEqual(4);
+      let treated = 0;
+      rules.forEach(ruleDir => {
+        const languages = fs.readdirSync(`${dstPath}/${ruleDir}`);
+        expect(languages.length).toBeGreaterThanOrEqual(1);
+        languages.forEach(file => {
+          const actual = fs.readFileSync(`${dstPath}/${ruleDir}/${file}`).toString();
+          const expectedPath = path.join(__dirname, 'resources', 'metadata', ruleDir, file);
+          const expected = fs.readFileSync(expectedPath).toString();
+          expect(actual).toBeValid(expected, expectedPath);
```
arseniy-sonar: It is odd to have to pass both `expectedPath` and the `expected` value read from the file by that path. Why do you not read the file in the check itself?

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -65,14 +65,21 @@ def validate_parameters(rule_language: LanguageSpecificRule):
                          'tsql': 'sql',
                          'xml': 'xml'
                          }
+def highlight_name(rule_language: LanguageSpecificRule):
+  if (rule_language.language in HIGHLIGHTED_LANGUAGES):
+    return HIGHLIGHTED_LANGUAGES[rule_language.language]
+  return rule_language.language
```
arseniy-sonar, line 71: To be explicit about the acceptable languages, I think it is better to extend `HIGHLIGHTED_LANGUAGES` to all the supported languages with identity mappings e.g., `'kotlin': 'kotlin'`. This way if it is not in the map you know it is not supported.
This will also enable us to make sure we properly map all the supported languages as listed in the https://github.com/SonarSource/rspec/blob/master/supported_languages.adoc

As I understand the current contents of `HIGHLIGHTED_LANGUAGES` correspond to the source languages supported by the highlighting engine used in the frontend page. It might be useful to mark with a comment which language will be properly highlighted, and which is not supported.

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -65,14 +65,21 @@ def validate_parameters(rule_language: LanguageSpecificRule):
...
-          raise RuleValidationError(f'''Rule {rule_language.id} has non highlighted code example in section "{name}".
-Use [source,{HIGHLIGHTED_LANGUAGES[rule_language.language]}] or [source,text] before the opening '----'.''')
+  for h2 in descr.findAll('h2'):
+    name = h2.text.strip()
+    if name.startswith('Compliant') or name.startswith('Noncompliant'):
+      pre = h2.parent.pre
+      if not pre or not pre.has_attr('class') or pre['class'][0] != u'highlight' or not pre.code or not pre.code.has_attr('data-lang'):
+        raise RuleValidationError(f'''Rule {rule_language.id} has non highlighted code example in section "{name}".
+Use [source,{highlight_name(rule_language)}] or [source,text] before the opening '----'.''')
+      elif pre.code['data-lang'] != highlight_name(rule_language) and pre.code['data-lang'] != 'text':
```
arseniy-sonar, line 83: IMO this is too strict: a rule description for language A might include code snippets for language B (for comparison, or as part of an ecosystem, e.g. a rule about Android/Java would mention the XML config). I think it is enough to check that the annoted source language is from the fixed set of languages.

arseniy-sonar (changes requested):
Looks good, and thank you for this titanic effort!
In addition to the comments that I added outside of this review (due to the sheer amount of modified files, I had to write comments on specific commits):
- Please update the new rule templates to use a language annotation: https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_template/single_language/language_specific/rule.adoc and https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_template/multi_language/language_specific/rule.adoc
- Please also mention the source languages for code snippets in the documentation (in particular the new requirement to specify the source language for every code snippet, and where to find the list of languages to choose from): https://github.com/SonarSource/rspec/blob/master/docs/description.adoc

rspec-tools/rspec_tools/validation/description.py
marco-antognini-sonarsource: That's something I've missed when reviewing it earlier. As it stands, we are also missing a test: it should be possible to use `C` highlighting for `cfamily` rules. Currently, CI will fail with a validation error.

frontend/src/deployment/__tests__/description.test.ts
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: done

rspec-tools/rspec_tools/validation/description.py
frederic-tingaud-sonarsource: That makes sense. Done

rspec-tools/rspec_tools/validation/description.py
frederic-tingaud-sonarsource: Good points from both of you. Fixed

arseniy-sonar (approved):

arseniy-sonar:
A quick poll conducted in LT slack in the past 2.5 hours indicated the 12:3 preference in favor of these changes, as opposed to highlighting with an implicit source language matching the rule target language.

A quick poll for people writing RSPECs (i.e. committers to https://github.com/SonarSource/rspec/tree/master/rules):
In the weekly I announced that a soon-to-be-merged PR will require every code snippet to have a source language explicitly specified. [@Pavel](https://sonarsource.slack.com/team/UQ3N1R7RV) challenged this change.
Since the arguments on both sides feel subjective to me, I would like to take a vote among people that write rule descriptions:
Which one do you prefer?

1. Specifying the source language explicitly for each code snippet ([source,java]) in the rule description (and have CI make sure that it is the case, with text as a backup for cases where the source language cannot be specified)
2. Having the source language assumed to be the language of the rule (the folder in the rule directory) and have the possibility to override that language ([source,java]) in the rare cases when the language does not match

Argument for (1): transparency - the rule specification is in pure asciidoc and will be rendered with proper highlighting by tools outside our ecosystem, also the two consumers of the rules we have now (rule-api, and rspec frontend page) will not need to implement the logic implied by (2).

Argument for (2): lower maintenance cost (brevity) - the rule specification is shorter - for each code snippet it contains 1 line less for the same end result (the implicit gaps being filled by our tools). Also newjoiners will not have to know about this additional detail - source language - as long as they write rules with code snippets that involve only the language the rule is specified for.

option 1: 12 votes
option 2: 3 votes

PR #772, opened 2022-02-02
Modify rule S6328: Add exceptions for methods throwing an exception

quentin-jaquier-sonarsource:
Refers to https://jira.sonarsource.com/browse/SONARJAVA-3887.


rules/S5808/java/rule.adoc
```diff
@@ -86,6 +86,10 @@ public class MyPermissionEvaluator implements PermissionEvaluator {
 } 
 ----
 
+== Exceptions
+
+No issue is reported when the method is throwing an exception as it might be used to indicate a strong decision.
```
dorian-burihabwa-sonarsource, line 91:
```suggestion
No issue is reported when the method throws an exception as it might be used to indicate a strong decision.
```

dorian-burihabwa-sonarsource (changes requested):

PR #775, opened 2022-02-02
Modify rule S128: Document fall through exception for Java

dorian-burihabwa-sonarsource:
SONARJAVA-3893


quentin-jaquier-sonarsource (approved):
LGTM

PR #777, opened 2022-02-03
Reduce amount of debug output in validate_asciidoc.sh

marco-antognini-sonarsource:
It can be hard to spot actual issues in the log with so much debug output. This PR removes most of it.

### DIFF
--- a/ci/validate_asciidoc.sh
+++ b/ci/validate_asciidoc.sh
@@ -38,7 +38,6 @@ do
     continue
   fi
   dir=${dir%*/}
-  echo "${dir##*/}"
 
   subdircount=$(find "$dir" -maxdepth 1 -type d | wc -l)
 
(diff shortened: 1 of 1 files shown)


ci/validate_asciidoc.sh
```diff
@@ -58,21 +57,17 @@ do
...
       if test -f "$RULE"; then
-        echo "$RULE exists."
         TMP_ADOC="$language/tmp.adoc"
         echo "== Description" > "$TMP_ADOC"
         cat "$RULE" >> "$TMP_ADOC"
         if asciidoctor --failure-level=WARNING -o /dev/null "$TMP_ADOC"; then
-            if asciidoctor -a rspecator-view --failure-level=WARNING -o /dev/null "$TMP_ADOC"; then
-                echo "$RULE syntax is fine"
-            else
+            if ! asciidoctor -a rspecator-view --failure-level=WARNING -o /dev/null "$TMP_ADOC"; then
```
frederic-tingaud-sonarsource, line 70: Since you are doing this, do you think it would be much work to just create the files in the loop, then run asciidoctor once on all generated files as described in https://jira.sonarsource.com/browse/RULEAPI-734?
It would make it wayyy faster

marco-antognini-sonarsource: yes, I agree but I wanted to keep this as a baby step for now. :-)

frederic-tingaud-sonarsource (approved):

frederic-tingaud-sonarsource:
We should still have one echo of `$affected_rules` to make sure we don't lose too much debugability

frederic-tingaud-sonarsource (changes requested):
We should still have one echo of `$affected_rules` to make sure we don't lose too much debugability

frederic-tingaud-sonarsource (approved):
Interestingly, it does prove that we don't have expansion of `rules/*` (inside the variable itself), as ShellCheck said.

marco-antognini-sonarsource:
> Interestingly, it does prove that we don't have expansion of `rules/*` (inside the variable itself), as ShellCheck said.

Very well observed. I tested this again locally and realized I made a mistake the other day when testing it. Sorry for the confusion. :(

PR #781, opened 2022-02-04
Use full clone for tests to enable proper new-code detection on SonarQube

arseniy-sonar:
~Excluding the file from coverage analysis for now to avoid SQ blocking PRs that do not touch the code at all. (for some reason, since this file was merged despite the failing QG on SQ, other PRs that target master are assumed to have changed this file and therefore also fail QG)~

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -88,7 +88,7 @@ validate_asciidoc_task:
     <<: *CONTAINER_DEFINITION
     dockerfile: ci/Dockerfile
     cpu: 1
-    memory: 1G
+    memory: 2G
   env:
     CIRRUS_CLONE_DEPTH: 0
   asciidoc_tests_script:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,7 +7,7 @@ env:
   SONAR_SCANNER_VERSION: 4.6.1.2450
   SONAR_SCANNER_HOME: ${HOME}/.sonar/sonar-scanner-${SONAR_SCANNER_VERSION}-linux
   PATH: ${SONAR_SCANNER_HOME}/bin:$PATH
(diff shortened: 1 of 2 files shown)


frederic-tingaud-sonarsource (changes requested):
I'm not sure about that. The other PR is failing because you merged the master in it and the analysis is on the most recent changelist and not the whole branch.
It should be possible to silence the error by squashing the PR.

arseniy-sonar:
Indeed, probably there is an issue with SQ configuration or the warning it displays (Shallow clone detected during the analysis. Some files will miss SCM information. This will affect features like auto-assignment of issues. Please configure your build to disable shallow clone.)
Will have a look after I investigate the frontend_tests hanging.

.cirrus.yml
```diff
@@ -7,7 +7,7 @@ env:
   SONAR_SCANNER_VERSION: 4.6.1.2450
   SONAR_SCANNER_HOME: ${HOME}/.sonar/sonar-scanner-${SONAR_SCANNER_VERSION}-linux
   PATH: ${SONAR_SCANNER_HOME}/bin:$PATH
-  CIRRUS_CLONE_DEPTH: 1
+  CIRRUS_CLONE_DEPTH: 0
```
marco-antognini-sonarsource, line 10: 0 is the default value, so removing the line should work too. But if you prefer to keep it for explicitness that's fine I would say.

However, the duplicated lines L82 + L93 should probably be removed.

(I'm not sure it's worth the headache to set it to `1` for jobs that currently work fine...)

marco-antognini-sonarsource: Nice fix BTW :-)

marco-antognini-sonarsource (approved):
LGTM

frederic-tingaud-sonarsource (approved):

PR #782, opened 2022-02-04
Test to accelerate validation

frederic-tingaud-sonarsource:
Run asciidoctor on all files at once instead of one by one. That allows us to considerably reduce the time it takes. On the current CI, we go from ~25 minutes to ~5 minutes to analyze all adocs.

There are two visible downsides:
* It is no longer clear which file is failing. This is mitigated by the fact that I renamed each `tmp.adoc` to contain the name of the rule and the language `tmp_S1234_cfamily.adoc`. The real problem should be handled by asciidoctor, but so far, the Issue on their side is stuck: https://github.com/asciidoctor/asciidoctor/issues/3414
* The analysis fails after the first error.

These downsides were considered reasonable because:
* For small changelists they are transparent (if you modify one rule, you know which file is touched and you probably don't have a lot of errors)
* For big changelists they bring a huge time benefit which greatly compensates the potential time loss of having to run tests locally when things go really bad.

### DIFF
--- a/ci/validate_asciidoc.sh
+++ b/ci/validate_asciidoc.sh
@@ -65,7 +65,7 @@ do
       fi
       RULE="$language/rule.adoc"
       if test -f "$RULE"; then
-        TMP_ADOC="$language/tmp.adoc"
+        TMP_ADOC="$language/tmp_$(basename "${dir}")_${language##*/}.adoc"
         echo "== Description" > "$TMP_ADOC"
         cat "$RULE" >> "$TMP_ADOC"
       else
@@ -76,7 +76,7 @@ do
 
     # Check that all adoc are included
     find "$dir" -name "*.adoc" -execdir sh -c 'grep -h "include::" "$1" | grep -v "rule.adoc" | sed "s/include::\(.*\)\[\]/\1/" | xargs -r -I@ realpath "$PWD/@"' shell {} \; > included
-    find "$dir" -name "*.adoc" ! -name 'rule.adoc' ! -name 'tmp.adoc' -exec sh -c 'realpath $1' shell {} \; > created
+    find "$dir" -name "*.adoc" ! -name 'rule.adoc' ! -name 'tmp*.adoc' -exec sh -c 'realpath $1' shell {} \; > created
     orphans=$(comm -1 -3 <(sort -u included) <(sort -u created))
     if [[ -n "$orphans" ]]; then
         printf 'ERROR: These adoc files are not included anywhere:\n-----\n%s\n-----\n' "$orphans"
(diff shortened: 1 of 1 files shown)


ci/validate_asciidoc.sh
```diff
@@ -95,6 +86,19 @@ do
   fi
 done
 
+if asciidoctor --failure-level=WARNING -o /dev/null rules/*/*/tmp*.adoc; then
+    if ! asciidoctor -a rspecator-view --failure-level=WARNING -o /dev/null rules/*/*/tmp*.adoc; then
```
arseniy-sonar: This is odd `if` chaining - one is negated but not the other. It would be easier to read for me if both conditions were not-negated, or both negated.

ci/validate_asciidoc.sh
```diff
@@ -95,6 +86,19 @@ do
...
+        echo "ERROR: malformed asciidoc files in rspecator-view"
+        exit_code=1
+    fi
+else
+  echo "ERROR: malformed asciidoc files"
+  exit_code=1
```
arseniy-sonar: I am not sure the trick with `exit_code` is necessary if you exit on the first error anyway. Since the check will get cleaned up at the end anyway, I think you don't need to remove the files yourself, you can just `exit 1`

arseniy-sonar (changes requested):

ci/validate_asciidoc.sh
frederic-tingaud-sonarsource: It was just a move of the original code. but I agree it's a bit complex to read. I changed it.

ci/validate_asciidoc.sh
frederic-tingaud-sonarsource: I removed the file removal.
I think I like that exit_code thing though, because it makes it clear where the script will exit and it allows us to keep a unique message for failure / success. If we add new checks toward the end in the future, we'll be less likely of missing them in case of earlier failure.

ci/validate_asciidoc.sh
```diff
@@ -65,27 +65,18 @@ do
       fi
       RULE="$language/rule.adoc"
       if test -f "$RULE"; then
-        TMP_ADOC="$language/tmp.adoc"
+        TMP_ADOC="$language/tmp_$(basename "${dir}")_${language##*/}.adoc"
```
arseniy-sonar, line 68: I think it is worth adding a comment here explaining the workaround, just like you do in the PR description
> the Issue on their side is stuck: https://github.com/asciidoctor/asciidoctor/issues/3414

arseniy-sonar (approved):

PR #783, opened 2022-02-04
Quick-fix for frontend testing's bad mock of fs.existsSync

marco-antognini-sonarsource:


marco-antognini-sonarsource:
The scope is intentionally very small to unblock things. We may want to inspect and update other mocks to ensure they are properly reset...

frontend/src/deployment/__tests__/pullRequestIndexing.test.ts
```diff
@@ -32,10 +32,14 @@ beforeEach(() => {
...
-    return fname.includes('rules/S');
+  jest.spyOn(fs, 'existsSync').mockImplementation((fname) => {
+    return fname.replace(/\\/g, '/').includes('rules/S');
   });
-})
+});
+
+afterAll(() => {
+  jest.restoreAllMocks();
+});
```
marco-antognini-sonarsource, line 42: This may actually not be required -- I can't find it in the doc, but it looks like `jest.restoreAllMocks()` is called from somewhere between tests. If I change the mock implementation above to return false, only the `pullRequestIndexing.test.ts` fails.

Using `jest.spyOn` may be the only thing needed, but given the lack of proof in the doc, we may want to keep this too.

frederic-tingaud-sonarsource: it makes sense. In that case, we might want to use the same pattern for the line above with Git.Clone

arseniy-sonar: What bothers me a little is the asymmetry: you install the mock in `beforeEach` and clear it in `afterAll`. Perhaps it makes no difference here since this test suite contains a single test, but I think it should be either `beforeEach`/`afterEach` or `beforeAll`/`afterAll`

marco-antognini-sonarsource: Yes to both of you -- initially I kept this as small as possible to integrate this as fast as possible and unblock other things. There may be other similar cases in different test files that work for now but may fail in the future.

arseniy-sonar (approved):

frontend/src/deployment/__tests__/pullRequestIndexing.test.ts
arseniy-sonar: I took the liberty to replace `afterAll` with `afterEach`

frederic-tingaud-sonarsource (approved):
🤞

PR #784, opened 2022-02-04
Bump memory limit for metadata validation task

arseniy-sonar:
1G is not enough for pipenv to install all the dependencies (at least the deps for #771)


arseniy-sonar:
@frederic-tingaud-sonarsource , I let you merge this PR to unblock #771 (which is stuck on [metadata validation](https://cirrus-ci.com/task/6182239800131584))

PR #789, opened 2022-02-07
Modify rule S6323: Add Python as covered language

nils-werner-sonarsource:
Relates to https://jira.sonarsource.com/browse/SONARPY-977


andrea-guarino-sonarsource:
> Relates to https://jira.sonarsource.com/browse/SONARPHP-1287

Looks like it's referring to the wrong ticket

rules/S6323/python/rule.adoc
```diff
@@ -0,0 +1,32 @@
+include::../description.adoc[]
+
+== Noncompliant Code Example
+[source,python]
+----
+re.search(r"Jack|Peter|", "John") # Noncompliant - returns true
+re.search(r"Jack||Peter", "John") # Noncompliant - returns true
```
andrea-guarino-sonarsource, line 7: `re.search` doesn't return a boolean, so I don't think it's correct to say `returns true`. Instead it will be return a match object or `None` otherwise https://docs.python.org/3/library/re.html#re.search
Maybe we could explicitly say that it will match the empty string?

andrea-guarino-sonarsource (changes requested):
Looks good, just a small comment about `re.search` return type

andrea-guarino-sonarsource (approved):

PR #792, opened 2022-02-08
Modify rule S107: Update code example for csharp to include rule behavior in the case of constructors.

mary-georgiou-sonarsource:
Update code example for rule S107 to reflect the case where the parameters of a constructor intended for the base class do not count in the total sum of parameters.

### DIFF
--- a/rules/S107/csharp/rule.adoc
+++ b/rules/S107/csharp/rule.adoc
@@ -6,7 +6,7 @@ With a maximum number of 4 parameters:
 
 [source,csharp]
 ----
-public void doSomething(int param1, int param2, int param3, string param4, long param5) 
+public void doSomething(int param1, int param2, int param3, string param4, long param5)
 {
 ...
 }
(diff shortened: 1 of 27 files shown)


arseniy-sonar (changes requested):
@mary-georgiou-sonarsource , please update the PR title according to our [naming convention](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule), and also add the `dotnet` label before merging.

arseniy-sonar (approved):
Thank you.

rules/S107/csharp/rule.adoc
```diff
@@ -16,12 +16,38 @@ public void doSomething(int param1, int param2, int param3, string param4, long
...
+	}
+}
+
+public TestClass(int param1, int param2, int param3, string param4, long param5)
+{
+	public void TestClass(int param1, int param2, int param3, string param4, long param5) // Noncompliant
+	{
+	...
+	}
+}
```
andrei-epure-sonarsource, line 48: I don't understand what `TestClass` is ? what's the return value of the function?

and what's the inner `TestClass`?

andrei-epure-sonarsource (changes requested):

rules/S107/csharp/rule.adoc
mary-georgiou-sonarsource: Apologies, the keyword `void` is a "typo".
My understanding is to leave only the first part of the example with the comment?
I agree the non-compliant example is too much.

rules/S107/csharp/rule.adoc
```diff
@@ -16,12 +16,30 @@ public void doSomething(int param1, int param2, int param3, string param4, long
...
 {
 ...
 }
 ----
 
+== Exceptions
+
+[source,csharp]
+----
+public BaseClass(int param1)
```
andrei-epure-sonarsource, line 29:
```suggestion
public class BaseClass(int param1)
```

rules/S107/csharp/rule.adoc
```diff
@@ -16,12 +16,30 @@ public void doSomething(int param1, int param2, int param3, string param4, long
...
+{
+...
+}
+
+public ExtensionClass : BaseClass
```
andrei-epure-sonarsource, line 34:
```suggestion
public class ExtensionClass : BaseClass
```

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM.

PR #795, opened 2022-02-08
Investigate branch fetching

marco-antognini-sonarsource:


marco-antognini-sonarsource:
sonarqube warning about missing reference seems to be solved with this quick patch.

PR #797, opened 2022-02-08
Modify rule S6328: Add Python as covered language

nils-werner-sonarsource:
Relates to https://jira.sonarsource.com/browse/SONARPY-981


guillaume-dequenne-sonarsource (approved):
Looks good, as mentioned in the comment, I'd update the documentation link for a more recent Python version and maybe double check that the behavior is consistent across Python versions.

PR #799, opened 2022-02-08
Modify rule S6395: Add Python as covered language

nils-werner-sonarsource:
Relates to https://jira.sonarsource.com/browse/SONARPY-983


rules/S6331/php/rule.adoc
```diff
@@ -14,7 +14,7 @@ include::../description.adoc[]
 "/foo\(\)/"  // Matches 'foo()'
 ----
 
-ifdef::env-github,rspecator-view[]
+ifdef::env-github,rspecator-view[S6395]
```
arseniy-sonar, line 17:
```suggestion
ifdef::env-github,rspecator-view[]
```
I wonder what you are trying to express here

nils-werner-sonarsource: Thank you for raising this. It was a change by accident. Maybe I had the focus on this line and was looking for rule S6395. Was then committed without intention. Nice catch.

rules/S6395/python/rule.adoc
```diff
@@ -0,0 +1,32 @@
...
+[source,python]
+----
+r"number\d{2}"
+r"(?:number)?\d{2}"
+----
+
+
+== Exceptions
+
+Do not report an issue if the non-capturing group is an alternation.
```
guillaume-dequenne-sonarsource, line 25: The use of imperative seems a bit weird here. What about `This rule does not report an issue if...`?
I notice this is the same in the PHP rule description.

guillaume-dequenne-sonarsource (approved):
LGTM for Python. Minor suggestions for some description wording.

PR #800, opened 2022-02-08
RULEAPI-744 automatically fill the template source tags with the current language

frederic-tingaud-sonarsource:


rspec-tools/rspec_tools/create_rule.py
```diff
@@ -65,6 +65,8 @@ class RuleCreator:
   ID_COUNTER_BRANCH: Final[str] = 'rspec-id-counter'
   ID_COUNTER_FILENAME: Final[str] = 'next_rspec_id.txt'
   TEMPLATE_PATH: Final[Path] = Path(__file__).parent.parent.joinpath('rspec_template')
+  LANGUAGE_NAMES: Final[dict[str, str]] = {
+      'cfamily': 'cpp', 'plsql': 'sql', 'tsql': 'sql'}
```
arseniy-sonar, line 69: IMO, to avoid desync with https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_tools/validation/description.py#L51 there should be a single source of truth for both these uses.
Also, please add this list to the [RULEAPI-714](https://jira.sonarsource.com/browse/RULEAPI-714) ticket.

arseniy-sonar (changes requested):
Looking good!

PR #802, opened 2022-02-08
Modify rule S6004: Add code highlighting

marco-antognini-sonarsource:


marco-antognini-sonarsource:
Highlighting in those sections isn't supported yet.

PR #809, opened 2022-02-10
Update metadata for JS quickfixes

saberduck:


vilchik-elena (approved):
I think we can add S1940 in the same PR

PR #827, opened 2022-02-16
Modify rule S1172: Match the behaviour of TypeScript compiler for JS/TS

yassin-kammoun-sonarsource:
We now consider unused parameters regardless of whether they appear before used parameters. Furthermore, we ignore intentionally ignored parameters, that is, parameters whose names start with an underscore.


vilchik-elena (changes requested):
I believe just example with `_` is not enough, we need to describe the reasoning behind it

rules/S1172/javascript/rule.adoc
```diff
@@ -18,39 +18,36 @@ function doSomething(b) {
...
+[source,javascript]
+----
+function doSomething(a, b, c) {
+  compute(arguments);
+}
 ----
 
-Further, when ``++arguments++`` is used in the function body, no parameter is reported as unused. 
-
+Also the rule ignores all parameters whose name starts with an underscore (``++_++``).
```
vilchik-elena, line 44: @yassin-kammoun-sonarsource wdyt?

yassin-kammoun-sonarsource: I would add a comma after `Also`, and the suggestion is on the wrong line, but still, I like it.

yassin-kammoun-sonarsource: Never mind, I missed your commit.

vilchik-elena (approved):

PR #833, opened 2022-02-18
Modify rule S107: Synchronize exception for VB.NET with C#

pavel-mikula-sonarsource:
Same change as in #792


mary-georgiou-sonarsource (approved):
LGTM

PR #844, opened 2022-02-22
Add message for omitted attributes.

nils-werner-sonarsource:
To follow the message standard we have set up, I would suggest formulating a second message for this rule as well, highlighting the absence of an attribute/block.

### DIFF
--- a/rules/S6403/message.adoc
+++ b/rules/S6403/message.adoc
@@ -1,3 +1,5 @@
 === Message
 
 Make sure creating a GCP SQL instance without requiring TLS is safe here.
+
+Omitting {parameter} allows unencrypted connection to the database. Make sure it is safe here.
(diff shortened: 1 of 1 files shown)


rules/S6403/message.adoc
```diff
@@ -1,3 +1,5 @@
 === Message
 
 Make sure creating a GCP SQL instance without requiring TLS is safe here.
+
+Omitting {parameter} allows unencrypted connection to the database. Make sure it is safe here.
```
hendrik-buchwald-sonarsource, line 5:
```suggestion
Omitting {parameter} allows unencrypted connections to the database. Make sure it is safe here.
```

hendrik-buchwald-sonarsource (approved):
LGTM. Just would recommend using plural here, since singular sounds a bit strange.

PR #847, opened 2022-02-23
RULEAPI-748: Introduce GitHub Action to update quickfix status

marco-antognini-sonarsource:
This can be tested at https://github.com/marco-antognini-sonarsource/rspec/actions/workflows/update_quickfix_status.yml
When testing, the generated PRs will be on my fork as well.

For examples:
 * https://github.com/marco-antognini-sonarsource/rspec/pull/2
 * https://github.com/marco-antognini-sonarsource/rspec/pull/3

Mind you, the formatting of the metadata file is slightly altered (empty arrays are one-liners and `\/` is replaced by a simple `/`, which is equivalent AFAICT according to [JSON's spec](https://www.json.org/json-en.html)).

```diff
diff --git a/rules/S5954/cfamily/metadata.json b/rules/S5954/cfamily/metadata.json
index ea5c6c070e..2a9528357f 100644
--- a/rules/S5954/cfamily/metadata.json
+++ b/rules/S5954/cfamily/metadata.json
@@ -3,7 +3,7 @@
   "type": "CODE_SMELL",
   "status": "ready",
   "remediation": {
-    "func": "Constant\/Issue",
+    "func": "Constant/Issue",
     "constantCost": "2min"
   },
   "tags": [
@@ -13,12 +13,8 @@
     "clumsy"
   ],
   "extra": {
-    "replacementRules": [
-
-    ],
-    "legacyKeys": [
-
-    ]
+    "replacementRules": [],
+    "legacyKeys": []
   },
   "defaultSeverity": "Minor",
   "ruleSpecification": "RSPEC-5954",
@@ -28,5 +24,5 @@
     "Sonar way",
     "MISRA C++ 2008 recommended"
   ],
-  "quickfix": "unknown"
-}
+  "quickfix": "covered"
+}
\ No newline at end of file
```

### DIFF
--- a/.github/workflows/update_quickfix_status.yml
+++ b/.github/workflows/update_quickfix_status.yml
@@ -1,0 +1,52 @@
+name: Update quick fix status
+
+on: 
+  workflow_dispatch:
+    inputs:
+      rule:
+        description: 'ID of an existing rule (e.g., S1234).'
+        required: true
+        type: string
+      language:
+        description: 'Language to be updated for the given rule, (e.g., cfamily)'
+        required: true
+        type: string
+      status:
+        description: 'The new status for quick fix (e.g., covered)'
+        required: true
+        type: choice
+        options:
+          - covered
+          - partial
+          - targeted
+          - infeasible
+          - unknown
+
+jobs:
+  update_quickfix_status:
+    name: Update quick fix status
+    runs-on: ubuntu-20.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        persist-credentials: true
+        ref: master
+        path: 'rspec'
+
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'
+
+    - name: 'Install Pipenv'
+      run: |
+        pip install pipenv
+
+    - name: 'Install rspec-tools'
+      working-directory: 'rspec/rspec-tools'
+      run: pipenv install
+
+    - name: 'Update quickfix status'
+      working-directory: 'rspec/rspec-tools'
+      run: pipenv run rspec-tools update-quickfix-status --user ${{ github.actor }} --rule "${{ github.event.inputs.rule }}" --language "${{ github.event.inputs.language }}" --status "${{ github.event.inputs.status }}"
--- a/docs/metadata.adoc
+++ b/docs/metadata.adoc
@@ -1,3 +1,7 @@
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+endif::[]
 = Rule Metadata
 
 This document describes how `+metadata.json+` should be structured.
@@ -20,4 +24,9 @@ A `covered` rule is still not guaranteed to provide a quick fix for a particular
 For instance, on one hand, if a rule detects two functions that are dangerous to use, `A` and `B`, and `A` has an obvious replacement (and therefore a quick fix) while `B` does not, the field should be set to `partial`.
 
 On the other hand, if a quick fix could be easily proposed for both `A` and `B`, but the fix location might be inside a macro expansion, or in a different file from the issue location (and hence not feasible given the current SonarLint capabilities), this should not prevent the rule from being tagged as `covered`.
+====
+
+[TIP]
+====
+You can update the quickfix field using this GitHub Workflow: https://github.com/SonarSource/rspec/actions/workflows/update_quickfix_status.yml
 ====
--- a/rspec-tools/rspec_tools/cli.py
+++ b/rspec-tools/rspec_tools/cli.py
@@ -5,11 +5,11 @@ from pathlib import Path
 import click
 from rspec_tools.checklinks import check_html_links
 from rspec_tools.errors import RuleNotFoundError, RuleValidationError
-from rspec_tools.create_rule import create_new_rule, add_language_to_rule
 from rspec_tools.rules import RulesRepository, LanguageSpecificRule
 from rspec_tools.validation.metadata import validate_rule_metadata
 from rspec_tools.validation.description import validate_section_names, validate_section_levels, validate_parameters, validate_source_language
 from rspec_tools.coverage import update_coverage_for_all_repos, update_coverage_for_repo, update_coverage_for_repo_version
+import rspec_tools.create_rule
 
 from rspec_tools.notify_failure_on_slack import notify_slack
 
@@ -38,7 +38,7 @@ def check_links(d):
 def create_rule(languages: str, user: Optional[str]):
   '''Create a new rule.'''
   token = os.environ.get('GITHUB_TOKEN')
-  create_new_rule(languages, token, user)
+  rspec_tools.create_rule.create_new_rule(languages, token, user)
 
 @cli.command()
 @click.option('--language', required=True)
@@ -47,7 +47,17 @@ def create_rule(languages: str, user: Optional[str]):
 def add_lang_to_rule(language: str, rule: str, user: Optional[str]):
   '''Add a new language to rule.'''
   token = os.environ.get('GITHUB_TOKEN')
-  add_language_to_rule(language, rule, token, user)
+  rspec_tools.create_rule.add_language_to_rule(language, rule, token, user)
+
+@cli.command()
+@click.option('--language', required=True)
+@click.option('--rule', required=True)
+@click.option('--status', required=True)
+@click.option('--user', required=False)
+def update_quickfix_status(language: str, rule: str, status: str, user: Optional[str]):
+  '''Update the status of quick fix for the given rule/language'''
+  token = os.environ.get('GITHUB_TOKEN')
+  rspec_tools.create_rule.update_rule_quickfix_status(language, rule, status, token, user)
 
 @cli.command()
 @click.argument('rules', nargs=-1, required=True)
--- a/rspec-tools/rspec_tools/cli.py
+++ b/rspec-tools/rspec_tools/cli.py
@@ -1,36 +1,41 @@
 import os
-from typing import Optional
 from pathlib import Path
+from typing import Optional
 
 import click
-from rspec_tools.checklinks import check_html_links
-from rspec_tools.errors import RuleNotFoundError, RuleValidationError
-from rspec_tools.rules import RulesRepository, LanguageSpecificRule
-from rspec_tools.validation.metadata import validate_rule_metadata
-from rspec_tools.validation.description import validate_section_names, validate_section_levels, validate_parameters, validate_source_language
-from rspec_tools.coverage import update_coverage_for_all_repos, update_coverage_for_repo, update_coverage_for_repo_version
-import rspec_tools.create_rule
 
+import rspec_tools.create_rule
+import rspec_tools.modify_rule
+from rspec_tools.checklinks import check_html_links
+from rspec_tools.coverage import (update_coverage_for_all_repos,
+                                  update_coverage_for_repo,
+                                  update_coverage_for_repo_version)
+from rspec_tools.errors import RuleValidationError
 from rspec_tools.notify_failure_on_slack import notify_slack
+from rspec_tools.rules import LanguageSpecificRule, RulesRepository
+from rspec_tools.validation.description import (validate_parameters,
+                                                validate_section_levels,
+                                                validate_section_names,
+                                                validate_source_language)
+from rspec_tools.validation.metadata import validate_rule_metadata
+
+
+def _fatal_error(message: str):
+  click.echo(message, err=True)
+  raise click.Abort(message)
 
 @click.group()
 @click.option('--debug/--no-debug', default=False)
 def cli(debug):
     'Tools automating RSPEC workflows.'
 
-@cli.command()
-@click.option('--rule', help='Validate only the rule matching the provided ID.')
-def validate(rule):
-  '''Validate rules.'''
-  # TODO
-  if rule == '42':
-    raise RuleNotFoundError(rule)
 
 @cli.command()
 @click.option('--d', required=True)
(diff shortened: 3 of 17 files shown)


rspec-tools/tests/test_create_rule.py
```diff
@@ -81,6 +82,57 @@ def read_counter_file(repo):
...
+def test_update_quickfix_status_branch(rule_creator: RuleCreator, mock_rspec_repo: Repo):
+  '''Test update_quickfix_status_branch'''
+  rule_number = '100'
+  language = 'cfamily'
+  status = 'covered'
+
+  metadata_path = Path(mock_rspec_repo.working_dir, 'rules', f'S{rule_number}', language, 'metadata.json')
+  mock_rspec_repo.git.checkout('master')
+  initial_metadata = json.loads(metadata_path.read_text())
+  assert status != initial_metadata.get('quickfix', 'unknown')
```
arseniy-sonar, line 94: I feel this assertion, as well as the one on l. 101 is too weak. As you are controlling the test inputs, you should know whether `quickfix` is present in the `initial_metadata` and `new_metadata` dictionaries, and you can assert that it is what you think it is.

rspec-tools/tests/test_create_rule.py
```diff
@@ -81,6 +82,57 @@ def read_counter_file(repo):
...
+
+  branch = rule_creator.update_quickfix_status_branch('Some title', rule_number, language, status)
+  mock_rspec_repo.git.checkout(branch)
+  new_metadata = json.loads(metadata_path.read_text())
+  for key in initial_metadata:
+    if key == 'quickfix':
+      assert status == new_metadata.get('quickfix', 'unknown')
+    else:
+      assert initial_metadata[key] == new_metadata[key]
```
arseniy-sonar, line 103: Perhaps also assert that no other file is changed

rspec-tools/tests/test_create_rule.py
```diff
@@ -81,6 +82,57 @@ def read_counter_file(repo):
...
+
+  rule_creator.update_quickfix_status_pull_request(mockGithub, 100, 'cfamily', 'covered', 'label-fraicheur', 'testuser')
+  ghRepoMock.create_pull.assert_called_once();
+  assert ghRepoMock.create_pull.call_args.kwargs['title'].startswith('Modify rule S100')
+  pullMock.add_to_assignees.assert_called_with('testuser')
+  pullMock.add_to_labels.assert_called_with('label-fraicheur')
+
+
+@patch('rspec_tools.create_rule.RuleCreator')
+def test_update_rule_quickfix_status(mockRuleCreator):
```
arseniy-sonar, line 124: A docstring?

rspec-tools/rspec_tools/create_rule.py
```diff
@@ -144,6 +162,25 @@ def create_new_rule_branch(self, rule_number: int, languages: Iterable[str]) ->
...
+      self._update_quickfix_status(metadata_path, status)
+      self.repository.git.add('--all')
+      self.repository.index.commit(title)
+    self.repository.git.push('origin', branch_name)
+    return branch_name
+
+  def _update_quickfix_status(self, file: Path, status: str):
+    metadata = json.loads(file.read_text())
+    metadata['quickfix'] = status
+    file.write_text(json.dumps(metadata, indent=2))
```
arseniy-sonar, line 182: I'd be cautious with ignoring the "/" vs. "\\/" syntax. While they are equivalent w.r.t. JSON spec, they might be treated differently by the tools we use to work with JSON. I think this topic deserves clarification as a separate task, not as a part of this PR.
For example, here's what [rule-api does](https://github.com/SonarSource/sonar-rule-api/blob/master/src/main/java/com/sonarsource/ruleapi/services/RuleFilesService.java#L119) when GSON the library it uses to work with JSON outputs just "/":
```
gson.toJson(metadata).replace("/", "\\/")
```

arseniy-sonar (changes requested):
This looks like a really useful action!
Besides the comments below, some documentation is missing for this new action. You can simply mention it in the [section about the quickfix field](https://github.com/SonarSource/rspec/blob/master/docs/metadata.adoc#quickfix-field)

arseniy-sonar:
BTW, I do not have a button to trigger the "update quickfix status" action on your repository.

marco-antognini-sonarsource:
> BTW, I do not have a button to trigger the "update quickfix status" action on your clone.

Ah, that's a bummer... I just added you as a collaborator on my fork, hoping that would allow you to trigger it.

arseniy-sonar:
Thanks, I was able to run a few tests. Optionally, you could also validate that the new quick-fix status is different from the status quo.

rspec-tools/rspec_tools/create_rule.py
```diff
@@ -237,6 +264,19 @@ def create_new_rule_pull_request(self, github_api: Callable[[Optional[str]], Git
...
 
+  def update_quickfix_status_pull_request(self, github_api: Callable[[Optional[str]], Github], rule_number: int, language: str, status: str, label: str, user: Optional[str]):
+    title = f'Modify rule S{rule_number}: mark quick fix as "{status}"'
+    branch_name = self.update_quickfix_status_branch(title, rule_number, language, status)
+    click.echo(f'Created rule branch {branch_name}')
+    return self._create_pull_request(
+      github_api,
+      branch_name,
+      title,
+      f'You can preview this rule [here](https://sonarsource.github.io/rspec/#/rspec/S{rule_number}/{language}) (updated a few minutes after each push).',
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: Hum.. I've mixed feeling there -- I agree it is unlikely to be useful all the time, but still, it could be handy to some people. For example, the PR starts as simply updating the quick fix status but later someone spots a typo in the RSPEC while implementing the quick fix and updates the PR to include the fix for the typo. At the very least, I don't think it hurts to keep it.

rspec-tools/tests/test_create_rule.py
marco-antognini-sonarsource: You mean, using this instead?

```suggestion
  assert status != initial_metadata.get('quickfix')
```

marco-antognini-sonarsource (commented):
Thanks for the feedback

rspec-tools/rspec_tools/cli.py
```diff
@@ -5,7 +5,7 @@
 import click
 from rspec_tools.checklinks import check_html_links
 from rspec_tools.errors import RuleNotFoundError, RuleValidationError
-from rspec_tools.create_rule import create_new_rule, add_language_to_rule
+from rspec_tools.create_rule import create_new_rule, add_language_to_rule, update_rule_quickfix_status
```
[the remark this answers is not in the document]
arseniy-sonar: Sure they all operate on rules that's what the whole `rspec-tools` project is for. I think `rule_factory` might be associated with the factory design pattern and would be misleading. They do perform git commands and PR creation, but IMO that is an implementation detail, rather than the categorical factor. I don't mind the refactoring size if you make it on top of what's already in the PR.

rspec-tools/rspec_tools/create_rule.py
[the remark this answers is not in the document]
arseniy-sonar: At the moment we don't preview changes to a rule see [RULEAPI-655](https://jira.sonarsource.com/browse/RULEAPI-655).
From this standpoint, this link is misleading, and that's why I think it is doing a disservice to the users.
At least it should be renamed to something like "See the original rule [here]"

rspec-tools/tests/test_create_rule.py
arseniy-sonar: That's an equivalently weak version for this case, since if the `quickfix` field is not present in `initial_metadata` and if the `quickfix` is bound to a value different from `status` the assertion would be satisfied and I am not sure which one you are asserting here.
I'd go for two assertions.
```suggestion
  assert 'quickfix' in initial_metadata
  assert status != initial_metadata['quickfix']
```

rspec-tools/rspec_tools/create_rule.py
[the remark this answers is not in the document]
marco-antognini-sonarsource: Right, I did not consider this indeed.

rspec-tools/tests/test_create_rule.py
marco-antognini-sonarsource: Right, that's better indeed! I forgot that `dict.get` never raises an exception...

rspec-tools/rspec_tools/cli.py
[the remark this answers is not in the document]
marco-antognini-sonarsource: Fair enough. I'll see how to improve things.

rspec-tools/rspec_tools/create_rule.py
marco-antognini-sonarsource: Raised [RULEAPI-750](https://jira.sonarsource.com/browse/RULEAPI-750)

rspec-tools/tests/test_create_rule.py
marco-antognini-sonarsource: Improved with 00c8e4ef9b49268715434a7280ee415ae96cf499.

rspec-tools/tests/test_create_rule.py
marco-antognini-sonarsource: Improved with 00c8e4ef9b49268715434a7280ee415ae96cf499.

rspec-tools/tests/test_create_rule.py
marco-antognini-sonarsource: Improved with 00c8e4ef9b49268715434a7280ee415ae96cf499.

rspec-tools/rspec_tools/create_rule.py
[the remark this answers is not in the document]
marco-antognini-sonarsource: Improved with 00c8e4ef9b49268715434a7280ee415ae96cf499.

rspec-tools/rspec_tools/create_rule.py
marco-antognini-sonarsource: Improved with 00c8e4ef9b49268715434a7280ee415ae96cf499.

rspec-tools/rspec_tools/cli.py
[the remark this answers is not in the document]
marco-antognini-sonarsource: I think I've addressed your other comments. I've left this out for now. Once we are happy about the functional changes and testing, I'll start this.

marco-antognini-sonarsource (commented):
An update before doing the refactoring.

.github/workflows/update_quickfix_status.yml
```diff
@@ -0,0 +1,52 @@
+name: Update quick fix status
```
marco-antognini-sonarsource, line 1: Forgot about:

> You can simply mention it in the [section about the quickfix field](https://github.com/SonarSource/rspec/blob/master/docs/metadata.adoc#quickfix-field)

(Making this a thread to avoid forgetting about it.)

marco-antognini-sonarsource: Done with db90dee.

docs/metadata.adoc
```diff
@@ -1,3 +1,7 @@
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+endif::[]
```
arseniy-sonar, line 4: This is something that wasn't properly moved when the documentation was split up. I hope you don't mind me inserting it directly into your PR :pray: 
Feel free to remove or use different emojis

rspec-tools/tests/test_create_rule.py
```diff
@@ -82,26 +85,89 @@ def read_counter_file(repo):
...
-  rule_number = '100'
+def test_update_quickfix_status_branch1(rule_creator: RuleCreator, mock_rspec_repo: Repo):
+  '''Test update_quickfix_status_branch when quickfix field is not present in language-specific metadata'''
+  rule_number = 100
   language = 'cfamily'
-  status = 'covered'
 
-  metadata_path = Path(mock_rspec_repo.working_dir, 'rules', f'S{rule_number}', language, 'metadata.json')
+  sub_path = Path('rules', f'S{rule_number}', language, 'metadata.json')
+  metadata_path = Path(mock_rspec_repo.working_dir) / sub_path
```
arseniy-sonar, line 94: Interesting syntax :+1:

arseniy-sonar (approved):

docs/metadata.adoc
marco-antognini-sonarsource: Ha, looks better! 👍

rspec-tools/rspec_tools/cli.py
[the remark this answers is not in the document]
marco-antognini-sonarsource: I've now pushed the refactoring -- see the commits after 750c650. Hopefully, it's still reviewable... Before merging, I'll squash the commits into two: the new functionality and the refactoring.

arseniy-sonar (approved):
I love the new structure!
This is a huge change indeed. I checked it commit-by-commit, but I am too impatient to follow every detail. One thing that might be improved is the name (yeah, I know, naming :sweat_smile: ) of the `conftest.py` file. I understand from its content that it stores the common configuration for tests, like various mock structures. But from the name, I assumed it would be testing configuration.
How about "test_config.py" or "common_test_config.py" ?

marco-antognini-sonarsource:
> One thing that might be improved is the name (yeah, I know, naming 😅 ) of the `conftest.py` file. I understand from its content that it stores the common configuration for tests, like various mock structures. But from the name, I assumed it would be testing configuration.

I agree, it's probably not the best name... but I'm not sure we can change it as it's a special name w.r.t. pytest. https://docs.pytest.org/en/latest/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

PR #867, opened 2022-03-07
S1988 Delete never implemented duplicated rule

loic-joly-sonarsource:
Removed as discussed in [CPP-1071](https://jira.sonarsource.com/browse/CPP-1071)


loic-joly-sonarsource:
> @loic-joly-sonarsource in the repo there are 3 references to this rule in comments and links. I think we should drop them.

@abbas-sabra-sonarsource: I agree on the principle (and I did not think to check about it), but since those references are in comments, I was not sure what was best... Let me know what you think...

abbas-sabra-sonarsource:
> > @loic-joly-sonarsource in the repo there are 3 references to this rule in comments and links. I think we should drop them.
> 
> @abbas-sabra-sonarsource: I agree on the principle (and I did not think to check about it), but since those references are in comments, I was not sure what was best... Let me know what you think...

These are Jira links that were referring to the rule. if we were on Jira we would just drop them, and I think we  should do the same here

PR #869, opened 2022-03-07
Modify S4423: Fix messages for Terraform and Cloudformation

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-379](https://jira.sonarsource.com/browse/SONARIAC-379)


pierre-loup-tristant-sonarsource:
Thanks for the feedback, it should be fixed now.

PR #870, opened 2022-03-07
Deprecate rule S3400: deprecate rule for CFamily (CPP-3456)

frederic-tingaud-sonarsource:
It has too many exceptions and brings little value.


loic-joly-sonarsource (commented):
According to the doc, you should also remove the tag for the C++ version of this rule

frederic-tingaud-sonarsource:
I referred to the wrong documentation (we might want to remove or change docs/deprecation.adoc). Fixed

PR #871, opened 2022-03-08
Enable rules removal

arseniy-sonar:
To address the CI failure on #867


rspec-tools/rspec_tools/rules.py
```diff
@@ -89,4 +90,7 @@ def rules(self) -> Generator[GenericRule, None, None]:
     return (GenericRule(child) for child in self.rules_path.glob('S*') if child.is_dir())
     
   def get_rule(self, ruleid: str):
+    rulepath = self.rules_path.joinpath(ruleid)
+    if not rulepath.is_dir():
+      raise RuleNotFoundError('Cannot find rule ' + ruleid + ' in ' + str(self.rules_path))
     return GenericRule(self.rules_path.joinpath(ruleid))
```
marco-antognini-sonarsource, line 96:
```suggestion
    return GenericRule(rulepath)
```

ci/validate_metadata.sh
```diff
@@ -7,10 +7,16 @@ base="$(git merge-base FETCH_HEAD HEAD)"
...
+  echo "Change in the tools, revalidating all rules"
 else
-  git diff --name-only "${base}" -- rules/ | sed -Ee 's#rules/(S[0-9]+)/.*#\1#' | sort -u | mapfile -t affected_rules
+  git diff --name-only "${base}" -- rules/ | # Get all the changes in rules
+    sed -Ee 's#(rules/S[0-9]+)/.*#\1#' | # extract the rule directories
+    sort -u | # deduplicate
+    while IFS= read -r rule; do [[ -d "$rule" ]] && echo "$rule" || true; done |  # filter non-deleted rules
+    sed 's#rules/##' | # get rule ids
+    mapfile -t affected_rules # store them in the `affected_rules` array
+  echo "Validating ${affected_rules[@]}"
```
marco-antognini-sonarsource, line 19: `git diff --diff-filter=d` (lower case `d` excludes Deleted files) will achieve the same as the `while` command you wrote.

marco-antognini-sonarsource (changes requested):
Looks good. A suggestion to simplify the shell script.

ci/validate_metadata.sh
arseniy-sonar: I'm afraid that won't achieve the same result.
Let's say someone accidentally removes a file from a rule, but not the whole rule. The rule becomes invalid and should normally be flagged by the validation but it `git diff --diff-filter=d` will skip the rule if no other file in it was touched.

arseniy-sonar: i.e.: remove `rules/S100/cfamily/metadta.json`, and have no other file touched in `rules/S100/*`.

marco-antognini-sonarsource: Hum.. you're right. Too bad.

marco-antognini-sonarsource (approved):

PR #874, opened 2022-03-08
Modify rule S5332: Update issue message

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-385](https://jira.sonarsource.com/browse/SONARIAC-385)


hendrik-buchwald-sonarsource (approved):
The implementation ticket here in the first message also seems to be wrong. Otherwise LGTM.

PR #878, opened 2022-03-09
CPP-3597 Remove CFamily legacy rules id

mpaladin:


arseniy-sonar (changes requested):
I think it is important to still be able to find these rules using legacy rule ids.
If you agree, I would kindly ask you to put these removed rule ids into metadata.json/`extra`/`legacyKeys`, or wait with merging this PR until I get the opportunity to do it.

arseniy-sonar (approved):
The extra/legacyKeys already contain the removed sqKey, sometimes it is on the outer level metadata.json (i.e. no language specific).

abbas-sabra-sonarsource (approved):

PR #880, opened 2022-03-09
Modify rule S6252: Update message and title

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-390](https://jira.sonarsource.com/browse/SONARIAC-390)


rules/S6252/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Unversioned or suspended versioned S3 bucket is security-sensitive",
+  "title": "Unversioned or suspended versioned S3 buckets are security-sensitive",
```
loris-s-sonarsource, line 2:
```suggestion
  "title": "Disabling versioning of S3 buckets is security-sensitive",
```

rules/S6252/message.adoc
```diff
@@ -1,4 +1,4 @@
 === Message
 
 Make sure using [unversioned|suspended versioned] S3 bucket is safe here.
```
loris-s-sonarsource, line 3:
```suggestion
Make sure using [unversioned|versioning-suspended] S3 bucket is safe here.
```

The docs actually mention "versioning-suspended": https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html

rules/S6252/ask-yourself.adoc
```diff
@@ -1,6 +1,5 @@
 == Ask Yourself Whether
 
-* The S3 bucket stores sensitive information that is required to be preserved on the long term.
-* The S3 bucket grants write permission to many users.
+* The bucket stores information that require high availability.
```
loris-s-sonarsource, line 3: I think that the initial intention of the "long-term" message was maybe for standard purposes, incident response, or operational safety.

 I think that the "high availability" suggestion is great, and for user-friendliness, I would say that it may be worth adding small details like that data needs to be preserved or restored, in case of unintended user actions, security incidents, and application failures. (Since the main use of versioning would be for restoration IMO)

loris-s-sonarsource (approved):
LGTM, medium suggestions

rules/S6252/ask-yourself.adoc
pierre-loup-tristant-sonarsource: My intention here was to align the rule with the content of [S6412](https://sonarsource.github.io/rspec/#/rspec/S6412).

I am ok to add more details, keep "long-term" or any other improvement, as long as it impacts both rule.
Can you open a separate PR and make the change suggestion?

PR #883, opened 2022-03-10
Modify rule S6275: update issue message and code examples

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-393](https://jira.sonarsource.com/browse/SONARIAC-393)


rules/S6275/cloudformation/rule.adoc
```diff
@@ -17,11 +18,12 @@ Resources:
       Encrypted: false  # Sensitive
 ----
 
+[source,cloudformation]
 ----
 AWSTemplateFormatVersion: '2010-09-09'
 Resources:
   Ec2Volume:
-    Type: AWS::EC2::Volume  # Sensitive if "encrypted by default" is disabled
+    Type: AWS::EC2::Volume  # Sensitive as encryption is disabled by default
```
hendrik-buchwald-sonarsource, line 26: I think the default value can be specified, so it is not clear if encryption is enabled or not when this parameter is missing.

hendrik-buchwald-sonarsource, line 26: I think the default value can be specified, so it is not clear if encryption is enabled or not when this parameter is missing.

pierre-loup-tristant-sonarsource: Do you know where this default value can be set? I change it mainly cause the sentence was hard to understand.

hendrik-buchwald-sonarsource: For Terraform, it can be set for example with this module: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_encryption_by_default

I agree that the previous sentence was not great. Maybe something like `Sensitive if not encrypted by default`?

pierre-loup-tristant-sonarsource: Ok I see. But what about the message then? _Omitting "{parameter}" disables volumes encryption. Make sure it is safe here._. It feels incorrect as someone could [ebs_encryption_by_default](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_encryption_by_default).

What about assuming that [ebs_encryption_by_default](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_encryption_by_default) is not use, accept that we have a gap and create a ticket to support [ebs_encryption_by_default](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_encryption_by_default) configuration?

hendrik-buchwald-sonarsource: True. Okay, sounds good to me.

pierre-loup-tristant-sonarsource: @hendrik-buchwald-sonarsource ok I take care of creating the ticket. Can you please approve this PR if there is no other comments?

PR #893, opened 2022-03-10
Modify rule S6332: Update issue message

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-403](https://jira.sonarsource.com/browse/SONARIAC-403)


rules/S6332/cloudformation/message.adoc
```diff
@@ -0,0 +1,4 @@
+=== Message 
+
+* Make sure that using unencrypted EFS file systems is safe here.
+* Omitting "Encrypted" disables EFS file systems encryption. Make sure it is safe here.
```
loris-s-sonarsource, line 4:
```suggestion
* Omitting the "Encrypted" attribute disables EFS file system encryption. Make sure it is safe here.
```

rules/S6332/terraform/message.adoc
```diff
@@ -0,0 +1,4 @@
+=== Message 
+
+* Make sure that using unencrypted EFS file systems is safe here.
+* Omitting "encrypted" disables EFS file systems encryption. Make sure it is safe here.
```
loris-s-sonarsource, line 4:
```suggestion
* Omitting the "encrypted" attribute disables EFS file system encryption. Make sure it is safe here.
```

rules/S6332/cloudformation/message.adoc
pierre-loup-tristant-sonarsource: I was referring to the fact that all other issue messages share the pattern `Omitting "{parameter}" ...`, I thought it was better for this PR to stick to the existing pattern.
In general I am ok switching to another pattern like the one you suggested `Omitting the "{parameter}" attribute ...`

loris-s-sonarsource: Ah ok, I see! Yes, we should do it in a separate future PR

PR #899, opened 2022-03-14
Modify rule S4226: document exception

costin-zaharia-sonarsource:


rules/S4226/csharp/rule.adoc
```diff
@@ -68,6 +68,10 @@ namespace MyLibrary
 }
 ----
 
+== Exceptions
+
+* Extension methods added for clases decorated with `System.CodeDom.Compiler.GeneratedCodeAttribute` attribute.
```
costin-zaharia-sonarsource, line 73: Can be merged since the fix is in progress.

pavel-mikula-sonarsource (approved):
LGTM

PR #906, opened 2022-03-15
Remove redirecting link to cognitive complexity white-paper

loic-joly-sonarsource:
Use the direct link to the pdf instead. Currently, both links are equivalent, but the redirection might not work in the future.


pierre-loup-tristant-sonarsource (approved):
LGTM

PR #872, opened 2022-03-08
Modify rule S6317: Add message, highlighting and the list of supported privilege escalation vector details

pierre-loup-tristant-sonarsource:
Ticket: [SONARIAC-384](https://jira.sonarsource.com/browse/SONARIAC-384)

PR #933, opened 2022-04-05
Modify rule S6317: revert #872

pierre-loup-tristant-sonarsource:
Reverts SonarSource/rspec#872


loris-s-sonarsource (approved):
LGTM

pierre-loup-tristant-sonarsource:
This revert never made it to master and the rule description was updated and shipped with sonar-iac 1.9.

PR #941, opened 2022-04-08
RULEAPI-753: Use modern ids and coalesce the coverage for legacy ids

arseniy-sonar:


rspec-tools/tests/test_coverage.py
```diff
@@ -13,11 +14,15 @@
 
 def clear_working_dir(repo_dir):
   for f in os.listdir(repo_dir):
-    if f != '.git':
-      if os.path.isdir(repo_dir / f):
+    if os.path.isdir(repo_dir / f):
+      if f != '.git':
```
marco-antognini-sonarsource, line 18: Previously, `.git` was never removed. Now it can if it's a file. (This can happen when using git-worktree, for example.) This particular change might be a bit dangerous. If not necessary, I'd keep the old version of `clear_working_dir`.

rspec-tools/tests/test_coverage.py
```diff
@@ -26,7 +31,8 @@ def clear_working_dir(repo_dir):
                 'date': '2020-03-03 10:00:00',
                 'files': [['sonarpedia.json', JSTS_SONARPEDIA],
                           ['rules/Sonar_way_profile.json', '{}'],
-                          ['rules/S100.json', '{}'], ['rules/S1145.json', '{}'],
+                          # MethodName is a lagacy key for S100
+                          ['rules/MethodName.json', '{}'], ['rules/S1145.json', '{}'],
```
marco-antognini-sonarsource, line 35: It may be good to have a test to ensure `MethodName` isn't part of the coverage, as opposed to the already existing test `assert 'S100' in cov['JAVASCRIPT']`.

marco-antognini-sonarsource (changes requested):
Looks nice! Some minor comments.

rspec-tools/rspec_tools/coverage.py
```diff
@@ -58,10 +58,35 @@ def canonicalize(language):
...
+  if 'extra' in metadata and 'legacyKeys' in metadata['extra']:
+    ret.extend(metadata['extra']['legacyKeys'])
+  return ret
+
+def read_canonical_rule_ids(rules_dir):
+  '''
+  Map all the keys identifying a rule to its modern key (which is also its directory name).
+  '''
+  print('Collecting the rule-id synonyms from ' + str(rules_dir))
+  rules_dir.resolve()
```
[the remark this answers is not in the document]
arseniy-sonar: indeed. thanks. removed.

rspec-tools/rspec_tools/coverage.py
```diff
@@ -58,10 +58,35 @@ def canonicalize(language):
...
+  canonical_id = {}
+  (_, rule_dirs, _) = os.walk(rules_dir).__next__()
```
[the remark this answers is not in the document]
arseniy-sonar: Looks reasonable. I didn't like the `os.walk` line. I used your version.

rspec-tools/tests/test_coverage.py
arseniy-sonar: Today I've learned about `git worktree` :)
Thanks, reverted.

rspec-tools/tests/test_coverage.py
arseniy-sonar: Indeed. Added to all three tests that test the presence of S100

arseniy-sonar:
@marco-antognini-sonarsource 
I also piggy-backed mocking of the temporary RSPEC repo for create_rule and modify_rule tests, as they were failing on my machine otherwise. Let me know if you think I should push this change separately

marco-antognini-sonarsource (approved):
Thanks for the update, LGTM!

> I also piggy-backed mocking of the temporary RSPEC repo for create_rule and modify_rule tests, as they were failing on my machine otherwise. Let me know if you think I should push this change separately

I think this deserves its own commit. But no need to go through another PR. I'll let you squash & split commits as necessary before merging. 👍

PR #943, opened 2022-04-11
Modify rule S5131: Add PHP text

hendrik-buchwald-sonarsource:
This is a draft for the Progressive Education Framework.

### DIFF
--- a/rules/S5131/php/laravel/rule.adoc
+++ b/rules/S5131/php/laravel/rule.adoc
@@ -47,19 +47,7 @@ $response = response($input, 200)->header('Content-Type', 'text/plain');
 
 === Pitfalls
 
-Be aware that there are more content-types than `text/html` that allow to execute JavaScript code in a browser and thus are prone to Cross-Site Scripting vulnerabilities.
-The following content-types are known to be affected:
-
- * application/mathml+xml
- * application/rdf+xml
- * application/vnd.wap.xhtml+xml
- * application/xhtml+xml
- * application/xml
- * image/svg+xml
- * text/html
- * text/rdf
- * text/xml
- * text/xsl
+include::../../common/pitfalls/content-types.adoc[]
 
 === Going the extra mile
 
--- a/rules/S5131/php/laravel/rule.adoc
+++ b/rules/S5131/php/laravel/rule.adoc
@@ -22,11 +22,21 @@ Thus, the response is not vulnerable to reflected Cross-Site Scripting.
 
 [source,php]
 ----
+$response = response(json_encode(['data' => $input]), 200); // Noncompliant
+----
+
+[source,php]
+----
 $response = response()->json(['data' => $input]);
 ----
 
 It is also possible to set the content-type manually using the `header` method of the `Response` class.
 For example, setting the content-type to `text/plain` allows to safely reflect user input since browsers will not try to parse and execute the response.
+
+[source,php]
+----
+$response = response($input, 200); // Noncompliant
+----
 
 [source,php]
 ----
--- a/rules/S5131/php/laravel/rule.adoc
+++ b/rules/S5131/php/laravel/rule.adoc
@@ -37,6 +37,20 @@ $response = response($input, 200)->header('Content-Type', 'text/plain');
 
 === Pitfalls
 
+Be aware that there are more content-types than `text/html` that allow to execute JavaScript code in a browser and thus are prone to Cross-Site Scripting vulnerabilities.
+The following content-types are known to be affected:
+
+ * application/mathml+xml
+ * application/rdf+xml
+ * application/vnd.wap.xhtml+xml
+ * application/xhtml+xml
+ * application/xml
+ * image/svg+xml
+ * text/html
+ * text/rdf
+ * text/xml
+ * text/xsl
+
 === Going the extra mile
 
 include::../../resources.adoc[]
--- a/rules/S5131/php/laravel/rule.adoc
+++ b/rules/S5131/php/laravel/rule.adoc
@@ -10,14 +10,27 @@ include::../../threat.adoc[]
 
 == How to fix it?
 
+In case the response consists of HTML code, it is highly recommended to use a template engine like https://laravel.com/docs/9.x/blade[Blade] to generate it.
+A proper template engine does not only separate the view from the business logic and thus results in a more maintainable application.
+It also automatically encodes variable output and thus drastically decreases the chance of introducing Cross-Site Scripting vulnerabilities.
+
+If you do not intend to send HTML code to the clients, the vulnerability can be resolved by telling them what data they are receiving.
+For example, you can use the `json` method of the `Response` class to safely return JSON messages.
+The method automatically sets the content-type to `application/json`.
+This tells the browser that the response does not contain HTML code and should not be parsed and interpreted as HTML.
+Thus, the response is not vulnerable to reflected Cross-Site Scripting.
+
 [source,php]
 ----
-
+$response = response()->json(['data' => $input]);
 ----
+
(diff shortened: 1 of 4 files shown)


rules/S5131/php/laravel/rule.adoc
```diff
@@ -10,20 +10,47 @@ include::../../threats.adoc[]
 
 == How to fix it?
 
+In case the response consists of HTML code, it is highly recommended to use a template engine like https://laravel.com/docs/9.x/blade[Blade] to generate it.
```
loris-s-sonarsource, line 13: While I agree, if the customer has a specific use case for really sending HTML from code, we are basically telling them to change the whole architecture of the program, is it reasonable in this case?

rules/S5131/php/laravel/rule.adoc
```diff
@@ -10,20 +10,47 @@ include::../../threats.adoc[]
...
+ * application/rdf+xml
+ * application/vnd.wap.xhtml+xml
+ * application/xhtml+xml
+ * application/xml
+ * image/svg+xml
+ * text/html
+ * text/rdf
+ * text/xml
+ * text/xsl
+
```
loris-s-sonarsource, line 53: This section is shared between multiple PHP descriptions, and I think it could really fit into a common text, for all languages actually. I think I could use it for C# too

rules/S5131/php/native/rule.adoc
```diff
@@ -10,20 +10,53 @@ include::../../threats.adoc[]
...
 == How to fix it?
 
+User input embedded in HTML code should be HTML-encoded to prevent the injection of additional code.
+PHP provides the built-in function `htmlspecialchars` to do this.
+
 [source,php]
 ----
-
+echo '<h1>' . htmlspecialchars($input) . '</h1>';
 ----
```
loris-s-sonarsource, line 19: The end result of having two code samples would be a diff. So, I agree It wouldn't make sense in a sequential configuration to add a simple concatenation sample, but it would be worth adding in the other case

rules/S5131/php/native/rule.adoc
```diff
@@ -10,20 +10,53 @@ include::../../threats.adoc[]
...
 === How does this work?
 
 === Pitfalls
 
+By default, `htmlspecialchars` does not encode single quotes. Make sure to set the option `ENT_QUOTES` to encode single quotes if required by the context.
+
+[source,php]
+----
+echo "<img src='" . htmlspecialchars($input, ENT_QUOTES) . "'>";
+----
```
loris-s-sonarsource, line 39: I think it would be worth having a subsection title there, like "PHP-specific Encoding pitfalls"

rules/S5131/php/laravel/rule.adoc
loris-s-sonarsource: "Content-Type Pitfalls"

rules/S5131/php/symfony/rule.adoc
```diff
@@ -10,20 +10,53 @@ include::../../threats.adoc[]
 
 == How to fix it?
 
+In case the response consists of HTML code, it is highly recommended to use a template engine like https://twig.symfony.com/[Twig] to generate it.
+A proper template engine does not only separate the view from the business logic and thus results in a more maintainable application.
```
loris-s-sonarsource, line 14: My comment is the same that the previous one; this is architectural advice, and maybe not adapted in the rule description. It would 100% fit in the "Going The Extra Mile" section

rules/S5131/php/laravel/rule.adoc
loris-s-sonarsource: It would 100% fit in the "Going The Extra Mile" section

hendrik-buchwald-sonarsource: Maybe it makes more sense in another section. My thought process was the following: I don't think it is very likely that someone actually wants to return HTML here. Blade is included in Laravel, so there is really no reason not to use it to generate the HTML code, and I would say every Laravel developer that has written a web application is aware of that. But just for the rare chance that someone really wants to return HTML code manually for some odd reason, I mentioned it so that they know what direction to take.
I was considering also mentioning `htmlspecialchars` but I consider it a bad practice to generate the HTML code manually in a Laravel application, so I decided against it. I don't want to confuse the reader by showing solutions that I would not recommend them to use.

rules/S5131/php/laravel/rule.adoc
hendrik-buchwald-sonarsource: Yep, I agree, I wanted to move this to a shared directory.

loris-s-sonarsource: Here are some other types: https://github.com/BlackFan/content-type-research/blob/master/XSS.md

hendrik-buchwald-sonarsource: Interesting, `multipart/x-mixed-replace` works in FF. The other 2 we left out on purpose [here](https://github.com/SonarSource/security-expected-issues/blob/master/rule-modeling/common/browsers.md), since they did not meet the requirements in that doc (if I remember correctly).

rules/S5131/php/native/rule.adoc
hendrik-buchwald-sonarsource: Yes, in a diff form it would be nice. I will add it as a separate box for now with the idea that it will be a diff-box later.

rules/S5131/php/native/rule.adoc
hendrik-buchwald-sonarsource: Okay, done.

rules/S5131/php/laravel/rule.adoc
pierre-loup-tristant-sonarsource: I follow @hendrik-buchwald-sonarsource here. Since there is a built-in template engine, the best practice is to use it.

pierre-loup-tristant-sonarsource (changes requested):
The content looks good. 
My main suggestion is to move most of the fix explanation to "How does it work"

rules/S5131/php/native/rule.adoc
```diff
@@ -10,20 +10,51 @@ include::../../threats.adoc[]
...
+echo $input; // Noncompliant
+----
 
+[source,php]
+----
+header('Content-Type: text/plain');
+echo $input;
 ----
 
 === How does this work?
```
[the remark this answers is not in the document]
hendrik-buchwald-sonarsource: Do you have an example what you mean?

rules/S5131/php/laravel/rule.adoc
```diff
@@ -10,20 +10,45 @@ include::../../threats.adoc[]
...
 
+In case the response consists of HTML code, it is highly recommended to use a template engine like https://laravel.com/docs/9.x/blade[Blade] to generate it.
+A proper template engine does not only separate the view from the business logic and thus results in a more maintainable application.
+It also automatically encodes variable output and thus drastically decreases the chance of introducing Cross-Site Scripting vulnerabilities.
+
+If you do not intend to send HTML code to the clients, the vulnerability can be resolved by telling them what data they are receiving.
+For example, you can use the `json` method of the `Response` class to safely return JSON messages.
+The method automatically sets the content-type to `application/json`.
+This tells the browser that the response does not contain HTML code and should not be parsed and interpreted as HTML.
+Thus, the response is not vulnerable to reflected Cross-Site Scripting.
```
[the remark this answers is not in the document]
hendrik-buchwald-sonarsource: Done.

hendrik-buchwald-sonarsource:
@pierre-loup-tristant-sonarsource should be the case, did I miss it somewhere?

pierre-loup-tristant-sonarsource:
Sorry I missed it. Everything looks good.

PR #944, opened 2022-04-11
Modify rule S4650: Activate rule by default

yassin-kammoun-sonarsource:


rules/S4650/css/metadata.json
```diff
@@ -21,8 +21,5 @@
   "ruleSpecification": "RSPEC-4650",
   "sqKey": "S4650",
   "scope": "Main",
-  "defaultQualityProfiles": [
```
vilchik-elena, line 24: parent has not quality profiles, rule will not become activated

vilchik-elena (approved):

PR #945, opened 2022-04-11
Modify rule S5131(C#): Education Framework

loris-s-sonarsource:
First draft on `rules/S5131/csharp/dotnet_core_mvc/razor/rule.adoc`

The suggestion of this PR is to move every common text in a `common` folder, to make it clean. And the goal would be to have a lot of common text, so that the actual recommendations end up concise

- [x] Razor
- [x] System.Web


rules/S5131/csharp/dotnet_core_mvc/razor/rule.adoc
```diff
@@ -0,0 +1,212 @@
...
+include::../../common/fix/data_encoding.adoc[]
+
+==== In the View, Model, or ViewModel 
+
+To prevent cross-site scripting directly from code, two concepts are worth mentioning:
+
+First, that the compiler considers `IHtmlContent` to be a developer-defined HTML content. This means that:
+
+* Razor does not HTML-encode it at runtime to avoid double-encoding.
+* Any data that is not programmatically HTML encoded before it is sent to Razor is unsafe.
```
loris-s-sonarsource, line 24:
```suggestion
* Any data that is not programmatically HTML encoded before being sent to Razor is unsafe.
```

rules/S5131/csharp/dotnet_core_mvc/razor/rule.adoc
```diff
@@ -0,0 +1,212 @@
...
+----
+public class ViewModel 
+{
+    public string ThirdPartyString     { get; private set; }
+    public IHtmlContent ThirdPartyHtml { get; private set; }
+
+    public ViewModel(string thirdPartyData)
+    {
+        this.ThirdPartyString = thirdPartyData;
+        this.ThirdPartyHtml   = new HtmlString(thirdPartyData); # NonCompliant
```
loris-s-sonarsource, line 43:
```suggestion
        this.ThirdPartyHtml   = new HtmlString(thirdPartyData); // NonCompliant
```

rules/S5131/csharp/dotnet_core_mvc/razor/rule.adoc
```diff
@@ -0,0 +1,212 @@
...
+}
+@thirdPartyString 
+----
+
+==== Special Cases
+
+Note: Including third-party data in Special Razor Form tags is considered safe
+for some of them:
+
+* The `asp-route-{value}` attribute is automatically encoded by the `BeginWriteTagHelperAttribute()` function.
```
loris-s-sonarsource, line 153:
```suggestion
* The `asp-route-{value}` attribute is automatically encoded by the Runtime, with the `BeginWriteTagHelperAttribute()` function.
```

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+To prevent Cross-Site Scripting from happening in this situation, 
+use the HTML, JavaScript and URL encoders. +
+They are accessible via the `System.Text.Encodings.Web` namespace.
+
+[source,csharp]
+----
+var builder = WebApplication.CreateBuilder(args);
+var app = builder.Build();
+
+app.MapGet("/send/{thirdPartyData}",
```
hendrik-buchwald-sonarsource, line 26: It was said in the coffee break meeting that we might not support this type of controllers. If we do not support it (yet), we maybe should not have code samples for it (yet), because they will never match the code of the customer then.

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+[source,csharp]
+----
+using System.Text.Encodings.Web;
+
+var builder = WebApplication.CreateBuilder(args);
+var app = builder.Build();
+var encoder = HtmlEncoder.Default;
+
+app.MapGet("/send/{thirdPartyData}", 
+    (string thirdPartyData) => $"HTTP Response: {encoder.Encode(thirdPartyData)}");
```
hendrik-buchwald-sonarsource, line 47: While the code is valid, I wonder how often this will be what the developer really wants. If the developer uses a framework already that provides a template engine, and the template engine is not used, maybe the response is not supposed to contain HTML code (and instead JSON or something like that).

rules/S5131/csharp/razor/rule.adoc
```diff
@@ -10,23 +10,189 @@ include::../../threats.adoc[]
...
+@{
+    object thirdPartyObject = ViewBag.ThirdPartyString; // Value is read from ViewBagAll argument => can cause FP
+    string thirdPartyString = ViewBag.ThirdPartyString as string;
+}
+@thirdPartyString 
+----
+
+==== Special Cases
+
+Note: Including third-party data in Special Razor Form tags is considered safe
```
hendrik-buchwald-sonarsource, line 150: Is "Special Razor Form" a name (because of the casing)?

rules/S5131/csharp/razor/rule.adoc
```diff
@@ -10,23 +10,189 @@ include::../../threats.adoc[]
...
+for some of them:
+
+* The `asp-route-{value}` attribute is automatically encoded by the Runtime, with the `BeginWriteTagHelperAttribute()` function.
```
hendrik-buchwald-sonarsource, line 153:
```suggestion
* The `asp-route-{value}` attribute is automatically encoded by the runtime, with the `BeginWriteTagHelperAttribute()` function.
```

rules/S5131/csharp/razor/rule.adoc
```diff
@@ -10,23 +10,189 @@ include::../../threats.adoc[]
...
+* HTML Attributes are automatically encoded by the Runtime, with the `AddHtmlAttributeValue()` function.
```
hendrik-buchwald-sonarsource, line 154:
```suggestion
* HTML Attributes are automatically encoded by the runtime, with the `AddHtmlAttributeValue()` function.
```

hendrik-buchwald-sonarsource:
Creating the `common` directory and sharing texts inside of a rule seems like a good idea to me. No need to have exact copies of a text in multiple files or slightly rewording the same information.

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+include::../../threats.adoc[]
+
+== How to fix it?
+
+include::../../common/fix/intro.adoc[]
+
+include::../../common/fix/data_encoding.adoc[]
+
+To prevent Cross-Site Scripting from happening in this situation, 
+use the HTML, JavaScript and URL encoders. +
```
hendrik-buchwald-sonarsource, line 18: A small detail: while URL encoding can prevent XSS, I would not consider an encoding that is intended to prevent XSS. So I would probably not include it as an encoding here.

rules/S5131/common/fix/intro.adoc
```diff
@@ -0,0 +1,17 @@
+The root of this problem lies in the application of the AppSec principle which
+states that third-party services should not be trusted.
+
+As a programmer of web services, this principle is quite easy to apply: If the 
+code programmatically receives data from a third-party service, validate and/or
+sanitize the data to make it safe.
+
+As an end-user, the only way unwanted code from running in a browser is to
```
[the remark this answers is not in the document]
loris-s-sonarsource:
```suggestion
As an end-user, the only way to prevent unwanted code from running in a browser is to
```

rules/S5131/common/pitfalls/deny-lists.adoc
```diff
@@ -0,0 +1,16 @@
...
+The creation of deny-lists is a typical mechanism to stop a particular attack
+without fixing the underlying problem. Such filters, considered a gimmick by
+hackers, are found either in code or in web application firewalls.
+
+Unfortunately, creating a deny list is a common trap that should be avoided at
+all costs, as malicious users can easily change their exploit payloads and get
+back into the application.
+
+If software architecture or short time-to-market forces the development of such
+a list, consider that it is worth developing a global encoding strategy for
```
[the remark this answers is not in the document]
loris-s-sonarsource: Ahh yes! Actually, this section was to be deleted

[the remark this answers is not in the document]
loris-s-sonarsource: done

rules/S5131/csharp/aspnetcore.http/rule.adoc
loris-s-sonarsource: Yes, I think that I should just drop this altogether and use the code I made for the razor sample

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+var app = builder.Build();
+var encoder = HtmlEncoder.Default;
+
+app.MapGet("/send/{thirdPartyData}", 
+    (string thirdPartyData) => $"HTTP Response: {encoder.Encode(thirdPartyData)}");
+
+app.MapGet("/",
+    (HttpContext context) => {
+        var thirdPartyData = context.Request.QueryString["thirdPartyData"];
+        var thirdPartyData = encoder.Encode(thirdPartyData);
```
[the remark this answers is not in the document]
loris-s-sonarsource: 👍 good eyes!

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+include::../../common/extra-mile/csp.adoc[]
+
+:blazor-csps: https://docs.microsoft.com/en-us/aspnet/core/blazor/security/content-security-policy?view=aspnetcore-6.0
+:aspdotnet-core-middleware: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-6.0
+:csp-nonces-owin: https://vcsjones.dev/content-security-policy-nonces-in-asp-net-and-owin/
+
+Implementing content security policies differs from one architecture to another.
+
+When client-side code is processed by Blazor::
```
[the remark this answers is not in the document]
loris-s-sonarsource: Thanks for the feedback , I thought that a normal list would feel too much wordy

rules/S5131/csharp/razor/rule.adoc
loris-s-sonarsource: Not at all, thanks for pointing it out
```suggestion
Note: Including third-party data in special Razor form tags is considered safe
```

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+* {blazor-csps}[the library supports it by default].
+When CSP requirements are low::
+* {aspdotnet-core-middleware}[the ASP.Net Core Middleware provides header control].
+When code requirements enforce optimal CSP, including content security policies nonces::
+* The Open Web Interface for . NET (OWIN) {csp-nonces-owin}[can be used for a simple implementation].
+
+
+include::../../common/resources/docs.adoc[]
```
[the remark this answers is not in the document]
loris-s-sonarsource: Strange, in the rendering tests I did, it seems to work fine

rules/S5131/csharp/aspnetcore.http/rule.adoc
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+app.Run();
+----
+
+=== Pitfalls
+
+include::../../common/pitfalls/deny-lists.adoc[]
+
+If the application needs to handle rich content, the preferred option is to use
```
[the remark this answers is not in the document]
loris-s-sonarsource: actually that is a good point, this is somehow related to the pitfall but it looks unrelated, I will look into it

rules/S5131/common/pitfalls/deny-lists.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+The creation of deny lists is a typical mechanism to stop a particular attack
+without fixing the underlying problem. Such filters, considered a gimmick by
+hackers, are found either in code or in web application firewalls.
+
+Unfortunately, creating a deny list is a common trap that should be avoided at
+all costs, as malicious users can easily change their exploit payloads and get
+back into the application.
+
+===== Error-prone Use Case
```
loris-s-sonarsource, line 11:
```suggestion
===== Error-prone use case
```

rules/S5131/common/fix/data_encoding.adoc
```diff
@@ -0,0 +1,7 @@
+Protect well-meaning users from the actions of malicious users by ensuring that
+user-supplied data cannot break out of its context when injected into a
+response. +
+To do this, encode the data according to its context. The following table lists
+how user input should be handled depending on its output context.
+
+Note: Encoding data does not change the content from the end user's point of view, but makes it uninterpretable by the end browser.
```
loris-s-sonarsource, line 7:
```suggestion
Protect well-meaning users from attacks by ensuring that third-party data cannot compromise the integrity of the services you provide. XSS attacks are facilitated when the web service inserts third-party data without controlling it first: This data breaks out of the user context.

Third-party data sources include the application's user base and also third-party services. Any user can be compromised or malicious from one moment to the next, and any third-party provider can be compromised.

The end user's browser distinguishes between two different forms of text:
- Text that is intended to be displayed to humans.
- Text that is intended to be interpreted as code by the browser.

As long as the third-party data stays un-encoded ("unescaped"), the browser considers it as both of the above forms of text. This situation makes this data harmful because the third-party data could mix HTML and JavaScript.

Therefore, to protect users, you should encode all third-party data according to its context.

- If the third-party data needs to be inserted into text that is directly intended for humans, it must be HTML-encoded.
- If the third-party data needs to be inserted into text that is to be used directly in JavaScript code, it must be JavaScript-encoded.
```

loris-s-sonarsource: This new suggestion could also partly fit in the rationale

rules/S5131/csharp/razor/rule.adoc
```diff
@@ -10,23 +10,248 @@ include::../../threats.adoc[]
 
 == How to fix it?
 
+include::../../common/fix/intro.adoc[]
```
loris-s-sonarsource, line 13: to remove

pierre-loup-tristant-sonarsource:
For the sake of simplicity and to wrap up the work I took the freedom of removing content that I think could go into "going the extra mile" or "pitfalls". 

As an example, I removed the part that mentions "AntiXssEncoder.HtmlEncode" because it would require explaining a lot of legacy and backward compatibility design choices from the past. (See [this thread](https://stackoverflow.com/questions/63149066/what-is-difference-between-webutility-htmlencode-and-antixssencoder-htmlencode))> In the end I think it could be useful for people using legacy version but for most people don't need to know all of this and should just use `System.Web.HttpUtility.HtmlEncode`.

rules/S5131/csharp/razor/rule.adoc
```diff
@@ -11,7 +11,7 @@ include::../../threats.adoc[]
 == How to fix it?
 
 The following code is vulnerable to Cross-Site Scripting because auto-escaping of special HTML characters has been disabled.
-Enable auto-escaping by moving the dynamic HTML content to the template engine.
+The recommended way to fix this code is to move the HTML content to the template and to only inject the dynamic value. Therefore, no need to disable auto-escaping.
```
hendrik-buchwald-sonarsource, line 14:
```suggestion
The recommended way to fix this code is to move the HTML content to the template and to only inject the dynamic value. Therefore, it is not necessary to disable auto-escaping.
```

rules/S5131/csharp/razor/rule.adoc
```diff
@@ -10,30 +10,94 @@ include::../../threats.adoc[]
...
+Implementing content security policies differs from one architecture to another.
+
+* When client-side code is processed by Blazor
+** {blazor-csps}[the library supports it by default].
+* When CSP requirements are low
+** {aspdotnet-core-middleware}[the ASP.Net Core Middleware provides header control].
+* When code requirements enforce optimal CSP, including content security policies nonces
+** The Open Web Interface for . NET (OWIN) {csp-nonces-owin}[can be used for a simple implementation].
+
+
```
pierre-loup-tristant-sonarsource, line 89: What we have in this section is not mature enough in my opinion:
- It's lacking "in practice" advice to enable CSP
- It's not explaining the benefit and limitations of CSP

I would suggest removing it for now and to provide additional content late.

PR #946, opened 2022-04-11
Modify rule S6377: Enhance the Noncompliant / Compliant examples

agigleux:


rules/S6377/java/rule.adoc
```diff
@@ -2,21 +2,32 @@ include::../description.adoc[]
...
 
 [source,java]
 ----
-DOMValidateContext valContext = new DOMValidateContext(new KeyValueKeySelector(), nl.item(0));
-valContext.setProperty("org.jcp.xml.dsig.secureValidation", Boolean.TRUE);
+NodeList signatureElement = doc.getElementsByTagNameNS(XMLSignature.XMLNS, "Signature");
+
+XMLSignatureFactory fac = XMLSignatureFactory.getInstance("DOM");
+DOMValidateContext valContext = new DOMValidateContext(new KeyValueKeySelector(), signatureElement.item(0));
+valContext.setProperty("org.jcp.xml.dsig.secureValidation", Boolean.TRUE); // Compliant
```
loris-s-sonarsource, line 27: the "Compliant" comment is deprecated

```suggestion
valContext.setProperty("org.jcp.xml.dsig.secureValidation", Boolean.TRUE);
```

rules/S6377/description.adoc
```diff
@@ -11,5 +11,3 @@ To protect the application from these vulnerabilities, the ``org.jcp.xml.dsig.se
 * Forbids Reference URIs of type ``http``, ``https``, or ``file``
 * Does not allow a ``RetrievalMethod`` element to reference another ``RetrievalMethod`` element
 * Forbids RSA or DSA keys less than 1024 bits
-
-In addition, the ``jdk.xml.dsig.secureValidationPolicy`` Security Property can be used to control and fine-tune the restrictions listed previously or add additional restrictions.
```
loris-s-sonarsource, line 15: Agreed on the deletion. We do not demonstrate it, but demonstrating would be out-of-scope for this description.

rules/S6377/java/rule.adoc
```diff
@@ -2,21 +2,32 @@ include::../description.adoc[]
 
 == Noncompliant Code Example
 
-The Java XML Digital Signature API doesn't use a strong signature validation mode by default (except when the application runs with a security manager):
```
loris-s-sonarsource, line 5: +1, agreed on the deletion

rules/S6377/java/rule.adoc
```diff
@@ -2,21 +2,32 @@ include::../description.adoc[]
...
 ----
-DOMValidateContext valContext = new DOMValidateContext(new KeyValueKeySelector(), nl.item(0)); // Noncompliant
+NodeList signatureElement = doc.getElementsByTagNameNS(XMLSignature.XMLNS, "Signature");
+
+XMLSignatureFactory fac = XMLSignatureFactory.getInstance("DOM");
+DOMValidateContext valContext = new DOMValidateContext(new KeyValueKeySelector(), signatureElement.item(0)); // Noncompliant
+XMLSignature signature = fac.unmarshalXMLSignature(valContext);
+
+// The Java XML Digital Signature API doesn’t use a strong signature validation mode by default (except when the application runs with a security manager)
+boolean signatureValidity = signature.validate(valContext);
```
loris-s-sonarsource, line 14: Oh, honestly I think that the code was short but I'm not sure about adding all of this. Now that you did it I would say that it should stay.

However, I suggest we move the comment to the very top of the description, as Ann suggested

rules/S6377/description.adoc
```diff
@@ -1,7 +1,7 @@
 XML signature validations work by parsing third-party data that cannot be trusted until it is actually validated.
 
 As with any other parsing process, unrestricted validation of third-party XML signatures can lead to security vulnerabilities. In this case, threats range from denial of service to confidentiality breaches. +
```
loris-s-sonarsource, line 3:
```suggestion
As with any other parsing process, unrestricted validation of third-party XML signatures can lead to security vulnerabilities. In this case, threats range from denial of service to confidentiality breaches.

By default, the Java XML Digital Signature API does not apply restrictions on XML signature validation, unless the application runs with a security manager. +
```

I suggest we take the comment from the noncompliant section and modify it to add it here.

rules/S6377/description.adoc
```diff
@@ -1,7 +1,7 @@
...
-To protect the application from these vulnerabilities, the ``org.jcp.xml.dsig.secureValidation`` attribute enforces the following restrictions:
+To protect the application from these vulnerabilities, the DOMValidateContext's ``org.jcp.xml.dsig.secureValidation`` attribute from the Java XML Digital Signature API enforces the following restrictions:
```
loris-s-sonarsource, line 4: Honestly, I'd say that adding the namespace instead of paraphrasing could be even more concise and clear:

```suggestion
To protect the application from these vulnerabilities, set the ``org.jcp.xml.dsig.secureValidation`` attribute with the``javax.xml.crypto.dsig.dom.DOMValidateContext.setProperty`` method. +
This attribute ensures that the code enforces the following restrictions:
```

WDYT?

rules/S6377/java/rule.adoc
loris-s-sonarsource:
```suggestion
NodeList signatureElement = doc.getElementsByTagNameNS(XMLSignature.XMLNS, "Signature");

XMLSignatureFactory fac = XMLSignatureFactory.getInstance("DOM");
DOMValidateContext valContext = new DOMValidateContext(new KeyValueKeySelector(), signatureElement.item(0)); // Noncompliant
XMLSignature signature = fac.unmarshalXMLSignature(valContext);

boolean signatureValidity = signature.validate(valContext);
```

I think removing it will also ease code reading (The suggestion removes the comment line).

rules/S6377/java/rule.adoc
```diff
@@ -2,21 +2,32 @@ include::../description.adoc[]
...
 ----
 
 == Compliant Solution
 The Java XML Digital Signature API offers a secure validation mode to protect against various https://docs.oracle.com/en/java/javase/14/security/java-xml-digital-signature-api-overview-and-tutorial.html#GUID-8618C294-3BFE-45C3-9A1E-C4629E337E68[security issues]. +
```
loris-s-sonarsource, line 18:
```suggestion
```

I didn't see it in my first review. We should remove this line since we already list the security issues in the description (it's the same list).

loris-s-sonarsource (changes requested):
Nice! We are almost there.

Also, I suggest changing the content of `message.adoc` to what you specified in the Jira Ticket.

> Set the ‘org.jcp.xml.dsig.secureValidation’ property to true on the ‘DOMValidateContext’ to validate this XML signature securely.

rules/S6377/description.adoc
agigleux: I really prefer this version:

To protect the application from these vulnerabilities, the DOMValidateContext's ``org.jcp.xml.dsig.secureValidation`` attribute from the Java XML Digital Signature API enforces the following restrictions:

loris-s-sonarsource: Ok! My opinion is that not having the method to actually add it might slow down the fix.

Here are two more arguments:

1. It's a duplicate explanation since the message already states:
    > Set the ‘org.jcp.xml.dsig.secureValidation’ property to true on the ‘DOMValidateContext’ to validate this XML signature securely.
2. Writing both "Java XML Digital Signature API" and "org.jcp.xml.dsig.secureValidation" is also a duplicate of the same meaning

PR #955, opened 2022-04-19
Modify rule S6212: Fix typos

hendrik-buchwald-sonarsource:


rules/S6212/java/rule.adoc
```diff
@@ -3,7 +3,7 @@ In Java 10 https://openjdk.java.net/jeps/286[Local-Variable Type Inference] was
 While it is not always possible or cleaner to use this new way of declaring a variable, when the type on the left is the same as the one on the right in an assignment, using the ``++var++`` will result in a more concise code.
 
 
-This rule reports an issue when the expected type of the variable is the same as the returned type of assigned expression and the type can be easily inferred by the reader, either when the tye is already mentioned in the name or the initializer, or when the expression is self-explained.
+This rule reports an issue when the expected type of the variable is the same as the returned type of assigned expression and the type can be easily inferred by the reader, either when the type is already mentioned in the name or the initializer, or when the expression is self-explanatory.
```
dorian-burihabwa-sonarsource, line 6:
```suggestion
This rule reports an issue when the expected type of the variable is the same as the return type of the assigned expression and the type can be easily inferred by the reader (either when the type is already mentioned in the name or initializer, or when the expression is self-explanatory).
```

dorian-burihabwa-sonarsource (changes requested):
The changes look good to me but I suggest we go further and break down this long sentence as suggested.

rules/S6212/java/rule.adoc
hendrik-buchwald-sonarsource: It is a pretty long sentence, though using parenthesis here would be against the [new styling guide](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc#parentheses).

dorian-burihabwa-sonarsource: Ok then, I will accept the change as-is 👍🏿

PR #961, opened 2022-04-21
Modify rule S5131: DTL autoescapeing globably disabled

pierre-loup-tristant-sonarsource:
I also experimented with putting the code examples in tables to better simulate diff views.

### DIFF
--- a/rules/S5131/python/dtl/rule.adoc
+++ b/rules/S5131/python/dtl/rule.adoc
@@ -10,9 +10,7 @@ include::../../threats.adoc[]
 
 == How to fix it?
 
-The following code is vulnerable to Cross-Site Scripting because Jinja auto-escaping of special HTML characters has been disabled.
-Enable auto-escaping by moving the dynamic HTML content to the template engine.
-
+The following code is vulnerable to Cross-Site Scripting because auto-escaping of special HTML characters has been disabled. The recommended way to fix this code is to move the HTML content to the template and to only inject the dynamic value. Therefore, no need to disable auto-escaping. 
 
 [cols="a,a"]
 |===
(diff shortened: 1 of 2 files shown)


rules/S5131/python/dtl/rule.adoc
```diff
@@ -51,6 +51,24 @@ Template engines are used by web applications to build HTML content. Template fi
 
 In Django applications, the engine's auto-escaping feature is enabled by default. XSS vulnerabilities arise when an untrusted value is injected in the template and that auto-escaping is disabled using ``++{% autoescape false %}++`` or ``++|safe++`` filter. This is often the case when a piece of dynamic HTML is generated from the Python code and used in a template variable.
 
+Auto-escaping may be disabled at application level causing XSS vulnerabilities even if ``++{% autoescape false %}++`` or ``++|safe++`` are not being used.
```
pierre-loup-tristant-sonarsource, line 54: I have the feeling this should go to the "How to fix it?" section but I also think it's a secondary case.

rules/S5131/python/dtl/rule.adoc
```diff
@@ -39,76 +35,145 @@ def hello(request):
...
 ----
 <!doctype html>
 <h1>Hello {{ name }}</h1>
 ----
 
+|===
+
 === How does this work?
 
 Template engines are used by web applications to build HTML content. Template files contain static HTML as well as template language instruction. These instructions allow, for example, to insert dynamic values in the document as the template is rendered. Template engines can auto-escape HTML special characters of dynamic values in order to prevent XSS vulnerabilities.
```
hendrik-buchwald-sonarsource, line 58:
```suggestion
Template engines are used by web applications to build HTML content. Template files contain static HTML as well as template language instructions. These instructions allow, for example, to insert dynamic values in the document as the template is rendered. Template engines can auto-escape HTML special characters of dynamic values in order to prevent XSS vulnerabilities.
```

rules/S5131/python/dtl/rule.adoc
```diff
@@ -39,76 +35,145 @@ def hello(request):
...
 
 In Django applications, the engine's auto-escaping feature is enabled by default. XSS vulnerabilities arise when an untrusted value is injected in the template and that auto-escaping is disabled using ``++{% autoescape false %}++`` or ``++|safe++`` filter. This is often the case when a piece of dynamic HTML is generated from the Python code and used in a template variable.
```
hendrik-buchwald-sonarsource, line 60: The `that` in the second sentence seems out of place.

rules/S5131/python/dtl/rule.adoc
```diff
@@ -39,76 +35,145 @@ def hello(request):
...
+----
+
+|===
+
+
+
+
+
+
+When HTML templates are rendered, this option should be set to ``++True++``
```
hendrik-buchwald-sonarsource, line 102: I am not sure about this sentence. For one, the default value is likely `True`, so this option does not necessarily have to be set to `True` again. Also, I do not think it is clear what `When HTML templates are rendered` means. I know what it means but I am not sure it will be clear to the normal reader.
Maybe it would help to quickly mention before where it is fine to disable the auto-escaping?

rules/S5131/python/dtl/rule.adoc
```diff
@@ -39,76 +35,145 @@ def hello(request):
...
+
 === Pitfalls
 
 Although auto-escaping drastically decreases the chance of introducing Cross-Site Scripting vulnerabilities, there are still specific cases where vulnerabilities can occur.
 
 ==== Unquoted variables in attributes
```
hendrik-buchwald-sonarsource, line 108: All these sections will be replaced by the shared common sections, right?

rules/S5131/python/jinja/rule.adoc
```diff
@@ -10,8 +10,11 @@ include::../../threats.adoc[]
 
 == How to fix it?
 
-The following code is vulnerable to Cross-Site Scripting because Jinja auto-escaping of special HTML characters has been disabled.
+The following code is vulnerable to Cross-Site Scripting because auto-escaping of special HTML characters has been disabled. Enable auto-escaping by moving the dynamic HTML content to the template engine.
```
hendrik-buchwald-sonarsource, line 13: The last sentence says that moving the HTML content will enable auto-escaping. That is not true though, auto-escaping still has to be enabled by the developer. Moving the HTML content is simply a requirement to disable it.

hendrik-buchwald-sonarsource (approved):
LGTM in general, but some small comments remain.

rules/S5131/python/dtl/rule.adoc
pierre-loup-tristant-sonarsource: I will simply remove it.

rules/S5131/python/dtl/rule.adoc
pierre-loup-tristant-sonarsource: I will only keep the one about "Variables in script blocks" because the Django template provides a special filter for this purpose.

PR #972, opened 2022-04-27
Modify rule S6245: Add python as covered language

nils-werner-sonarsource:


rules/S6245/python/rule.adoc
```diff
@@ -0,0 +1,48 @@
...
+bucket = s3.Bucket(self,"MyEncryptedBucket",
+    encryption=s3.BucketEncryption.S3_MANAGED       # Compliant
+)
+
+# Altarnativly with a KMS key managed by the user.
+
+bucket = s3.Bucket(self,"MyEncryptedBucket",
+    encryptionKey=access_key                        # Compliant
+)
+----
```
marco-bearzi-sonarsource, line 34: Maybe another compliant code where the key will be defined automatically could be a nice addition : 
```
bucket = s3.Bucket(self,"MyEncryptedBucket",
    encryption=s3.BucketEncryption.KMS       # Compliant
)
```
I'm not sure about this given the scope of the MMF, but adding a resource policy denying any unencrypted operation may also produce the same effect : 
```
bucket = s3.Bucket(self,"MyEncryptedBucket")
bucket.add_to_resource_policy(iam.PolicyStatement(
		effect=iam.Effect.DENY,
        principals=["*"],
        actions=["*"],
        resources=["*"],
		conditions={"NumericEquals" : {"s3:encryption": s3.BucketEncryption.UNENCRYPTED}}
	)
)
```

rules/S6245/python/rule.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+Server-side encryption is not used:
+
+[source,python]
+----
+bucket = s3.Bucket(self,"MyUnencryptedBucket",
+    encryption=s3.BucketEncryption.UNENCRYPTED       # Sensitive
+)
+----
+The default value of `encryption` is `KMS` if `encryptionKey` is set. Otherwise, if both parameter are absent the bucket is unencrypted.
```
marco-bearzi-sonarsource, line 17: Small mistake : "parameters" instead of "parameter"

rules/S6245/python/rule.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+Server-side encryption with Amazon S3-Managed Keys is used:
+
+[source,python]
+----
+bucket = s3.Bucket(self,"MyEncryptedBucket",
+    encryption=s3.BucketEncryption.S3_MANAGED       # Compliant
+)
+
+# Altarnativly with a KMS key managed by the user.
```
marco-bearzi-sonarsource, line 29: Small mistake : "Alternatively" instead of "Altarnativly"

marco-bearzi-sonarsource (commented):
A few minor comments

rules/S6245/python/rule.adoc
nils-werner-sonarsource, line 29:
```suggestion
# Alternatively with a KMS key managed by the user.
```

rules/S6245/python/rule.adoc
nils-werner-sonarsource, line 17:
```suggestion
The default value of `encryption` is `KMS` if `encryptionKey` is set. Otherwise, if both parameters are absent the bucket is unencrypted.
```

pierre-loup-tristant-sonarsource (approved):
LGTM.
Suggestions: 
- Remove `# Compliant` comment as this is the current convention.
- Use a more neutral resource name like "example". I know it's not enforced everywhere.

PR #975, opened 2022-04-28
Modify rule S5131: Add context specific patch instructions for Flask and Django

pierre-loup-tristant-sonarsource:


hendrik-buchwald-sonarsource (approved):
LGTM

PR #978, opened 2022-05-03
Modify rule S5131: align with other rules structure

pierre-loup-tristant-sonarsource:


rules/S5131/python/jinja/rule.adoc
```diff
@@ -57,57 +57,11 @@ In Flask applications, the Jinja auto-escaping feature is enabled by default. XS
...
-<a href="{{ url_for('index') }}">...</a>
-----
-|===
-
 ==== Variables in script blocks
 
-If ``++name++`` is untrusted, JavaScript code can be injected.
-
-Use ``++tojson++`` filter to insert a data structure in the JavaScript code at render time.
+Although auto-escaping drastically decreases the chance of introducing Cross-Site Scripting vulnerabilities, there are still specific cases where vulnerabilities can occur.
```
loris-s-sonarsource, line 62:
```suggestion
Although auto-escaping drastically decreases the chance of introducing cross-site scripting vulnerabilities, there are still specific cases where vulnerabilities can occur.
```

PR #987, opened 2022-05-12
Adapt images of educational rule descriptions

roberto-orlandi-sonarsource:


rules/S5131/rationale.adoc
```diff
@@ -1,11 +1,11 @@
...
-A user with malicious intent carefully crafts the link beforehand. Here is an example:
-
-image:common/images/url.png[]
+A user with malicious intent carefully crafts the link beforehand.
+// Here is an example:
 
-After creating this link, the attacker must use phishing techniques to ensure that his target users click on the link.
+// image:common/images/url.png[]
+// After creating this link, the attacker must use phishing techniques to ensure that his target users click on the link.
+After creating such link, the attacker must use phishing techniques to ensure that his target users click on the link.
```
christophe-zurn-sonarsource, line 11: Is the change of `After creating this link` -> `After creating such link` really needed?

roberto-orlandi-sonarsource: In my opinion yes. It refers to the link mentioned in line 6. Basically this is what we read after the modification I did:
```
A user with malicious intent carefully crafts the link beforehand.
After creating such link, the attacker must use phishing techniques to ensure that his target users click on the link.
```

PR #992, opened 2022-05-16
RULEAPI-760: Add a new language identifier for kubernetes rules

loris-s-sonarsource:
# Note
Full name: "Add a new language identifier for kubernetes YAML misconfiguration detection rules"

# References
- Jira: [**RULEAPI-760**](https://jira.sonarsource.com/browse/RULEAPI-760)
- RSPEC: https://github.com/SonarSource/rspec/pull/992
- sonar-rule-api: https://github.com/SonarSource/sonar-rule-api/pull/193
- rules.sonarsource.com: https://github.com/SonarSource/rules.sonarsource.com/pull/145


supported_languages.adoc
```diff
@@ -1 +1 @@
-`abap`, `apex`, `cfamily`, `cobol`, `csharp`, `css`, `flex`, `go`, `html`, `java`, `javascript`, `kotlin`, `php`, `pli`, `plsql`, `python`, `rpg`, `ruby`, `rust`, `scala`, `secrets`, `solidity`, `swift`, `text`, `tsql`, `vb6`, `vbnet`, `cloudformation`, `terraform`, or `xml`
+`abap`, `apex`, `cfamily`, `cobol`, `csharp`, `css`, `flex`, `go`, `html`, `java`, `javascript`, `kotlin`, `php`, `pli`, `plsql`, `python`, `rpg`, `ruby`, `rust`, `scala`, `secrets`, `solidity`, `swift`, `text`, `tsql`, `vb6`, `vbnet`, `cloudformation`, `terraform`, `kubernetes yamls` or `xml`
```
arseniy-sonar, line 1: One of the uses of this list is the validation of subfolders in a rule folder. I.e., a rule folder `rules/S100` can have only language-specific subfolders, and only for the supported languages - that are listed here. So for Kubernetes that would make it look for `kubernetes yamls` subfolder (with a space).
Even though we live in 2022, I still strongly advise against spaces in the paths when there is no strong need for them.
Moreover to keep the style consistent with other languages, I suggest using only `[a-z0-9]` characters. How about just `kubernetes` as you used in other places?

arseniy-sonar (changes requested):
Regarding rspec-tools/rspec_tools/utils.py: `LANG_TO_SOURCE`, is there a highlighting mode in highlight.js supporting Kubernetes? Maybe just YAML?

arseniy-sonar:
Also, PTAL at the test failures, I think you can use them for guidance of what other places need to be updated with the new "language"

supported_languages.adoc
loris-s-sonarsource: Thanks for this. The "or" made me think that this list was just intended to be converted to html.

arseniy-sonar (approved):
Looks good :+1:

PR #1004, opened 2022-05-20
SONARSEC-3040 Update checks and documentation to fit new rule format

roberto-orlandi-sonarsource:


rspec-tools/rspec_tools/rules.py
```diff
@@ -63,7 +64,8 @@ def id(self) -> str:
 
   @property
   def specializations(self) -> Generator[LanguageSpecificRule, None, None]:
-    return (LanguageSpecificRule(child, self) for child in self.rule_path.iterdir() if child.is_dir())
+    return (LanguageSpecificRule(child, self) for child in self.rule_path.iterdir() if
+            child.is_dir() and child.name in utils.load_valid_languages())
```
christophe-zurn-sonarsource, line 68: Is there a unit test somewhere we could also add/update to ensure this behaves as expected ?

rspec-tools/rspec_tools/rules.py
```diff
@@ -1,5 +1,6 @@
 
 import json
+from . import utils
```
christophe-zurn-sonarsource, line 3: Regarding our quick talk about import, if I look in other files, the syntax used for importing local functions is `from rspec_tools.utils import ...` (see https://github.com/SonarSource/rspec/blob/4cd575af12dc869554f642fcd2d23b8150a4759c/rspec-tools/rspec_tools/create_rule.py#L11). I'd suggest to keep the same for consistency

ci/validate_asciidoc.sh
```diff
@@ -60,20 +62,23 @@ do
...
-        # We build this filename that describes the path to workaround the fact that asciidoctor will not tell
-        # us the path of the file in case of error.
-        # We can remove it if https://github.com/asciidoctor/asciidoctor/issues/3414 is fixed.
-        TMP_ADOC="$language/tmp_$(basename "${dir}")_${language##*/}.adoc"
-        echo "== Description" > "$TMP_ADOC"
-        cat "$RULE" >> "$TMP_ADOC"
+        if [[ ! "${ALLOWED_RULE_SUB_FOLDERS[*]}" == *"${language##*/}"* ]]; then
+          echo "ERROR: ${language##*/} is not a supported language"
+          exit_code=1
+        fi
```
christophe-zurn-sonarsource, line 68: I'd even suggest adding a check that if we are in an allowed subfolder, there is no `rule.adoc` file, otherwise we throw an error: that way we ensure that there will be no overlap between supported languages and allowed subfolder names

christophe-zurn-sonarsource (changes requested):
Overall changes look good. I think there could be a couple of improvements. Let me know if you want to discuss some of the comments I've left

rspec-tools/rspec_tools/rules.py
roberto-orlandi-sonarsource: good point. I got confused because PyCharm raises an error with this syntax :)

christophe-zurn-sonarsource (approved):
LGTM.
I let you check why there is failing check before merging

PR #1009, opened 2022-05-23
Modify rules S1774,S3358: Add exception for C++11 constexpr functions

tomasz-kaminski-sonarsource:


rules/S1774/cfamily/rule.adoc
```diff
@@ -1,5 +1,12 @@
 include::../rule.adoc[]
 
+== Exceptions
+
+For {cpp}11 mode only, the issue is not raised for ternary operators used inside `constexpr` functions. 
+In {cpp}11 such functions are limited to single statement, that is `return` so use of ternary operator
```
frederic-tingaud-sonarsource, line 6:
```suggestion
In {cpp}11 such functions are limited to a single return statement, so the use of a ternary operator
```

rules/S3358/cfamily/rule.adoc
```diff
@@ -22,6 +22,13 @@ int max(int p1, int p2, int p3) {
 }
 ----
 
+== Exceptions
+
+For {cpp}11 mode only, the issue is not raised for ternary operators used inside `constexpr` functions. 
+In {cpp}11 such functions are limited to single statement, that is `return` so use of ternary operator
```
frederic-tingaud-sonarsource, line 28:
```suggestion
In {cpp}11 such functions are limited to a single return statement, so the use of a ternary operator
```

frederic-tingaud-sonarsource (commented):
Proposing a slight rewording.

rules/S1774/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This does not convey wording. Being limited to `single return statement` can be read as having multiple statements, but only one `return`. But this function can have only one statement, that is also `return`.

tomasz-kaminski-sonarsource: Looks good.

frederic-tingaud-sonarsource (approved):

PR #1010, opened 2022-05-23
SONARSEC-3035 Adapt images of educational rule descriptions

roberto-orlandi-sonarsource:


rules/S5131/rationale.adoc
```diff
@@ -1,11 +1,11 @@
...
-A user with malicious intent carefully crafts the link beforehand. Here is an example:
-
-image:common/images/url.png[]
+A user with malicious intent carefully crafts the link beforehand.
+// Here is an example:
 
-After creating this link, the attacker must use phishing techniques to ensure that his target users click on the link.
+// image:common/images/url.png[]
+// After creating this link, the attacker must use phishing techniques to ensure that his target users click on the link.
+After creating such link, the attacker must use phishing techniques to ensure that his target users click on the link.
```
roberto-orlandi-sonarsource, line 11: @christophe-zurn-sonarsource I created again the branch, so your previous comment here got lost, sorry for that! You can still find it here though:
https://github.com/SonarSource/rspec/pull/987#pullrequestreview-971853344
Let me know if we still think it is not needed to mention "After creating such link"?

christophe-zurn-sonarsource: I still think the change is not necessary, as I believe the original content still transmits the same meaning, i.e. :
```
A user with malicious intent carefully crafts the link beforehand.
After creating this link, the attacker must use phishing techniques to ensure that his target users click on the link.
```
has the same meaning as your changes. Although I admit I might be missing a subtlety somewhere... I let you decide whether to revert to original or keep you current changes

christophe-zurn-sonarsource (approved):

PR #1025, opened 2022-06-01
Modify rule S3806: Also report include directives with a trailing space (CPP-3710)

loic-joly-sonarsource:


rules/S3806/cfamily/rule.adoc
```diff
@@ -1,11 +1,19 @@
-The path provided here doesn't match the actual path on this file system (e.g. the case is different). While this may work on a particular environment, this is not portable and may fail on a different environment.
+The way an `#include` directive finds an actual file is implementation-defined, and in practice, it slightly differs in different systems.
```
amelie-renard-sonarsource, line 1:
```suggestion
The way an `#include` directive finds an actual file is implementation-defined, and slightly differs in practice in different systems.
```

rules/S3806/cfamily/rule.adoc
```diff
@@ -1,11 +1,19 @@
...
+
+Therefore, a good practice is to identify the files to include in the most straightforward way possible to reduce the risk of inconsistent behaviors.
+
+This rule raises an issue when:
+
+* The case of the file in the `#include` directive does not match the case of the file on the disk (the inclusion would not work on a case-sensitive OS),
+* The file name in the `#include` directive contains trailing spaces (they would be ignored on Windows but considered on Unix).
```
amelie-renard-sonarsource, line 8:
```suggestion
* The case of the file in the `#include` directive does not match the case of the file on the disk (the inclusion would not work on a case-sensitive OS)
* The file name in the `#include` directive contains trailing spaces (they would be ignored on Windows but considered on Unix)
```

amelie-renard-sonarsource (changes requested):

PR #1027, opened 2022-06-02
[Education] Improve language

hendrik-buchwald-sonarsource:
These are (probably) not changes that will be part of the first release anymore, but I thought it still makes sense to update it so that we do not forget about it.


hendrik-buchwald-sonarsource:
Closing this PR since it is based on the wrong branch.

PR #1030, opened 2022-06-06
CPP-220 S3135 "sizeof" should not be called on pointer parameters

loic-joly-sonarsource:
Undeprecate S1913

### DIFF
--- a/rules/S1913/cfamily/metadata.json
+++ b/rules/S1913/cfamily/metadata.json
@@ -1,7 +1,7 @@
 {
   "title": "\"sizeof(sizeof(...))\" should not be used",
   "type": "BUG",
-  "status": "deprecated",
+  "status": "ready",
   "remediation": {
     "func": "Constant\/Issue",
     "constantCost": "20min"
@@ -10,9 +10,6 @@
 
   ],
   "extra": {
-    "replacementRules": [
-      "RSPEC-3135"
-    ],
     "legacyKeys": [
       "SizeofSizeof"
     ]
--- a/rules/S1913/cfamily/metadata.json
+++ b/rules/S1913/cfamily/metadata.json
@@ -19,7 +19,7 @@
   "sqKey": "S1913",
   "scope": "Main",
   "defaultQualityProfiles": [
-
+    "Sonar way"
   ],
   "quickfix": "unknown"
 }
--- a/rules/S3135/cfamily/metadata.json
+++ b/rules/S3135/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "\"sizeof\" should not be called on pointer parameters",
+  "title": "\"sizeof\" should not be called on pointer",
   "type": "BUG",
   "status": "ready",
   "remediation": {
--- a/rules/S3135/cfamily/metadata.json
+++ b/rules/S3135/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "\"sizeof\" should not be called on pointer",
+  "title": "\"sizeof\" should not be called on pointers",
   "type": "BUG",
   "status": "ready",
   "remediation": {
@@ -27,7 +27,7 @@
       467
     ],
     "CERT": [
-      "ARR01-C.",
+      "ARR01-C."
     ]
   },
   "defaultQualityProfiles": [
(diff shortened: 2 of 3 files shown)


rules/S3135/cfamily/rule.adoc
```diff
@@ -1,40 +1,19 @@
-``++sizeof++`` returns the size in bytes of a type. ``++alignof++`` (since {cpp}-11) returns the alignment, in bytes, required for any instance of a type. Using any other kind of value with ``++sizeof++``, or ``++alignof++``, is likely a mistake - what is returned is the size, or alignment, of the operand's type.
+`sizeof` returns the size in bytes of a type. One common usage pattern, especially in C, is to use `sizeof` to determine the size of an array. However, arrays decay to pointers when passed as arguments to a function, and if `sizeof` is applied to such an argument, it will return the size of the pointer, not of the array. A similar issue happens when the array is used in an arithmetic operation.
 
+This rule raises issues when `sizeof` is used to compute the array size of a pointer passed as a function argument or used in an arithmetic operation.
```
amelie-renard-sonarsource, line 3: I find the last part of the sentence a bit misleading: the problem is not when `sizeof` itself is part of an arithmetic operation but when it takes an arithmetic operation involving an array as an argument. 
What about something like:
```suggestion
This rule raises issues when:
* `sizeof` is used to compute the array size of a pointer passed as a function argument
* `sizeof` is called with an arithmetic operation involving an array
```

amelie-renard-sonarsource (changes requested):

rules/S3135/cfamily/metadata.json
```diff
@@ -29,13 +28,11 @@
     ],
     "CERT": [
       "ARR01-C.",
```
amelie-renard-sonarsource, line 30:
```suggestion
      "ARR01-C."
```

amelie-renard-sonarsource (changes requested):

rules/S1913/cfamily/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "\"sizeof(sizeof(...))\" should not be used",
   "type": "BUG",
-  "status": "deprecated",
+  "status": "ready",
```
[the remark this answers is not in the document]
loic-joly-sonarsource: Yes, we should.

rules/S3135/cfamily/rule.adoc
```diff
@@ -1,40 +1,22 @@
...
-
-
-* with parameters that have a ``++void++`` type. That is forbidden by both the C and {cpp} standards
-* on a pointer type variable. This is typically a mistake; usually the intent is to get the size, or alignment, of the pointed-to value instead
-* when the parameter is another ``++sizeof++``, or ``++alignof++`` call. ``++sizeof(sizeof(...))++`` indicates a misuse or misunderstanding of the ``++sizeof++`` construct
-* when side effects are intended in the parameter. The side effects don't actually occur; the expression is not evaluated because ``++sizeof++`` only acts on the type of the expression
-* when there are calculations within ``++sizeof++``, or ``++alignof++``; the result will be ``++sizeof(int)++``, or ``++alignof(int)++``, assuming the calculation's operands are ``++int++``s
-* when numeric constants are used with ``++sizeof++`` or ``++alignof++``. Such calls simply return ``++sizeof(int)++`` or ``++alignof(int)++`` respectively.
+* `sizeof` is used to compute the array size of a pointer passed as a function argument.
+* `sizeof` is called on the result of an arithmetic operation involving an array.
```
amelie-renard-sonarsource, line 6:
```suggestion
* `sizeof` is used to compute the array size of a pointer passed as a function argument
* `sizeof` is called on the result of an arithmetic operation involving an array
```

amelie-renard-sonarsource (changes requested):
Minor comment, and after that the PR can be merged

rules/S3135/cfamily/rule.adoc
loic-joly-sonarsource: I don't like these dots either, but I checked in the style guides that we use, and there is very little leeway for that: We cannot omit them because they form a full sentence...:
https://www.codot.gov/business/grants/safetygrants/assets/APStyleGuideCheatSheet.pdf
https://www.instructionalsolutions.com/blog/bulleted-list-punctuation

amelie-renard-sonarsource: [Styling guide](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc) is not precise on this subject.
Having a quick look at how rpsecs were currently written, we chose to keep the dots.

amelie-renard-sonarsource (approved):

PR #1033, opened 2022-06-08
RULEAPI-762: Fix link-probing cache: cache on failure and success

arseniy-sonar:


loic-joly-sonarsource (approved):
I'm not an expert with shell scripts... But LGTM.

PR #1044, opened 2022-06-13
Modify rule S4125: multiple updates

ilia-kebets-sonarsource:


rules/S4125/javascript/metadata.json
```diff
@@ -4,7 +4,7 @@
   "status": "ready",
   "remediation": {
     "func": "Constant\/Issue",
-    "constantCost": "5min"
+    "constantCost": "1min"
```
vilchik-elena, line 7: why? not sure it's always so easy to fix, as this operator is often misused (and not just mistyped)

vilchik-elena (changes requested):

rules/S4125/javascript/metadata.json
ilia-kebets-sonarsource: I thought the scope of our rule was only mistyped strings. As there are 8 valid values with no ambiguity between them, the issue looks pretty straightforward.

vilchik-elena (approved):

PR #1054, opened 2022-06-17
Modify rule S5542[kotlin]: Detect CBC mode when used with padding (APPSEC-30)

pierre-loup-tristant-sonarsource:
Specification ticket: [APPSEC-30](https://jira.sonarsource.com/browse/APPSEC-30)
Implementation ticket: [SONARKT-208](https://jira.sonarsource.com/browse/SONARKT-208)
Preview: [S5542](https://sonarsource.github.io/rspec/#/rspec/S5542)

### DIFF
--- a/rules/S5542/kotlin/rule.adoc
+++ b/rules/S5542/kotlin/rule.adoc
@@ -4,10 +4,11 @@ include::../description.adoc[]
 
 [source,kotlin]
 ----
-val c1 = Cipher.getInstance("AES") // Noncompliant: by default ECB mode is chosen
-val c2 = Cipher.getInstance("AES/ECB/NoPadding") // Noncompliant: ECB doesn't provide serious message confidentiality
+Cipher.getInstance("AES"); // Noncompliant: by default ECB mode is chosen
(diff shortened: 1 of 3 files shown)


rules/S5542/kotlin/rule.adoc
```diff
@@ -4,23 +4,24 @@ include::../description.adoc[]
 
 [source,kotlin]
 ----
-val c1 = Cipher.getInstance("AES") // Noncompliant: by default ECB mode is chosen
-val c2 = Cipher.getInstance("AES/ECB/NoPadding") // Noncompliant: ECB doesn't provide serious message confidentiality
+Cipher.getInstance("AES"); // Noncompliant: by default ECB mode is chosen
```
pedro-oliveira-sonarsource, line 7: Nothing wrong here as you can use both, but you could leave the semi-colon out so it's more Kotlin specific.

pedro-oliveira-sonarsource, line 7: Nothing wrong here as you can use both, but you could leave the semi-colon out so it's more Kotlin specific.
```suggestion
Cipher.getInstance("AES") // Noncompliant: by default ECB mode is chosen
```

pedro-oliveira-sonarsource (approved):
All good to me.

PR #1063, opened 2022-06-19
Modify rule S4225: Add VB.NET

Corniel:


rules/S4225/vbnet/rule.adoc
```diff
@@ -1,18 +1,18 @@
-include::../rule.adoc[]
+Creating an extension method that extends ``++Object++`` is not recommended because it makes the method available on _every_ type. Extensions should be applied at the most specialized level possible, and that is very unlikely to be ``++Object++``.
```
pavel-mikula-sonarsource, line 1: This rule structure is a bit messy. 
Please do this same change in C# version as well:
- move the introduction to `csharp/rule.adoc`, because there's the difference between `Object` and `object`. If the intro would be the same, than including `/rule.adoc` would be the correct way.
- move the C# example to `csharp/rule.adoc` file

pavel-mikula-sonarsource (changes requested):

rules/S4225/vbnet/rule.adoc
Corniel: @pavel-mikula-sonarsource Can you answer this?

pavel-mikula-sonarsource (changes requested):
Ah, sorry. I had my comments hanging here without posting them.

rules/S4225/vbnet/rule.adoc
pavel-mikula-sonarsource: Done, sorry. I've answered them and GitHub was showing them. I didn't notice that they are in "pending" state.

Corniel:
I'm willing to help out. But as there is zero documentation on how this works, there is no way I could even run it myself, I ask you to take the lead here. I have no problems with discovering how a black box works. Unless the person I'm trying to help, already knows...

pavel-mikula-sonarsource:
I'd say that it's fine if you do the changes blindly, without testing them. Just as you did for the VB part. I'll take care of validating / adjusting details.

The repo structure is documented in the readme https://github.com/SonarSource/rspec

As general concept:
* `rules/Sxxxx` is a common part of the rule. Whatever is here is meant to be shared by all languages
* `rules/Sxxxx/vbnet` starting point for a language-specific data. It can override stuff inherited from root (like metadata.json). It can include fragments of adoc files from root if it makes sense.

pavel-mikula-sonarsource:
To merge this, we need:
* update the C# part in a similar style as you did for VB
* remove namespace from VB.NET example
* add `title` to VB metadata.json to override it

pavel-mikula-sonarsource:
@Corniel do you plan to process this? Once it's done, we should be able to merge the rule.

Corniel:
@pavel-mikula-sonarsource As mentioned, I'm willing to put some effort in, but given the fact that Sonar (for whatever reason) does not allow contributors like me to have access to the full development kit, I think it should be someone with that full access to do most of the work.

pavel-mikula-sonarsource:
We don't have much more to do changes than what you have. Except the rule-api that is outside this repo.

rules/S4225/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
-{
-  
+﻿{
+  "title": "Extension methods should not extend \"object\""
```
pavel-mikula-sonarsource, line 2: Not needed - will be picked up from the root. I'll fix it.

pavel-mikula-sonarsource (approved):
LGTM

PR #1068, opened 2022-06-22
Modify rule S1788: Add exception for Redux reducers

victor-diez-sonarsource:
add exception to rule S1788 (JS) rule for Redux reducers


yassin-kammoun-sonarsource (changes requested):
Looks good overall. Just a minor thing in the exception snippet to address. Also, we need to add the `jsts` label to this PR. And while we are at it. I would rename the PR title to: `Modify rule S1788: Add exception for Redux reducers`

yassin-kammoun-sonarsource (approved):
LGTM!

PR #1072, opened 2022-06-23
CPP-2907 S1199: Rule is too strict for variables in case blocks

loic-joly-sonarsource:
Make the description more C++-oriented
Change the example (the previous one is now going to be compliant)

### DIFF
--- a/rules/S1199/cfamily/rule.adoc
+++ b/rules/S1199/cfamily/rule.adoc
@@ -1,8 +1,8 @@
-Nested code blocks can be used to create a new scope: Variables declared within that block cannot be accessed from the outside, and their lifetime end at the end of the block.
+Nested code blocks can be used to create a new scope: variables declared within that block cannot be accessed from the outside, and their lifetime end at the end of the block.
 
 While this might seem convenient, using this feature in a function often indicates that it has too many responsibilities and should be refactored into smaller functions.
 
-A nested code block is acceptable when it surrounds all the statements inside a `case` of a `switch` because it prevents variable declarations from polluting other `cases`.
+A nested code block is acceptable when it surrounds all the statements inside an alternative of a `switch` (a `case xxx:` or a `default:`) because it prevents variable declarations from polluting other `cases`.
 
 == Noncompliant Code Example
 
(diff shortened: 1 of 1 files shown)


rules/S1199/cfamily/rule.adoc
```diff
@@ -1,32 +1,36 @@
-include::../description.adoc[]
+Nested code blocks can be used to create a new scope: Variables declared within that block cannot be accessed from the outside, and their lifetime end at the end of the block.
```
amelie-renard-sonarsource, line 1:
```suggestion
Nested code blocks can be used to create a new scope: variables declared within that block cannot be accessed from the outside, and their lifetime end at the end of the block.
```

rules/S1199/cfamily/rule.adoc
```diff
@@ -35,34 +39,31 @@ void evaluate(OP op, Stack& stack) {
...
-    case OP::ADD: // Compliant
-      evaluateAdd(stack);
-      break;
-    /* ... */
+void f(Cache &c, int data) {
+  int value = getValue(c, data);
+
+  switch(value) {
+    case 1: 
+    { // Compliant, limits the scope or result
```
amelie-renard-sonarsource, line 58:
```suggestion
    { // Compliant, limits the scope of "result"
```

amelie-renard-sonarsource (changes requested):
Minor changes otherwise LGTM!

rules/S1199/cfamily/rule.adoc
```diff
@@ -1,32 +1,36 @@
-include::../description.adoc[]
+Nested code blocks can be used to create a new scope: Variables declared within that block cannot be accessed from the outside, and their lifetime end at the end of the block.
+
+While this might seem convenient, using this feature in a function often indicates that it has too many responsibilities and should be refactored into smaller functions.
+
+A nested code block is acceptable when it surrounds all the statements inside a `case` of a `switch` because it prevents variable declarations from polluting other `cases`.
```
amelie-renard-sonarsource, line 5:
```suggestion
A nested code block is acceptable when it surrounds all the statements inside a `case` or a `default` label of a `switch` because it prevents variable declarations from polluting other `cases`.
```

PR #1076, opened 2022-06-27
Modify rule S5415: CPP-3695 Improve the RSPEC

loic-joly-sonarsource:


rules/S5415/cfamily/rule.adoc
```diff
@@ -1,38 +1,38 @@
...
-This can happen when calling ``++std::move++`` with const argument or with non-movable types.
-
-
-This can also happen when passing the result of ``++std::move++`` as a const reference argument.  In this case, no object will be moved since the result is in a context where calling the move constructor is not possible. It makes sense to call ``++std::move++`` on a function argument when:
-
-* The result of ``++std::move++`` is passed as an r-value reference.
-* The result of ``++std::move++`` is passed by copy.
+* when `std::move` is called on an object which does not provide a specific move constructor and will resort to copying when requested to move.
+* when calling `std::move` with a const argument.
+* when passing the result of `std::move` as a const reference argument. In this case, no object will be moved since it's impossible to call the move constructor from within the function. `std::move` should only be used when the argument is passed by value or by r-value reference.
```
abbas-sabra-sonarsource, line 7:
```suggestion
* When `std::move` is called on an object which does not provide a specific move constructor and will resort to copying when requested to move.
* When calling `std::move` with a const argument.
* When passing the result of `std::move` as a const reference argument. In this case, no object will be moved since it's impossible to call the move constructor from within the function. `std::move` should only be used when the argument is passed by value or by r-value reference.
```

abbas-sabra-sonarsource (approved):
LGTM

PR #1078, opened 2022-06-28
Modify rule S6263: Add info about scope of rule and correct upper-case tag to be lower-case.

johann-beleites-sonarsource:
Add info about scope of rule and correct upper-case tag to be lower-case.


dorian-burihabwa-sonarsource (approved):

dorian-burihabwa-sonarsource (changes requested):
Thanks for the changes here but it looks like there are other occurrences of unnecessarily capitalized "Long" in the rule description.

rules/S6263/java/rule.adoc
```diff
@@ -1,5 +1,6 @@
 In AWS, Long-term access keys will be valid until you manually revoke them. This makes them highly sensitive as any exposure can have serious consequences and should be used with care.
```
dorian-burihabwa-sonarsource:
```suggestion
In AWS, long-term access keys will be valid until you manually revoke them. This makes them highly sensitive as any exposure can have serious consequences and should be used with care.
```

PR #1086, opened 2022-06-30
Modify rule S6244: Fix an unsupported uppercase tag

alban-auzeill:


chrislain-razafimahefa-sonarsource:
Added the 'Java' label.

PR #1088, opened 2022-07-01
Create rule S6245: Add JS as covered language (APPSEC-41)

pedro-oliveira-sonarsource:
https://github.com/SonarSource/security-expected-issues/pull/722


pedro-oliveira-sonarsource:
We should change the message 'Omitting "{argument_name}" disables server-side encryption. Make sure it is safe here.' as the encryption is already disabled. Something like 'Omitting "{argument_name}" implies no bucket object server-side encryption. Make sure it is safe here.'

PR #1090, opened 2022-07-04
SONARSEC-3109 S5131 XSS rule should contain context-specific patches

christophe-zurn-sonarsource:


roberto-orlandi-sonarsource (approved):
LGTM, I just left a small comment, feel free to ignore it if you disagree!

PR #1093, opened 2022-07-04
Modify rule S6252: Add JS as covered language

pedro-oliveira-sonarsource:
https://github.com/SonarSource/security-expected-issues/pull/724


pedro-oliveira-sonarsource:
We should change the message 'Omitting "{parameter}" disables S3 bucket versioning. Make sure it is safe here.' as the encryption is already disabled. Something like 'Omitting "{parameter}" implies no bucket versioning. Make sure it is safe here.'

pierre-loup-tristant-sonarsource (approved):
LGTM.

PR #1104, opened 2022-07-13
Replace remediation 'mn' with 'min'

pavel-mikula-sonarsource:
Some plugins contains manual workaround to replace "mn" with "min". We can fix this directly in the source now.

Most of the rules already use "min"

### DIFF
--- a/rspec-tools/rspec_tools/validation/rule-metadata-schema.json
+++ b/rspec-tools/rspec_tools/validation/rule-metadata-schema.json
@@ -258,7 +258,7 @@
   "definitions": {
     "time": {
       "type": "string",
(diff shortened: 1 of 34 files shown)


rspec-tools/rspec_tools/validation/rule-metadata-schema.json
```diff
@@ -258,7 +258,7 @@
   "definitions": {
     "time": {
       "type": "string",
-      "pattern": "^[ ]*[0-9]+[ ]*(mn|min|h|d)$"
+      "pattern": "^[ ]*[0-9]+[ ]*(min|h|d)$"
```
pavel-mikula-sonarsource, line 261: Nice

PR #1110, opened 2022-07-20
Map rules to OWASP ASVS 4

agigleux:


pedro-oliveira-sonarsource (commented):
**CWE-20**
Some Rule IDs are falling to more broad ASVS IDs. E.g. S2076 is overwhelmed with ASVS IDs but 5.3.8 is quite on the point.

**CWE-200**
Some unnecessary assignments as the CWE is pretty generic, therefore it matches with a lot of ASVS IDs. E.g S5332 is being assigned with 7.3.3.

**CWE-256**
Maybe we can add 2.10.3? Not sure if 2.7.5 is accurate to our rule implementation, due to it being specific regarding out-of-band verifier (It's still regarding the same issue though).

**CWE-272**
Although CWE-272 is related to 10.2.2 and S6405, I would not say 10.2.2 is related to S6405.

**CWE-284**
Do not think 1.10.1 is adequate to what we cover on this rule. Perhaps the section 4 has more related ASVS IDs.

**CWE-285**
Similar to the previous one. I would say ASVS section 4 is the more related. Perhaps both the 13's can be kept but they are to specific for S2809.

**CWE-287**
Review the section 2 ASVS elements attribution as some are more specific and are falling in an unrelated rule ID e.g. S6293 with 2.3.3. I would probably also remove 9.2.3 as the CWE is to generic.

**CWE-295**
One of the mapped ASVS IDs 9.1.4 seems to not exist.

**CWE-304**
S5832 is being linked to 2.5.5 which I would say it's not correct. Perhaps 2.5.5 is to specific to keep it on the CWE-304 mapping.

**CWE-308**


**CWE-319**
Would remove 5.2.6, 8.3.1

**CWE-326**
It is a generic CWE, therefore, gets assigned a high number of ASVS IDs causing mismatches e.g S3345 with 6.2.5. Would review the rules assigned to CWE-326 individually one by one or shorten the ASVS IDs mapping to generic ones as well.

**CWE-390**
Would remove 11.1.8 over items within the 7.4 section.

**CWE-770**
Would remove  11.1.3., 11.1.4, and 13.4.1. And would unmap S5693 with 13.2.4

**CWE-943**
Would add 5.3.5. The assignment of these ASVS IDs need to be reviewed as the CWE is very generic and the ASVS IDs are more specific e.g S2077 assigned to 5.3.7.

**CWE-98**
Would add 5.3.9.

**Notes:**

- See https://github.com/SonarSource/languages-experimental-tooling/pull/34 for an updated version of the mapping with some of the above suggestions.

- Some rules seem to lack mapping even though they have CWEs with ASVS IDs mapped, e.g. S2819, S6336 or S6358. Did not look into it, but perhaps are the ones with the metadata.json file outside the root folder?

agigleux:
I took into account all your comments.

S2819, S6336 or S6358 have now an associated ASVS id.

For CWE-20, I kept only  5.1.3 and 5.1.4
For CWE-200, I kept nothing to limit the noise.

Here is the list of all the rules impacted: https://github.com/SonarSource/languages-experimental-tooling/blob/master/bubble/appsec/asvs/RuleToASVS-Mapping-Generated.csv

pierre-loup-tristant-sonarsource (changes requested):
I think `Pipfile` and `Pipfile.lock` should not be removed

rules/S3371/java/metadata.json
```diff
@@ -37,6 +37,14 @@
     ],
     "OWASP Top 10 2021": [
       "A5"
+    ],
+    "ASVS 4.0": [
+      "5.1.3",
+      "5.1.4",
+      "5.3.10",
```
pierre-loup-tristant-sonarsource, line 44:
```suggestion
```

pierre-loup-tristant-sonarsource (changes requested):
I fixed an incorrect mapping regarding LDAP injection and XPath injections as this was obvious. I think we should fix that, make sure the checks are passing and then merge as is.

pedro-oliveira-sonarsource (approved):
LGTM for the first round.
A hardening ticket has been created [APPSEC-63](https://sonarsource.atlassian.net/browse/APPSEC-63?atlOrigin=eyJpIjoiZmEyYzU4ODMwYmM3NGJiYWI2ZDAyOTQ1MDZkZDViNzkiLCJwIjoiaiJ9) .

PR #1113, opened 2022-07-21
Modify rule S6423: Swap arguments in LogError call in compliant solution

martin-strecker-sonarsource:
The LogError call in the compliant section is compiling, but it picks the wrong overload:

`LogError(ILogger, Exception, String, Object[])` should be called, but 
`LogError(ILogger, String, Object[])` is called instead.

Fixes https://community.sonarsource.com/t/67647/2


andrei-epure-sonarsource (approved):
LGTM!

PR #1115, opened 2022-07-22
Modify rule S5768: Remove broken link

martin-strecker-sonarsource:
The link to the document is no longer available on SAP. The document linked was probably this one:

https://fdocuments.in/document/best-practice-guidecustom-abap-code-transition-to-sap-hana.html?page=1

The document gives only high-level performance advice. The rule violation is not part of the document. The link can be removed.


andrei-epure-sonarsource (approved):
LGTM

PR #1119, opened 2022-07-25
Prohibit non-ASCII characters in rule metadata.json files

arseniy-sonar:
Triggered by the deployment failure that was caused by an invisible Unicode character in a rule's metadata.json.
This PR implements three conceptual changes:
- make the deployment parse error more informative
- prohibit the use of non-ASCII characters in the metadata.json files
- remove the existing non-ASCII characters from the existing rules


frontend/src/deployment/metadata.ts
```diff
@@ -117,7 +117,12 @@ function getRuleMetadata(srcDir: string, language?: string) {
     }
     const languageFile = path.join(srcDir, language, 'metadata.json');
     if (fs.existsSync(languageFile)) {
-      return JSON.parse(fs.readFileSync(languageFile, 'utf8'));
+      try {
+        return JSON.parse(fs.readFileSync(languageFile, 'utf8'));
+      } catch (error) {
+        console.error("When parsing the JSON file " + languageFile);
```
marco-antognini-sonarsource, line 123:
```suggestion
        console.error('When parsing the JSON file ' + languageFile);
```
That should address the code smell.

But maybe it's also worth printing the exception? I'm not sure how it will be displayed at the moment.

rules/S4225/vbnet/metadata.json
```diff
@@ -1,3 +1,2 @@
-﻿{
-  "title": "Extension methods should not extend \"Object\""
```
marco-antognini-sonarsource, line 2: Is this change intentional? It seems unrelated at first sight.

marco-antognini-sonarsource (changes requested):
It's also nice to make quoting more consistent. 👍

rules/S4225/vbnet/metadata.json
arseniy-sonar: Oups, no. thanks for catching this !

frontend/src/deployment/metadata.ts
arseniy-sonar: The error will be printed together with the stacktrace just below this, because it is rethrown.
This change address the issue that when this exception happens it does not say which file has broken syntax.

marco-antognini-sonarsource (approved):
LGTM

marco-antognini-sonarsource (approved):
Thanks for the test, looks good.

PR #1121, opened 2022-07-26
Modify rule S2083[C#]: Educational content (APPSEC-49)

pierre-loup-tristant-sonarsource:
Specification ticket: [APPSEC-49](https://sonarsource.atlassian.net/browse/APPSEC-49)

### DIFF
--- a/rules/S2083/common/fix/how-does-this-work.adoc
+++ b/rules/S2083/common/fix/how-does-this-work.adoc
@@ -1,0 +1,6 @@
+The universal way to prevent path injection is to validate paths constructed from untrusted data. 
+
+The validation should be done as follow:
+
+1. Resolve the canonical path of the file by using methods like java.io.File/getCanonicalPath. This will resolve relative path or path components like `../` and removes any ambiguity regarding the file's location. 
+2. Check that the canonical path is within the directory where the file should be located.
--- a/rules/S2083/common/fix/how-does-this-work.adoc
+++ b/rules/S2083/common/fix/how-does-this-work.adoc
@@ -2,5 +2,5 @@ The universal way to prevent path injection is to validate paths constructed fro
 
 The validation should be done as follow:
 
-1. Resolve the canonical path of the file by using methods like java.io.File/getCanonicalPath. This will resolve relative path or path components like `../` and removes any ambiguity regarding the file's location. 
+1. Resolve the canonical path of the file by using methods like `{canonicalization_function}`. This will resolve relative path or path components like `../` and removes any ambiguity regarding the file's location. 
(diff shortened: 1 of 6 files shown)


rules/S2083/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,6 @@
+The universal way to prevent path injection is to validate paths constructed from untrusted data. 
+
+The validation should be done as follow:
+
+1. Resolve the canonical path of the file by using methods like java.io.File/getCanonicalPath. This will resolve relative path or path components like `../` and removes any ambiguity regarding the file's location.
```
loris-s-sonarsource, line 5: I have an idea for this one. I think that this generic text is good, but I think that using a java library is a shame, since we are specifically doing framework-specific fixes.

So here is a suggestion:

-> In `how-to-fix-it` files, just before the include of `code-rationale.adoc`, we create an asciidoc variable called `canonicalization_function`
-> The `code-rationale.adoc` file will just call it.

My next suggestions will give you an idea of what it will look like. I tested it and it works

rules/S2083/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,46 @@
+=== How to fix it in .NET
+
+include::../../common/fix/code-rationale.adoc[]
```
loris-s-sonarsource, line 3:
```suggestion

:canonicalization_function: System.IO.Path.GetFullPath
include::../../common/fix/code-rationale.adoc[]
```

rules/S2083/java/how-to-fix-it/java-se.adoc
```diff
@@ -1,6 +1,6 @@
 === How to fix it in Java SE
```
loris-s-sonarsource, line 2:
```suggestion

:canonicalization_function: java.io.File/getCanonicalPath
```

rules/S2083/common/fix/how-does-this-work.adoc
loris-s-sonarsource, line 5:
```suggestion
1. Resolve the canonical path of the file by using methods like {canonicalization_function}. This will resolve relative path or path components like `../` and removes any ambiguity regarding the file's location. 
```

loris-s-sonarsource (approved):
LGTM! I made this suggestion about adding a variable, but this can be merged without it, imho

rules/S2083/common/fix/how-does-this-work.adoc
pierre-loup-tristant-sonarsource: You are right it should be updated. I will take care of it.

PR #1124, opened 2022-07-28
Modify rule S1116: Add exception paragraph

francoismora:
Modify rule S1116: Add exception paragraph

### DIFF
--- a/rules/S1116/javascript/message.adoc
+++ b/rules/S1116/javascript/message.adoc
@@ -2,19 +2,3 @@
 
 Remove this extra semicolon.
 
-== Exceptions
-
-This rule does not apply when the semicolon is after a line break and before ``++(++`` or ``++[++``.
-
-----
-const hello = 'Hello'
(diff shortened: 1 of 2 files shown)


rules/S1116/javascript/message.adoc
```diff
@@ -2,3 +2,19 @@
 
 Remove this extra semicolon.
 
+== Exceptions
+
+This rule does not apply when the semicolon is after a line break and before ``++(++`` or ``++[++``.
```
saberduck, line 7: if possible we should explain why we added the exception

saberduck (changes requested):
can you pls add the reason for exception (why)

rules/S1116/javascript/message.adoc
francoismora: Updated

francoismora:
Fixes SonarSource/SonarJS#3250

saberduck (approved):

PR #1125, opened 2022-07-28
Modify rule S3358: Add an exception for JSX expressions.

francoismora:
Modify rule S3358: Add an exception for JSX expressions.

### DIFF
--- a/rules/S3358/javascript/rule.adoc
+++ b/rules/S3358/javascript/rule.adoc
@@ -40,7 +40,7 @@ endif::env-github,rspecator-view[]
 
 == Exceptions
 
-This rule does not apply in JSX expressions.
+This rule does not apply in JSX expressions to support conditional rendering and conditional attributes.
 
 ----
 return (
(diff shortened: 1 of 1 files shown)


rules/S3358/javascript/rule.adoc
```diff
@@ -37,3 +37,22 @@ include::../highlighting.adoc[]
 
 include::../comments-and-links.adoc[]
 endif::env-github,rspecator-view[]
+
+== Exceptions
+
+This rule does not apply in JSX expressions.
```
saberduck, line 43: we should briefly explain the reason behind exception (eg. this is common pattern to nest ternaries)

saberduck (changes requested):

rules/S3358/javascript/rule.adoc
francoismora: Updated.

francoismora: Fixes SonarSource/SonarJS#3244

saberduck (approved):

PR #1126, opened 2022-07-28
Modify rule S1117: Remove rule for JavaScript/TypeScript from profile Sonar way

francoismora:
Modify rule S1117: Remove rule for JavaScript/TypeScript from profile Sonar way


francoismora:
Fixes SonarSource/SonarJS#3242

saberduck (approved):

PR #1127, opened 2022-07-28
CPP-40 Make it clear that not using #pragma once is almost a stylistic choice.

loic-joly-sonarsource:


rules/S1852/cfamily/rule.adoc
```diff
@@ -1,10 +1,10 @@
 ``++#pragma once++`` is a preprocessor directive meant to ensure that a file is only included once in a compilation unit. However, there are several reasons to avoid it:
 
-* While it is widely implemented, it's not universally supported.
-* It's not part of the standard, so its behavior may vary from compiler to compiler.
-* If the same file exists in two different locations, it will _not_ prevent the second file from being included in the compilation.
+* It is not part of the standard, which prevents its use in some contexts.
+* Even if it is supported by virtually all compilers, since its behavior is not defined, it may differ between compilers, especially for some corner cases when determining if two files are identical (for instance in presence of symbolic links).
+* Its semantic is slightly different from the semantic of include guards. For instance, if a file is duplicated in two different locations, `#pragma once` will not prevent multiple inclusion of this file.
```
abbas-sabra-sonarsource, line 5:
```suggestion
* It is not part of the standard, which prevents its use in some contexts.
* Even if it is supported by virtually all compilers, since its behavior is not defined, it may differ between compilers, especially for some corner cases when determining if two files are identical (for instance, in the presence of symbolic links).
* Its semantic is slightly different from the semantic of an include guard. For instance, if a file is duplicated in two different locations, `#pragma once` will not prevent multiple inclusion of this file.
```

rules/S1852/cfamily/rule.adoc
```diff
@@ -1,10 +1,10 @@
...
 
-For these reasons, include guards, or include guards along with ``++#pragma once++`` should be used instead.
+Note: There used to be a build performance improvement when using `#pragma once` instead of include guards, because include guards required to parse the full file to get the `#endif` matching the `#if`. But most modern compilers specifically detect the include guard pattern and use a dedicated optimization that makes it about as fast as `#pragma once`.
```
abbas-sabra-sonarsource, line 7:
```suggestion
Note: There used to be a build performance improvement when using `#pragma once` instead of an include guard because include guards are required to parse the full file to get the `#endif` matching the `#if`. But most modern compilers specifically detect the include guard pattern and use a dedicated optimization that makes it about as fast as `#pragma once`.
```

abbas-sabra-sonarsource (commented):
minor comments

rules/S1852/cfamily/rule.adoc
loic-joly-sonarsource: I don't think your suggestion worked well, so I tried something else.

abbas-sabra-sonarsource (approved):

PR #1128, opened 2022-07-29
Fix CWE mapping

pierre-loup-tristant-sonarsource:


pedro-oliveira-sonarsource (approved):
LGTM.

rules/S2077/metadata.json
```diff
@@ -27,10 +27,8 @@
   "scope": "Main",
   "securityStandards": {
     "CWE": [
-      564,
-      89,
       20,
-      943
```
martin-strecker-sonarsource, line 33: What is the reasoning behind the removal of https://cwe.mitre.org/data/definitions/943? Is it because https://cwe.mitre.org/data/definitions/89 is basically the same but a better fit because SQL only?

pierre-loup-tristant-sonarsource: Exactly [CWE-943](https://cwe.mitre.org/data/definitions/943) is a child of [CWE-89](https://cwe.mitre.org/data/definitions/89)

pierre-loup-tristant-sonarsource: Sorry, a parent. Therefore or covers other kinds of vulnerabilities like XPath injection that have nothing to do with SQL injection.

martin-strecker-sonarsource: Thank you for the confirmation. This unblocks https://github.com/SonarSource/sonar-dotnet/pull/6015

PR #1130, opened 2022-08-02
Modify rules S6424 and S6420: Change titles

martin-strecker-sonarsource:
Some new rules for Azure Functions don't follow the *X shou [not] Y* naming scheme.

### DIFF
--- a/rules/S6420/csharp/metadata.json
+++ b/rules/S6420/csharp/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "Client instances should not be created repeatedly on each Azure Function invocation",
+  "title": "Client instances should not be recreated on each Azure Function invocation",
   "type": "CODE_SMELL",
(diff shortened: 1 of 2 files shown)


costin-zaharia-sonarsource (approved):

rules/S6420/csharp/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Reuse client instances rather than creating new ones with each Azure Function invocation",
+  "title": "Client instances should not be created repeatedly on each Azure Function invocation",
```
costin-zaharia-sonarsource, line 2: maybe: "recreated" instead of "created repeatedly"

costin-zaharia-sonarsource: or, on a positive note, "Client instances should be reused between Azure Function invocations"

PR #1146, opened 2022-08-04
[APPSEC-59] Modify rule S5146: Educational content (Java)

pedro-oliveira-sonarsource:
Specification ticket: [APPSEC-59](https://sonarsource.atlassian.net/browse/APPSEC-59).

### DIFF
--- a/rules/S5146/common/fix/how-does-this-work.adoc
+++ b/rules/S5146/common/fix/how-does-this-work.adoc
@@ -1,0 +1,6 @@
+=== How does this work?
+
+In case the application strictly requires redirecting based on user-controllable data, this could be done using the following alternatives:
+
+1. Using an allow-list approach, in case the destination URLs are limited;
+2. Adding a customized page to which users are redirected, warning about the imminent action and requiring manual authorization to proceed.
--- a/rules/S5146/common/fix/how-does-this-work.adoc
+++ b/rules/S5146/common/fix/how-does-this-work.adoc
@@ -2,5 +2,5 @@
 
 In case the application strictly requires redirecting based on user-controllable data, this could be done using the following alternatives:
 
-1. Using an allow-list approach, in case the destination URLs are limited;
+1. Using an allow-list approach, in case the destination URLs are limited.
 2. Adding a customized page to which users are redirected, warning about the imminent action and requiring manual authorization to proceed.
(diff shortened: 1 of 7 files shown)


rules/S5146/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,6 @@
+=== How does this work?
+
+In case the application strictly requires redirecting based on user-controllable data, this could be done using the following alternatives:
+
+1. Using an allow-list approach, in case the destination URLs are limited;
```
hendrik-buchwald-sonarsource, line 5: I think a normal dot would fit better here than a semicolon.

```suggestion
1. Using an allow-list approach, in case the destination URLs are limited.
```

rules/S5146/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -12,10 +11,10 @@ h| Non-compliant code example
...
 using System.Web.Mvc;
 
-public class HelloController : Controller
+public class ExampleController : Controller
 {
     [HttpGet]
-    public void Hello(string url)
+    public void Redirect(string url)
     {
         Response.Redirect(url);
```
hendrik-buchwald-sonarsource, line 19: Just noticed that the .NET part was missing the noncompliant comment.

```suggestion
        Response.Redirect(url); // Noncompliant
```

rules/S5146/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,42 @@
...
+----
+protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+  String location = req.getParameter("url");
+
+  List<String> allowedHosts = new ArrayList<String>();  
+  allowedHosts.add("https://example.com/");
+  allowedHosts.add("https://www.example.com/");
+
+  if (allowedHosts.contains(location))
+    resp.sendRedirect(location);
```
hendrik-buchwald-sonarsource, line 36: The compliant and noncompliant code samples use different approaches to redirect (setHeader vs sendRedirect). Since it is not relevant for this rule which approach is used, maybe the code samples would be easier to understand if both used `sendRedirect`.

hendrik-buchwald-sonarsource (approved):
LGTM! Left some minor comments, mostly about previous stuff that is not related to the Java part.

rules/S5146/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,41 @@
...
+}
+----
+[source,java]
+----
+protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+  String location = req.getParameter("url");
+  resp.setStatus(302);
+  resp.setHeader("Location", location); // Noncompliant
+}
+----
```
pedro-oliveira-sonarsource, line 23:
```suggestion
```

PR #1154, opened 2022-08-08
Modify rule S3353: Add sample with "var" and interpolated string replacements

martin-strecker-sonarsource:
Part of SonarSource/sonar-dotnet/issues/5868

### DIFF
--- a/rules/S3353/csharp/rule.adoc
+++ b/rules/S3353/csharp/rule.adoc
@@ -17,6 +17,18 @@ public bool Seek(int[] input)
   return false;
 }
 ----
+or
+
+[source,csharp]
+----
+public class Sample
+{
+  public void Method()
+  {
+    var context = $"{nameof(Sample)}.{nameof(Method)}";  // Noncompliant
+  }
+}
+----
 
 == Compliant Solution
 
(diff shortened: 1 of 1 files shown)


rules/S3353/csharp/rule.adoc
```diff
@@ -35,6 +47,20 @@ public bool Seek(int[] input)
...
+or
+
+[source,csharp]
+----
+public class Sample
+{
+  public void Method()
+  {
+    const string context = $"{nameof(Sample)}.{nameof(Method)}";  // C# 10 and above
+    const string context = nameof(Sample) + "." + nameof(Method); // lower versions of C#
```
andrei-epure-sonarsource, line 59: I'm not sure we want to promote this string concatenation because it makes the code less readable

I would simply not raise for interpolated strings when C# 10 is not used.

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

andrei-epure-sonarsource:
(should be merged after the final PR for the rule gets merged, to be sure they're in sync)

andrei-epure-sonarsource (approved):

PR #1173, opened 2022-08-23
[APPSEC-84] Modify rule S5145: Educational content (Common)

pedro-oliveira-sonarsource:
Specification ticket: [APPSEC-84](https://sonarsource.atlassian.net/browse/APPSEC-84).


rules/S5145/common/resources/standards.adoc
```diff
@@ -0,0 +1,6 @@
+=== Standards
+
+* https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/[OWASP Top 10 2021 Category A9] - Security Logging and Monitoring Failures
+* https://owasp.org/www-project-top-ten/2017/A1_2017-Injection[OWASP Top 10 2017 Category A1] - Injection
```
pierre-loup-tristant-sonarsource, line 4: Maybe it would make sense to update OWASP top 10 2017 and 2021 mapping because I think it fits both Injection and "Security Logging and Monitoring Failures".

rules/S5145/impact.adoc
```diff
@@ -0,0 +1,14 @@
+=== What is the potential impact?
+
+If an attacker can insert arbitrary data into a log file, the integrity of the chain of events being recorded is compromised. +
```
pierre-loup-tristant-sonarsource, line 3: Maybe you can explain that log entries are usually separated by new lines. It's not uncommon to be able to add arbitrary data to a log entry. What is the problem here is the ability to add new lines to forge artificial entries.
User should be able to understand this.

rules/S5145/impact.adoc
```diff
@@ -0,0 +1,14 @@
...
+Other attacks, such as cross-site scripting (XSS) or code injection, could also occur if the logged data is fed to other vulnerable application components.
+
+==== Log Forge
+
+An attacker inserts bogus data into a log file to conceal his malicious activities. This obscures the content for an incident response team to trace the origin of the breach as the indicators of compromise (IoCs) lead to fake application events.
```
pierre-loup-tristant-sonarsource, line 8: Here again, explain how new line characters make it able to forge log entries.

rules/S5145/impact.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+==== Code Execution
+
+Sometimes the logged data is consumed by a different downstream component, which improperly assumes its safety due to a trusted origin. +
+This can lead to various nefarious attacks, such as code injection, as data is left unsanitized and consumed by critical application functionality. +
+A notorious vulnerability, also known as Log4Shell, found in the popular Java logging framework Log4j, is an example of such a vector.
```
pierre-loup-tristant-sonarsource, line 14: Again, I am pretty sure the rule will not protect you against such attacks.

rules/S5145/impact.adoc
```diff
@@ -0,0 +1,14 @@
+=== What is the potential impact?
+
+If an attacker can insert arbitrary data into a log file, the integrity of the chain of events being recorded is compromised. +
+Other attacks, such as cross-site scripting (XSS) or code injection, could also occur if the logged data is fed to other vulnerable application components.
```
pierre-loup-tristant-sonarsource, line 4: True but I am pretty sure the rule will not protect you against such attacks.

pierre-loup-tristant-sonarsource (changes requested):
I think the content should focus on the log forging attack vector. Therefore it should explain the role of newline characters.

rules/S5145/impact.adoc
pedro-oliveira-sonarsource, line 3:
```suggestion
If an attacker can insert arbitrary data into a log file, the integrity of the chain of events being recorded is compromised. This occurs because the attacker is able to inject the log entry separator of the target system, commonly newlines and thus insert artificial log entries.  +
```

rules/S5145/impact.adoc
pedro-oliveira-sonarsource, line 8:
```suggestion
An attacker, able to create independent log entries by injecting the log separator, inserts bogus data into a log file to conceal his malicious activities. This obscures the content for an incident response team to trace the origin of the breach as the indicators of compromise (IoCs) lead to fake application events.
```

rules/S5145/common/resources/standards.adoc
pedro-oliveira-sonarsource: I think we can drop the 2021 A9, nowhere in the OWASP page mentions injection, rather it scopes lack of scope, event reaction or improper storage.

rules/S5145/impact.adoc
pedro-oliveira-sonarsource: Kept the mention of the other attacks, but added some clarification over the coverage

pierre-loup-tristant-sonarsource (approved):
LGTM

PR #1175, opened 2022-08-23
[APPSEC-85] Modify rule S5145: Educational content (Java)

pedro-oliveira-sonarsource:
Specification ticket: [APPSEC-85](https://sonarsource.atlassian.net/browse/APPSEC-85).

### DIFF
--- a/rules/S5145/common/fix/how-does-this-work.adoc
+++ b/rules/S5145/common/fix/how-does-this-work.adoc
@@ -1,0 +1,4 @@
+=== How does this work?
+
+Data being used for logging should be content restricted, typed, and sanitized. +
+Sanitization mainly revolves around stripping carriage return (CR) and line feed (LF) characters, but further actions could be required based on the application context and the logged data usage.
--- a/rules/S5145/java/how-to-fix-it/java-se.adoc
+++ b/rules/S5145/java/how-to-fix-it/java-se.adoc
@@ -28,7 +28,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
   
   String data = request.getParameter("data");
   if(data != null){
-    data = data.replaceAll("[\n\r\t]", "_");
+    data = data.replaceAll("[\n\r]", "_");
     logger.log(Level.INFO, "Data: {0} ", data);
   }
 }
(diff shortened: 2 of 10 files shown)


rules/S5145/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,4 @@
+=== How does this work?
+
+Data being used for logging should be content restricted, typed, and sanitized. +
+Sanitization mainly revolves around stripping carriage return (CR) and line feed (LF) characters, but further actions could be required based on the application context and the logged data usage.
```
loris-s-sonarsource, line 4: I am wondering if we shouldnt also talk a little about Log4Shell, to emphasize sanitization in general. WDYT?
Something along the lines of
- For example, Log4Shell
- Is Log injection vuln
- Could have been avoided if we sanitized by allowing only certain ranges of characters

I am aware that it's really a "vendor point of view", and that reality isn't as simple as that haha, but WDYT ?

loris-s-sonarsource (approved):
LGTM! One minor comment

rules/S5145/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+[cols="a"]
+|===
+h| Non-compliant code example
+|
+[source,java]
+----
+private static final Logger logger = Logger.getLogger("Logger");
+
+protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
loris-s-sonarsource, line 13:
```suggestion
protected void doGet(HttpServletRequest request) throws IOException {
```
Response isn't used here

rules/S5145/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+----
+h| Compliant solution
+|
+[source,java]
+----
+private static final Logger logger = Logger.getLogger("Logger");
+
+protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
loris-s-sonarsource, line 27:
```suggestion
protected void doGet(HttpServletRequest request) throws IOException {
```
Same

PR #1176, opened 2022-08-23
[APPSEC-87] Modify rule S5334: Educational content (Common)

hendrik-buchwald-sonarsource:


rules/S5334/common/resources/articles.adoc
```diff
@@ -0,0 +1,4 @@
+=== Articles & Blog Posts
+
+* https://blog.sonarsource.com/moodle-remote-code-execution/[SonarSource, Evil Teacher: Code Injection in Moodle]
```
hendrik-buchwald-sonarsource, line 3: Moodle is PHP but I still think it is relevant to understand the problem.

rules/S5334/impact.adoc
```diff
@@ -0,0 +1,43 @@
...
+==== Denial of service and data leaks
+
+In this scenario, the attack aims to disrupt the organization's activities and
+profit from data leaks.
+
+The attackers:
+
+* download the internal server's data, most likely to sell it
+* modify data, send malware
+* stop services or exhaust resources (with fork bombs for example)
```
pedro-oliveira-sonarsource, line 23:
```suggestion
* stop services or exhaust resources (e.g., fork bombs)
```

rules/S5334/impact.adoc
```diff
@@ -0,0 +1,43 @@
+=== What is the potential impact?
+
+In the context of a web application vulnerable to code injection: +
+After detecting the injection, attackers inject data into the vulnerable text
+field to execute malicious instructions.
```
pedro-oliveira-sonarsource, line 5:
```suggestion
After detecting the injection point, attackers inject data into the vulnerable parameter to execute malicious instructions.
```

rules/S5334/impact.adoc
```diff
@@ -0,0 +1,43 @@
...
+The attackers:
+
+* download the internal server's data, most likely to sell it
+* modify data, send malware
+* stop services or exhaust resources (with fork bombs for example)
+
+This threat is particularly insidious if the attacked organization does not
+maintain a disaster recovery plan (DRP).
+
+==== Root privilege escalation and pivot
```
pedro-oliveira-sonarsource, line 28:
```suggestion
==== Privilege escalation and pivoting
```

rules/S5334/common/fix/pre-approved-list.adoc
```diff
@@ -0,0 +1,16 @@
...
+allowed values before executing it. If a user input does not fully match an
+entry in this list, it should be rejected because it is considered unsafe.
+
+Depending on the number of instructions you want the application to support and
+their complexity, the list can be either a regular expression or an array. If
+you use regular expressions, choose simple ones to avoid ReDOS attacks. For
+example, you can accept only a specific set of instructions by using
+`^(send|receive)\(\);$`.
+
+*Important note*: The application must do validation on the server side. Not on
```
pedro-oliveira-sonarsource, line 14:
```suggestion
*Important note*: The application must do server-side validation. Not on
```

pedro-oliveira-sonarsource (approved):

rules/S5334/impact.adoc
hendrik-buchwald-sonarsource: I am not in favor of this change. When we defined the [styling guide](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc) we decided against using abbreviations like `e.g.`. Reading the styling guide again, this is not too clear though. I think I might have to update the guide in that regard.

Optimally, the text should also not use the brackets according to the styling guide. I left them in because the text is copied from the OS command injection rule and I did not want to deviate too much.

rules/S5334/impact.adoc
hendrik-buchwald-sonarsource: I am in favor of leaving this as it is. Both versions are fine to me by themselves but we also use this title in the OS command injection rule and I think it would be better not to have 2 slightly different titles for the same thing.

PR #1187, opened 2022-08-24
[APPSEC-76] Modify rule S5135[C#]: Update to Education Framework

loris-s-sonarsource:


rules/S5135/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,52 @@
+=== How to fix it in .Net
```
hendrik-buchwald-sonarsource, line 1:
```suggestion
=== How to fix it in .NET
```

rules/S5135/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,52 @@
...
+|
+[source,csharp]
+----
+public class Example : Controller
+{
+    [HttpPost]
+    public ActionResult Deserialize(HttpPostedFileBase inputFile) 
+    {
+        ExpectedType expectedObject = null;
+        XmlSerializer serializer    = new XmlSerializer(typeof(ExpectedObject));
```
hendrik-buchwald-sonarsource, line 34: Should this be `ExpectedType` instead of `ExpectedObject`?

I am really unsure if this can be considered more secure than the non-compliant solution. Aren't both examples using the same `Type` object?

hendrik-buchwald-sonarsource (commented):
I am not sure about the code samples yet, see comments.

Since "common" is not merged yet, the tests are still failing. So I would recommend to rebase it once "common" is merged, so that we can see if there is a problem with some of the includes or not.

rules/S5135/csharp/how-to-fix-it/dotnet.adoc
loris-s-sonarsource: Yes, to be honest I took the existing code samples and I assumed that the hidden meaning in this compliant solution was that GetType was not to be trusted. I'm going to quickly investigate this

loris-s-sonarsource: Ok, so actually I misunderstood the previous code sample. The unsafe behavior is to actually control the XmlSerializer type.

I don't know about you, but I'm not sure about the real-world application of this... It looks slightly boring

loris-s-sonarsource: The commit adds text into "`how does this work?`"

hendrik-buchwald-sonarsource:
The compliant code sample still looks vulnerable to me. The type is fixed in the `XmlSerializer` constructor but is this really enough to prevent deserialization attacks here? As the common text states, it can be enough to control the properties of one specific class.

loris-s-sonarsource:
> The compliant code sample still looks vulnerable to me. The type is fixed in the XmlSerializer constructor but is this really enough to prevent deserialization attacks here? As the common text states, it can be enough to control the properties of one specific class.

@hendrik-buchwald-sonarsource Yeah, agreed. At the same time we emphasized this in the common text:
> As a last resort, create a list of approved and safe classes that the
application should be able to deserialize. +
If the untrusted class does not match an entry in this list, it should be
rejected because it is considered unsafe.

My last commit also emphasizes it. Do you still think something is off?

rules/S5135/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,65 @@
...
+|
+[source,csharp]
+----
+public class Example : Controller
+{
+    [HttpPost]
+    public ActionResult Deserialize(HttpPostedFileBase inputFile) 
+    {
+        ExpectedType expectedObject = null;
+        XmlSerializer serializer    = new JsonSerializer(typeof(expectedObject));
```
hendrik-buchwald-sonarsource, line 33: Can a `JsonSerializer` object be stored in a `XmlSerializer` variable?

hendrik-buchwald-sonarsource (approved):
I can't say for certain if the compliant example can be considered secure but overall it looks good to me.

rules/S5135/csharp/rule.adoc
```diff
@@ -1,51 +1,17 @@
...
-* https://www.sans.org/top25-software-errors/#cat2[SANS Top 25] - Risky Resource Management
+== Why is this an issue?
+
+include::../rationale.adoc[]
+
+include::../impact.adoc[]
+
+== How to fix it?
+
+include::how-to-fix-it/java-se.adoc[]
```
loris-s-sonarsource, line 9:
```suggestion
include::how-to-fix-it/dotnet.adoc[]
```

PR #1189, opened 2022-08-25
[APPSEC-80] Modify rule S5334: Educational content (C#)

hendrik-buchwald-sonarsource:
I decided against using the "pre-approved" fix because it is not realistic. Instead I tried to come up with a way to get user input into dynamic code without making it part of the code. I will rework the Java code sample and do the same there.


rules/S5334/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,70 @@
+=== How to fix it in .NET
+
+include::../../common/fix/code-rationale.adoc[]
```
pedro-oliveira-sonarsource, line 3: Should this be here? Not resolving.
```suggestion
```

pedro-oliveira-sonarsource (approved):
Just an include issue on dotnet.adoc. LGTM.

PR #1190, opened 2022-08-26
[APPSEC-73] Modify rule S6096: Educational content (C#)

pedro-oliveira-sonarsource:
Specification ticket: [APPSEC-73](https://sonarsource.atlassian.net/browse/APPSEC-73).

### DIFF
--- a/rules/S6096/common/fix/code-rationale.adoc
+++ b/rules/S6096/common/fix/code-rationale.adoc
@@ -1,0 +1,1 @@
+The following code is vulnerable to Zip Slip as it is constructing a path using untrusted data. This path is then used to copy a file without being validated first. Therefore, it can be leveraged by an attacker to overwrite arbitrary files.
--- a/rules/S6096/common/fix/code-rationale.adoc
+++ b/rules/S6096/common/fix/code-rationale.adoc
@@ -1,1 +1,1 @@
-The following code is vulnerable to Zip Slip as it is constructing a path using untrusted data. This path is then used to copy a file without being validated first. Therefore, it can be leveraged by an attacker to overwrite arbitrary files.
+The following code is vulnerable to Zip Slip as it is constructing a path using an archive entry name. This path is then used to copy a file without being validated first. Therefore, it can be leveraged by an attacker to overwrite arbitrary files.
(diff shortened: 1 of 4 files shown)


rules/S6096/common/fix/code-rationale.adoc
```diff
@@ -0,0 +1 @@
+The following code is vulnerable to Zip Slip as it is constructing a path using untrusted data. This path is then used to copy a file without being validated first. Therefore, it can be leveraged by an attacker to overwrite arbitrary files.
```
pierre-loup-tristant-sonarsource, line 1:
```suggestion
The following code is vulnerable to Zip Slip as it is constructing a path using an archive entry name. This path is then used to copy a file without being validated first. Therefore, it can be leveraged by an attacker to overwrite arbitrary files.
```

rules/S6096/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,7 @@
+The universal way to prevent Zip Slip is to validate the paths constructed from untrusted data.
```
pierre-loup-tristant-sonarsource, line 1:
```suggestion
The universal way to prevent Zip Slip is to validate the paths constructed from untrusted archive entry names. 
```

rules/S6096/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,7 @@
+The universal way to prevent Zip Slip is to validate the paths constructed from untrusted data. 
+
+The validation should be done as follow:
+
+1. Resolve the canonical path of the file by using methods like `{canonicalization_function1}` or `{canonicalization_function2}`. This will resolve relative path or path components like `../` and removes any ambiguity regarding the file's location. 
+2. Check that the canonical path is within the directory where the file should be located.
+3. Ensure the target directory path ends with a forward slash to prevent partial path traversal, for example, */base/dirmalicious* starts with */base/dir* but not */base/dir/*.
```
pierre-loup-tristant-sonarsource, line 7: "Not with" is ambiguous here
```suggestion
3. Ensure the target directory path ends with a forward slash to prevent partial path traversal, for example, */base/dirmalicious* starts with */base/dir* but not start with */base/dir/*.
```

pierre-loup-tristant-sonarsource (approved):
LGTM. 
Suggestion: instead of untrusted data I would use "achieve file entry".

PR #1194, opened 2022-08-29
checklinks.py Fix broken Accept header

martin-strecker-sonarsource:
There is a typo in the accept header of the script. Most servers seem to be fine with it, but https://www.alibabacloud.com/help/en/key-management-service/latest/secrets-manager-overview returned

`HTTP Error 406: Not Acceptable`

with the broken accept-header. I updated the header using the F12 tools from google chrome.

The status code and Accept header can be checked via curl:
```sh
curl --insecure -H "Accept:text/html" -sSL -D - https://www.alibabacloud.com/help/en/key-management-service/latest/secrets-manager-overview -o /dev/null

curl --insecure -H "Accept:broken" -sSL -D - https://www.alibabacloud.com/help/en/key-management-service/latest/secrets-manager-overview -o /dev/null
```


frederic-tingaud-sonarsource (approved):
Thanks for tracking this!

PR #1195, opened 2022-08-29
Modify S5147(Common text): Update to the education framework (APPSEC-94)

loris-s-sonarsource:
Related to [**APPSEC-94**](https://sonarsource.atlassian.net/browse/APPSEC-94).

### DIFF
--- a/rules/S5147/common/fix/dangerous-operators.adoc
+++ b/rules/S5147/common/fix/dangerous-operators.adoc
@@ -10,8 +10,8 @@ and functions in both untrusted input and internal code, such as:
 
 Depending on your use case, you should first try all the ways to use regular
 API calls before using any of these operators. +
-For example, using a `$where` operator is unnecessarily complex when only a simple search is
-required. It also leads to performance problems.
+For example, using a `$where` operator is unnecessarily complex when only a
+simple search is required. It also leads to performance problems.
 
 **Note**: https://www.mongodb.com/docs/manual/reference/operator/query/where/#javascript-enablement[Server-side scripting can be disabled].
 
--- a/rules/S5147/common/fix/dangerous-operators.adoc
+++ b/rules/S5147/common/fix/dangerous-operators.adoc
@@ -1,0 +1,23 @@
+==== Operators are to be classified as dangerous
+
+Because these operators run JavaScript, restrict the use of dangerous operators
+and functions in both untrusted input and internal code, such as:
+
+* `$where`
+* `$function`
+* `$accumulator`
+* `mapReduce`
+
+Depending on your use case, you should first try all the ways to use regular
+API calls before using any of these operators. +
+For example, a `$where` operator is excessive when only a simple search is
+required. It also leads to performance problems.
+
+**Note**: https://www.mongodb.com/docs/manual/reference/operator/query/where/#javascript-enablement[Server-side scripting can be disabled].
+
+Regular operators can also lead to data leaks. +
+If no operators are needed, you should generally reject user input containing
+them. If some operators are necessary, you should restrict their use.
+
+
+
--- a/rules/S5147/common/fix/dangerous-operators.adoc
+++ b/rules/S5147/common/fix/dangerous-operators.adoc
@@ -1,7 +1,12 @@
 ==== Operators are to be classified as dangerous
 
-Because these operators run JavaScript, restrict the use of dangerous operators
-and functions in both untrusted input and internal code, such as:
+As a rule of thumb if no operators are needed, you should generally reject user
+input containing them. If some operators are necessary, you should restrict
+their use.
+
+Some operators execute JavaScript, and their use should be restricted for both
+untrusted input and internal code. +
+These operators include:
 
 * `$where`
 * `$function`
@@ -16,8 +21,7 @@ simple search is required. It also leads to performance problems.
 **Note**: https://www.mongodb.com/docs/manual/reference/operator/query/where/#javascript-enablement[Server-side scripting can be disabled].
 
 Regular operators can also lead to data leaks. +
-If no operators are needed, you should generally reject user input containing
-them. If some operators are necessary, you should restrict their use.
-
-
+For example, attackers can use "comparison query operators" in their attack
+data to trick the backend database into giving hints about sensitive
+information or entirely giving it out.
 
--- a/rules/S5147/common/fix/dangerous-operators.adoc
+++ b/rules/S5147/common/fix/dangerous-operators.adoc
@@ -10,7 +10,7 @@ and functions in both untrusted input and internal code, such as:
 
 Depending on your use case, you should first try all the ways to use regular
 API calls before using any of these operators. +
-For example, a `$where` operator is excessive when only a simple search is
+For example, using a `$where` operator is unnecessarily complex when only a simple search is
 required. It also leads to performance problems.
 
 **Note**: https://www.mongodb.com/docs/manual/reference/operator/query/where/#javascript-enablement[Server-side scripting can be disabled].
--- a/rules/S5147/common/fix/dangerous-operators.adoc
+++ b/rules/S5147/common/fix/dangerous-operators.adoc
@@ -8,7 +8,7 @@ and functions in both untrusted input and internal code, such as:
 * `$accumulator`
 * `mapReduce`
 
-Depending on your use case, you should first try all the ways to use regular
+Depending on your use case, you should first try using regular
 API calls before using any of these operators. +
 For example, using a `$where` operator is unnecessarily complex when only a
 simple search is required. It also leads to performance problems.
--- a/rules/S5147/common/fix/validation.adoc
+++ b/rules/S5147/common/fix/validation.adoc
@@ -1,4 +1,4 @@
-==== Pre-Approved list
+==== Validation
 
 As a rule of thumb, the best approach to protect against injections is to
 systematically ensure that untrusted data cannot break out of an interpreted
@@ -19,11 +19,10 @@ application to be able to use in a query. +
 If a user input does not match an entry in this list, it should be rejected
 because it is considered unsafe.
 
-The list can be either a regex string, arrays, or validators on specific ranges
-of characters. If you use regexes, choose simple regexes to avoid ReDOS
+The list can be either a regex string, an array, or validators on specific
+ranges of characters. If you use regexes, choose simple regexes to avoid ReDOS
 attacks.
 
 *Important note*: The application must do validation on the server side. Not on
 client-side front-ends.
-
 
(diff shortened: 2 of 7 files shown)


rules/S5147/common/fix/validation.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+For NoSQL injections, the cleanest way to do so is to validate the input before
+using it in a NoSQL query.
+
+First, restrict the use of operators. +
+If no operators are required, then reject user input that contains any. If some
+operators are needed, restrict their use to avoid data leaks.
+
+**Note**: Even if the operator is strictly trusted, concatenating untrusted
+data to the expression associated with it makes the query vulnerable.
```
pedro-oliveira-sonarsource, line 15: Maybe we could also refer to avoid specific sensitive operators.

Example 'where' in MongoDB:
query = { $where: this.username == '${username}' }
you would be able to inject plain JavaScript syntax, e.g || 'a'=='a

rules/S5147/common/fix/validation.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+Else, create a list of authorized and secure strings that you want the
+application to be able to use in a query. +
+If a user input does not match an entry in this list, it should be rejected
+because it is considered unsafe.
+
+The list can be either a regex string, an array, or validators on specific
```
pedro-oliveira-sonarsource, line 22: Due to the wording, it comes across to me as if you're doing a string comparison to all these elements instead of a pattern match for regexs for example.

rules/S5147/impact.adoc
```diff
@@ -0,0 +1,31 @@
+=== What is the potential impact?
+
+In the context of a web application that is vulnerable to NoSQL injection: +
+After discovering the injection, attackers inject data into the vulnerable
+field to execute malicious commands in the affected databases.
+
+Below are some real-world scenarios that illustrate the various impacts of an
+attacker exploiting the vulnerability.
+
+==== Identity spoofing and data manipulation
```
pedro-oliveira-sonarsource, line 10: Because you mention data deletion on the next one, might be best to be specific on this one as well.
```suggestion
==== Identity spoofing and data leakage
```

rules/S5147/impact.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+A malicious database query enables privilege escalation or direct data leakage
+from one or more databases. This threat is the most widespread impact.
+
+==== Data deletion and denial of service
+
+The malicious query makes it possible for the attacker to delete data in the
+affected databases. +
+This threat is particularly insidious if the attacked organization does not
+maintain a disaster recovery plan (DRP).
```
pedro-oliveira-sonarsource, line 20: Just to be clear regarding the DoS.
```suggestion
maintain a disaster recovery plan (DRP) as missing data can disrupt the regular
operations of an organization.
```

rules/S5147/rationale.adoc
```diff
@@ -0,0 +1,3 @@
+NoSQL injections occur in an application when the application retrieves data
+from a user or a third-party service and inserts it into a database query
+without sanitizing it first.
```
pedro-oliveira-sonarsource, line 3:
```suggestion
NoSQL injections occur when an application retrieves untrusted data and inserts
it into a database query without sanitizing it first.
```

pedro-oliveira-sonarsource (changes requested):
Just some suggestions.

rules/S5147/common/fix/validation.adoc
```diff
@@ -0,0 +1,28 @@
+==== Validation
+
+As a rule of thumb, the best approach to protect against injections is to
+systematically ensure that untrusted data cannot break out of an interpreted
+context.
+
+For NoSQL injections, the cleanest way to do so is to validate the input before
+using it in a NoSQL query.
```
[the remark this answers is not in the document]
loris-s-sonarsource: Ah yeah, maybe we should write about Neo4J somewhere. For now we don't support it so I don't see a point in adding it now

rules/S5147/common/fix/validation.adoc
```diff
@@ -0,0 +1,28 @@
...
+Else, create a list of authorized and secure strings that you want the
+application to be able to use in a query. +
+If a user input does not match an entry in this list, it should be rejected
+because it is considered unsafe.
+
+The list can be either a regex string, an array, or validators on specific
+ranges of characters. If you use regexes, choose simple regexes to avoid ReDOS
+attacks.
+
+*Important note*: The application must do validation on the server side. Not on
```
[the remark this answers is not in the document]
loris-s-sonarsource: I don't see the difference between both versions, except wording

rules/S5147/common/fix/validation.adoc
```diff
@@ -0,0 +1,28 @@
+==== Validation
+
+As a rule of thumb, the best approach to protect against injections is to
+systematically ensure that untrusted data cannot break out of an interpreted
+context.
+
+For NoSQL injections, the cleanest way to do so is to validate the input before
+using it in a NoSQL query.
+
+First, restrict the use of operators. +
```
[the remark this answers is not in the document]
loris-s-sonarsource: Unfortunately, we don't support them yet, so it's not really in scope for now

rules/S5147/common/fix/validation.adoc
loris-s-sonarsource: Yeah, at first I thought about listing especially bad operators, and even creating a subsection about it (https://www.mongodb.com/docs/manual/core/server-side-javascript/)

But it seemed overkill at the time, and also a kind of validation. Do you think it makes sense to still create this subsection? It would list bad operators, mention the server-side scripting and the fact that we can disable it

rules/S5147/common/fix/validation.adoc
loris-s-sonarsource: which elements are you talking about ? Because yeah, the goal would be to provide user input validation

rules/S5147/common/fix/validation.adoc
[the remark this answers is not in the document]
loris-s-sonarsource: Maybe we should create a ticket about it, but I think we actually need to see which NoSQL libraries would actually be useful in the future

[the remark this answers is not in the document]
loris-s-sonarsource: I created a ticket for this: [**APPSEC-113**](https://sonarsource.atlassian.net/browse/APPSEC-113)

rules/S5147/common/fix/dangerous-operators.adoc
```diff
@@ -0,0 +1,23 @@
...
+and functions in both untrusted input and internal code, such as:
+
+* `$where`
+* `$function`
+* `$accumulator`
+* `mapReduce`
+
+Depending on your use case, you should first try all the ways to use regular
+API calls before using any of these operators. +
+For example, a `$where` operator is excessive when only a simple search is
```
loris-s-sonarsource, line 13:
```suggestion
For example, using a `$where` operator is unnecessarily complex when only a simple search is
```

rules/S5147/common/fix/dangerous-operators.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+* `$where`
+* `$function`
+* `$accumulator`
+* `mapReduce`
+
+Depending on your use case, you should first try all the ways to use regular
+API calls before using any of these operators. +
+For example, using a `$where` operator is unnecessarily complex when only a simple search is
+required. It also leads to performance problems.
```
loris-s-sonarsource, line 14:
```suggestion
For example, using a `$where` operator is unnecessarily complex when only a
simple search is required. It also leads to performance problems.
```

rules/S5147/common/fix/dangerous-operators.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+Because these operators run JavaScript, restrict the use of dangerous operators
+and functions in both untrusted input and internal code, such as:
+
+* `$where`
+* `$function`
+* `$accumulator`
+* `mapReduce`
+
+Depending on your use case, you should first try all the ways to use regular
```
pedro-oliveira-sonarsource, line 11:
```suggestion
Depending on your use case, you should first try using regular
```

rules/S5147/common/fix/dangerous-operators.adoc
```diff
@@ -0,0 +1,23 @@
...
+API calls before using any of these operators. +
+For example, using a `$where` operator is unnecessarily complex when only a
+simple search is required. It also leads to performance problems.
+
+**Note**: https://www.mongodb.com/docs/manual/reference/operator/query/where/#javascript-enablement[Server-side scripting can be disabled].
+
+Regular operators can also lead to data leaks. +
```
pedro-oliveira-sonarsource, line 18: Seems it requires further description.

rules/S5147/impact.adoc
```diff
@@ -0,0 +1,32 @@
...
+After discovering the injection point, attackers insert data into the vulnerable
+field to execute malicious commands in the affected databases.
+
+Below are some real-world scenarios that illustrate the various impacts of an
+attacker exploiting the vulnerability.
+
+==== Identity spoofing and data leakage
+
+A malicious database query enables privilege escalation or direct data leakage
+from one or more databases. This threat is the most widespread impact.
```
pedro-oliveira-sonarsource, line 13: I think this one could get a bit more improved.
Perhaps instead of 'Identity Spoofing' say 'Privilege Escalation' and on the description of it, elaborate on how the 'malicious database query enables privilege escalation', perhaps talking about database misconfigurations.

rules/S5147/impact.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+==== Data deletion and denial of service
+
+The malicious query makes it possible for the attacker to delete data in the
+affected databases. +
+This threat is particularly insidious if the attacked organization does not
+maintain a disaster recovery plan (DRP) as missing data can disrupt the regular
+operations of an organization.
+
+==== Chaining DB injections with other vulnerabilities
```
pedro-oliveira-sonarsource, line 23:
```suggestion
==== Chaining NoSQL injections with other vulnerabilities
```

pedro-oliveira-sonarsource (approved):
Just some minor suggestions, LGTM.

rules/S5147/impact.adoc
loris-s-sonarsource: This one is actually really about simple queries, so I think it could be unnecessarily complex to talk about DB misconfigs here tbh

loris-s-sonarsource: I will mention the simple queries for clarity's sake

PR #1196, opened 2022-08-29
Modify rule S5147(Java): Add Educational content (APPSEC-95)

loris-s-sonarsource:
Related to [**APPSEC-95**](https://sonarsource.atlassian.net/browse/APPSEC-95).


rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
```diff
@@ -0,0 +1,49 @@
...
+protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws UnknownHostException
+{
+    String input = req.getParameter("input");
+    
+    MongoClient mongoClient = new MongoClient();
+    DB database             = mongoClient.getDB("exampleDatabase");
+    DBCollection collection = database.getCollection("exampleCollection");
+    BasicDBObject query     = new BasicDBObject();
+
+    query.put("$where", "this.field == \"" + input + "\""); // Noncompliant
```
pedro-oliveira-sonarsource, line 21: Non-compliant code is not flagging.

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
```diff
@@ -0,0 +1,49 @@
...
+----
+|===
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+In the previous example, the untrusted data doesn't need validation for its use
+case. Moving it out of a `$where` expression into a proper field query is
+enough.
```
pedro-oliveira-sonarsource, line 48: We could give a bit more emphasis to these types of solutions in the how-to-fix-it section referring to them as safe built-in methods or something.

pedro-oliveira-sonarsource (changes requested):
Have a look at the non-compliant code example.

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
```diff
@@ -0,0 +1,49 @@
+=== How to fix it in Legacy Mongo Java Driver
+
+The following code is vulnerable to NoSQL injections because untrusted data is
+interpreted by the `$where` expression.
```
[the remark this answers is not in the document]
loris-s-sonarsource: I'm going to slightly adjust your suggestion, to add the technical impact:

```suggestion
The following code is vulnerable to NoSQL injections because untrusted data is
concatenated to the `$where` operator. In this case, the untrusted data will be
interpreted as code, instead of a regular string.
```

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+[cols="a"]
+|===
+h| Non-compliant code example
+|
+[source,java]
+----
```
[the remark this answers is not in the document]
loris-s-sonarsource: I'm not sure about adding imports... It's likely to end up as a big chuck of lines in the end, and IMHO this could be more of a clutter.

Since this code will be shown to people that have the exact same library, I'm not sure I would do this. But this can be debatable. WDYT?

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
loris-s-sonarsource: Yes, I think we could definitely use a part about "why you shouldn't use operators". I keep this in mind.

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
```diff
@@ -0,0 +1,56 @@
...
+protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws UnknownHostException
+{
+    String input = req.getParameter("input");
+
+    MongoClient mongoClient = new MongoClient();
+    DB database             = mongoClient.getDB("ExampleDatabase");
+    DBCollection collection = database.getCollection("exampleCollection");
+    BasicDBObject query     = new BasicDBObject();
+
+    query.put("field", input);
```
pedro-oliveira-sonarsource, line 40: Since the put method is not supported, we should provide a different compliant solution (if there is any).

pedro-oliveira-sonarsource (changes requested):
Just the fix on the compliant solution already discussed.

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
loris-s-sonarsource:
```suggestion
    query.append("field", input);
```

rules/S5147/java/how-to-fix-it/mongo-java-driver.adoc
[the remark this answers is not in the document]
loris-s-sonarsource: I added imports in the end, it should be alright 👍

pedro-oliveira-sonarsource (approved):
LGTM.

PR #1198, opened 2022-08-29
Modify rule S2091(Java): Add Educational content (APPSEC-99)

loris-s-sonarsource:
Related to [APPSEC-99](https://sonarsource.atlassian.net/browse/APPSEC-99).

### DIFF
--- a/rules/S2091/java/how-to-fix-it/java-se.adoc
+++ b/rules/S2091/java/how-to-fix-it/java-se.adoc
@@ -1,7 +1,7 @@
 === How to fix it in Java SE
 
-The following code is vulnerable to Xpath injections because untrusted data is
-injected in an Xpath query without prior validation.
+The following non-compliant code is vulnerable to Xpath injections because untrusted data is
+concatenated to an Xpath query without prior validation.
 
 [cols="a"]
 |===
@@ -49,7 +49,7 @@ public boolean authenticate(HttpServletRequest req, XPath xpath, Document doc) t
 
 include::../../common/fix/validation.adoc[]
 
-In the example, a sanitization mechanism is applied to an XPath Factory as a
-resolver to resolve the untrusted data to the request in a safe way, similarly
-to parametrized SQL queries.
+In the example, a parameterized XPath query is created as an XPath Factory
+resolver to insert untrusted data to the request in a safe way, similarly to
+parametrized SQL queries.
 
--- a/rules/S2091/java/how-to-fix-it/java-se.adoc
+++ b/rules/S2091/java/how-to-fix-it/java-se.adoc
@@ -48,3 +48,8 @@ public boolean authenticate(HttpServletRequest req, XPath xpath, Document doc) t
 === How does this work?
 
 include::../../common/fix/validation.adoc[]
+
+In the example, a sanitization mechanism is applied to an XPath Factory as a
+resolver to resolve the untrusted data to the request in a safe way, similarly
+to parametrized SQL queries.
+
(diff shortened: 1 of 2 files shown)


rules/S2091/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,55 @@
+=== How to fix it in Java SE
+
+The following code is vulnerable to Xpath injections because untrusted data is
+injected in an Xpath query without prior validation.
```
pedro-oliveira-sonarsource, line 4:
```suggestion
concatenated to an Xpath query without prior validation.
```

rules/S2091/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,55 @@
...
+|===
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+In the example, a sanitization mechanism is applied to an XPath Factory as a
+resolver to resolve the untrusted data to the request in a safe way, similarly
+to parametrized SQL queries.
+
```
pedro-oliveira-sonarsource, line 55: Can you call it 'parameterized XPath queries instead on 'a sanitization mechanism' like they do in [OWASP XPATH_Injection](https://owasp.org/www-community/attacks/XPATH_Injection) .

rules/S2091/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,55 @@
...
+}
+
+----
+|===
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+In the example, a sanitization mechanism is applied to an XPath Factory as a
```
pedro-oliveira-sonarsource, line 52:
```suggestion
In the compliant solution example, a sanitization mechanism is applied to an XPath Factory as a
```

rules/S2091/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,55 @@
+=== How to fix it in Java SE
+
+The following code is vulnerable to Xpath injections because untrusted data is
```
pedro-oliveira-sonarsource, line 3: I have been doing this as well on my rules but maybe we should start specifying since the 'following code' expression also captures de compliant solution.
```suggestion
The following non-compliant code is vulnerable to Xpath injections because untrusted data is
```

pedro-oliveira-sonarsource (approved):
LGTM. Some minor suggestions.

rules/S2091/java/how-to-fix-it/java-se.adoc
loris-s-sonarsource:
```suggestion
In the example, a parameterized XPath query is created as an XPath Factory
resolver to insert untrusted data to the request in a safe way, similarly to
parametrized SQL queries.

```

rules/S2091/java/how-to-fix-it/java-se.adoc
loris-s-sonarsource: Hmm, don't worry about this IMHO. This is going to be displayed as a diff so I think we can keep this pattern

rules/S2091/java/how-to-fix-it/java-se.adoc
loris-s-sonarsource: same as the "non-compliant" solution. IMHO we can trust the diff view to make it clear.

PR #1201, opened 2022-08-30
[APPSEC-91] Modify rule S5144: Educational content (Common)

pedro-oliveira-sonarsource:
Specification ticket: [APPSEC-91](https://sonarsource.atlassian.net/browse/APPSEC-91).


rules/S5144/common/fix/code-rationale.adoc
```diff
@@ -0,0 +1,2 @@
+The following code is vulnerable to SSRF as it performs an HTTP request to a
+URL defined by user-controllable data.
```
loris-s-sonarsource, line 2:
```suggestion
The following code is vulnerable to SSRF as it performs an HTTP request to a
URL defined by untrusted data.
```

rules/S5144/impact.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+SSRF usually results in unauthorized actions or data disclosure in the
+vulnerable application or on a different system it can reach. Conditional to
+what is reachable, remote command execution can be achieved, although it often
+requires chaining with further exploitations.
+
+Information disclosure is SSRF's core outcome. Depending on the extracted data,
+an attacker can perform a variety of different actions that can range from low
+to critical severity.
+
```
loris-s-sonarsource, line 11:
```suggestion
Information disclosure is SSRF's core outcome. Depending on the extracted data,
an attacker can perform a variety of different actions that can range from low
to critical severity.

Below are some real-world scenarios that illustrate the various effects of an
attacker exploiting the vulnerability.

```

rules/S5144/impact.adoc
```diff
@@ -0,0 +1,27 @@
...
+==== Local file read to host takeover
+
+An attacker manipulates an application into performing a local request for a
+sensitive file, such as *~/.ssh/id_rsa*, by using the File URI scheme
```
loris-s-sonarsource, line 15:
```suggestion
sensitive file, such as `~/.ssh/id_rsa`, by using the File URI scheme
```

rules/S5144/impact.adoc
```diff
@@ -0,0 +1,27 @@
...
+*file://*. +
```
loris-s-sonarsource, line 16:
```suggestion
`file://`. +
```

rules/S5144/impact.adoc
```diff
@@ -0,0 +1,27 @@
...
+Once in possession of the SSH keys, the attacker establishes a remote
+connection to the system hosting the web application.
+
+==== Internal Network Reconnaissance
+
+An attacker enumerates internal accessible ports from the affected server or
+others to which the server can communicate by iterating over the port field in
+the URL *http://127.0.0.1:{port}*. +
```
loris-s-sonarsource, line 24:
```suggestion
the URL `http://127.0.0.1:{port}`. +
```

rules/S5144/impact.adoc
```diff
@@ -0,0 +1,27 @@
...
+Taking advantage of other supported URL schemas, for example,
+*gopher://127.0.0.1:3306*, an attacker would be able to connect to a database
```
loris-s-sonarsource, line 26:
```suggestion
`gopher://127.0.0.1:3306`, an attacker would be able to connect to a database
```

rules/S5144/rationale.adoc
```diff
@@ -0,0 +1,8 @@
+Server-Side Request Forgery (SSRF) occurs when attackers can coerce a server to
+perform arbitrary requests on their behalf. +
+
+A SSRF vulnerability can either be basic or blind depending on whether the
+fetched data is directly returned in the web application's response. +
+
+It is a common remote entry point for attackers on their victim's organization
+infrastructure.
```
loris-s-sonarsource, line 8: These two paragraphs are interesting, but talk about the exploitation techniques and it is out of scope. Maybe the intention implied by the mention of "blind" could be to warn users about the fact that attackers may exploit this vulnerability even without textual input.

I suggest:
* we add the warning above in the impact section (Like `**Note**: attackers may exploit this vulnerability even without textual input.`)
* And we add the common remote entry point part in the impact.adoc's introduction

rules/S5144/impact.adoc
loris-s-sonarsource, line 26: Suggestion: That we add accuracy here, and specify that this situation depends on the vulnerable systems.

loris-s-sonarsource (changes requested):
Overall Looks good and needs a few feedback changes before merging

rules/S5144/rationale.adoc
pedro-oliveira-sonarsource: I am not following on how categorizing the types of SSRF is describing the exploitation techniques. See [S5131](https://github.com/SonarSource/rspec/blob/feature/education/rules/S5131/rationale.adoc).

loris-s-sonarsource: Ok, after giving it a thought, here is what I think about this.

About the paragraph mentioning blind and regular ones: I think it doesn't give more added value because it doesn't go into details on how this is something developers should have in mind. And going too much into details would be out of scope since it is likely to slip into exploitation explanations. 

Thus, keeping in mind that we target developers (and keeping in mind that a developer might ask `"ok, why should I care?"`, that's why I would suggest either:
1. not mentioning it to avoid information overload
2. making a 180° turn and say:
```
Depending on the application code, attackers may be able to exploit this
vulnerability without text output. This vulnerability would be referred to
as "blind SSRF." +
This means that the absence of textual output is not the same as the
absence of a vulnerability.
```
This could keep the "fun fact", and mentioning `attackers` adds clarity and give added-value without going to much into details. 

The why could be: `Developers won't be able to say "well, nothing's output so this is fine 🔥 ☕ 🐶 🔥"`

---

About the second text, after giving it more thought, I think it is pretty subjective and shouldn't be there without proper statistics to support what we say. So I would suggest:
1. Not mentioning it
2. Turning it into this another sentence:

```
This entry point is commonly used by remote attackers to pivot on their
victim's organization infrastructure. 
```

loris-s-sonarsource (changes requested):
Multiple suggestions:

1. I suggest removing the csharp and java text, because it's clutter for future readers of this PR
2. There's a pending suggestion for wording with "untrusted data", that the team agreed upon in the past
3. And I added a comment to clarify my thoughts about the rationale

rules/S5144/rationale.adoc
pedro-oliveira-sonarsource: Removed the mention for the common entrypoint and adjusted the SSRF categorisation description to be more developer focus by applying the last sentence on your suggestion.

rules/S5144/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,7 @@
+Untrusted data, such as GET or POST request content, should always be
+considered tainted. An application should avoid performing server-side requests
+based on taintable data.
+
+When such a feature is strictly necessary, SSRF could be mitigated by applying
+an allow-list of trustable domains or by providing an API which indexes the
+allowed domains, example *http://example.com?url=1*.
```
loris-s-sonarsource, line 7:
```suggestion
allowed domains, such as `http://example.com?url=1`.
```
technical bits should be highlighted with the code syntax. And "example" seems incomplete here

rules/S5144/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,15 @@
+=== How to fix it in ASP.NET
```
loris-s-sonarsource, line 1: This file should be deleted

rules/S5144/rationale.adoc
```diff
@@ -0,0 +1,7 @@
+Server-Side Request Forgery (SSRF) occurs when attackers can coerce a server to
+perform arbitrary requests on their behalf. +
+
+An SSRF vulnerability can either be basic or blind, depending on whether the
+server's fetched data is directly returned in the web application's response. +
+The absence of the corresponding response for the coerced request on
+the application does not imply the absence of vulnerability.
```
loris-s-sonarsource, line 7: I think this is not completely clear if we don't mention attackers. I don't think the answer "why should I care" from a developer is completely answered if we don't face them with facts.

Here from my POV, we are emphasizing that " the application does not imply the absence of vulnerability. " but IMHO this doesn't give a reaction to people. What gives them reactions IMHO is to mention actual attacks.


My advice in general:
I think that neutrally saying "it could be like this or like that" doesn't bring reactions in general if we don't face people with the answer to "why should I care?".

I would suggest something like this
```suggestion
An SSRF vulnerability can either be basic or blind, depending on whether the
server's fetched data is directly returned in the web application's response. +
Attackers know about this fact and are equipped to exploit this vulnerability
in both cases.

**The absence of the corresponding response for the coerced
request on the application does not imply the absence of vulnerability.**
```

loris-s-sonarsource (changes requested):
Almost LGTM, will look good after this one

And I think we should add articles (slightly out of scope but provided by R&D):
* https://blog.assetnote.io/2021/01/13/blind-ssrf-chains/

rules/S5144/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,7 @@
+Untrusted data, such as GET or POST request content, should always be
```
loris-s-sonarsource, line 1: It's missing the title:
```
=== How does this work?
```

rules/S5144/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,7 @@
...
+considered tainted. An application should avoid performing server-side requests
+based on taintable data.
+
+When such a feature is strictly necessary, SSRF could be mitigated by applying
```
loris-s-sonarsource, line 5: Why don't you use our already existing patterns?

Like this one: https://github.com/SonarSource/rspec/blob/e499ee902159f5710c59c60f22797c92f30c357a/rules/S5147/common/fix/pre-approved-list.adoc

loris-s-sonarsource (approved):
LGTM

PR #1208, opened 2022-09-02
Rule S2259: Add VB.NET, improve C#

pavel-mikula-sonarsource:
Related to https://github.com/SonarSource/sonar-dotnet/issues/5861

### DIFF
--- a/rules/S2259/csharp/rule.adoc
+++ b/rules/S2259/csharp/rule.adoc
@@ -11,7 +11,7 @@ public void Method()
 }
 ----
 
-== Compliant Code Example
+== Compliant Solution
 
 [source,csharp]
 ----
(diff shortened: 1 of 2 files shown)


rules/S2259/csharp/rule.adoc
```diff
@@ -1,28 +1,34 @@
...
 == Exceptions
 
-Calls to extension methods are not reported because they can still operate on ``++null++`` values.
-
+Calls to extension methods are not reported because they can still operate on `null` values.
 
-To create a custom null validation method declare an attribute with name ``++ValidatedNotNullAttribute++`` and mark the parameter that is validated for null in your method declaration with it:
+To create a custom null validation method declare an attribute with name `ValidatedNotNullAttribute` and mark the parameter that is validated for null in your method declaration with it:
 
+[source,csharp]
```
martin-strecker-sonarsource, line 31: Should we add `[CallerMemberName]` in the sample?

martin-strecker-sonarsource (approved):
The sample could be improved though.

rules/S2259/csharp/rule.adoc
pavel-mikula-sonarsource: I'd keep it simple. The name is not the focus of the rule and could drag the attention sideways.

PR #1217, opened 2022-09-06
[APPSEC-122] Modify rule S6287: Educational content (Common)

pedro-oliveira-sonarsource:
Specification ticket: [APPSEC-122](https://sonarsource.atlassian.net/browse/APPSEC-122).


rules/S6287/impact.adoc
```diff
@@ -0,0 +1,20 @@
+=== What is the potential impact?
+
+Session Cookie Injection makes it possible for an attacker to force a victim
+into browsing within an application session to which they also have access.
```
loris-s-sonarsource, line 4: # Multiple linked suggestions

##  Introductory impact

This sentence is slightly confusing me near the end. 

Before all, I would suggest we use the same turns of phrase as in the xss rule. e.g. we introduce the subject of phishing in the rationale, and then say:

```suggestion
A well-intentioned user opens a malicious link that injects a cookie in the web
application. This cookie forces the user into unknowingly browsing into a session
that isn't theirs.
```

And then I would suggest using this premise in the following subsection. 
For example, by starting with `"If the application does X this way, Y wan happen"`

This could allow us to get rid of the introductory `An attacker manipulates a user into browsing an application using a session cookie[...]`

##  Threat examples

* In the "Sensitive Data Disclosure", section, thanks to the introductory stuff I suggest, we could talk specific handling of personal data, and talk about sensitive stuff (credit cards, names, addresses, application-specific sensitive stuff)

* In the Vulnerability chaining, maybe we could also mention some little XSS impacts, not a lot but just enough to make think about impacts

WDYT?

rules/S6287/rationale.adoc
```diff
@@ -0,0 +1,7 @@
+Session Cookie Injections occur in a web application when it assigns session
+cookies to users using untrusted
+data.
```
loris-s-sonarsource, line 3:
```suggestion
Session Cookie Injections occur in a web application when it assigns session
cookies to users using untrusted data.
```

rules/S6287/rationale.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+The injection might occur via a GET parameter, and the payload, for example,
+`https://example.com?cookie=injectedcookie`, delivered using phishing
+techniques.
```
loris-s-sonarsource, line 7: I would also use what's already been done in the XSS rule for consistency:

```
When well-intentioned users open a link to a page that is vulnerable to
session cookie injection, they are exposed to attacks that target their own
browser.

A user with malicious intent carefully crafts the link beforehand, and after
creating this link, the attacker must use phishing techniques to ensure that
his target users click on the link.
```

wdyt?

loris-s-sonarsource (changes requested):
A few suggestions to add stuff, it's looking good already 👍

rules/S6287/rationale.adoc
pedro-oliveira-sonarsource: I think moving a little outside of those constant structures is good as long as it is correct, or the readers will keep reading the same thing over and over again.

rules/S6287/impact.adoc
```diff
@@ -0,0 +1,22 @@
...
+that isn't theirs.
+
+Below are some real-world scenarios that illustrate the various effects of
+an attacker exploiting the vulnerability.
+
+==== Sensitive data disclosure
+
+A victim introduces sensitive data within the attacker's application session that
+can later be retrieved by them. This can lead to a variety of implications depending
+on what type of data is disclosed.
```
loris-s-sonarsource, line 14: Why not giving an example and talk about data sensitivity, such as personal data?
A text like the one below could actually give an example and still keep text concise, since we only have like 3 lines:

```
For example, leaks could involve strictly confidential user data or organizational data.
```

rules/S6287/rationale.adoc
loris-s-sonarsource: Well, the suggested text is correct; and this structure could also be used only in phishing-related rules, so for now that's only the XSS rule and this one. 

To be clear with my initial intention, it was not to overwrite everything but to merge and adjust them and give added value in terms of education. The goal of the XSS text is to add more pedagogy.

I think both suggestions (the PR one and the XSS one) are valid on their own, but I still think that we miss out on user-friendly education here by not merging them. And since it's already validated by the team, why not use it?

rules/S6287/impact.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+==== Vulnerability chaining
+
+An attacker not only manipulates a user into browsing an application using a session
+cookie of their control but also successfully detects and exploits a self-XSS
+on the target application. The victim browses the vulnerable page using the attacker's
+session and is affected by the XSS, which can then be used for a wide range of attacks
+including credential stealing using mirrored login pages.
```
loris-s-sonarsource, line 22:
```suggestion
on the target application. +
The victim browses the vulnerable page using the attacker's session and is
affected by the XSS, which can then be used for a wide range of attacks
including credential stealing using mirrored login pages.
```

I suggest a line break to make the paragraph breathable

loris-s-sonarsource (approved):
LGTM, with some suggestions and discussions

rules/S6287/rationale.adoc
pedro-oliveira-sonarsource: What is the added pedagogy value from the XSS text compared to what was written?

rules/S6287/impact.adoc
pedro-oliveira-sonarsource: Was following along what was written in for LDAP "The impact of this vulnerability depends on how vital LDAP servers are to the organization.".

rules/S6287/rationale.adoc
loris-s-sonarsource: @pedro-oliveira-sonarsource 

> What is the added pedagogy value from the XSS text compared to what was written?
The "story-telling" writing style of the XSS is its educative added value, because it adds continuity and user-friendliness to the text.

> I think this is extremely vague 
The goal of the XSS text is to introduce the impact in a user-friendly way. Here, after reading this, the question that could be asked is "what attacks ?" And the impacts section aims to actually answer this. 

> I think it lacks concision for this rule.
This is actually the same size as the suggestion of the PR, so I'm not sure I understand. And since the initial goal is to introduce impacts, maybe we have a misalignment.

rules/S6287/rationale.adoc
```diff
@@ -0,0 +1,10 @@
+Session Cookie Injections occur when a web application assigns session cookies
+to users using untrusted data.
+
+Session cookies are used by web applications to identify users; thus,
+controlling these enable control over the identity of the users within the
+application.
```
loris-s-sonarsource, line 6:
```suggestion
Session cookies are used by web applications to identify users. Thus,
controlling these enable control over the identity of the users within the
application.
```

rules/S6287/rationale.adoc
loris-s-sonarsource: Solved in internal discussion

rules/S6287/impact.adoc
loris-s-sonarsource: 👍

PR #1225, opened 2022-09-06
Modify rule S2631(C#): Update text to Education Framework structure (APPSEC-127)

loris-s-sonarsource:


rules/S2631/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,45 @@
+=== How to fix it in .NET
+
+The following non-compliant code is vulnerable to Regex Denial of Service
+because untrusted data is used as a pattern to scan a string without prior
```
pedro-oliveira-sonarsource, line 4:
```suggestion
because untrusted data is used as a regex to scan a string without prior
```

rules/S2631/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+----
+|===
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+In the example, `Regex.Escape` escapes metacharacters and escape sequences that
```
pedro-oliveira-sonarsource, line 43:
```suggestion
In the compliant solution example, `Regex.Escape` escapes metacharacters and escape sequences that
```

pedro-oliveira-sonarsource: Same suggestion as on the other PR, we could write something similar to what is on this method description https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.escape?view=net-6.0.

pedro-oliveira-sonarsource (approved):
LGTM, just a minor suggestion.

PR #1245, opened 2022-09-12
Modify rule S5332: Adding AWS CDK support for the Python language

gaetan-ferry-sonarsource:
Implementation ticket: https://sonarsource.atlassian.net/browse/SONARPY-1119

### DIFF
--- a/rules/S5332/python/rule.adoc
+++ b/rules/S5332/python/rule.adoc
@@ -59,10 +59,10 @@ from aws_cdk import (
 
 elbv2.ApplicationListener(
     self,
-    "listener-https-explicit-const",
+    "listener-http-explicit-const",
     load_balancer=lb,
-    protocol=elbv2.ApplicationProtocol.HTTP,
-    port=443,
+    protocol=elbv2.ApplicationProtocol.HTTP, #Sensitive
+    port=8081,
     open=True
 )
 ----
--- a/rules/S5332/python/rule.adoc
+++ b/rules/S5332/python/rule.adoc
@@ -88,6 +88,40 @@ loadBalancer.add_listener(
 )
 ----
 
+For https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_elasticache/CfnReplicationGroup.html[aws_cdk.aws_elasticache.CfnReplicationGroup]:
+
+[source,python]
+----
+from aws_cdk import (
+    aws_elasticache as elasticache
+)
+
+elasticache.CfnReplicationGroup(
+    self,
+    "unencrypted-explicit",
+    replication_group_description="a replication group",
+    automatic_failover_enabled=False,
+    transit_encryption_enabled=False, # Sensitive
+    cache_subnet_group_name="test",
+    engine="redis",
+    engine_version="3.2.6",
+    num_cache_clusters=1,
+    cache_node_type="cache.t2.micro"
+)
+
+elasticache.CfnReplicationGroup( # Sensitive, encryption is disabled by default
+    self,
+    "unencrypted-implicit",
+    replication_group_description="a test replication group",
+    automatic_failover_enabled=False,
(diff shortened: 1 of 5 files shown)


rules/S5332/python/rule.adoc
```diff
@@ -21,6 +22,279 @@ import smtplib
 smtp = smtplib.SMTP("smtp.example.com", port=587) # Sensitive
 ----
 
+For https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_elasticloadbalancingv2/ApplicationLoadBalancer.html[aws_cdk.aws_elasticloadbalancingv2.ApplicationLoadBalancer]
```
hendrik-buchwald-sonarsource, line 25: I think it makes sense to have a colon after this sentence. The same applies of course to all the code samples below.

```suggestion
For https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_elasticloadbalancingv2/ApplicationLoadBalancer.html[aws_cdk.aws_elasticloadbalancingv2.ApplicationLoadBalancer]:

```

rules/S5332/python/rule.adoc
```diff
@@ -21,6 +22,279 @@ import smtplib
...
+)
+
+lb.add_listener( # Sensitive
+    "Listener-default",
+    port=80,
+    open=True
+)
+lb.add_listener(
+    "Listener-http-explicit",
+    protocol=elbv2.ApplicationProtocol.HTTP, #Sensitive
```
hendrik-buchwald-sonarsource, line 46: Would be good to be consistent here and use `# Sensitive`. I would recommend a find-replace in this document.

```suggestion
    protocol=elbv2.ApplicationProtocol.HTTP, # Sensitive
```

rules/S5332/python/rule.adoc
```diff
@@ -21,6 +22,279 @@ import smtplib
...
+)
+
+lb = elbv2.ApplicationLoadBalancer(
+    self,
+    "LB",
+    vpc=vpc,
+    internet_facing=True
+)
+
+lb.add_listener( # Sensitive
```
hendrik-buchwald-sonarsource, line 39: I am not 100% sure if the comment should be here or in the line `port=80`. Is the absence of `protocol` considered to be the problem? In that case, the comment is right where it is.

rules/S5332/python/rule.adoc
```diff
@@ -39,6 +313,322 @@ smtp.starttls(context=context)
 smtp_ssl = smtplib.SMTP_SSL("smtp.gmail.com", port=465) # Compliant
 ----
 
+For https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_elasticloadbalancingv2/ApplicationLoadBalancer.html[aws_cdk.aws_elasticloadbalancingv2.ApplicationLoadBalancer]
```
hendrik-buchwald-sonarsource, line 316:
```suggestion
For https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_elasticloadbalancingv2/ApplicationLoadBalancer.html[aws_cdk.aws_elasticloadbalancingv2.ApplicationLoadBalancer]:

```

hendrik-buchwald-sonarsource (changes requested):
I think it looks good in general! I have left some comments with a few changes.

A tip for the future though: I don't think it would be necessary here to have a code sample for every service. A lot of them are very similar. If the only thing that is different is the parameter name, we can convey this information through the message that is displayed for the issue.
No need to remove the code samples now that they are here

hendrik-buchwald-sonarsource:
Also, for RSPEC PRs please always add `Implementation ticket: XXX` in the first comment, unless there is no related implementation ticket.

hendrik-buchwald-sonarsource (approved):
LGTM

PR #1251, opened 2022-09-13
[APPSEC-63] Modify Rules: Fix ASVS assignments

pedro-oliveira-sonarsource:
[APPSEC-63](https://sonarsource.atlassian.net/browse/APPSEC-63)


rules/S3361/metadata.json
```diff
@@ -37,12 +37,7 @@
       "8.2.8"
     ],
     "ASVS 4.0": [
-      "2.8.1",
-      "2.8.6",
-      "3.3.1",
```
loris-s-sonarsource, line 42: >  | **3.3.1** | Verify that logout and expiration invalidate the session token, such that the back button or a downstream relying party does not resume an authenticated session, including across relying parties. ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | ✓ | ✓ | ✓ | 613 | 7.1 |

I think this one might make sense, except if you consider that we need to interpret the first "and" as a "strong logical and"

loris-s-sonarsource (approved):
Great ! LGTM.

For future reference, I cloned the ASVS repo and used the following command to review this PR:

```bash
clear && pbpaste | grep -Eo "[0-9\.]*" |xargs -I {} grep {} ASVS/5.0/en/ -R  --color=auto
```

rules/S3361/metadata.json
pedro-oliveira-sonarsource: 3.3.1 was removed as it refers to insufficient session expiration, i.e., lack of session token validity termination, whereas S3361 is specific to insufficient session timeout, i.e., the period definition of which a session token is valid.

PR #1256, opened 2022-09-14
Modify All Current Education Rules: Support intuitive view

loris-s-sonarsource:
I recommend you use the commits for the review. At some point, a '$' file is created, I deleted it after that.


pedro-oliveira-sonarsource (approved):
LGTM. A PR will be created to fix the OR operand character escape (e.g. \|).

PR #1260, opened 2022-09-15
Modify rule S107: Add an exception for TypeScript constructors

yassin-kammoun-sonarsource:


rules/S107/javascript/rule.adoc
```diff
@@ -20,6 +20,23 @@ function doSomething(param1, param2, param3, param4) {
 }
 ----
 
+== Exceptions
+
+The rule ignores constructors where parameters are all parameter properties:
```
ilia-kebets-sonarsource, line 25:
```suggestion
The rule ignores constructors where *all* parameters are properties:
```

ilia-kebets-sonarsource (approved):
LGTM, what do you think about my rephrasing?

rules/S107/javascript/rule.adoc
yassin-kammoun-sonarsource: As discussed, let's use TypeScript terminology, that is, parameter properties. Still, it makes sense to put "all" in bold. Thanks for suggesting that.

PR #1263, opened 2022-09-15
Modify Rule S6258(IAC): Add additionnal util recommendation

loris-s-sonarsource:


rules/S6258/recommended.adoc
```diff
@@ -1,3 +1,8 @@
 == Recommended Secure Coding Practices
 
-Enable the logging capabilities of this component.
+Enable the logging capabilities of this component. Depending on the component, 
+new permissions might be required by the logging storage components. +
+For example, AWS Application Load Balancer Access Logs
+https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html[require
+an additional bucket policy].
```
pedro-oliveira-sonarsource, line 7: If we could personalize depending on which Cloud platform the rule is being flagged on I would be up to it but not sure we should be giving an AWS example when it flags for Azure or GCP. WDYT?

rules/S6258/recommended.adoc
```diff
@@ -1,3 +1,8 @@
 == Recommended Secure Coding Practices
 
-Enable the logging capabilities of this component.
+Enable the logging capabilities of this component. Depending on the component, 
+new permissions might be required by the logging storage components. +
```
pedro-oliveira-sonarsource, line 4:
```suggestion
Enable the logging capabilities of this component. Applying write
permissions in the logging storage resource may be required. +
```

rules/S6258/recommended.adoc
pedro-oliveira-sonarsource, line 7:
```suggestion
```

pedro-oliveira-sonarsource (changes requested):
Some suggestions.

rules/S6258/recommended.adoc
loris-s-sonarsource: Slack discussion: Other permissions may be required, more than write

rules/S6258/recommended.adoc
loris-s-sonarsource: disregarded, as per slack convo

loris-s-sonarsource: Yeah, I thought about that, and I still wrote this because I would emphasize the users to directly go see their docs. Most of the time it won't flag for a specific AWS ALB, so it's just an example here

loris-s-sonarsource:
```suggestion
You should consult the official documentation to enable logging for the impacted components.
For example, AWS Application Load Balancer Access Logs
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html[require
an additional bucket policy].
```

pedro-oliveira-sonarsource (approved):
LGTM.

PR #1272, opened 2022-09-19
Drop PHP profiles PSR-2 and Drupal

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! I saw the jira ticket, indeed the goal is to remove PSR-2 and Drupal profiles (and I guess put back "Sonar Way" as the default in most cases ?).
Also, couldn't this be the opportunity to remove the rule S1117 from the default profile ?

PR #1282, opened 2022-09-26
Modify rule S2589: Make sure Python code raises expected issues

andreaguarino:


rules/S2589/python/rule.adoc
```diff
@@ -4,30 +4,26 @@ include::../description.adoc[]
 
 [source,python]
 ----
-a = True;
-if a:  # Noncompliant
-    doSomething()
+def f(b):
+    a = True
+    if a:  # Noncompliant
+        doSomething()
```
guillaume-dequenne-sonarsource, line 10: While you're changing these snippets, I think it may be interesting to change the camel case to snake case to make the snippet more relatable to Python developers (though I believe we might have many snippets using camel cases and consistency might be better...but I'm not sure we're currently very consistent though...). Feel free to ignore this, though.

guillaume-dequenne-sonarsource (approved):
LGTM, left a minor suggestion unrelated to the change.

PR #1292, opened 2022-09-28
Modify rule S107: Add exception for Angular component constructors

yassin-kammoun-sonarsource:


ilia-kebets-sonarsource (approved):
LGTM!

PR #1302, opened 2022-09-30
provide missing quickfixes information

alban-auzeill:


dorian-burihabwa-sonarsource (approved):
Thanks for completing the gaps

PR #1322, opened 2022-10-07
Modify Rule S1226: Add exception for argument reassignments inside if statements

victor-diez-sonarsource:


rules/S1226/javascript/rule.adoc
```diff
@@ -13,6 +13,23 @@ function MyClass(name, strings) {
 }
 ----
 
+== Exceptions
+
+There is a common pattern is Javascript to overwrite certain parameters depending on other optional arguments being passed. For example, a callback is by convention always passed in the last position, meaning that if a parameter is a previous position was not passed, the callback will be passed in its position instead.
+
+Because of this pattern, parameter reassignments which are inside an `if` statement block will not raise an issue.
```
yassin-kammoun-sonarsource, line 20:
```suggestion
Therefore, the rule ignores parameter reassignments that are inside an `if` statement block.
```

rules/S1226/javascript/rule.adoc
```diff
@@ -13,6 +13,23 @@ function MyClass(name, strings) {
 }
 ----
 
+== Exceptions
+
+There is a common pattern is Javascript to overwrite certain parameters depending on other optional arguments being passed. For example, a callback is by convention always passed in the last position, meaning that if a parameter is a previous position was not passed, the callback will be passed in its position instead.
```
yassin-kammoun-sonarsource, line 18:
```suggestion
There is a common pattern in JavaScript to overwrite certain parameters depending on other parameters that are optional. For example, a callback is by convention always passed in the last position. If a parameter in a previous position was not passed, the callback will be passed in its position instead.
```

rules/S1226/javascript/rule.adoc
```diff
@@ -13,6 +13,23 @@ function MyClass(name, strings) {
...
+
+Because of this pattern, parameter reassignments which are inside an `if` statement block will not raise an issue.
+
+[source,javascript]
+----
+// Use the initialState as a default value
```
yassin-kammoun-sonarsource, line 24: I am not sure to understand how this comment is related to the function. Let's just drop it.

yassin-kammoun-sonarsource (changes requested):

rules/S1226/javascript/rule.adoc
```diff
@@ -13,6 +13,22 @@ function MyClass(name, strings) {
 }
 ----
 
+== Exceptions
+
+There is a common pattern in Javascript to overwrite certain parameters depending on other parameters that are optional. For example, a callback is by convention always passed in the last position. If a parameter in a previous position was not passed, the callback will be passed in its position instead.
```
yassin-kammoun-sonarsource, line 18:
```suggestion
There is a common pattern in JavaScript to overwrite certain parameters depending on other parameters that are optional. For example, a callback is by convention always passed in the last position. If a parameter in a previous position was not passed, the callback will be passed in its position instead.
```

yassin-kammoun-sonarsource (approved):

PR #1325, opened 2022-10-10
Modify rule S4421: Remove EXTRACT from the list of deprecated functions.

mary-georgiou-sonarsource:
EXTRACT was removed from the rule for now, as it was creating false positives.


andrei-epure-sonarsource (approved):
LGTM

PR #1327, opened 2022-10-10
Modify rule S6333: Creating public APIs is security sensitive - apigateway v1 python examples

gaetan-ferry-sonarsource:


loris-s-sonarsource (approved):
LGTM

PR #1329, opened 2022-10-10
Modify rule S2083(Python): Update text to Education Framework structure (APPSEC-200)

loris-s-sonarsource:


rules/S2083/python/how-to-fix-it/flask.adoc
```diff
@@ -0,0 +1,33 @@
...
+----
+@app.route('/example')
+def example():
+    my_file = request.args['my_file']
+    return send_from_directory('static', my_file)
+----
+
+=== How does this work?
+
+include::../../common/fix/how-does-this-work.adoc[]
```
pierre-loup-tristant-sonarsource, line 28: This renders the following content:

> The universal way to prevent path injection is to validate paths constructed from untrusted data.
> 
> The validation should be done as follow:
> 
> Resolve the canonical path of the file by using methods like `send_from_directory`. This will resolve relative path or path components like ../ and removes any ambiguity regarding the file’s location.
> 
> Check that the canonical path is within the directory where the file should be located.
> 
> Ensure the target directory path ends with a forward slash to prevent partial path traversal, for example, /base/dirmalicious starts with /base/dir but does not start with /base/dir/.

It's incorrect because:
- `send_from_directory` doesn't resolve the canonical path.
- it already checks that the sent file is within the specified directory

This part should simply explain that `send_from_directory` will restrict the file path to a specific folder and therefore it will prevent path traversal from happening.

pierre-loup-tristant-sonarsource (changes requested):
Please update the "How does this work?" section so that its explains how `send_from_directory` prevent path traversal.

docs/allowed_framework_names.adoc
```diff
@@ -21,6 +21,7 @@
 * Laravel
 * Symfony
 // Python
+* Python Standard Library
```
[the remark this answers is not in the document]
loris-s-sonarsource: Good point, forgot it after removing the standard library code

rules/S2083/common/pitfalls/python-path-join.adoc
```diff
@@ -0,0 +1,12 @@
+==== os.path.join(path, \*paths)
+
+This function should not be used as a validator.
+
+The standard library states:*"if a component is an absolute path, all previous
```
loris-s-sonarsource, line 5:
```suggestion
The standard library states: *"if a component is an absolute path, all previous
```

rules/S2083/common/fix/function-based-validation.adoc
```diff
@@ -0,0 +1,13 @@
+==== Use functions that combine feature and validation
+
+The cleanest way to fix this vulnerability is to use a function that combines:
+
+* the retrieval of a file
+* restrict the retrieval of a file to a specific folder
+
+Built-in APIs from libraries like {auto_canonicalization_function} benefit from
```
loris-s-sonarsource, line 8:
```suggestion
Built-in APIs from libraries like `{auto_canonicalization_function}` benefit from
```

rules/S2083/python/how-to-fix-it/flask.adoc
```diff
@@ -0,0 +1,48 @@
...
+    my_file = request.args['my_file']
+    return send_from_directory('static', my_file)
+----
+
+=== How does this work?
+
+The universal way to prevent path injection is to validate paths constructed
+from untrusted data. 
+
+:auto_canonicalization_function: `send_from_directory`
```
loris-s-sonarsource, line 42:
```suggestion
:auto_canonicalization_function: send_from_directory
```

rules/S2083/common/fix/function-based-validation.adoc
```diff
@@ -0,0 +1,13 @@
+==== Use functions that combine feature and validation
```
pierre-loup-tristant-sonarsource, line 1: What about "Use secure-by-design API"

rules/S2083/common/fix/self-validation.adoc
```diff
@@ -1,4 +1,4 @@
-The universal way to prevent path injection is to validate paths constructed from untrusted data. 
+==== Validate data yourself
```
pierre-loup-tristant-sonarsource, line 1: What about "Validation of canonical path"

rules/S2083/common/fix/function-based-validation.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+* the retrieval of a file
+* restrict the retrieval of a file to a specific folder
+
+Built-in APIs from libraries like `{auto_canonicalization_function}` benefit from
+quality maintenance regarding features, security, or hardening. It is usually
+better to use these solutions than to develop your own.
+
+Behind the scenes, this function protects against both regular and partial path
+injection.
```
pierre-loup-tristant-sonarsource, line 13: Isn't it a bit too sujective?
> The cleanest way
> Benefit from quality maintenance

Here is what I would suggest
```suggestion
A common way to fix this vulnerability is to use a secure-by-design API that combines:

* the retrieval of a file
* restrict the retrieval of a file to a specific folder

Built-in APIs like `{auto_canonicalization_function}` were designed to prevent path traversal. It is usually
better to use these solutions than to develop your own.

Behind the scenes, these functions protects against both regular and partial path
injection.
```

rules/S2083/common/fix/self-validation.adoc
```diff
@@ -1,4 +1,4 @@
-The universal way to prevent path injection is to validate paths constructed from untrusted data.
```
pierre-loup-tristant-sonarsource, line 1: I think this sentence should stay with the "Validation of canonical path" content (see my other comment at the beginning of "How does this work?").

rules/S2083/python/how-to-fix-it/flask.adoc
```diff
@@ -0,0 +1,48 @@
...
+@app.route('/example')
+def example():
+    my_file = request.args['my_file']
+    return send_from_directory('static', my_file)
+----
+
+=== How does this work?
+
+The universal way to prevent path injection is to validate paths constructed
+from untrusted data.
```
pierre-loup-tristant-sonarsource, line 40: To me, it's a bit strange to start by explaining that the universal way to fix the vuln is to validate the constructed path. No validation is required when using `send_from_directory`, devs just uses a safe-by-design API. Only at the end of the section, do users understand that the validation happens "Behind the scenes". I find it hard to reconnect with this introduction sentence.

pierre-loup-tristant-sonarsource (approved):
LGTM. Please, have a look at the improvements I suggested before merging.

rules/S2083/common/pitfalls/python-path-join.adoc
```diff
@@ -0,0 +1,12 @@
...
+This function should not be used as a validator.
+
+The standard library states: *"if a component is an absolute path, all previous
+components are discarded, and linking continues from the component with the
+absolute path."*
+
+This means that including untrusted data in any of the path parameters can lead
+to a full or partial path traversal vulnerability.
+
+
```
pierre-loup-tristant-sonarsource, line 12: This is one of the pitfalls Denis described in his blog post. https://blog.sonarsource.com/10-unknown-security-pitfalls-for-python/
It is probably worth mentioning it.

rules/S2083/common/fix/function-based-validation.adoc
loris-s-sonarsource:
```suggestion
==== Use secure-by-design APIs
```

rules/S2083/python/how-to-fix-it/flask.adoc
loris-s-sonarsource:
```suggestion
:auto_canonicalization_function: send_from_directory

The universal method to prevent path injection is to validate paths created
from untrusted data. This can be done either manually or automatically,
depending on whether the library includes a data sanitization feature and the
required function.

Here, {auto_canonicalization_function} can be considered a secure-by-design API.
```

rules/S2083/common/fix/function-based-validation.adoc
loris-s-sonarsource: Arf, I'm not really comfortable with this suggestion, but I think I'm onto something

rules/S2083/common/fix/function-based-validation.adoc
```diff
@@ -0,0 +1,13 @@
+==== Use functions that combine feature and validation
+
+The cleanest way to fix this vulnerability is to use a function that combines:
+
+* the retrieval of a file
+* restrict the retrieval of a file to a specific folder
```
loris-s-sonarsource, line 6:
```suggestion
* restrict the retrieval of a file to a specific folder
* A feature -- such as file download, or file deletion.
```

rules/S2083/common/fix/function-based-validation.adoc
```diff
@@ -0,0 +1,13 @@
+==== Use functions that combine feature and validation
+
+The cleanest way to fix this vulnerability is to use a function that combines:
```
loris-s-sonarsource, line 3:
```suggestion
Some libraries contain APIs these three capabilities:
```

rules/S2083/common/fix/function-based-validation.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+* the retrieval of a file
+* restrict the retrieval of a file to a specific folder
+
+Built-in APIs from libraries like `{auto_canonicalization_function}` benefit from
+quality maintenance regarding features, security, or hardening. It is usually
+better to use these solutions than to develop your own.
```
loris-s-sonarsource, line 10:
```suggestion
They can be referred to as "secure-by-design" APIs. Using this type of library,
such as '{auto_canonicalization_function}', brings multiple layers of security
to the code while keeping the code base shorter.
```

rules/S2083/common/pitfalls/python-path-join.adoc
loris-s-sonarsource:
```suggestion
==== os.path.join(path, \*paths)

This function should not be used as a validator.

The standard library states: *"if a component is an absolute path, all previous
components are discarded, and linking continues from the component with the
absolute path."*

This means that including untrusted data in any of the path parameters can lead
to a full or partial path traversal vulnerability.

https://blog.sonarsource.com/10-unknown-security-pitfalls-for-python/[This Sonar blog post] talks about this issue.
```

rules/S2083/common/fix/self-validation.adoc
```diff
@@ -1,4 +1,4 @@
-The universal way to prevent path injection is to validate paths constructed from untrusted data. 
+==== Validate data yourself
 
 The validation should be done as follow:
```
loris-s-sonarsource, line 3:
```suggestion
==== Validate data yourself

==== Self-validation of canonical paths

If it is not possible to use secure-by-design APIs that do this automatically, validation should be done as follows:
```

PR #1332, opened 2022-10-11
Modify rule S5883: Update PHP text to education framework structure[APPSEC-199]

gaetan-ferry-sonarsource:


rules/S5883/php/how-to-fix-it/core.adoc
```diff
@@ -0,0 +1,96 @@
...
+injection of subsequent ones.
+
+It is also important not to combine both `escapeshellarg` and `escapeshellcmd`.
+Indeed, a call to `escapeshellcmd` on a complete command line will void any
+escaping previously added with `escapeshellarg`.
+
+For that reason, it is not possible to prevent an argument injection issue in
+the `mail` function with `escapeshellarg`. Indeed, `mail` internally relies on
+`escapeshellcmd` for escaping purposes. In that case, a whitelist of explicitely
+allowed additional arguments should be used.
```
pedro-oliveira-sonarsource, line 94:
```suggestion
When building a system command from user-submitted data is unavoidable, using
the `escapeshellarg` sanitizing function should be preferred. It ensures that
the provided data will be considered a single argument and prevents the
injection of subsequent ones.

It is also important not to combine both `escapeshellarg` and `escapeshellcmd`.
Indeed, a call to `escapeshellcmd` on a complete command line will void any
escaping previously added with `escapeshellarg`.

Therefore, it is impossible to prevent an argument injection issue in
the `mail` function with `escapeshellarg`. Indeed, `mail` internally relies on
`escapeshellcmd` for escaping purposes. In that case, an allowlist of explicitly
trusted additional arguments should be used.
```

pedro-oliveira-sonarsource (changes requested):
Some suggestions.

pedro-oliveira-sonarsource (approved):
LGTM.

pedro-oliveira-sonarsource:
This branch was polluted by me with commits meant to a different PR thus the commits after the LGTM comment.

PR #1337, opened 2022-10-13
Modify rule S1186(Java): Add exception for @Pointcut annotation (SONARJAVA-3668)

johann-beleites-sonarsource:


rules/S1186/java/rule.adoc
```diff
@@ -33,7 +33,11 @@ public void doSomethingElse() {
 
 == Exceptions
 
-Default (no-argument) constructors are ignored when there are other constructors in the class, as are empty methods in abstract classes.
+This does not raise an issue in the following cases:
+
+* Default (no-argument) constructors when there are other constructors in the class
```
dorian-burihabwa-sonarsource, line 38:
```suggestion
* Default (no-argument) constructors when there are other constructors in the class
* Non-public default constructors
```

dorian-burihabwa-sonarsource (approved):

rules/S1186/java/rule.adoc
dorian-burihabwa-sonarsource: After offline discussion, the RPSEC can be merged as-is

PR #1341, opened 2022-10-14
Modify rule S1694 (CS): correct code comment in sample

martin-strecker-sonarsource:
The comment is misleading as a private constructor is too restrictive.


pavel-mikula-sonarsource (approved):
LGTM

pavel-mikula-sonarsource:
@martin-strecker-sonarsource Something went probably wrong with the SQ part, try pushing empty commit

PR #1344, opened 2022-10-14
Modify rule S2222:  Improve rule description

martin-strecker-sonarsource:
Changes based on the discussion in
https://discuss.sonarsource.com/t/10639

### DIFF
--- a/rules/S2222/csharp/description.adoc
+++ b/rules/S2222/csharp/description.adoc
@@ -1,5 +1,4 @@
-A lock must be released along all execution paths if held or acquired and then released within the same method.
-If a lock is held or acquired and then released within a method, it must be released along all execution paths.
+If a lock is held or acquired and then released within a method, it should be released along all execution paths.
 
 Failing to do so could make an application deadlock-prone because there is a risk of the lock not being released.
 
--- a/rules/S2222/csharp/description.adoc
+++ b/rules/S2222/csharp/description.adoc
@@ -1,5 +1,4 @@
 A lock must be released along all execution paths if held or acquired and then released within the same method.
-
 If a lock is held or acquired and then released within a method, it must be released along all execution paths.
 
 Failing to do so could make an application deadlock-prone because there is a risk of the lock not being released.
--- a/rules/S2222/csharp/description.adoc
+++ b/rules/S2222/csharp/description.adoc
@@ -1,4 +1,4 @@
-If a lock is known to be held or acquired, and then released within a method, then it must be released along all execution paths of that method.
+If a lock is acquired and released within a method, then it must be released along all execution paths of that method.
 
 Failing to do so will expose the conditional locking logic to the method's callers and hence be deadlock-prone.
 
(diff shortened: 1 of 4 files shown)


rules/S2222/csharp/description.adoc
```diff
@@ -1,5 +1,5 @@
-If a lock is known to be held or acquired, and then released within a method, then it must be released along all execution paths of that method.
+If a lock is acquired and released within a method, then it must be released along all execution paths of that method.
```
martin-strecker-sonarsource, line 1: Source is the "main" file for the rule:
https://github.com/SonarSource/rspec/blob/master/rules/S2222/description.adoc

rules/S2222/csharp/description.adoc
```diff
@@ -1,5 +1,5 @@
...
 
-Failing to do so will expose the conditional locking logic to the method's callers and hence be deadlock-prone.
+Failing to do so could make an application deadlock-prone because there is a risk of the lock not being released.
```
martin-strecker-sonarsource, line 3: A mixture of Toms suggestions:
* “Failing to do so could make an application deadlock-prone”
* "If your unlock is conditional there is a risk that it won’t be released."

rules/S2222/csharp/metadata.json
```diff
@@ -1,3 +1,4 @@
 {
+  "title": "Locks should be released on all paths",
```
martin-strecker-sonarsource, line 2: Title change was suggested by Ann (with Tom approving it).
This changes the title for CSharp only. If we do it like this, we need to change it for VB too. If we do it in the shared file, we would change the title for Java too.

pavel-mikula-sonarsource (changes requested):

rules/S2222/csharp/description.adoc
```diff
@@ -1,5 +1,6 @@
-If a lock is known to be held or acquired, and then released within a method, then it must be released along all execution paths of that method.
+A lock must be released along all execution paths if held or acquired and then released within the same method.
+If a lock is held or acquired and then released within a method, it must be released along all execution paths.
```
martin-strecker-sonarsource, line 2: Which version of the sentence do you like best?

rules/S2222/csharp/description.adoc
```diff
@@ -1,5 +1,6 @@
-If a lock is known to be held or acquired, and then released within a method, then it must be released along all execution paths of that method.
+A lock must be released along all execution paths if held or acquired and then released within the same method.
```
martin-strecker-sonarsource, line 1: The tile uses `should be`. Should we use that phrase also?

rules/S2222/csharp/metadata.json
martin-strecker-sonarsource: I wasn't sure how to handle cross-bubble changes so I kept the change small. I put the change in the main file now.

pavel-mikula-sonarsource: There's no strict guideline for those. Let's just do what makes sense, best to our knowledge. It can always be improved later.

rules/S2222/csharp/description.adoc
pavel-mikula-sonarsource: second one. Gives a context and then action needed. Sounds easier to understand to me.

rules/S2222/csharp/description.adoc
pavel-mikula-sonarsource: Sure, why not..

pavel-mikula-sonarsource (approved):
LGTM

PR #1349, opened 2022-10-18
Modify rule S5146(PHP): Update text to Education Framework structure (APPSEC-193)

pedro-oliveira-sonarsource:
Adding internal route examples for Laravel and Symfony.


rules/S5146/common/fix/how-does-this-work.adoc
```diff
@@ -1,7 +1,16 @@
 === How does this work?
 
-In case the application strictly requires redirecting based on user-controllable data, this could be done using the following alternatives:
+Built-in framework methods should be preferred as, more often than not, these
+provide additional security considerations. Usually, these built-in methods are
```
gaetan-ferry-sonarsource, line 4: "Considerations" does not seem appropriate here. Maybe use "mechanisms", "constraints", "safer security options defaults", "checks"

pedro-oliveira-sonarsource: 'mechanisms' was applied.

PR #1353, opened 2022-10-20
Modify S5348: Update description, message and highlighting to match the rule implementation

mary-georgiou-sonarsource:
[SONARPLI-236](https://sonarsource.atlassian.net/browse/SONARPLI-236)

### DIFF
--- a/rules/S5348/pli/rule.adoc
+++ b/rules/S5348/pli/rule.adoc
@@ -1,10 +1,8 @@
-Whenever a CICS command is used with a ``++NOHANDE++`` or ``++RESP++`` option, the default CICS exception handling is disabled. The correct approach then is to ensure that every possible exception is handled correctly directly in the code, and to do this, you need to examine the ``++RESP++`` value or the ``++EIBRESP++`` field value.
+Whenever a CICS command is used with a ``++NOHANDE++`` or ``++RESP++`` option, the default CICS exception handling is disabled. The correct approach then is to ensure that every possible exception is handled correctly directly in the code and to do this, you need to examine the ``++RESP++`` value or the ``++EIBRESP++`` field value.
 
+It is possible to compare the ``++RESP++`` and ``++EIBRESP++`` field values to hard-coded numbers or variables containing numeric values; however, this makes the code difficult to read and maintain. It is recommended to use instead the ``++DFHRESP++`` built-in translator function, which enables the use of the equivalent symbolic values.
 
-It is possible to compare the ``++RESP++`` and ``++EIBRESP++`` field values to hardcoded numbers or variables containing numeric values; however this, makes the code difficult to read and maintain. It is recommended to use instead the ``++DFHRESP++`` built-in translator function, which enables the use of the equivalent symbolic values.
-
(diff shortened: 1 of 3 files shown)


rules/S5348/pli/rule.adoc
```diff
@@ -1,10 +1,12 @@
...
+Whenever a CICS command is used with a ``++NOHANDE++`` or ``++RESP++`` option, the default CICS exception handling is disabled. The correct approach then is to ensure that every possible exception is handled correctly directly in the code, and to do this, you need to examine the ``++RESP++`` value or the ``++EIBRESP++`` field value.
 
 
-It is possible to compare their value to hardcoded numbers corresponding to their binary value, however this makes the code difficult to read and maintain. It is recommended to use instead the DFHRESP built-in translator function which enables the use of equivalent symbolic values.
+It is possible to compare the ``++RESP++`` and ``++EIBRESP++`` field values to hardcoded numbers or variables containing numeric values; however this, makes the code difficult to read and maintain. It is recommended to use instead the ``++DFHRESP++`` built-in translator function, which enables the use of the equivalent symbolic values.
```
andrei-epure-sonarsource, line 4: It is possible to compare the ``++RESP++`` and ``++EIBRESP++`` field values to hardcoded numbers or variables containing numeric values; however, this makes the code difficult to read and maintain. It is recommended to use instead the ``++DFHRESP++`` built-in translator function, which enables the use of the equivalent symbolic values.

andrei-epure-sonarsource, line 4:
```suggestion
It is possible to compare the ``++RESP++`` and ``++EIBRESP++`` field values to hard-coded numbers or variables containing numeric values; however, this makes the code difficult to read and maintain. It is recommended to use instead the ``++DFHRESP++`` built-in translator function, which enables the use of the equivalent symbolic values.
```

rules/S5348/pli/rule.adoc
```diff
@@ -1,10 +1,12 @@
-Whenever a CICS command is used with a ``++NOHANDE++`` or ``++RESP++`` option, the default CICS exception handling is disabled. The correct approach is then to make sure that every possible exception is handled properly directly in the code. This is done by examining the ``++RESP++`` value or the ``++EIBRESP++`` field value.
+Whenever a CICS command is used with a ``++NOHANDE++`` or ``++RESP++`` option, the default CICS exception handling is disabled. The correct approach then is to ensure that every possible exception is handled correctly directly in the code, and to do this, you need to examine the ``++RESP++`` value or the ``++EIBRESP++`` field value.
```
andrei-epure-sonarsource, line 1:
```suggestion
Whenever a CICS command is used with a ``++NOHANDE++`` or ``++RESP++`` option, the default CICS exception handling is disabled. The correct approach then is to ensure that every possible exception is handled correctly directly in the code and to do this, you need to examine the ``++RESP++`` value or the ``++EIBRESP++`` field value.
```

andrei-epure-sonarsource (approved):
LGTM , some nitpicks

PR #1355, opened 2022-10-20
Modify rule S3649(PHP): Update text to the Education Framework structure (APPSEC-190)

loris-s-sonarsource:

PR #1356, opened 2022-10-21
Modify rule S3649(Python): Update text to Education Framework structure (APPSEC-189)

pedro-oliveira-sonarsource:


pedro-oliveira-sonarsource:
secure-by-design.adoc is being done in this [PR](https://github.com/SonarSource/rspec/pull/1355).

pedro-oliveira-sonarsource:
A hardening [ticket](https://sonarsource.atlassian.net/browse/APPSEC-215) to add Django content was created.

rules/S3649/python/how-to-fix-it/sqlite3.adoc
```diff
@@ -0,0 +1,51 @@
...
+The following code is an example of an overly simple data retrieval function.
+It is vulnerable to SQL injection because user-controlled data is inserted
+directly into a query string: The application assumes that incoming data
+always has a specific range of characters and ignores that some characters may
+change the query logic to a malicious one.
+
+In this particular case, the query can be exploited with the following string: 
+
+----
+' OR 1='1
```
gaetan-ferry-sonarsource, line 12: This payload does not work with SQLite that is strongly typed.
```
sqlite> select 1 where 1='1';
sqlite>
```
```suggestion
' OR '1'='1
```

rules/S3649/python/how-to-fix-it/sqlite3.adoc
```diff
@@ -0,0 +1,51 @@
...
+----
+
+Using the UNION clause, an attacker would be able to perform queries against
```
gaetan-ferry-sonarsource, line 15:
```suggestion
Using the UNION clause, an attacker would also be able to perform queries against
```

gaetan-ferry-sonarsource (changes requested):
LGTM in general, with a few suggestions and fixes.

Two other things not in the comments:

1. The rule.adoc does not include the two how to fix files. It raises a testing error.
2. I think there is something missing regarding the prepared statement fix. It is not possible to parameterize the SQL keywords or table names. I think we could add a paragraph in the prepared statement how to fix with this and propose solutions. A pitfall could probably work too.

pedro-oliveira-sonarsource:
>     2. I think there is something missing regarding the prepared statement fix. It is not possible to parameterize the SQL keywords or table names. I think we could add a paragraph in the prepared statement how to fix with this and propose solutions. A pitfall could probably work too.

Think creating a hardening ticket for this would be appropriate.

gaetan-ferry-sonarsource (approved):
LGTM

PR #1359, opened 2022-10-21
Modify rule S2076: Change text to education framework structure (Py,JS,PHP) [APPSEC-186]

gaetan-ferry-sonarsource:


loris-s-sonarsource (approved):
LGTM

PR #1362, opened 2022-10-24
Modify rule S107: Add exceptions to description for Java (SONARJAVA-4072)

leonardo-pilastri-sonarsource:


rules/S107/java/rule.adoc
```diff
@@ -29,6 +29,8 @@ Methods annotated with :
 * Bean constructor injection with ``++@org.springframework.beans.factory.annotation.Autowired++``
 * CDI constructor injection with ``++@javax.inject.Inject++``
 * ``++@com.fasterxml.jackson.annotation.JsonCreator++``
+* Micronaut's annotations (like ``++@io.micronaut.http.annotation.Get++``)
+* Single constructors of classes annotated as Spring components (like ``++@org.springframework.stereotype.Component++`` )
```
chrislain-razafimahefa-sonarsource, line 34: There is an extra space before the closing parenthesis.

chrislain-razafimahefa-sonarsource: I find 'Single constructors....' a bit less clear than 'Classes with a single constructor annotated....'

chrislain-razafimahefa-sonarsource (changes requested):
Just a few suggestions to consider...

rules/S107/java/rule.adoc
leonardo-pilastri-sonarsource: Maybe I should separate this last exception from the list then, because the list is titled "methods annotated with", but for this last exception the annotation is on the class, and I want to be specific that for those classes only the single constructor will be ignored, but not the other class methods.

chrislain-razafimahefa-sonarsource: Right, I forgot about the 'methods annotated with'. 
So, I agree, it is a good idea to create a new section for class annotations.

rules/S107/java/rule.adoc
```diff
@@ -29,9 +29,13 @@ Methods annotated with :
 * Bean constructor injection with ``++@org.springframework.beans.factory.annotation.Autowired++``
 * CDI constructor injection with ``++@javax.inject.Inject++``
 * ``++@com.fasterxml.jackson.annotation.JsonCreator++``
+* Micronaut's annotations (like ``++@io.micronaut.http.annotation.Get++``)
 
 may have a lot of parameters, encapsulation being possible. Such methods are therefore ignored.
 
+Also, if a class annotated as a Spring component (like ``++@org.springframework.stereotype.Component++``) has a single
+constructor, that constructor will be considered ``+@Autowired+`` and ignored by the rule.
```
chrislain-razafimahefa-sonarsource, line 37: Sounds good. Except that you forgot the two '+' around 'Autowired'.

chrislain-razafimahefa-sonarsource (changes requested):
One last change to consider.

PR #1372, opened 2022-11-03
Modify rule S6096(JavaScript): Update text to the Education Framework structure (APPSEC-231)

pedro-oliveira-sonarsource:


rules/S6096/javascript/how-to-fix-it/nodejs.adoc
```diff
@@ -0,0 +1,94 @@
...
+[source,javascript,diff-id=1,diff-type=noncompliant]
+----
+const AdmZip = require("adm-zip");
+
+const targetDirectory = "/example/directory/";
+
+app.get('/example', (req, res) => {
+    const path = targetDirectory + req.query.filename;
+
+    const zip = new AdmZip(path); // Noncompliant
```
gaetan-ferry-sonarsource, line 19: I don't really get what you are trying to do here. This looks a lot like a traditional path traversal vulnerability. Shouldn't this be a read of an uploaded zip file?
```suggestion
    const zip = new AdminZip(req.files.foo.tempFilePath); // Noncompliant
```
Also, I don't get why this specific line is marked as non compliant (except if this is for a normal path traversal).

rules/S6096/javascript/how-to-fix-it/nodejs.adoc
```diff
@@ -0,0 +1,94 @@
...
+    const zipEntries = zip.getEntries();
+
+    zipEntries.forEach(function (zipEntry) {
+        var writer = fs.createWriteStream(zipEntry.entryName);
```
gaetan-ferry-sonarsource, line 23: To me, this should be the non compliant line in this case.
```suggestion
        var writer = fs.createWriteStream(zipEntry.entryName); // Noncompliant
```

rules/S6096/javascript/how-to-fix-it/nodejs.adoc
```diff
@@ -0,0 +1,94 @@
...
+[source,javascript,diff-id=1,diff-type=compliant]
+----
+const AdmZip = require("adm-zip");
+
+const targetDirectory = "/example/directory/";
+
+app.get('/example', (req, res) => {
+    const canonicalPath = path.normalize(targetDirectory + req.query.filename);
+
+    if (canonicalPath.startsWith(targetDirectory)) {
```
gaetan-ferry-sonarsource, line 40: I think this is fixing the old school path traversal issue and not the zip slip one.

rules/S6096/javascript/how-to-fix-it/nodejs.adoc
```diff
@@ -0,0 +1,94 @@
...
+        const zip = new AdmZip(canonicalPath);
+	    const zipEntries = zip.getEntries();
+
+    	zipEntries.forEach(function (zipEntry) {
+            var writer = fs.createWriteStream(zipEntry.entryName);
+            writer.write(zipEntry.getData().toString("utf8"));
+	    });
```
gaetan-ferry-sonarsource, line 47: To me the zip slip issue is still there and exploitable here. Maybe refer to the existing rspec for an appropriate solution. The key is to check the entry name before extraction, not the archive name.

rules/S6096/javascript/how-to-fix-it/nodejs.adoc
```diff
@@ -0,0 +1,94 @@
...
+    }
```
gaetan-ferry-sonarsource, line 81: Same remarks as above. The issue is not in the right location.

gaetan-ferry-sonarsource (changes requested):
I think there is a misunderstanding of the issue to raise here. The code examples focus on a classical path traversal issue while the rule is specifically about zip slip.

pedro-oliveira-sonarsource:
Think it's required to add the following?
const fs = require('fs');
const path = require('path');

gaetan-ferry-sonarsource (approved):
LGTM

PR #1373, opened 2022-11-03
Modify rule S5135: Change text to the education framework format [PHP][APPSEC-230]

gaetan-ferry-sonarsource:


rules/S5135/common/fix/integrity-check.adoc
```diff
@@ -3,8 +3,20 @@
...
+    MAC on the server side before deserialization.
 
-**Note**: Be sure to store the HMAC secret securely.
+Depending on the situation, two MAC computation modes can be used.
+
+If the same application will be responsible for the MAC computing and
+validation, a symmetric signature algorithm can be used. In that case, HMAC
+should be preferred, with a strong underlying hash algorithm such as SHA-256.
+
+If multiple components have to validate the serialized data, an asymetric
```
pedro-oliveira-sonarsource, line 17:
```suggestion
If multiple parties have to validate the serialized data, an asymetric
```

pedro-oliveira-sonarsource: Can be something else, but I think a word that distinguishes between entities and not application components would be better here.

pedro-oliveira-sonarsource (approved):

PR #1375, opened 2022-11-04
Modify rule S3871: Remove Java, add VB.NET

pavel-mikula-sonarsource:
Adding VB.NET for https://github.com/SonarSource/sonar-dotnet/pull/5822

Java was removed, as it is not implemented by java analyzer and inherited C# text (with System.Exception). This is an import mess from Jira.
I confirmed with Java Bubble that it should be removed, see https://sonarsource.slack.com/archives/CFVNB3S59/p1667565788831769


Corniel (approved):
LGTM

rules/S3871/csharp/rule.adoc
```diff
@@ -1,18 +1,29 @@
-include::../rule.adoc[]
+The point of having custom exception types is to convey more information than is available in standard types. But custom exception types must be `public` for that to work.
 
-ifdef::env-github,rspecator-view[]
+If a method throws a non-public exception, the best you can do on the caller's side is to `catch` the closest `public` base of the class. That is, you lose all that custom information you created the exception type to pass.
```
martin-strecker-sonarsource, line 3:
```suggestion
If a method throws a non-public exception, the best you can do on the caller's side is to `catch` the closest `public` base of the class. However, you lose all that custom information the exception type passes.
```

rules/S3871/exceptions.adoc
```diff
@@ -0,0 +1,3 @@
+== Exceptions
+
+This rule ignores Exception types that are not derived directly from `System.Exception+`, `System.SystemException`, or `System.ApplicationException`.
```
martin-strecker-sonarsource, line 3:
```suggestion
This rule ignores Exception types that are not derived directly from `System.Exception`, `System.SystemException`, or `System.ApplicationException`.
```

rules/S3871/vbnet/rule.adoc
```diff
@@ -0,0 +1,32 @@
+The point of having custom exception types is to convey more information than is available in standard types. But custom exception types must be `Public` for that to work.
+
+If a method throws a non-public exception, the best you can do on the caller's side is to `Catch` the closest `Public` base of the class. That is, you lose all that custom information you created the exception type to pass.
```
martin-strecker-sonarsource, line 3: The changes in the C# version should be applied here too.

martin-strecker-sonarsource (changes requested):
There is an `+` leftover in one of the exceptions exceptions.

rules/S3871/csharp/rule.adoc
pavel-mikula-sonarsource: The whole sentence is a bit weird. I tried to rephrase it

martin-strecker-sonarsource (approved):
LGTM

PR #1378, opened 2022-11-04
Modify rule S6258: Add new Terraform code sample

hendrik-buchwald-sonarsource:
Implementation ticket: https://sonarsource.atlassian.net/browse/SONARIAC-418

The new implementation does not raise an issue for AWS provider >= 4, so the issues reported for this rule will slowly dry out. Since there are already a lot of code samples for this rule, I decided to combine the compliant solution for both versions.


loris-s-sonarsource (approved):
LGTM

PR #1379, opened 2022-11-04
Modify rule S6255: Add new Terraform code sample

hendrik-buchwald-sonarsource:
Implementation ticket: https://sonarsource.atlassian.net/browse/SONARIAC-416

### DIFF
--- a/rules/S6255/terraform/rule.adoc
+++ b/rules/S6255/terraform/rule.adoc
@@ -20,7 +20,7 @@ It's recommended to enable S3 MFA delete, note that:
 
 == Sensitive Code Example
 
-A versioned S3 bucket doesn't have MFA delete enabled:
+A versioned S3 bucket does not have MFA delete enabled for AWS provider version 3 or below:
 
 [source,terraform]
 ----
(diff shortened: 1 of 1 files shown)


rules/S6255/terraform/rule.adoc
```diff
@@ -50,6 +65,24 @@ resource "aws_s3_bucket" "example" { # Compliant
...
+  bucket = "example"
+}
+
+resource "aws_s3_bucket_versioning" "example" {
+  bucket = aws_s3_bucket.example.id
+  versioning_configuration {
+    status = "Enabled"
+    mfa_delete = "Enabled"
+  }
+  mfa = "${var.MFA}"
```
loris-s-sonarsource, line 82: minor - we should replace the variable with "true" here for clarity

loris-s-sonarsource (approved):
LGTM -- 1 minor suggestion

rules/S6255/terraform/rule.adoc
hendrik-buchwald-sonarsource: I am not sure I understand what you mean, could you clarify?

loris-s-sonarsource: I didn't understand what the mfa variable was about, so I assumed it was a boolean. But no, it's the actual MFA code, right ?

hendrik-buchwald-sonarsource: Ah, yea, it is the actual MFA code. I could specify a fake one but I did not want to show a hard-coded MFA token as an example.

loris-s-sonarsource: Ok, got it. My train of thought was that using a variable could be ambiguous, but it looks like it doesn't matter here. We can dismiss my comment

PR #1380, opened 2022-11-04
Modify rule S6245: Add new Terraform code sample

hendrik-buchwald-sonarsource:
Implementation ticket: https://sonarsource.atlassian.net/browse/SONARIAC-417

The new implementation does not raise an issue for AWS provider >= 4, so the issues reported for this rule will slowly dry out. At some point, we can deprecate and remove this rule unless the detection logic gets improved.

### DIFF
--- a/rules/S6245/terraform/rule.adoc
+++ b/rules/S6245/terraform/rule.adoc
@@ -8,26 +8,46 @@ include::../recommended.adoc[]
 
 Server-side encryption is not used:
 
+[source,terraform]
 ----
-resource "aws_s3_bucket" "mynoncompliantbucket" {
-  bucket = "mynoncompliantbucket"
+resource "aws_s3_bucket" "example" {
+  bucket = "example"
 }
 ----
 
 == Compliant Solution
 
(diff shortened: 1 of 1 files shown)


loris-s-sonarsource (approved):
LGTM - I just feel weird with the wording "Server-side encryption is not used:". I think we should prefix it with "In case", "with", or "If".

See how it looks like: https://next.sonarqube.com/sonarqube/coding_rules?q=S6245&open=terraform%3AS6245

rules/S6245/terraform/rule.adoc
```diff
@@ -8,31 +8,51 @@ include::../recommended.adoc[]
 
 Server-side encryption is not used:
 
+[source,terraform]
 ----
-resource "aws_s3_bucket" "mynoncompliantbucket" {
-  bucket = "mynoncompliantbucket"
+resource "aws_s3_bucket" "example" {
```
loris-s-sonarsource, line 13: Just noticed there's no "sensitive" comment

hendrik-buchwald-sonarsource: Good catch!

PR #1384, opened 2022-11-08
Modify rule S5147[JS/TS]: Change text to the education framework format (APPSEC-233)

pierre-loup-tristant-sonarsource:
Ticket: [APPSEC-233](https://sonarsource.atlassian.net/browse/APPSEC-233)


rules/S5147/common/fix/code-rationale.adoc
```diff
@@ -0,0 +1 @@
+The following code is vulnerable to a NoSQL injection because the database query is constructed using untrusted JavaScript objects that are serialized from user inputs.
```
gaetan-ferry-sonarsource, line 1: I would probably use build in place of construct, which seems clearer to me.
I don't think *serialized* is the right word to use here. Unserialize seems more relevant. Maybe something more generic like "extracted" "obtained" or "retrieved".

I would also add a small explanation about the input typing, that induces the issue.

```suggestion
The following code is vulnerable to a NoSQL injection because the database query is built using untrusted JavaScript objects that are extracted from user inputs.

Here the application assumes the user-submitted parameters are always strings, while they might contain more complex structures. An array or dictionary input might tamper with the expected query behavior.
```

rules/S5147/javascript/how-to-fix-it/mongodb.adoc
```diff
@@ -0,0 +1,45 @@
...
+==== Noncompliant code example
+
+[source,javascript,diff-id=1,diff-type=noncompliant]
+----
+const { MongoClient } = require('mongodb');
+
+function (req, res) {
+    let query = { user: req.query.user, city: req.query.city };
+
+    MongoClient.connect(url, (err, db) => {
```
gaetan-ferry-sonarsource, line 14: This function signature is deprecated. The Promises version of the function should probably be used instead.
It actually raises a code smell in SonarQube and SonarLint.

rules/S5147/javascript/how-to-fix-it/mongodb.adoc
```diff
@@ -0,0 +1,45 @@
...
+==== Compliant solution
+
+[source,javascript,diff-id=1,diff-type=compliant]
+----
+const { MongoClient } = require('mongodb');
+
+function (req, res) {
+    let query = { user: req.query.user.toString(), city: req.query.city.toString() };
+
+    MongoClient.connect(url, (err, db) => {
```
gaetan-ferry-sonarsource, line 31: Same thing here. The callback form of the function is deprecated. Promises should be used instead.

rules/S5147/common/fix/plain-string-values.adoc
```diff
@@ -0,0 +1,10 @@
+==== Use only plain string values 
+
+With MongoDB, NoSQL injection can arise when attackers are able to inject objects in the query instead of plain string values. For example, using the object `{ $ne: "" }` in a field of a `find` query, will return every entries where the field is not empty.
+
+Some JavaScript applicative web servers enable "extended" syntax that serializes url query parameters into JavaScript objects or arrays. This is allow attacker to control all the fileds of an object.
```
gaetan-ferry-sonarsource, line 5:
```suggestion
Some JavaScript application servers enable "extended" syntax that serializes URL query parameters into JavaScript objects or arrays. This allows attackers to control all the fields of an object.
```

rules/S5147/common/fix/plain-string-values.adoc
```diff
@@ -0,0 +1,10 @@
+==== Use only plain string values 
+
+With MongoDB, NoSQL injection can arise when attackers are able to inject objects in the query instead of plain string values. For example, using the object `{ $ne: "" }` in a field of a `find` query, will return every entries where the field is not empty.
```
gaetan-ferry-sonarsource, line 3:
```suggestion
With MongoDB, NoSQL injection can arise when attackers are able to inject objects in the query instead of plain string values. For example, using the object `{ $ne: "" }` in a field of a `find` query, will return every entry where the field is not empty.
```

rules/S5147/common/fix/plain-string-values.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+Some JavaScript applicative web servers enable "extended" syntax that serializes url query parameters into JavaScript objects or arrays. This is allow attacker to control all the fileds of an object.
+In express.js, this "extended" syntax https://expressjs.com/en/4x/api.html#express.urlencoded[is enabled by default].
+
+Before using any untrusted value in a MongoDB query, make sure it is a plain string and not JavaScript objects or arrays.
```
gaetan-ferry-sonarsource, line 8:
```suggestion
Before using any untrusted value in a MongoDB query, make sure it is a plain string and not a JavaScript object or an array.
```

rules/S5147/common/pitfalls/code-execution.adoc
```diff
@@ -0,0 +1,7 @@
+==== Code execution
+
+When untrusted data are used within query operators such as `$where`, `$accumulator` or `$function` it usually results in JavaScript code execution vulnerabilities.
```
gaetan-ferry-sonarsource, line 3:
```suggestion
When untrusted data is used within query operators such as `$where`, `$accumulator`, or `$function` it usually results in JavaScript code execution vulnerabilities.
```

rules/S5147/common/pitfalls/code-execution.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+Therefore, untrusted values should not be used inside these query operators unless they are validated against a whitelist.
```
gaetan-ferry-sonarsource, line 5: Maybe I would not give any fix information there. Will prevent keeping the two rules in sync in the future if we need to change the main reco.

```suggestion
Therefore, untrusted values should not be used inside these query operators unless they are properly validated.
```

gaetan-ferry-sonarsource (changes requested):
LGTM in general. Some typo and phrasing suggestions. There is also a code example that uses a deprecated function signature which leads to a code smell in SonarQube and SonarLint. I think it would be nice to use a supported signature.

rules/S5147/javascript/how-to-fix-it/mongodb.adoc
pierre-loup-tristant-sonarsource: The thing is, we don't support the promised version and it leads to a FN.
See https://sonarsource.atlassian.net/browse/SONARSEC-3420

rules/S5147/common/fix/code-rationale.adoc
pierre-loup-tristant-sonarsource: You are right, serialized is not the right word in this context.

PR #1383, opened 2022-11-07
Modify rule S5334: Changing text to the education framework format[PHP][APPSEC-236]

gaetan-ferry-sonarsource:

PR #1387, opened 2022-11-08
Modify rule S5334: Change text to the education framework format [Javascript][APPSEC-235]

gaetan-ferry-sonarsource:


rules/S5334/javascript/how-to-fix-it/node.adoc
```diff
@@ -0,0 +1,35 @@
...
+The following code is vulnerable to arbitrary code execution because it dynamically
+runs JavaScript code built from untrusted data.
+
+==== Noncompliant code example
+
+[source,javascript,diff-id=1,diff-type=noncompliant]
+----
+let operation = req.query.operation
+eval(`product_${operation}()`) // Noncompliant
+res.send("OK")
```
pierre-loup-tristant-sonarsource, line 12: For me, it's obvious here that `req` is an [express.js Request](https://expressjs.com/en/api.html#req) but I wonder if it's the case for others.
Some other injection rules like S2083 or S5147 for JS/TS embed code examples within a function receiving `req` and `res`. The reason for that is so that developer can easily identify the function as an express.js middleware.
Let me know what you think.
 
```suggestion
function (req, res) {
    let operation = req.query.operation
    eval(`product_${operation}()`) // Noncompliant
    res.send("OK")
}
```

rules/S5334/javascript/how-to-fix-it/mongo.adoc
```diff
@@ -0,0 +1,72 @@
...
+perform advanced queries. In most cases, the same result can be achieved by
+combining classical, safer, operators, or by performing additional data
+filtering on the application side.
+
+Note that MongoDB servers implement a sandboxed JavaScript engine to run the
+operator-related code. Without other vulnerabilities in the database engine,
+the execution of arbitrary code is thus not possible through the exploitation
+of evaluation operators. However, a successful injection could still lead to
+the compromise of the MongoDB server hosted data. Moreover, the discovery by
+an attacker of a sandbox escaping attack payload should also be considered.
```
pierre-loup-tristant-sonarsource, line 54: Not sure all users are familiar with the notion of payload. What about, 
```suggestion
an attacker of a way to escape the sandbox should also be considered.
```

pierre-loup-tristant-sonarsource (approved):
LGTM. Have a look at the two suggestion I made before merging.

pierre-loup-tristant-sonarsource:
@gaetan-ferry-sonarsource, since https://github.com/SonarSource/rspec/pull/1383 was merged, please update this PR (merge or rebase) and check that the validate_asciidoc check passes correctly before merging to [feature/education](https://github.com/SonarSource/rspec/tree/feature/education).

rules/S5334/javascript/how-to-fix-it/node.adoc
gaetan-ferry-sonarsource: As this is not really express related, I did not wanted to include any express routing snippet. But this seems like a good trade-off. Thanks.

PR #1391, opened 2022-11-14
Modify rule S2260: Add Docker as new language (RULEAPI-773)

nils-werner-sonarsource:


loris-s-sonarsource (approved):
LGTM. 

Small concern about the use of wording related to "Docker", because this is not the only existing container runtime, but it's very likely to be the most widespread so I would keep this as is.

rules/S2260/docker/rule.adoc
```diff
@@ -0,0 +1,16 @@
+When the Dockerfile parser fails, it is possible to record the failure as a violation on the file. This way, not only it is possible to track the number of files that do not parse but also to easily find out why they do not parse.
```
arseniy-sonar, line 1:
```suggestion
When the Dockerfile parser fails, it is possible to record the failure as a violation on the file. This way, not only is it possible to track the number of files that do not parse but also to easily find out why they do not parse.
```

arseniy-sonar (approved):

PR #1393, opened 2022-11-14
Modify rule S2078: Change text to the education framework format [Python][APPSEC-269]

pedro-oliveira-sonarsource:


pedro-oliveira-sonarsource:
Please suggest an alternative for the allowed_framework file, do you think "Python LDAP" would be ok? my issue is that it's very generic and not library-specific as you have other ones.

egon-okerman-sonarsource:
> Please suggest an alternative for the allowed_framework file, do you think "Python LDAP" would be ok? my issue is that it's very generic and not library-specific as you have other ones.

IMO the framework is called "python-ldap" on PyPI so the most correct thing would be to leave it like that. Calling it "Python LDAP" is indeed too confusing, and even things like "PyLDAP" are just a different framework.
Looking at the frameworks we currently allow, it's e.g. not immediately clear to me which library "Python SQLite" is supposed to be (looking it up, it's sqlite3.) Maybe best to involve someone who's spent more time on the education topic to decide this.

rules/S2078/python/metadata.json
```diff
@@ -0,0 +1,28 @@
+{
```
egon-okerman-sonarsource, line 1: Can this CERT standard be moved to `S2078/metadata.json`? I don't see any other differences between this and the file above

rules/S2078/python/how-to-fix-it/python-ldap.adoc
```diff
@@ -0,0 +1,49 @@
...
+    user = ldap_connection.search_s("dc=example,dc=org", ldap.SCOPE_SUBTREE, search_filter)
+
+    return user[0]
+----
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+In the compliant solution example, the `ldap.filter.escape_filter_chars`
```
egon-okerman-sonarsource, line 48: Validation talks about escaping characters for DNs, which is different from filters. I would also put `ldap.filter.escape_dn_chars` in the example and the documentation here. I think it could be a nasty pitfall for developers otherwise.

egon-okerman-sonarsource (changes requested):

rules/S2078/python/how-to-fix-it/python-ldap.adoc
pedro-oliveira-sonarsource: Agree.

rules/S2078/python/metadata.json
pedro-oliveira-sonarsource: This should be empty (copy paste issues), nice catch :).

pedro-oliveira-sonarsource:
What do you think about the library name? Should we go for `Python-ldap` over `python-ldap`?

rules/S2078/common/fix/validation.adoc
```diff
@@ -1,26 +1,22 @@
...
-if it is available for your use case. An alternative is to
-validate the input before using it in an LDAP query.
-
-Input validation should be primordially done using native library sanitizers.
-Alternatively, it can be implemented using an allow-list approach by creating a list of
-authorized and secure strings you want the application to
-use in a query. +
-If a user input does not match an entry in this list, it should be rejected
+Another approach is to sanitize the input before using it in an LDAP query.
+Input sanitization should be primordially done using native libraries.
```
egon-okerman-sonarsource, line 8: I don't think "primordially" is the best word, could be "primarily" instead

rules/S2078/common/fix/validation.adoc
```diff
@@ -1,26 +1,22 @@
...
+
+Alternatively, validation can be implemented using an allow-list approach by
```
egon-okerman-sonarsource, line 10: Nitpick: "allow-list" should be ["allowlist"](https://en.wiktionary.org/wiki/allowlist). Unfortunately it is spelled as "allow-list" everywhere in the rspec repo.

rules/S2078/common/fix/validation.adoc
```diff
@@ -1,26 +1,22 @@
...
 because it is considered unsafe.
 
-*Important note*: The application must validate on the server side. Not on
-client-side front-ends.
+*Important note*: The application must sanitize and validate on the
+server-side. Not on client-side front-ends.
 
-As a last resort, the most fundamental security mechanism is the restriction of
-LDAP metacharacters.
+The most fundamental security mechanism is the restriction of LDAP
```
egon-okerman-sonarsource, line 18: I think removing "as a last resort" changes the importance of this recommendation. I don't disagree with it necessarily, but would like to understand why it was removed.

egon-okerman-sonarsource:
> What do you think about the library name? Should we go for `Python-ldap` over `python-ldap`?

IMO python-ldap is better. If someone really disagrees with it, it can always be changed later.

egon-okerman-sonarsource (commented):
Just some small fixes are necessary in the common validation text, otherwise LGTM!

rules/S2078/common/fix/validation.adoc
pedro-oliveira-sonarsource: In my opinion here, when we say "the most fundamental security mechanism", this is not an alternative suggestion, it is stating the core fix of the problem. And 'as a last resort' to me sounds like we provided all the suggestions.

egon-okerman-sonarsource: That sounds good to me

PR #1396, opened 2022-11-15
[APPSEC-271] Modify rule S2091: Change text to the education framework format [Python]

egon-okerman-sonarsource:


rules/S2091/python/how-to-fix-it/python.adoc
```diff
@@ -0,0 +1,42 @@
...
+==== Compliant solution
+
+[source,python,diff-id=1,diff-type=compliant]
+----
+from lxml.etree import ElementTree
+
+def authenticate(tree: ElementTree, username: str, password: str) -> bool:
+    expression = "./users/user[@name='$username' and @pass='$password']"
+
+    return tree.xpath(expression, username=username, password=password) is not None
```
pierre-loup-tristant-sonarsource, line 34: As we discussed maybe we should only mention switching to `lxml` in a later section. Therefore switch to a compliant code example that demonstrates the regex validation approach.

rules/S2091/python/how-to-fix-it/python.adoc
```diff
@@ -0,0 +1,42 @@
...
+    expression = "./users/user[@name='" + username + "'][@pass='" + password + "']"
+
+    return tree.find(expression) is not None
+----
+
+////
+The above example would be exploitable by using e.g. `username=”admin][.!=”` and `password=”][.!='XXXXX”`
+This creates the following query: `./users/user[@name='admin][.!='][@pass='][.!='XXXXX']`,
+where `[.!='][@pass=']` and `[.!='XXXXX']` should match on everything, resulting in authentication as admin.
+////
```
pierre-loup-tristant-sonarsource, line 23: Not sure it's worth showing the payload.

rules/S2091/python/how-to-fix-it/python.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+    return tree.xpath(expression, username=username, password=password) is not None
+----
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+The Python standard library does not offer a method to create parameterized queries. In the example, the compliant solution instead uses
+the `ElementTree.xpath()` method defined by the https://pypi.org/project/lxml/[lxml library] to create a parameterized query, which prevents injection.
```
pierre-loup-tristant-sonarsource, line 42: Same comment here.

rules/S2091/python/how-to-fix-it/lxml.adoc
```diff
@@ -0,0 +1,34 @@
...
+    expression = "./users/user[@name=$username and @pass=$password]"
+
+    return tree.xpath(expression, username=username, password=password) is not None
+----
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+In the example, the username and password are passed as https://lxml.de/xpathxslt.html#:~:text=The%20xpath()%20method%20has%20support%20for%20XPath%20variables%3A[XPath variables] rather than concatenated to the XPath query. By using a parameterized query, injection is successfully prevented.
```
pierre-loup-tristant-sonarsource, line 34: Should be reworked to mention `parameterized queries` in a separate section.

pierre-loup-tristant-sonarsource (changes requested):
As discussed, the `common/fix/validation.adoc` should be reworked to mention `parameterized queries` in a separate section that can also be reused for Java.

It's not in the scope of the ticket but we probably have to adjust what is included in the C# rule description, so that it doesn't mention OWASP ESAPI. 

Regarding parameterize query for C# either we don't mention it or we suggest using[XPathExpression](https://learn.microsoft.com/en-us/dotnet/standard/data/xml/user-defined-functions-and-variables?redirectedfrom=MSDN). I would rather go for not mentioning since the API documentation is a bit useless without a code example.

rules/S2091/python/how-to-fix-it/python.adoc
egon-okerman-sonarsource: I can take this out, but it's in a comment, so it's really just for future reference for us

rules/S2091/python/how-to-fix-it/python.adoc
```diff
@@ -26,17 +26,28 @@ where `[.!='][@pass=']` and `[.!='XXXXX']` should match on everything, resulting
 
 [source,python,diff-id=1,diff-type=compliant]
 ----
-from lxml.etree import ElementTree
+from xml.etree import ElementTree
+import re
 
 def authenticate(tree: ElementTree, username: str, password: str) -> bool:
```
pierre-loup-tristant-sonarsource, line 32: Currently no issue is raise on this code example because:
- `username` and `password` are only function parameters and therefore are not detected as sources
-  Apparently, `tree.find` is not resolved with the correct method id because type inference has limitations (see [SONARSEC-2590](https://sonarsource.atlassian.net/browse/SONARSEC-2590)). Therefore it's not resolved as a sink.

Here I suggest: 
- Updating the function to make it a Flask entry point like we do in other rule
- Create the `tree` variable using `tree = ElementTree.parse` so that the methodId of the sink is correctly resolved.

```suggestion
import re
from flask import request
from xml.etree import ElementTree

tree = ElementTree.parse('ressources/users.xml')

@app.route('/authenticate')
def authenticate():
    
    username = request.args['username']
    password = request.args['password']
```

You probably also need to adjust what the method returns.

pierre-loup-tristant-sonarsource (changes requested):
Please rework the code examples, the syntax you used for the regular expression is not supported in Python 3.8 (at least). I also think noncompliant code examples should raise the expected issue and it's currently not the case.

The change on the common part looks good. Nice job, I think it's way clearer now.

rules/S2091/python/how-to-fix-it/python.adoc
pierre-loup-tristant-sonarsource: Ok, works for me.

rules/S2091/python/how-to-fix-it/python.adoc
```diff
@@ -26,17 +26,28 @@ where `[.!='][@pass=']` and `[.!='XXXXX']` should match on everything, resulting
...
-    expression = "./users/user[@name='$username' and @pass='$password']"
+    if re.match(/^[a-zA-Z0-9]*$/, username) is None or re.match(/^[a-zA-Z0-9]*$/, username) is None:
```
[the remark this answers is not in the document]
egon-okerman-sonarsource: You're right! I accidentally copied this over from a regex tester, my bad.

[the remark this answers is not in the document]
pierre-loup-tristant-sonarsource: I thought it was a fancy new syntax I didn't know.

rules/S2091/common/fix/validation.adoc
```diff
@@ -1,40 +1,15 @@
...
 
-As a rule of thumb, the best approach to protect against injections is to
-systematically ensure that untrusted data cannot break out of the initially
-intended logic.
+In case XPath parameterized queries are not available, the most secure way to protect against injections is to validate the input before using it in an XPath query.
 
-For XPath injections, the cleanest way to do so is to use parameterized queries
-if it is available for your use case. Else, the most secure way to do so is to
-validate the input before using it in an XPath query.
+IMPORTANT: The application must do this validation server-side. Validating this client-side is insecure.
```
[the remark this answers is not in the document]
egon-okerman-sonarsource: I spent some time looking it up yesterday, asciidoctorj does support admonitions. Still, I'll revert it and maybe it's something we can investigate for all rules instead.

rules/S2091/python/how-to-fix-it/python.adoc
egon-okerman-sonarsource: Alright, this relates to what we discussed on Wednesday about the simplicity of code samples. I'll change both this and the lxml example.

pierre-loup-tristant-sonarsource (changes requested):

egon-okerman-sonarsource:
Oh my bad, it seems my changes weren't pushed before. Everything should be fixed.

pierre-loup-tristant-sonarsource (approved):
LGTM

PR #1401, opened 2022-11-17
Modify rule S5144[Python]: Change text to the education framework format (APPSEC-285)

pierre-loup-tristant-sonarsource:
Ticket: [APPSEC-285](https://sonarsource.atlassian.net/browse/APPSEC-285)


pierre-loup-tristant-sonarsource:
I have the feeling the "How does this work?" could be improved. For example, it suggests restricting domains to a white list but what about webhook which should allow arbitrary URL as long as they are external?

pedro-oliveira-sonarsource (approved):
Regarding the code-rationale.adoc content 'The following code is vulnerable to SSRF as it performs an HTTP request to a URL defined by untrusted data' , when you use a payload such as `file:///etc/passwd`, does it still trigger an HTTP request?

pierre-loup-tristant-sonarsource:
@pedro-oliveira-sonarsource, should we mention the risk of local file being accessed using the `file://` scheme? `urllib.request` allows this scheme by default.

Also, I am asking again if you think "How does this work?" should be improved. For example, it suggests restricting domains to a white list but what about webhook which should allow arbitrary URL as long as they are external?

pierre-loup-tristant-sonarsource:
@pedro-oliveira-sonarsource following the discussion we had yesterday I update Python `urllib` and .Net code example so that the URL scheme is validated. I also update the common "How does this work?" to mention that both domains and schemes should be validated. I guess it's more consistent now. Let me know what you think.

pedro-oliveira-sonarsource (changes requested):
Just some suggestions.

- Do not forget to also remove the space after `ExampleContoller` in the NonCompliant code.
> `public class ExampleController : Controller`

- The sink used in dotnet.adoc is [obsolete](https://learn.microsoft.com/en-us/dotnet/core/compatibility/networking/6.0/webrequest-deprecated), we should replace it with HttpClient.

pierre-loup-tristant-sonarsource:
@pedro-oliveira-sonarsource following the discussion we add this morning as the standup I created [APPSEC-302](https://sonarsource.atlassian.net/browse/APPSEC-302) and suggest sticking with the current content that only mention validated the hostname and scheme.

Regarding the C# code example, I created [APPSEC-301](https://sonarsource.atlassian.net/browse/APPSEC-301)

rules/S5144/csharp/how-to-fix-it/dotnet.adoc
```diff
@@ -28,16 +28,17 @@ public class ExampleController : Controller
...
+    private readonly string[] allowedSchemes = { "https" };
+    private readonly string[] allowedDomains = { "trusted1.example.com", "trusted2.example.com" };
 
     [HttpGet]
     public IActionResult ImageFetch(string location)
     {
         Uri uri = new Uri(location);
 
-        if (!allowedUrls.Contains(uri.Host))
+        if (!allowedUrls.Contains(uri.Host) && !allowedSchemes.Contains(uri.Scheme))
```
pedro-oliveira-sonarsource, line 41:
```suggestion
        if (!allowedDomains.Contains(uri.Host) && !allowedSchemes.Contains(uri.Scheme))
```

pedro-oliveira-sonarsource (approved):
LGTM. Just 2 adjustments to do.

rules/S5144/csharp/how-to-fix-it/dotnet.adoc
pierre-loup-tristant-sonarsource: @pedro-oliveira-sonarsource I only see this one. what's the second?

PR #1403, opened 2022-11-18
[APPSEC-272] Modify rule S2091 (PHP): Change text to the education framework format

egon-okerman-sonarsource:


pedro-oliveira-sonarsource (approved):

PR #1406, opened 2022-11-21
Modify rule S2631: Change text to the education framework format [Python][APPSEC-288]

pedro-oliveira-sonarsource:


rules/S2631/common/fix/validation.adoc
```diff
@@ -1,16 +1,17 @@
-==== Sanitization and validation
+==== Sanitization and Validation
 
-The easiest way to protect against regex injections is to escape metacharacters,
- preferably with native functions. +
-The escape function instructs the regular expression engine to interpret these
-characters literally and not as metacharacters.
+Metacharacters escape using native functions is a solution against regex
+injection. + The escape function sanitizes the input so that the regular expression
```
egon-okerman-sonarsource, line 4: The `+`'es in this document don't act how they're supposed to (they can be seen in the rendered document.) I think that to function correctly as a line break, you have to put the next sentence on a new line.

rules/S2631/python/rule.adoc
```diff
@@ -0,0 +1,31 @@
+== Why is this an issue?
+
+include::../rationale.adoc[]
+
+include::../impact.adoc[]
```
egon-okerman-sonarsource, line 5: This wasn't changed in this PR, but rationale and impact have some terms (e.g. backtracking and "catastrophic backtracking situations") surrounded in code tags. I think [quotation marks should be used instead](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc#literals).

rules/S2631/python/how-to-fix-it/python.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+[source,python,diff-id=1,diff-type=compliant]
+----
+from flask import request
+import re
+
+@app.route('/lookup')
+def lookup():
+  regex = request.args.get('regex')
+  data = request.get('data')
```
egon-okerman-sonarsource, line 32: Not sure if this is intentional, but this line changed compared to the noncompliant code sample.

egon-okerman-sonarsource (commented):
Some small fixes are necessary, after that should be good to go.

rules/S2631/python/rule.adoc
pedro-oliveira-sonarsource: I have removed them as I am not aware of the quotation mark alternative, can you apply the change to one of them to have a look at?

egon-okerman-sonarsource (approved):
LGTM!

PR #1407, opened 2022-11-21
Modify rule S1116: Change RSPEC to match current implementation (CPP-3875)

alejandro-alvarez-sonarsource:


rules/S1116/cfamily/rule.adoc
```diff
@@ -7,6 +7,12 @@ include::../description.adoc[]
 void doSomething() {
   ;                                                       // Noncompliant - was used as a kind of TODO marker
```
marco-antognini-sonarsource, line 8: Can you remove some of the horizontal whitespaces? One has to scroll to read it on https://rules.sonarsource.com/cpp/RSPEC-1116 or in the IDE.

marco-antognini-sonarsource (approved):
We have a different guideline for the PR's title in this repo. See https://github.com/SonarSource/rspec#to-modify-an-existing-rule. I've updated it. (FWIW, there is no clear convention where to put the Jira ticket id, and you can keep the PR number in the commit title when merging if you want.)

LGTM, one tiny nit. No need for another round of review.

PR #1408, opened 2022-11-22
[APPSEC-289] Modify rule S2631 (PHP): Change text to the education framework format

pedro-oliveira-sonarsource:


rules/S2631/php/how-to-fix-it/core.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+The following noncompliant code is vulnerable to Regex Denial of Service (ReDoS)
+because untrusted data is used as a regex to scan a string without prior
+sanitization or validation.
+
+==== Noncompliant code example
+
+[source,php,diff-id=1,diff-type=noncompliant]
+----
+function lookup(string $data): bool {
```
egon-okerman-sonarsource, line 11: I would remove `$data` from args and use `$data = $_GET["data"]` instead. This way, we raise an issue when scanning this code sample in SQ (currently I don't believe we would.)

pedro-oliveira-sonarsource: Followed this approach you used here https://github.com/SonarSource/rspec/pull/1403/files .

My first code sample was without the function and both coming from $GET, but the rule itself only cares about whether the regex definition is tainted or not that is why I changed it. But I do not mind going back to this:

```
$regex = $_GET["regex"];
$data = $_GET["data"];

preg_grep($regex, $data); // Noncompliant

$regex = $_GET["regex"];
$data = $_GET["data"];

preg_grep(preg_quote($regex), $data);
```

egon-okerman-sonarsource: You're right, `$data` does not matter here, I was just dumb 😅. I'll approve this PR since there are no issues otherwise.

egon-okerman-sonarsource (approved):
LGTM

rules/S2631/php/how-to-fix-it/core.adoc
egon-okerman-sonarsource: And FYI, I think I'll have to revisit #1403 , since using a function there will actually prevent SQ from raising issues.

PR #1414, opened 2022-11-23
Create rule S2116(Java): "hashCode" and "toString" should not be called on array instances

alban-auzeill:


johann-beleites-sonarsource (approved):
LGTM!

PR #1416, opened 2022-11-23
Modify rule S5945: Add constexpr exception and fix typos (CPP-3873)

marco-antognini-sonarsource:
Related to CPP-3873 & SonarSource/sonar-cpp#2263.

I started fixing typos and went on to add the exception for `constexpr`. @alejandro-alvarez-sonarsource, This should be merged once your PR is merged.

### DIFF
--- a/rules/S5945/cfamily/rule.adoc
+++ b/rules/S5945/cfamily/rule.adoc
@@ -46,7 +46,6 @@ This rule will not report these uses of C-style arrays:
 
  * in `extern "C"` code, since those arrays are often required here for compatibility with external code;
  * in `constexpr` contexts, since using `std::string` is not straightforward at all because the string cannot escape the `constexpr` scope;
- * in the arguments of `main`, since their types are mandated by the standard;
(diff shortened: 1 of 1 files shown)


rules/S5945/cfamily/rule.adoc
```diff
@@ -31,16 +31,21 @@ void f() {
...
 
 
 == Exceptions
 
-This rule will not report the use of C-style arrays in ``++extern "C"++`` code (since those arrays are often required here for compatibility with external code) and in the arguments of ``++main++``.
+This rule will not report these uses of C-style arrays:
+
+ * in `extern "C"` code, since those arrays are often required here for compatibility with external code;
+ * in `constexpr` contexts, since using `std::string` is not straightforward at all because the string cannot escape the `constexpr` scope;
+ * in the arguments of `main`, since their types are mandated by the standard;
```
abbas-sabra-sonarsource, line 47: I'm okay with `constexpr` and const array as they are a good learning point for the user and they might wonder why the issue isn't raised.
For the `main`, I think it is pure implementation details, I don't see the value of bothering the user with such information.

marco-antognini-sonarsource: Fair point; removed with 3abf9e2.

PR #1421, opened 2022-11-24
[APPSEC-293] Modify rule S6287 (JS/TS): Change text to the education framework format 

egon-okerman-sonarsource:


rules/S6287/javascript/how-to-fix-it/express.adoc
```diff
@@ -0,0 +1,44 @@
...
+import cookieParser from "cookie-parser";
+
+const app = express();
+app.use(cookieParser());
+
+app.get("/checkcookie", (req, res) => {
+    if (req.cookies["connect.sid"] === undefined) {
+        const cookie = req.query.cookie;
+        res.cookie("connect.sid", cookie); // Noncompliant
+    }
```
pedro-oliveira-sonarsource, line 19: Any specific reason for the lack of else on the non-compliant, or the reverse? If not either do the else here as well, or remove the else on the compliant code example for consistency and better diff results.

pedro-oliveira-sonarsource (approved):
LGTM. Just a minor suggestion.

rules/S6287/javascript/how-to-fix-it/express.adoc
egon-okerman-sonarsource: Didn't pay attention that it can be removed in the compliant example, it's fixed.

PR #1423, opened 2022-11-24
[APPSEC-287] Modify rule S5144 (JS/TS): Change text to the education framework format

pedro-oliveira-sonarsource:


rules/S5144/javascript/how-to-fix-it/node.adoc
```diff
@@ -0,0 +1,41 @@
...
+----
+const axios = require('axios');
+
+async function example(request) {
+    const schemesList = ["http:", "https:"];
+    const domainsList = ["trusted1.example.com", "trusted2.example.com"];
+
+    let url = new URL(request.query.url);
+
+    if (schemesList.includes(url.protocol) && domainsList.includes(url.hostname)) {
```
egon-okerman-sonarsource, line 29:
```suggestion

const ALLOWED_SCHEMES = ["http:", "https:"];
const ALLOWED_DOMAINS = ["trusted1.example.com", "trusted2.example.com"];

async function example(request) {
    let url = new URL(request.query.url);

    if (ALLOWED_SCHEMES.includes(url.protocol) && ALLOWED_DOMAINS.includes(url.hostname)) {
```
Is it possible to make this change? It would look cleaner IMO, but not sure if this causes problems with detection.

rules/S5144/javascript/how-to-fix-it/node.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+The following code is vulnerable to SSRF as it opens a URL defined by untrusted data.
+
+==== Noncompliant code example
+
+[source,python,diff-id=1,diff-type=noncompliant]
+----
+const axios = require('axios');
+
+async function example(request) {
```
egon-okerman-sonarsource, line 11: Since it's JS/TS, it might be nice to define the type of the argument `request`, but I'll leave this up to you to decide.

egon-okerman-sonarsource (changes requested):
I made some suggestions on code style changes. Feel free to ignore, but do make sure the language of the code blocks are changed.

rules/S5144/javascript/how-to-fix-it/node.adoc
pedro-oliveira-sonarsource: Added the 'express' require.

egon-okerman-sonarsource: I think that would be a great change, but do make sure to change the import statements if you do. As far as I remember (I might be wrong here), require doesn't play well with TypeScript and `import express from "express"` should be used instead.

pedro-oliveira-sonarsource: I decided to keep it JS, but adjusted the axios request to a synchronous one.

rules/S5144/javascript/how-to-fix-it/node.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+const app = express();
+
+app.get('/example', (req, res) => {
+    axios.get(req.query.url).then(resp => { // Noncompliant
+        res.send("OK");
+    }).catch(err => {
+        console.error(err);
+        res.send("ERROR");
+    });
```
egon-okerman-sonarsource, line 20: Wrapping this in an express example is great, but I'd use async/await here over `then()` for legibility

```suggestion
app.get('/example', async (req, res) => {
    try {
        const resp = await axios.get(req.query.url); // Noncompliant
        res.send("OK");
    } catch (err) {
        console.error(err);
        res.send("ERROR");
    }
});
```

rules/S5144/javascript/how-to-fix-it/node.adoc
egon-okerman-sonarsource: Keeping it JS is fine, but I think asynchronous should really be preferred over synchronous for legibility, unless there is a good reason not to.

rules/S5144/javascript/how-to-fix-it/node.adoc
pedro-oliveira-sonarsource: Unaware I could specify this function as async, thank you!

PR #1425, opened 2022-11-25
[APPSEC-292] Modify rule S6287 (PHP): Change text to the education framework format 

egon-okerman-sonarsource:


pierre-loup-tristant-sonarsource (approved):
LGTM.
It not related to a change made with this PR this sentence should probably be update:

> An application should not blindly assign session cookies' values
based on taintable data

I am not sure the concept of "taintable data" exist. Tainted data does, although I don't really use it in rule description since it feels more like a SAST technology "jargon".

PR #1428, opened 2022-11-28
Rule S1123: Add VB.NET

pavel-mikula-sonarsource:
https://github.com/SonarSource/sonar-dotnet/issues/6459


rules/S1123/description-dotnet.adoc
```diff
@@ -0,0 +1 @@
+The `Obsolete` attribute can be applied with or without arguments, but marking something `Obsolete` without including advice as to why it's obsolete or on what to use instead will lead maintainers to waste time trying to figure those things out - every single time the warning is encountered.
```
martin-strecker-sonarsource, line 1: A few thoughts:
* Obsolete has three constructors (no parameter, message parameter, and message + isError parameter). So "arguments" is a bit ambiguous. Better *with or without message argument*.
* Wording: "waste time ... to figure out" and *every single time the warning is encountered* sound too informal to me. It doesn't fit the formal style of the e.g. Java description.
* Implementation: `[Obsolete(null)]` and `[Obsolete("   ")]` may also raise the rule.

rules/S1123/vbnet/rule.adoc
```diff
@@ -0,0 +1,44 @@
...
+(visible only on this page)
+
+include::message.adoc[]
+
+'''
+== Comments And Links
+(visible only on this page)
+
+include::../comments-and-links.adoc[]
+endif::env-github,rspecator-view[]
```
martin-strecker-sonarsource, line 44: Wasn't there an issue with newlines around includes? Does this cause problems?

pavel-mikula-sonarsource: There was. And this rspecator-view is not part of rule HTML output, it's only for the website viewer. And it's copied from C# version, so we're fine.

martin-strecker-sonarsource (approved):
LGTM

PR #1433, opened 2022-11-29
Modify S5594(XML - Android): Improve text understandability (APPSEC-134)

loris-s-sonarsource:


rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
-If an Android component is exported and no permissions are defined then other mobile apps can interact with it and perform potential unauthorized actions.
+Once an Android component has been exported, both it and non-exported
+components are exposed for untrusted use (more information about collateral damages on non-exported components
+https://hackerone.com/reports/951691[here].)
```
egon-okerman-sonarsource, line 3: I read through this report briefly, and it does look interesting, but I have a few notes.

* Wrt "both it and non-exported components are exposed for untrusted use": we speak about it as if it is a guarantee, but it is more of an example as what can happen in specific instances. The report requires an exported Activity that also starts other Activities through Intents without doing any verification. IMO we should not put this as the focus of this rule.
* Though the report is very clear on exploitation, it is not very clear on mitigation. Are the mitigations we define here clear in general, and do they prevent from the attacks mentioned in the article?

rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
...
+* Activity-aliases
+* Services
 
-It's highly recommended to implement restrictive permissions on exposed components.
+To do so, it is recommended to either set `exported` to `false`, add
+`android:readPermission` and `android:writePermission` attributes, or add a
+`<permission>` tag.
 
+**Warning**: The presence of intent filters will cause exported to be set to
+true by default.
```
egon-okerman-sonarsource, line 20:
```suggestion
**Warning**: The presence of intent filters will cause ``++exported++`` to be set to
``++true++`` by default.
```
For consistency with the rest of the document.

egon-okerman-sonarsource (changes requested):
A lot of good changes, but some things should be reprioritized/further looked into.

rules/S5594/xml/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Restrict access to exported components with appropriate permissions",
+  "title": "Exported components should be protected with permissions",
```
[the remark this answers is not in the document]
loris-s-sonarsource: Ok, what about this? (the intention behind my change is also to make the title compliant with our writing guidelines):
```suggestion
  "title": "Exported components should be restricted with appropriate permissions",
```

rules/S5594/xml/rule.adoc
loris-s-sonarsource: Oh yeah, good point. Maybe that's actually out of scope here even. WDYT if we change the whole thing to:

```
Once an Android component has been exported, it is exposed to
untrusted use and might also weaken other components.
```

rules/S5594/xml/metadata.json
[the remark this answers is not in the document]
egon-okerman-sonarsource: Sounds good!

rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
...
+* Activities
+* Activity-aliases
+* Services
 
-It's highly recommended to implement restrictive permissions on exposed components.
+To do so, it is recommended to either set `exported` to `false`, add
+`android:readPermission` and `android:writePermission` attributes, or add a
+`<permission>` tag.
 
+**Warning**: The presence of intent filters will cause ``++exported++`` to be set to
```
loris-s-sonarsource, line 19:
```suggestion
**Warning**: For Android versions lower than 12, The presence of intent filters will cause ``++exported++`` to be set to
```

rules/S5594/xml/rule.adoc
loris-s-sonarsource: yeah you are right, my version might actually end up too ambiguous

rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
-If an Android component is exported and no permissions are defined then other mobile apps can interact with it and perform potential unauthorized actions.
+Once an Android component has been exported, both it and non-exported
+components are exposed for untrusted use (more information about collateral damage on non-exported components
+https://hackerone.com/reports/951691[here].)
```
loris-s-sonarsource, line 3:
```suggestion
Once an Android component has been exported, it can be used by attackers to
launch malicious actions and might also give access to other components
otherwise non-exported.
```

rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
...
+* Activities
+* Activity-aliases
+* Services
 
-It's highly recommended to implement restrictive permissions on exposed components.
+To do so, it is recommended to either set `exported` to `false`, add
+`android:readPermission` and `android:writePermission` attributes, or add a
+`<permission>` tag.
 
+**Warning**: For Android versions lower than 12, The presence of intent filters will cause ``++exported++`` to be set to
```
loris-s-sonarsource, line 19:
```suggestion
**Warning**: When targeting Android versions lower than 12, The presence of intent filters will cause ``++exported++`` to be set to
```

rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
...
+* Activities
+* Activity-aliases
+* Services
 
-It's highly recommended to implement restrictive permissions on exposed components.
+To do so, it is recommended to either set `exported` to `false`, add
+`android:readPermission` and `android:writePermission` attributes, or add a
+`<permission>` tag.
 
+**Warning**: When targeting Android versions lower than 12, The presence of intent filters will cause ``++exported++`` to be set to
```
egon-okerman-sonarsource, line 19:
```suggestion
**Warning**: When targeting Android versions lower than 12, the presence of intent filters will cause ``++exported++`` to be set to
```
Small typo

rules/S5594/xml/rule.adoc
```diff
@@ -1,36 +1,55 @@
-If an Android component is exported and no permissions are defined then other mobile apps can interact with it and perform potential unauthorized actions.
+Once an Android component has been exported, it can be used by attackers to
+launch malicious actions and might also give access to other components
+otherwise non-exported.
```
egon-okerman-sonarsource, line 3:
```suggestion
launch malicious actions and might also give access to other components
that are not exported.
```
"other components otherwise non-exported" could be worded a bit differently

egon-okerman-sonarsource (approved):
LGTM, just left 2 small suggestions wrt to the wording

PR #1435, opened 2022-11-29
Modify S5542 [JS/TS]: Only raise for AES-CBC when used with padding (APPSEC-32)

egon-okerman-sonarsource:
Specification ticket: [APPSEC-32](https://sonarsource.atlassian.net/browse/APPSEC-32)
Implementation ticket: [SonarJS #3592](https://github.com/SonarSource/SonarJS/issues/3592)
[Preview](https://sonarsource.github.io/rspec/#/rspec/S5542)

### DIFF
--- a/rules/S5542/javascript/message.adoc
+++ b/rules/S5542/javascript/message.adoc
@@ -1,0 +1,13 @@
+=== Message
+
+==== ECB
+
+Use a secure cipher mode.
+
+==== CBC
+
+Use another cipher mode or disable padding.
+
+==== RSA
+
+Use a secure padding scheme.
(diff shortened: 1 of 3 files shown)


rules/S5542/javascript/message.adoc
```diff
@@ -0,0 +1,13 @@
+=== Message
+
+==== ECB
+
+Use a secure cipher mode.
+
+==== CBC
+
+Use another cipher mode or disable padding.
```
gaetan-ferry-sonarsource, line 9: I would stick with "secure" instead of "another" which is quite vague.
```suggestion
Use a secure cipher mode or disable padding.
```
Also, I would not recommend disabling padding. In most cases, this will not be possible at all and bring some confusion to the user. Moreover, CBC with no padding is still prone to practical attacks such as the "byte flipping" one.

To finish with some nitpicking, the established wording is "block cipher mode of operation" instead of "cipher mode". This might be a bit verbose though.

gaetan-ferry-sonarsource (changes requested):
I suggest some minor changes in the wording. Moreover, I have the feeling that setting CBC without padding as safe is wrong.

egon-okerman-sonarsource:
Closing this for now, as the cryptography-related rules will be restructured at a later point.

PR #1442, opened 2022-11-30
Modify rule S5945: Add exceptions for constexpr

alejandro-alvarez-sonarsource:


alejandro-alvarez-sonarsource:
Sorry, I forgot about #1416

PR #1445, opened 2022-12-01
Deprecate rule S4214 (replaced by S4200)

martin-strecker-sonarsource:
Reasoning is described [here](https://docs.google.com/document/d/1QXn3rWuAw5am9UG3dtjbg-ChHzX-N9N-XLgTUdpb_0w/edit?usp=sharing)

### DIFF
--- a/rules/S4214/metadata.json
+++ b/rules/S4214/metadata.json
@@ -21,7 +21,6 @@
   "sqKey": "S4214",
   "scope": "Main",
   "defaultQualityProfiles": [
-    "Sonar way"
   ],
   "quickfix": "unknown"
 }
--- a/rules/S4214/metadata.json
+++ b/rules/S4214/metadata.json
@@ -10,7 +10,7 @@
   ],
   "extra": {
     "replacementRules": [
-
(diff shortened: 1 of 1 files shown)


rules/S4214/metadata.json
```diff
@@ -1,17 +1,16 @@
...
   "title": "\"P\/Invoke\" methods should not be visible",
   "type": "CODE_SMELL",
-  "status": "ready",
+  "status": "deprecated",
   "remediation": {
     "func": "Constant\/Issue",
     "constantCost": "2min"
   },
   "tags": [
-    "pinvoke"
```
martin-strecker-sonarsource, line 10: This was the only rule with that tag. We should not add it to the replacement rule IMO. The replacement rule has the "pitfall" tag.

andrei-epure-sonarsource (approved):
LGTM

~before merging, could you ask on the LT slack what's the "correct" format for the replacement rule key?~ I saw you followed the format in https://github.com/SonarSource/rspec#to-deprecate-a-rule

PR #1446, opened 2022-12-01
Readme: Clarify rule deprecation

martin-strecker-sonarsource:
The replacementRule part isn't specifying the rule id format:
```json
  "extra": {
    "replacementRules": [
      "RSPEC-4200" // or "S4200" or "RSPEC-S4200"? 😕 
    ],
    "legacyKeys": [

    ]
  },
```

Also, a link to an existing example helps with cross-checking.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -159,7 +159,9 @@ In the `metadata.json` of the rule you want to deprecate:
 * Remove all tags
 * Remove all quality profiles
 * Change the status to `deprecated`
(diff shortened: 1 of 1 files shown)


arseniy-sonar (approved):
Indeed, thanks for the clarification!

arseniy-sonar:
A complete solution could also fix S1567 which uses the wrong format and added a metadata check to make sure no other format is used.
However, the documentation improvement is good on it's own, so I see no need to delay this PR for the sake of a complete solution.

PR #1456, opened 2022-12-05
Modify rule S3353: Mark quickfix for CSharp as partially covered

martin-strecker-sonarsource:
Marks the quick fix implemented by https://github.com/SonarSource/sonar-dotnet/pull/6043 as partially implemented. Some edge cases with e.g. multiple declarators are not supported. See also https://github.com/SonarSource/rspec/blob/master/docs/metadata.adoc


csaba-sagi-sonarsource (approved):
LGTM!

PR #1457, opened 2022-12-05
Modify Multiple Rules: Add consistency accross  S2076, S5334 and S5883 rules impact descriptions (APPSEC-340)

gaetan-ferry-sonarsource:


rules/S2076/impact.adoc
```diff
@@ -18,18 +18,18 @@ profit from data leaks.
 
 The attackers:
```
egon-okerman-sonarsource, line 19: I feel like this list doesn't read exactly right. This looks like a list of possibilities, so I'd put e.g. "The attackers could for example: [download, modify, stop, ...]

PR #1471, opened 2022-12-12
Modify rule S1612 (Java): Short lambdas are considered compliant

Swalkyn:
When a lambda function is shorter than its equivalent method reference, the code is considered compliant and no issue is raised.


Swalkyn:
Corresponding changes in sonar-java: https://github.com/SonarSource/sonar-java/pull/4238

alban-auzeill (approved):

PR #1475, opened 2022-12-15
Modify Rule S1191: removed "com.sun.*" packages check from rule implementation (SONARJAVA-4382)

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #1476, opened 2022-12-15
Modify rule S1168: Raise on default and ternaries

gregory-paidis-sonarsource:
Added the necessary changes to support the following cases:

```csharp
return default;
return default(T);
return condition ? null : someResult;
```

More can be found on [this PR](https://github.com/SonarSource/sonar-dotnet/pull/6457).

### DIFF
--- a/rules/S1168/csharp/rule.adoc
+++ b/rules/S1168/csharp/rule.adoc
@@ -9,11 +9,9 @@ public Result[] GetResults()
     return null; // Noncompliant
 }
 
-public IEnumerable<Result> GetResults()
+public IEnumerable<Result> GetResults(bool condition)
 {
     var results = GenerateResults();
-    var condition = RollDice();
-
     return condition 
         ? results 
         : null; // Noncompliant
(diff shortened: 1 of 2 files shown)


rules/S1168/csharp/rule.adoc
```diff
@@ -11,7 +11,12 @@ public Result[] GetResults()
...
 public IEnumerable<Result> GetResults()
 {
-    return null; // Noncompliant
+    var results = GenerateResults();
+    var condition = RollDice();
+
+    return condition 
+        ? results 
+        : null; // Noncompliant
 }
```
costin-zaharia-sonarsource, line 20: To simplify the example we could also have:
```
public IEnumerable<Result> GetResults(bool condition)
{
    var results = GenerateResults();
    return condition 
        ? results 
        : null; // Noncompliant
}
```
Make sure thought that it raises the issue before changing it.

rules/S1168/csharp/rule.adoc
```diff
@@ -38,7 +43,12 @@ public Result[] GetResults()
 
 public IEnumerable<Result> GetResults()
 {
-    return Enumerable.Empty<Result>();
+    var results = GenerateResults();
+    var condition = RollDice();
```
costin-zaharia-sonarsource, line 47: If you change the other one, this one will have to be changed as well.

costin-zaharia-sonarsource (changes requested):

costin-zaharia-sonarsource (approved):
LGTM!

PR #1479, opened 2022-12-19
Modify rule S3654: Improve on rule description (CPP-3734)

pdschbrt:
Adjust rule description: throwing an exception in a destructor is not necessarily undefined behavior. However, it is highly dangerous to throw in a destructor. This PR also adds an example on why that is.

### DIFF
--- a/rules/S3654/cfamily/rule.adoc
+++ b/rules/S3654/cfamily/rule.adoc
@@ -1,4 +1,4 @@
-Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.
 
 Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
 Destructors may still throw an exception if they are explicitly declared as ``noexcept(false)``.
--- a/rules/S3654/cfamily/rule.adoc
+++ b/rules/S3654/cfamily/rule.adoc
@@ -4,7 +4,7 @@ Destructors are usually (implicitly) declared as ``noexcept`` by default such th
 Destructors may still propagate an exception if they are explicitly declared as ``noexcept(false)``.
 However, even a destructor that is declared as ``noexcept(false)`` will call `std::terminate` if it throws during stack unwinding.
 
-A commonly used example that highlights the severity of the underlying problem is presented in what follows.
+A commonly used example that highlights the severity of the underlying problem is presented in what follows:
 The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all managed objects.
 Suppose a call to an object's destructor throws an exception.
 In that case, there are only two conceptional ways to proceed:
--- a/rules/S3654/cfamily/rule.adoc
+++ b/rules/S3654/cfamily/rule.adoc
@@ -7,7 +7,7 @@ However, even a destructor that is declared as ``noexcept(false)`` will call `st
 A commonly used example that highlights the severity of the underlying problem is presented in what follows:
 The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all managed objects.
 Suppose a call to an object's destructor throws an exception.
-In that case, there are only two conceptional ways to proceed:
+In that case, there are only two conceptual ways to proceed:
 
 1. Abort destruction. This will result in a partially destroyed object and possibly many more objects whose destructor has not been called. 
 2. Ignore the exception and proceed with destroying the remaining objects. However, if yet another exception is thrown while destroying the remaining objects, there would be two outstanding exceptions. This is disallowed by C++'s exception-handling mechanism and results in a call to ``std::terminate``.
--- a/rules/S3654/cfamily/rule.adoc
+++ b/rules/S3654/cfamily/rule.adoc
@@ -1,7 +1,7 @@
 Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.
 
 Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
-Destructors may still throw an exception if they are explicitly declared as ``noexcept(false)``.
+Destructors may still propagate an exception if they are explicitly declared as ``noexcept(false)``.
 However, if a destructor that is declared as ``noexcept(false)`` happens to be called during stack unwinding, ``std::terminate`` is called instead.
 
 A commonly used example that highlights the severity of the underlying problem is presented in what follows.
--- a/rules/S3654/cfamily/rule.adoc
+++ b/rules/S3654/cfamily/rule.adoc
@@ -1,4 +1,4 @@
-Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without being allowed to perform a clean shutdown.
 
 Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
 Destructors may still propagate an exception if they are explicitly declared as ``noexcept(false)``.
--- a/rules/S3654/cfamily/rule.adoc
+++ b/rules/S3654/cfamily/rule.adoc
@@ -1,6 +1,9 @@
 Throwing an exception from a destructor can lead to undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
 
-A commonly used example that highlights the severity of the underlying problem is presented in what follows. The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all of the managed objects. If a call to an object's destructor throws an exception, there are only two conceptional ways to proceed: (i) abort destruction which will result in a half-destroyed object and possibly many more objects whose destructor has not been called, or (ii) ignore the exception and proceed with destroying the remaining objects. However, if yet another exception is thrown while destroying the remaining objects, there would be two outstanding exceptions which is disallowed by C++ exception-handling mechanism and results in a call to ``std::terminate``. Both of these options are undesired.
+A commonly used example that highlights the severity of the underlying problem is presented in what follows. The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all managed objects. Suppose a call to an object's destructor throws an exception. In that case, there are only two conceptional ways to proceed:
+1. Abort destruction. This will result in a partially destroyed object and possibly many more objects whose destructor has not been called. 
+2. Ignore the exception and proceed with destroying the remaining objects. However, if yet another exception is thrown while destroying the remaining objects, there would be two outstanding exceptions. This is disallowed by C++ exception-handling mechanism and results in a call to ``std::terminate``. 
+Both of these options are undesired.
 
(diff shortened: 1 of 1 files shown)


loic-joly-sonarsource:
I don't think this is the full story. 

Since destructors are in many cases declared as `noexcept`, throwing from a destructor usually ends with calling `std::terminate`, even if only one exception is generated, so the scenario you describe can only happen if, in the first place, users violate this rule.

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,5 +1,6 @@
-Throwing an exception from a destructor results in undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor can lead to undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
 
+A commonly used example that highlights the severity of the underlying problem is presented in what follows. The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all of the managed objects. If a call to an object's destructor throws an exception, there are only two conceptional ways to proceed: (i) abort destruction which will result in a half-destroyed object and possibly many more objects whose destructor has not been called, or (ii) ignore the exception and proceed with destroying the remaining objects. However, if yet another exception is thrown while destroying the remaining objects, there would be two outstanding exceptions which is disallowed by C++ exception-handling mechanism and results in a call to ``std::terminate``. Both of these options are undesired.
```
abbas-sabra-sonarsource, line 3: We try to avoid long sentences and use list to make things easier to read:
```suggestion
A commonly used example that highlights the severity of the underlying problem is presented in what follows. The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all managed objects. Suppose a call to an object's destructor throws an exception. In that case, there are only two conceptional ways to proceed:
1. Abort destruction. This will result in a partially destroyed object and possibly many more objects whose destructor has not been called. 
2. Ignore the exception and proceed with destroying the remaining objects. However, if yet another exception is thrown while destroying the remaining objects, there would be two outstanding exceptions. This is disallowed by C++ exception-handling mechanism and results in a call to ``std::terminate``. 
Both of these options are undesired.
```
Also, you forgot the most common case; the destructor is noexcept. A lot of destructors are noexcept implicitly

abbas-sabra-sonarsource (commented):
You have to open a PR for sonar-cpp to update the metadata there also. You can use this branch when it is approved to create the sonar-cpp PR: https://github.com/SonarSource/sonar-cpp/blob/master/docs/RuleMetadata.adoc

pdschbrt:
Thanks for your feedback. Will update the description to include the point in implicit `noexcept`.

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
-Throwing an exception from a destructor results in undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying others objects.
 
+Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
+Destructors may still throw an exception if they are explicitly declared as ``noexcept(false)``.
+However, if a destructor that is declared as ``noexcept(false)`` happens to be called during stack unwinding, ``std::terminate`` is called instead.
```
abbas-sabra-sonarsource, line 5:
```suggestion
Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.

Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
Destructors may still throw an exception if they are explicitly declared as ``noexcept(false)``.
However, if a destructor that is declared as ``noexcept(false)`` happens to be called during stack unwinding, ``std::terminate`` is called instead.
```

arseniy-sonar:
Hi @pdschbrt 
Please follow the [PR description guideline](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule) (also pay attention to the generated commit message when you squash&merge it)

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
-Throwing an exception from a destructor results in undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying others objects.
```
abbas-sabra-sonarsource, line 1:
```suggestion
Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.
```

abbas-sabra-sonarsource (changes requested):

abbas-sabra-sonarsource:
@pdschbrt, rspec repo has different conventions for PR titles because it is shared between bubbles. have a look at modifying the rules convention:
https://github.com/SonarSource/rspec#to-modify-an-existing-rule
* You need to change the title and add a label.
* Also, can you add a link to the `sonar-cpp` rspec update PR so I can review/merge them together

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
-Throwing an exception from a destructor results in undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.
 
+Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
+Destructors may still throw an exception if they are explicitly declared as ``noexcept(false)``.
```
loic-joly-sonarsource, line 4:
```suggestion
Destructors may still propagate an exception if they are explicitly declared as ``noexcept(false)``.
```

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
...
+However, if a destructor that is declared as ``noexcept(false)`` happens to be called during stack unwinding, ``std::terminate`` is called instead.
```
loic-joly-sonarsource, line 5:
```suggestion
However, even a destructor that is declared as ``noexcept(false)`` will call `std::terminate` if it throws during stack unwinding.
```

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
...
 
-Thus destructors should never ``++throw++`` exceptions. Instead, they should catch and handle those thrown by the functions they call, and be ``++noexcept++``.
+A commonly used example that highlights the severity of the underlying problem is presented in what follows.
```
loic-joly-sonarsource, line 7:
```suggestion
A commonly used example that highlights the severity of the underlying problem is presented in what follows:
```

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
...
+The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all managed objects.
+Suppose a call to an object's destructor throws an exception.
+In that case, there are only two conceptional ways to proceed:
```
loic-joly-sonarsource, line 10:
```suggestion
In that case, there are only two conceptual ways to proceed:
```
??

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
-Throwing an exception from a destructor results in undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without neatly destroying other objects.
```
loic-joly-sonarsource, line 1: In the current sentence wording, it does not seem clear that destroying other objects is a good thing...

```suggestion
Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without being allowed to perform a clean shutdown.
```

pdschbrt: Thanks for your inputs, @loic-joly-sonarsource.

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
-Throwing an exception from a destructor results in undefined behavior, meaning that your program could be terminated abruptly without neatly destroying others objects.
+Throwing an exception from a destructor may result in a call to ``std::terminate``, meaning that your program could be terminated abruptly without being allowed to perform a clean shutdown.
 
+Destructors are usually (implicitly) declared as ``noexcept`` by default such that ``std::terminate`` is called when they throw an exception.
+Destructors may still propagate an exception if they are explicitly declared as ``noexcept(false)``.
+However, even a destructor that is declared as ``noexcept(false)`` will call `std::terminate` if it throws during stack unwinding.
 
-Thus destructors should never ``++throw++`` exceptions. Instead, they should catch and handle those thrown by the functions they call, and be ``++noexcept++``.
+A commonly used example that highlights the severity of the underlying problem is presented in what follows:
+The destructor of a standard container like ``++std::vector++`` automatically calls the destructors for all managed objects.
```
tomasz-kaminski-sonarsource, line 8: If destructor of type used as elements on the vector throws in the desturctor, the behavior is undefined: https://eel.is/c++draft/requirements#res.on.functions-2.4

tomasz-kaminski-sonarsource: There are more ways:
* delete the source of the program, and other manifestation of UB
* termination
This suggests that `std::vector` has stronger guarantee, that it actually have

rules/S3654/cfamily/rule.adoc
```diff
@@ -1,7 +1,20 @@
...
+Suppose a call to an object's destructor throws an exception.
+In that case, there are only two conceptual ways to proceed:
+
+1. Abort destruction. This will result in a partially destroyed object and possibly many more objects whose destructor has not been called. 
+2. Ignore the exception and proceed with destroying the remaining objects. However, if yet another exception is thrown while destroying the remaining objects, there would be two outstanding exceptions. This is disallowed by C++'s exception-handling mechanism and results in a call to ``std::terminate``.
```
tomasz-kaminski-sonarsource, line 13: Reading this. If you ignore an exception and continue with the destruction of remaining objects, you will not end up with two exceptions, as you ignored one. The phrasing seems to be unclear.

rules/S3654/cfamily/rule.adoc
pdschbrt: Indeed, I just wanted to give an example why this is a lost case and explain it on a conceptual level here.

This topic is huge; there are also exceptions (other than declaring a destructor as `noexcept(false)`) when destructors are _not_ implicitly `noexcept`. But I feel this would be too much for a reader of this rule.

rules/S3654/cfamily/rule.adoc
abbas-sabra-sonarsource: I agree this part needs clarification/can be misunderstood

rules/S3654/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Maybe:
```The destructor of a  container needs to call the destructors for all managed objects.```
And add clarification that for the standard container, the behavior is undefined.

rules/S3654/cfamily/rule.adoc
pdschbrt: I've updated it.

rules/S3654/cfamily/rule.adoc
pdschbrt: Thanks, I've integrated your suggestions.

tomasz-kaminski-sonarsource: Looks good.

abbas-sabra-sonarsource (approved):
LGTM

PR #1480, opened 2022-12-20
Modify rule S1144: Added exception for methods with specific annotations on parameters (SONARJAVA-4347)

leonardo-pilastri-sonarsource:


Wohops (approved):
LGTM!

PR #1483, opened 2022-12-22
SONARJAVA-4381 S1135: Update metadata to be explicit about main code …

irina-batinic-sonarsource:


Wohops (dismissed):

rules/S1135/metadata.json
```diff
@@ -20,7 +20,7 @@
   "defaultSeverity": "Info",
   "ruleSpecification": "RSPEC-1135",
   "sqKey": "S1135",
-  "scope": "All",
```
Wohops, line 23: Not sure that it's the correct place to change it. If you do change here, it will also impact other analyzers. Other analyzers MIGHT actually analyze tests. It's only the Java analyzer that does not.

I believe you should just change the metadata for java.

Wohops (changes requested):

Wohops (approved):
:+1:

PR #1484, opened 2022-12-23
Modify rule S3628: Raw string literals should be used

pdschbrt:


tomasz-kaminski-sonarsource:
Add exmaple showing that you can add prefix in raw string literal.

tomasz-kaminski-sonarsource (approved):
Merge it around the same time as a corresponding change in the repository.

PR #1485, opened 2022-12-23
Modify rule S3539: Access specifiers should not be redundant

pdschbrt:


pdschbrt:
Please find the pull-request for sync with `sonar-cpp` at https://github.com/SonarSource/sonar-cpp/pull/2342

PR #1490, opened 2023-01-05
Modify rule S4200: Improve documentation

martin-strecker-sonarsource:
Based on the discussion in https://github.com/SonarSource/sonar-dotnet/issues/6340 the wording of the rule description is changed to be less misunderstandable.


andrei-epure-sonarsource (approved):
LGTM

PR #1491, opened 2023-01-05
Modify rule S1527: Add mention about why we keep adopted future reserved words

ilia-kebets-sonarsource:
Relates to https://github.com/SonarSource/SonarJS/issues/2990

### DIFF
--- a/rules/S1527/description.adoc
+++ b/rules/S1527/description.adoc
@@ -21,4 +21,4 @@ The following words may be used as keywords in future evolutions of the language
 
 Use of these words as identifiers would produce an error in JavaScript ``++strict++`` mode code.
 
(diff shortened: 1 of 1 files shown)


rules/S1527/description.adoc
```diff
@@ -20,3 +20,5 @@ The following words may be used as keywords in future evolutions of the language
 * ``++yield++``
 
 Use of these words as identifiers would produce an error in JavaScript ``++strict++`` mode code.
+
+This list includes keywords that have been adopted by the latest implementations of ECMAScript. We have kept them in the rule because our analyzer runs on legacy JavaScript codebases as well.
```
vilchik-elena, line 24: I would try to be more depersonalized

```suggestion
This list includes also the keywords that have been adopted by the latest versions of ECMAScript. They are kept to support legacy JavaScript codebases.
```

vilchik-elena (approved):

PR #1497, opened 2023-01-09
Modify multiple rules: Clean up texts of MMF-2503

hendrik-buchwald-sonarsource:


rules/S4507/python/rule.adoc
```diff
@@ -20,6 +19,7 @@ def custom_config(config):
 
 Django's "settings.py" or "global_settings.py" configuration file
```
hendrik-buchwald-sonarsource, line 20:
```suggestion
Django's "settings.py" or "global_settings.py" configuration file:
```

PR #1498, opened 2023-01-11
Modify S2221: Fix title and rule formatting

martin-strecker-sonarsource:


rules/S2221/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+    "title": "\"Exception\" should not be caught"
```
martin-strecker-sonarsource, line 2: The default title is 

`"title": "\"Exception\" should not be caught when not required by called methods",`

The "when not required by called methods" is confusing. It probably refers to [Checked Exceptions in Java](https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/).

andrei-epure-sonarsource (approved):
LGTM

PR #1502, opened 2023-01-13
Fix comment in S6493 code example

jwakely:
I don't think "non-user-defined type" is what was intended here. If you consider any class or enum type to be "user-defined" then `std::string` *is* user-defined. If the term is being used to mean "not defined by the implementation" then that's not relevant here, because that isn't what determines whether it dangles in this code or not.


frederic-tingaud-sonarsource:
Thanks a lot for pointing that problem out @jwakely

PR #1503, opened 2023-01-16
Modify rule S4018: modify message to make it clear the rule is about type inference

zsolt-kolbay-sonarsource:
The rule's message and description were too vague. They didn't make it clear that the type parameters need to be used in the argument list in order to enable type inference when the method is called. Those texts in the rule were updated to reflect that.

Implementation ticket: SonarSource/sonar-dotnet/issues/6636

### DIFF
--- a/rules/S4018/message.adoc
+++ b/rules/S4018/message.adoc
@@ -1,4 +1,4 @@
 === Message
 
-Refactor this method to have parameters matching all the type parameters.
+Use all type parameters in the parameter list to enable type inference.
 
--- a/rules/S4018/message.adoc
+++ b/rules/S4018/message.adoc
@@ -1,4 +1,4 @@
 === Message
 
-Use all type parameters in the parameter list to enable type inference.
+Refactor this method to use all type parameters in the parameter list to enable type inference.
 
--- a/rules/S4018/rule.adoc
+++ b/rules/S4018/rule.adoc
@@ -13,6 +13,7 @@ namespace MyLibrary
   {
     public void MyMethod<T>()  // Noncompliant
     {
+        // this method can only be invoked by providing the type argument e.g. 'MyMethod<int>()'
     }
   }
 }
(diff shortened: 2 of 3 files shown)


rules/S4018/rule.adoc
```diff
@@ -1,4 +1,4 @@
-The best way to determine the type of a generic method is by inference based on the type of argument that is passed to the method. This is not possible when a parameter type is missing from the argument list.
+Type inference enables the call of a generic method without explicitly specifying its type arguments. This is not possible when a parameter type is missing from the argument list.
```
andrei-epure-sonarsource, line 2: to be more explicit, maybe you can add comments inside the samples to better explain visually the difference

(for me, the text only is very abstract)

```
    public void MyMethod<T>()  // Noncompliant
    {
      // this must be invoked providing the type argument e.g. `MyMethod<int>()`
     }
```

and

```
    public void MyMethod<T>(T param)
    {
      // type inference allows this to be invoked `MyMethod(arg)`
    }
```

WDYT?

andrei-epure-sonarsource (commented):
WDYT about giving a hint as comment in the sample?

andrei-epure-sonarsource:
@zsolt-kolbay-sonarsource please update the PR title to match the guideline : https://github.com/SonarSource/rspec#to-modify-an-existing-rule

>  Modify rule S9999: Foo bar baz

Add the link to the implementation ticket: https://github.com/SonarSource/rspec#3-create-an-implementation-ticket

This is necessary for the automation for https://sonarsource.github.io/rspec/

rules/S4018/message.adoc
```diff
@@ -1,4 +1,4 @@
 === Message
 
-Refactor this method to have parameters matching all the type parameters.
+Use all type parameters in the parameter list to enable type inference.
```
andrei-epure-sonarsource, line 3: What about still mentioning "refactor" because 

> Refactor this method to use all type parameters in the parameter list to enable type inference.

I find the message a bit milder, taking into account it wouldn't be a super easy task

PR #1508, opened 2023-01-23
Modify rule S5935: Fix dead link

zsolt-kolbay-sonarsource:
Some of the rules contain dead links, which causes the PR checks to fail.

### DIFF
--- a/rspec-tools/rspec_tools/checklinks.py
+++ b/rspec-tools/rspec_tools/checklinks.py
@@ -20,7 +20,8 @@ EXCEPTIONS = ['https://blogs.oracle.com/java-platform-group/diagnosing-tls,-ssl,
               # FIXME: RULEAPI-763 - investigate why these live links respond with 403
               'https://medium.com/@cowtowncoder/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062',
               'https://medium.com/@nyomanpradipta120/ssti-in-flask-jinja2-20b068fdaeee',
-              'https://elizarov.medium.com/coroutine-context-and-scope-c8b255d59055']
+              'https://elizarov.medium.com/coroutine-context-and-scope-c8b255d59055',
+              'https://www.ssi.gouv.fr/uploads/2014/11/RGS_v-2-0_B1.pdf']
 
 def show_files(filenames):
   for filename in filenames:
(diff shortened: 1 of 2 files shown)


rspec-tools/rspec_tools/checklinks.py
```diff
@@ -20,7 +20,8 @@
               # FIXME: RULEAPI-763 - investigate why these live links respond with 403
               'https://medium.com/@cowtowncoder/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062',
               'https://medium.com/@nyomanpradipta120/ssti-in-flask-jinja2-20b068fdaeee',
-              'https://elizarov.medium.com/coroutine-context-and-scope-c8b255d59055']
+              'https://elizarov.medium.com/coroutine-context-and-scope-c8b255d59055',
+              'https://www.ssi.gouv.fr/uploads/2014/11/RGS_v-2-0_B1.pdf']
```
martin-strecker-sonarsource, line 24: Please don't do this. We need to fix the root cause which seems to be a missing root certificate on the build machine:
https://stackoverflow.com/questions/52805115/certificate-verify-failed-unable-to-get-local-issuer-certificate

martin-strecker-sonarsource (changes requested):

rspec-tools/rspec_tools/checklinks.py
martin-strecker-sonarsource: What is odd is that the certifi dependency has the right root CA installed:
* Root CA of ssi.gouv.fr: CN = Certigna, O = Dhimyotis, C = FR, serial number: 18364802974209362175
* certifi version 2022.5.18.1: https://github.com/SonarSource/rspec/blob/master/rspec-tools/Pipfile.lock#L143-L149
* Matching certificate in certifi: https://github.com/certifi/python-certifi/blob/2022.05.18.1/certifi/cacert.pem#L694-L722

zsolt-kolbay-sonarsource: I tried to upgrade it, but `2022.12.7` is indeed the latest version, so that doesn't resolve it.
Any other ideas? I'm currently stuck with this one. The only solution that worked so far was the inelegantly adding the link to the list of exceptions.

martin-strecker-sonarsource: I would just fix the one broken link and give the task of fixing the other one to the owners of the check_link script. Ask in our bubble who that is.

zsolt-kolbay-sonarsource: I reverted the change.

martin-strecker-sonarsource (approved):
LGTM

martin-strecker-sonarsource:
@zsolt-kolbay-sonarsource I asked the PHP bubble to also have a look. Please wait until one of them has approved too.

nils-werner-sonarsource (approved):
LGTM!

PR #1509, opened 2023-01-23
Modify rule S1244: improve C# message for special floating point values (NaN, PositiveInfinity, ...)

antonioaversa:


pavel-mikula-sonarsource (approved):
LGTM.

This is a good opporotunity to merge directly from the review comment.

This needs to be merged AFTER https://github.com/SonarSource/sonar-dotnet/issues/6571

PR #1510, opened 2023-01-24
S3898: Add VB.NET

pavel-mikula-sonarsource:
https://github.com/SonarSource/sonar-dotnet/issues/6667


rules/S3898/csharp/rule.adoc
```diff
@@ -1,49 +1,28 @@
...
-  {
-    //...
-  }
+    public int Value { get; set; }
 
-  public bool Equals(MyStruct other)
-  {
-    throw new NotImplementedException();
-  }
+    public bool Equals(MyStruct other)
```
martin-strecker-sonarsource, line 21: Implementing `Equals<T>` without also implementing `object.Equals(object o)` and `object.GetHashCode()` is a violation of another rule.
From the docs:
> If you implement [IEquatable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-7.0), you should also override the base class implementations of [Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0#system-object-equals(system-object)) and [GetHashCode()](https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-7.0#system-object-gethashcode) so that their behavior is consistent with that of the [Equals(T)](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1.equals?view=net-7.0#system-iequatable-1-equals(-0)) method.

> For a value type, you should always implement [IEquatable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-7.0) and override [Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0#system-object-equals(system-object)) for better performance.

[Source](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-7.0)

I would add these methods to the compliant solution.

rules/S3898/vbnet/rule.adoc
```diff
@@ -0,0 +1,30 @@
...
+== Compliant Solution
+
+[source,vbnet]
+----
+Structure MyStruct
+    Implements IEquatable(Of MyStruct)
+
+    Public Property Value As Integer
+
+    Public Overloads Function Equals(other As MyStruct) As Boolean Implements IEquatable(Of MyStruct).Equals
```
martin-strecker-sonarsource, line 23: Same here. `Equals(object)` and `GetHashcode()` should be part of the compliant solution.

rules/S3898/description.adoc
```diff
@@ -0,0 +1 @@
+If you're using a `struct`, it is likely because you're interested in performance. But by failing to implement `IEquatable<T>` you're loosing performance when comparisons are made because without `IEquatable<T>`, boxing and reflection are used to make comparisons.
```
martin-strecker-sonarsource, line 1: In VB structs are called `Structure` and you implement `IEquatable(Of T)`

martin-strecker-sonarsource: Can we add a link to the "Notes for implementers" section to back up our claims?
https://learn.microsoft.com/dotnet/api/system.iequatable-1#notes-to-implementers

martin-strecker-sonarsource: Strictly speaking, IEquality is not about comparison but equality.

martin-strecker-sonarsource (changes requested):

rules/S3898/csharp/rule.adoc
pavel-mikula-sonarsource: That would be just a noise for this rule. The other rule should start complaining.

rules/S3898/description.adoc
pavel-mikula-sonarsource: Equality that is typicaly checked as part of comparison? :) I'll leave that one as is

rules/S3898/csharp/rule.adoc
martin-strecker-sonarsource: Okay, so we have https://sonarsource.github.io/rspec/#/rspec/S1206/csharp *"Equals(Object)" and "GetHashCode()" should be overridden in pairs* but it seems there seems to be no *override Equals(Object) and delegate to Equals(T) if Equals(T) is implemented* rule. Should we do a suggestion in the New rules backlog?

pavel-mikula-sonarsource: Sounds reasonable

rules/S3898/see.adoc
```diff
@@ -0,0 +1,4 @@
+
+== See
+
+* https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1
```
martin-strecker-sonarsource, line 4:
```suggestion
* https://learn.microsoft.com/dotnet/api/system.iequatable-1#notes-to-implementers[IEquatable<T> Interface (Microsoft Learn)]
```

martin-strecker-sonarsource (approved):
LGTM. One suggestion regarding the link. And https://github.com/SonarSource/rspec/pull/1510#discussion_r1085425802 should be answered.

rules/S3898/see.adoc
pavel-mikula-sonarsource: I intentionally provided link to the whole class, without the anchor. The displayed text can be improved

rules/S3898/csharp/rule.adoc
martin-strecker-sonarsource: New rule description https://github.com/SonarSource/sonar-dotnet/issues/6671

pavel-mikula-sonarsource: That should be nice new rule for the next onboarding batch :)

rules/S3898/csharp/rule.adoc
```diff
@@ -1,49 +1,29 @@
...
 [source,csharp]
 ----
 struct MyStruct  // Noncompliant
 {
-  private int i;
-  public int I
-  {
-    //...
-  }
+    public int Value { get; set; }
```
Corniel, line 9: get only is a more common scenario.

pavel-mikula-sonarsource: This is just to avoid an empty struct, it doesn't matter what exactly it is

Corniel: I get your reasoning. I would argue that `public int Value { get; }` is just a small improvement. That's all.

martin-strecker-sonarsource: Mutable structs make me shudder as well, but you would also have to add a constructor to an immutable struct to make it usable so `{ get; set; }` is a good compromise in this case.

Corniel: @martin-strecker-sonarsource that is a valid argument. The documentation improved any way, getting rid of some noise.

pavel-mikula-sonarsource: Like this one? :)
https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Point.cs,94

PR #1511, opened 2023-01-24
Modify Rule S4457(C#): Remove from Sonar way Profile

gregory-paidis-sonarsource:
After [this discussion](https://github.com/SonarSource/sonar-dotnet/issues/6132) and internal talk with the .NET Bubble, we decided to disable this rule by default.

https://sonarsource.github.io/rspec/#/rspec/S4457/csharp

### DIFF
--- a/rules/S4457/comments-and-links.adoc
+++ b/rules/S4457/comments-and-links.adoc
@@ -1,3 +1,3 @@
 === on 24 Jan 2023, 12:13:54 Grigorios Paidis wrote:
 
-To see more technical details, check https://stackoverflow.com/questions/56909648/split-async-method-into-two-for-code-analysis/56909963#56909963
+The rule is controversial and was thoroughly and correctly analyzed on https://stackoverflow.com/a/56909963[StackOverflow]. As a consequence, we removed it from SonarWay.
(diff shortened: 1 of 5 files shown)


rules/S4457/csharp/metadata.json
```diff
@@ -1,3 +1,5 @@
 {
-  
-}
+    "defaultQualityProfiles": [
```
martin-strecker-sonarsource, line 2: Only CSharp and VB have support for this rule. I would change the profile in the main metadata.json.

martin-strecker-sonarsource (changes requested):

rules/S4457/csharp/metadata.json
martin-strecker-sonarsource: Also: add a link to https://stackoverflow.com/a/56909963 in the comments

rules/S4457/comments-and-links.adoc
```diff
@@ -0,0 +1,3 @@
+=== on 24 Jan 2023, 12:13:54 Grigorios Paidis wrote:
+
+To see more technical details, check https://stackoverflow.com/questions/56909648/split-async-method-into-two-for-code-analysis/56909963#56909963
```
martin-strecker-sonarsource, line 3: Use the link from the "share" button:
```suggestion
The rule is controversial and was thoroughly and correctly analyzed on [StackOverflow](https://stackoverflow.com/a/56909963). As a consequence, we removed it from SonarWay.
```

martin-strecker-sonarsource (changes requested):

martin-strecker-sonarsource (approved):
LGTM.

PR #1519, opened 2023-01-25
Modify S6105: Move to progressive education format (APPSEC-422)

egon-okerman-sonarsource:
[AppSec POC PR#96](https://github.com/SonarSource/appsec-poc/pull/96)
[APPSEC-422](https://sonarsource.atlassian.net/browse/APPSEC-422)

[APPSEC-422]: https://sonarsource.atlassian.net/browse/APPSEC-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rules/S6105/impact.adoc
```diff
@@ -0,0 +1,19 @@
+=== What is the potential impact?
+
+An attacker can use this vulnerability to redirect a user from a trusted domain to a malicious domain controlled by the attacker. At that point, the attacker can perform various malicious actions, such as phishing.
```
gaetan-ferry-sonarsource, line 3:
```suggestion
An attacker can use this vulnerability to redirect a user from a trusted domain to a malicious domain controlled by the attacker. At that point, the attacker can perform various attacks, such as phishing.
```
To avoid "malicious" repetition.

rules/S6105/impact.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+Below are some scenarios that illustrate some impacts of an attacker exploiting the vulnerability.
+
+==== Phishing
+
+Suppose the attacker creates a malicious website that mirrors the interface of the trusted website. In that case, they can use the open redirect vulnerability to lead the user to this malicious site.
+
+Due to the similarity in the application appearance and the supposedly trustable hyperlink, the user struggles to identify that they are browsing on a malicious domain. From here, an attacker can capture the user's credentials, bypass Multi-Factor Authentication (MFA), and take over the user's account on the trusted website.
```
gaetan-ferry-sonarsource, line 11:
```suggestion
Due to the similarity in the application appearance and the supposedly trustable hyperlink, the user fails to identify that they are browsing on a malicious domain. From here, an attacker can capture the user's credentials, bypass Multi-Factor Authentication (MFA), and take over the user's account on the trusted website.
```
"struggle" sounds a bit strange here IMO

rules/S6105/impact.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+==== Malware distribution
+
+By leveraging the domain mirroring technique explained above, the attacker could also create a website that hosts malware. A user who is unaware of the redirection from a trusted website to this malicious website might then download and execute the attacker's malware. In the worst case, this can lead to a complete system compromise for the user.
+
+==== JavaScript injection (XSS)
+
+In certain circumstances, an attacker can use DOM-based open redirection to execute JavaScript code through the ``javascript:`` pseudo-protocol. This can lead to further exploitation in the trusted domain and has consequences such as the compromise of the user's account.
```
gaetan-ferry-sonarsource, line 19: As discussed previously, the `javascript:` handler is not required in general. Redirecting to a `data:` URL can also allow to turn the redirection into an XSS.

rules/S6105/javascript/how-to-fix-it/dom.adoc
```diff
@@ -0,0 +1,32 @@
+=== How to fix it in the DOM
```
gaetan-ferry-sonarsource, line 1: I don't think this is accepted by the RSPEC definition. Here "the DOM" should appear in the allowed frameworks list, which is not currently the case. This causes the adoc validation to fail in the pipeline.
However, I don't think it would be satisfying to add "the DOM" as an allowed framework. I don't think it would make much sense. I would replace with something more generic like "Web APIs" for example (https://developer.mozilla.org/en-US/docs/Web/API)
```suggestion
=== How to fix it in Web APIs
```

rules/S6105/javascript/how-to-fix-it/dom.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+include::../../common/fix/code-rationale.adoc[]
+
+==== Noncompliant code example
+
+The following example is vulnerable to open redirection through the following URL: `\https://example.com/redirect?url=https://evil.com`;
```
gaetan-ferry-sonarsource, line 7: I feel this is mostly a repetition of the code rationale. I would personally not include it here, but feel free to ignore.

rules/S6105/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,9 @@
+=== How does this work?
+
+Built-in framework methods should be preferred as these often provide additional security mechanisms. Usually, these built-in methods are engineered for internal page redirections. Thus, they might not solve the reader's use case.
```
gaetan-ferry-sonarsource, line 3: I think this part mostly applies to the server-side version of the vulnerability. Even in that case, I find it obscure what "Built-in framework methods" are. Moreover, while "internal page redirections" might make sense from a server-side point of view, I feel it is less relevant on the client-side.
Also, this is not something that is showcased in the code examples.

rules/S6105/common/fix/how-does-this-work.adoc
```diff
@@ -0,0 +1,9 @@
...
+
+In case the application strictly requires external redirections based on user-controllable data, the following should be done instead:
+
+1. Validating the `authority` part of the URL against a statically defined value (see Pitfalls.)
+2. Using an allowlist approach in case the destination URLs are multiple but limited.
+3. Adding a customized page to which users are redirected, warning about the imminent action and requiring manual authorization to proceed.
```
gaetan-ferry-sonarsource, line 9: I don't think this is relevant on the client side. Maybe a JavaScript-powered confirmation dialog is more pertinent in this context.

```suggestion
3. Adding a dynamic confirmation dialog, warning about the imminent action and requiring manual authorization to proceed to the actual redirection.
```

rules/S6105/common/pitfalls/starts-with.adoc
```diff
@@ -0,0 +1,7 @@
+==== The trap of ``String.startsWith`` and equivalents
+
+When validating untrusted URLs by checking if they start with a trusted scheme and authority pair `scheme://authority`, **ensure that the validation string contains a path separator `/` as the last character**.
```
gaetan-ferry-sonarsource, line 3:
```suggestion
When validating untrusted URLs by checking if they start with a trusted scheme and authority pair `scheme://authority`, **ensure that the validation string contains a `/` path separator as the last character**.
```
```suggestion
When validating untrusted URLs by checking if they start with a trusted scheme and authority pair `scheme://authority`, **ensure that the validation string contains a path separator character (i.e a '/') as the last character**.
```

rules/S6105/common/pitfalls/starts-with.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+If the validation string does not contain a terminating path separator, the Open Redirect vulnerability remains; only the exploitation technique changes. A validation such as `redirectUrl.startsWith("https://example.com")` or an equivalent with the regex `^https://example\.com.*` can be exploited with the following URL: `\https://example.com.malicious.io`.
```
gaetan-ferry-sonarsource, line 5: I don't really like this paragraph, it feels a bit too Mr Evil Hacker powered, showcasing the internals of the attack. I would go for something more formal.

```suggestion
Indeed, when the terminating path character is not present, attackers may be able to register a specific domain name that will contain the validation string as a prefix. In particular, the subdomain part of a given domain can be set to match the full domain name of the validation while still being attacker controlled.
For example, the domain `example.com.evil.example.org` will pass a prefix-based validation for domain `example.com` while located under the malicious `evil.example.org` one.
```

The suggestion is a bit messy but you probably see what I mean :)

rules/S6105/rationale.adoc
```diff
@@ -0,0 +1,5 @@
+Open redirection occurs when an application uses user-controllable data to redirect users to a URL.
```
gaetan-ferry-sonarsource, line 1:
```suggestion
Open redirection occurs when an application uses user-controllable data to build URLs used during redirects.
```

gaetan-ferry-sonarsource (changes requested):
Parts of the content seems a bit too server-side related. I think they could be made a bit more relevant to the client-side nature of DOM XSS.
Also an issue with allowed framework that prevent the pipeline to pass.
And a few rewording suggestion here and there.

rules/S6105/javascript/how-to-fix-it/dom.adoc
egon-okerman-sonarsource: It mostly serves as an example for the code sample below, so I'd prefer to keep it.

rules/S6105/javascript/how-to-fix-it/dom.adoc
egon-okerman-sonarsource: I spent some time thinking about what the best term would be. Though "Web APIs" makes sense, I feel a bit uncomfortable as:

* it is rather generic, vs things like Node.js, Express.js being straightforward.
* this rule only concerns `Document.location`, which is part of the Document Object Model.

However, I fully agree that "How to fix it in the DOM" can be improved. Perhaps "DOM API" would be a good middle ground?

rules/S6105/common/fix/how-does-this-work.adoc
egon-okerman-sonarsource: I agree that it is not demonstrated and perhaps is better left out. For the record, I wanted to imply that SPA frameworks such as React or Vue should be preferred over using your own client-side redirection.

rules/S6105/javascript/how-to-fix-it/dom.adoc
gaetan-ferry-sonarsource: Works for me

rules/S6105/common/pitfalls/starts-with.adoc
egon-okerman-sonarsource: Haha got it, I'll rewrite it

rules/S6105/common/fix/how-does-this-work.adoc
gaetan-ferry-sonarsource: Makes sense. Maybe something like this then:
```suggestion
Most client-side frameworks, such as `Vue.js` or `React.js`, provide built-in redirection methods. Those should be preferred as they often provide additional security mechanisms. However, these built-in methods are usually engineered for internal page redirections. Thus, they might not solve the reader's use case.
```

egon-okerman-sonarsource: Exactly! Thank you

gaetan-ferry-sonarsource (approved):
LGTM!

PR #1527, opened 2023-01-30
Modify rule S6390: Change text to the progressive education format (APPSEC-427)

gaetan-ferry-sonarsource:


rules/S6390/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+----
+
+=== How does this work?
+
+In most cases, it is discouraged to define a thread suspension time from
+user-input.
+
+If really necessary, the application should ensure that the provided suspension
```
egon-okerman-sonarsource, line 38:
```suggestion
If necessary, the application should ensure that the provided suspension
```

rules/S6390/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,48 @@
...
+time is below a safe limit. This one should be evaluated and set to the lowest
+possible time to ensure the application's operation and limit denial of service
+attacks.
```
egon-okerman-sonarsource, line 41:
```suggestion
time is below a safe limit. Such a limit should be evaluated and set to the lowest
possible time that ensures the application's operation and restricts denial of service
attacks.
```

egon-okerman-sonarsource (approved):
I've added two small suggestions on how to word some sentences better, but regardless of those changes LGTM :)

rules/S6390/java/how-to-fix-it/java-se.adoc
gaetan-ferry-sonarsource: I would prefer living the really here, even if grammarly does not agree. I really want to emphasize that you should **REALLY** consider not doing.

PR #1528, opened 2023-01-31
S5856: Add C# and VB.NET

pavel-mikula-sonarsource:
https://github.com/SonarSource/sonar-dotnet/issues/6691


rules/S5856/csharp/rule.adoc
```diff
@@ -0,0 +1,37 @@
+include::../description.adoc[]
+
+== Noncompliant Code Example
+
+[source,csharp]
+----
+public void DoSomething(string input)
+{
+    var regex = new Regex("[A");                // Noncompliant
```
martin-strecker-sonarsource, line 9: This already raises the Roslyn warning [RE0001](https://github.com/dotnet/roslyn/blob/main/src/Features/Core/Portable/EmbeddedLanguages/RegularExpressions/LanguageServices/AbstractRegexDiagnosticAnalyzer.cs). See e.g. [sharplab.io](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQAwAIpwHQBUCmAHgC7YBK+A5gK4A2AhgE4CihADo/gM5cCWA9gDsuAbmTIoAZkwx0AYXQBvZOlWZpUACzoAIvwDK/ALb5iAC16DKACiwZLbasQCUKtcqRqv6AProAvOiC+ADu6BSURNYARADaAILRziKqAPSp6CFMgpaU4cyohXAg4VRE6Lw81PglAKqCxPiMRpb0jcDosQC66FymbqoAvsiDQA===). Do we really want to reimplement that compiler warning?
I don't know if the warning is on by default or whether users needs to opt-in. This requires some investigation.

rules/S5856/csharp/rule.adoc
```diff
@@ -0,0 +1,37 @@
...
+    var match = Regex.Match(input, "[A");       // Noncompliant
```
martin-strecker-sonarsource, line 10: The other languages used a variety of different faulty regexes. Maybe we can spice the samples up a bit by doing the same.

martin-strecker-sonarsource: Corniel also added a test case where the "NoBacktracking" option was specified and the regex could not be evaluated without back tracking. I think this is an interesting case to add:
https://github.com/SonarSource/sonar-dotnet/blob/cac476984808fceb10aaae151540d015d9af83dc/analyzers/tests/SonarAnalyzer.UnitTest/RegularExpressions/RegexTreeTest.cs#L30-L34

martin-strecker-sonarsource (changes requested):
This basically duplicates RE0001 from Roslyn. Do we want to do this?

rules/S5856/csharp/rule.adoc
pavel-mikula-sonarsource: Yes, we want. The primary purpose of this effort is to put the scaffolding for regex rules in place. And implement 3 rather trivial rules. Later, we will embed the Java regex engine for more valuable rules.

Do you know when RE00001 was added and activated by default?

rules/S5856/csharp/rule.adoc
pavel-mikula-sonarsource: I would keep the regexes themselves simple, just to demonstrate the rule behavior. We can't cover much anyway

rules/S5856/csharp/rule.adoc
martin-strecker-sonarsource: > Later, we will embed the Java regex engine for more valuable rules.

Oh really? There is the [embedded language parser in Roslyn](https://github.com/dotnet/roslyn/blob/main/src/Features/Core/Portable/EmbeddedLanguages/RegularExpressions/RegexParser.cs) which is based on the [runtime regex parser](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs). Both of these are internal, though. [There are quite some differences](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options?view=netframework-4.8#ecmascript-matching-behavior) in the implementation of the .Net regex engine with respect to other regex engines.
Shouldn't we try to reuse these instead of embedding some java stuff?

martin-strecker-sonarsource: > Do you know when RE00001 was added and activated by default?

It was added late 2018 in VS 2019 preview: https://devblogs.microsoft.com/dotnet/visual-studio-2019-net-productivity/ I don't know much about the warning itself and there isn't much [documentation about it](https://github.com/dotnet/roslyn/blob/1c0f249e5405133f01ceb6ee0a1495400ab87db3/src/Features/RulesMissingDocumentation.md).

The rule is defined here: 
https://github.com/dotnet/roslyn/blob/bb11cf445557b977a26fcd10872553cd0a47a30b/src/Features/Core/Portable/EmbeddedLanguages/RegularExpressions/LanguageServices/AbstractRegexDiagnosticAnalyzer.cs

The interesting part reads like this:
```cs
var option = context.GetIdeAnalyzerOptions().ReportInvalidRegexPatterns;
    if (!option)
        return;
```

In [RulesMissingDocumentation.ms](https://github.com/dotnet/roslyn/blob/1c0f249e5405133f01ceb6ee0a1495400ab87db3/src/Features/RulesMissingDocumentation.md) it is placed next to the IDE warnings. Given these two indicators, I would say that this rule pretty much behaves like the other IDE rules.

pavel-mikula-sonarsource: We'll either cross-compile it (and they'll have to deal with .NET specific behavior, as they do for existing languages), or rewrite it on our side. 

In any case, we want to start somewhere.

PR #1533, opened 2023-02-01
Modify rule S3937: Add Java exception and improve rule description (SONARJAVA-4327)

johann-beleites-sonarsource:


kaufco (changes requested):
The validate_links check failed. I did rerun it in case this was only a network problem, but failed again. Could you please have a look?

johann-beleites-sonarsource:
> The validate_links check failed. I did rerun it in case this was only a network problem, but failed again. Could you please have a look?

Looking at the failing links, I think it is fine not to fix them in the scope of this PR.

PR #1534, opened 2023-02-01
S2198: Add C#

gregory-paidis-sonarsource:
Rule implementation: https://github.com/SonarSource/sonar-dotnet/issues/6658

### DIFF
--- a/rules/S2198/csharp/rule.adoc
+++ b/rules/S2198/csharp/rule.adoc
@@ -2,10 +2,10 @@ Certain mathematical comparisons will always return the same value, and should s
 
 These comparisons will always return ``++false++``:
 
-* comparing a ``++char++`` with a numeric constant that's outside the ``++char++`` range
-* comparing a ``++float++`` with a numeric constant that's outside the ``++float++`` range
-* comparing a ``++long++`` with a numeric constant that's outside the ``++long++`` range
-* comparing a ``++ulong++`` with a numeric constant that's outside the ``++ulong++`` range
+* comparing a `char` with a numeric constant that's outside the `char` range
+* comparing a `float` with a numeric constant that's outside the `float` range
+* comparing a `long` with a numeric constant that's outside the `long` range
+* comparing a `ulong` with a numeric constant that's outside the `ulong` range
 
 == Noncompliant Code Example
 
--- a/rules/S2198/csharp/rule.adoc
+++ b/rules/S2198/csharp/rule.adoc
@@ -1,19 +1,18 @@
 Certain mathematical comparisons will always return the same value, and should simply not be made. 
 
-These comparisons will always return ``++false++``:
+These comparisons will always return `false`:
 
-* comparing a `char` with a numeric constant that's outside the `char` range
-* comparing a `float` with a numeric constant that's outside the `float` range
-* comparing a `long` with a numeric constant that's outside the `long` range
-* comparing a `ulong` with a numeric constant that's outside the `ulong` range
+* Comparing a `char` with a numeric constant that is outside of the range of `char`.
+* Comparing a `float` with a numeric constant that is outside of the range of `float`.
+* Comparing a `long` with a numeric constant that is outside of the range of `long`.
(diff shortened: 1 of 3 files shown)


rules/S2198/csharp/rule.adoc
```diff
@@ -1,4 +1,21 @@
-include::../rule.adoc[]
+Certain mathematical comparisons will always return the same value, and should simply not be made. 
+
+These comparisons will always return ``++false++``:
+
+* comparing a ``++char++`` with a numeric constant that's outside the ``++char++`` range
+* comparing a ``++float++`` with a numeric constant that's outside the ``++float++`` range
+* comparing a ``++long++`` with a numeric constant that's outside the ``++long++`` range
+* comparing a ``++ulong++`` with a numeric constant that's outside the ``++ulong++`` range
```
martin-strecker-sonarsource, line 8:
```suggestion
These comparisons will always return ``++false++``:

* comparing a `char` with a numeric constant that's outside the `char` range
* comparing a `float` with a numeric constant that's outside the `float` range
* comparing a `long` with a numeric constant that's outside the `long` range
* comparing a `ulong` with a numeric constant that's outside the `ulong` range
```

martin-strecker-sonarsource (changes requested):

rules/S2198/csharp/rule.adoc
```diff
@@ -1,4 +1,21 @@
-include::../rule.adoc[]
+Certain mathematical comparisons will always return the same value, and should simply not be made. 
+
+These comparisons will always return ``++false++``:
```
martin-strecker-sonarsource, line 3:
```suggestion
These comparisons will always return `false`:
```

rules/S2198/csharp/rule.adoc
```diff
@@ -1,4 +1,21 @@
...
+
+* comparing a `char` with a numeric constant that's outside the `char` range
```
martin-strecker-sonarsource, line 5: The convention is to start with an upper letter and to end with a dot:
```suggestion
* Comparing a `char` with a numeric constant that's outside the `char` range.
```

rules/S2198/csharp/rule.adoc
```diff
@@ -1,4 +1,21 @@
...
+* comparing a `long` with a numeric constant that's outside the `long` range
+* comparing a `ulong` with a numeric constant that's outside the `ulong` range
+
+== Noncompliant Code Example
+
+[source,text]
+----
+float f = 42.0f;
+const double d = double.MaxValue;
+if (f <= d) { } // Noncompliant
```
martin-strecker-sonarsource, line 16: Simpler:
```suggestion
float f = 42.0f;
if (f <= double.MaxValue) { } // Noncompliant
```

martin-strecker-sonarsource (changes requested):

rules/S2198/csharp/rule.adoc
martin-strecker-sonarsource, line 5: Maybe better like so?
```suggestion
* Comparing a `char` with a numeric constant that is outside of the range of `char`.
```

rules/S2198/csharp/rule.adoc
```diff
@@ -1,4 +1,20 @@
-include::../rule.adoc[]
+Certain mathematical comparisons will always return the same value, and should simply not be made. 
+
+These comparisons will always return `false`:
```
martin-strecker-sonarsource, line 3:
```suggestion
These comparisons will always return the same value (either always `true` or always `false` depending on the kind of comparison):
```

martin-strecker-sonarsource (changes requested):
"always false" is not correct.

martin-strecker-sonarsource (approved):
LGTM. Please merge once 🟢 .

arseniy-sonar:
Hi @gregory-paidis-sonarsource 
Please, follow [the guideline](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule) for the commit and the PR summary next time.

PR #1536, opened 2023-02-01
Modify rule S3553: Add exception (SONARJAVA-4403)

johann-beleites-sonarsource:


johann-beleites-sonarsource:
Some links are failing - I had a look and I think it is okay to not fix them within the scope of this PR.

alban-auzeill (approved):

PR #1551, opened 2023-02-06
Modify S1656: Update quickfix status (SONARJAVA-3959)

irina-batinic-sonarsource:


rules/S1656/java/metadata.json
```diff
@@ -6,5 +6,6 @@
     "CERT": [
       "MSC12-C."
     ]
-  }
+  },
+  "quickfix": "partial"
```
chrislain-razafimahefa-sonarsource, line 10: Why is it 'partial'? By looking quickly at the code, you have covered all the possible issues, no?

chrislain-razafimahefa-sonarsource (changes requested):
I am not sure why the quickfix is 'partial'.

rules/S1656/java/metadata.json
irina-batinic-sonarsource: The case where we have: `int b = a = a = a;` is not covered, the quick fix is not offered. It is out of the scope of this ticket because the expression needs to be simplified first.

chrislain-razafimahefa-sonarsource (approved):
Well done!

PR #1564, opened 2023-02-08
Modify rule S6191 CPP-3876 Update the RSPEC with de-deprecation of these uses in C++23

loic-joly-sonarsource:


rules/S6191/cfamily/rule.adoc
```diff
@@ -1,7 +1,6 @@
-The behavior of volatile types in compound expressions that both read and write the same variable (for instance ``a{plus}{plus}`` or ``++a+=2++``, where ``++a++`` is a ``++volatile int++``) is not well defined, and is deprecated since {cpp}20.
+When a variable appears in a compound expression (for instance ``a{plus}{plus}`` or `a+=2`), the variable is accessed twice. This detail only matters for `volatile` variables, where any access is important. And it may not be obvious to all developers that two accesses happen while the variable is only named once. The standard was not explicit on this topic up until {cpp}23.
```
phil-nash-sonarsource, line 1: a. Is it worth expanding on "where any access is important"?
b. "The standard was not explicit on this topic up until C++23" - surely it was explicit in C++20, just in a different direction. Would it be clearer to acknowledge the C++20 perspective to clarify that this has been refined in 23?

phil-nash-sonarsource (commented):
Overall this reads well. The couple of comments I added, inline, are open to question and not, necessarily, showstoppers.

loic-joly-sonarsource:
@phil-nash-sonarsource: I tried a more verbose approach, that explains volatile in more detail. Let me know if you think it's any better.

rules/S6191/cfamily/rule.adoc
```diff
@@ -1,7 +1,12 @@
...
+In C++, it usually does not matter how many times you access a variable, as long as the variable value is the right one. This is not the case when the variable is in a memory region that is mapped to external hardware. In that case, for instance, several successive reads can yield different values (if the memory is updated by the hardware in-between) and several writes of the same value may be significant (some hardware trigger events each time a memory location is written to). 
+
+To specify that every read and write has an impact outside of the abstract machine of the language, it is possible to qualify the access to a variable as `volatile`. In that case, the program must perform all reads and writes that are specified, without optimizing anything away.
+
+When a variable appears in a compound expression (for instance ``a{plus}{plus}`` or `a+=2`), the variable is accessed twice. This detail only matters for `volatile` variables. And it may not be obvious to all developers that two accesses happen while the variable is only named once. The standard was not explicit on this topic up until {cpp}23.
```
philsquared, line 5: I think this paragraph should still be first. That way if you know the whole context you'll get what you need straight away. If it is not convincing on its own (because you're not so familiar with how volatile variables work) then reading on (ie the preceding two paragraphs here) nicely fills it out.

rules/S6191/cfamily/rule.adoc
```diff
@@ -1,7 +1,12 @@
-When a variable appears in a compound expression (for instance ``a{plus}{plus}`` or `a+=2`), the variable is accessed twice. This detail only matters for `volatile` variables, where any access is important. And it may not be obvious to all developers that two accesses happen while the variable is only named once. The standard was not explicit on this topic up until {cpp}23.
+In C++, it usually does not matter how many times you access a variable, as long as the variable value is the right one. This is not the case when the variable is in a memory region that is mapped to external hardware. In that case, for instance, several successive reads can yield different values (if the memory is updated by the hardware in-between) and several writes of the same value may be significant (some hardware trigger events each time a memory location is written to). 
+
+To specify that every read and write has an impact outside of the abstract machine of the language, it is possible to qualify the access to a variable as `volatile`. In that case, the program must perform all reads and writes that are specified, without optimizing anything away.
```
philsquared, line 3: second part of first sentence could be: "access to a variable may be qualified as `volatile`"

philsquared (changes requested):
Overall that reads great! I did have a small concern that it now takes a while to get to the meat of it by expanding on the background that the reader may already be familiar with. I think moving that third paragraph back to the start, as suggested inline, should be enough to counter that. There was one more comment along the lines of "omit needless words", too.

PR #1570, opened 2023-02-13
Modify rule S2445: Add C# description and samples

antonioaversa:
Implementation [#6750](https://github.com/SonarSource/sonar-dotnet/issues/6701)

Kept as draft as it is to be merged after [the implementation](https://github.com/SonarSource/sonar-dotnet/pull/6750).

## Narrowed scope for C#
The definition of the rule for C# has been narrowed during the review. 
Outcome of the offline discussion:
- it may make sense to have a lock on a field of a class different from the one where the lock is acquired, for example in a situation where locking objects are externalized and shared among multiple classes, for inter-class synchronization (e.g. a serial port lock);
- so, requiring the `readonly` field to also be `private` appears in the end too constraining;
- it also makes the rule too complex and non-atomic, and doesn't allow the user to opt in one behavior (lock on a `readonly` object), opting out the other (lock on a non-`private` object);
- you may also want to have it `protected`, instead of `private`, to allow methods of a derived type to synchronize with methods of the base type, and you don't want to be forced to use properties to do that.

For these reasons, we are changing the rule and narrowing its scope for C#: when locking on a field, that field should be `readonly`. Whether it's private or not, it won't matter for the implementation of this rule.


zsolt-kolbay-sonarsource (approved):
LGTM
One note: shouldn't this be applied to VB.NET as well? Locking works there exactly the same way as it does in C# (with the `SyncLock `keyword), and it seems like a simple rule.

antonioaversa:
@zsolt-kolbay-sonarsource 
> LGTM One note: shouldn't this be applied to VB.NET as well? Locking works there exactly the same way as it does in C# (with the `SyncLock `keyword), and it seems like a simple rule.
While only C# is mentioned in the task, it surely could. If you and @mary-georgiou-sonarsource both agree, I would invest some time extending the implementation to VB.NET as well.

mary-georgiou-sonarsource:
> @zsolt-kolbay-sonarsource
> 
> > LGTM One note: shouldn't this be applied to VB.NET as well? Locking works there exactly the same way as it does in C# (with the `SyncLock `keyword), and it seems like a simple rule.
> 
> While only C# is mentioned in the task, it surely could. If you and @mary-georgiou-sonarsource both agree, I would invest some time extending the implementation to VB.NET as well.

Di

antonioaversa:
> > @zsolt-kolbay-sonarsource
> > > LGTM One note: shouldn't this be applied to VB.NET as well? Locking works there exactly the same way as it does in C# (with the `SyncLock `keyword), and it seems like a simple rule.
> > 
> > 
> > While only C# is mentioned in the task, it surely could. If you and @mary-georgiou-sonarsource both agree, I would invest some time extending the implementation to VB.NET as well.
> 
> In case this rule is valid also for VB.NET (I haven't checked) - IMO, it's better to create another issue regarding the VB.NET version of the rule and implement the C# part in way that the implementation can be re-used for the VB.NET (by putting the common code in a base class).

Reporting the outcome of the discussion offline on the possibility of extending the rule to VB.NET.

We are going to have a dedicated GitHub issue to implement this rule for VB.NET, rather than extending this GitHub issue.

We are going to stick with the C#-specific implementation for the time being, since the effort to refactor the code into a base class is possibly 90% or more, of the effort to implement the [VB.NET](http://vb.net/) rule.

rules/S2445/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+
+The story is very similar for locking on a local variable or on a new instance; two different threads running the method in parallel would lock on two different object instances, completely undermining the synchronization.
+
+
+== Noncompliant Code Example
+
+[source,csharp]
+----
+private string color = "red";
```
mary-georgiou-sonarsource, line 11: I would use another type of object (e.g. a silly fake object ```new Color("red")```) - so that we can add also a simple example with a string object (interned string case).

rules/S2445/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,47 @@
...
+== Compliant Solution
+
+[source,csharp]
+----
+private string color = "red";
+private readonly object lockObj = new object();
+
+private void DoSomething()
+{
+  lock (lockObj)
```
mary-georgiou-sonarsource, line 38:
```suggestion
  // synchronizing access to "color"
  lock (lockObj) 
```

rules/S2445/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,47 @@
...
+  lock (color) // Noncompliant; lock is actually on object instance "red" referred to by the color variable
+  {
+    //...
+    color = "green"; // other threads now allowed into this block
+    // ...
+  }
+  lock (new object()) // Noncompliant this is a no-op.
+  {
+     // ...
+  }
```
mary-georgiou-sonarsource, line 24: please add also an example with a `string` lock where the instance is an interned string.

mary-georgiou-sonarsource (changes requested):
Round 1:
- I'd update the description to match better the C# case.
- Add in the example the case of the interned string
- Please add educational links in the RSPEC. For example: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock and https://learn.microsoft.com/en-us/dotnet/api/system.string.intern?view=net-7.0

rules/S2445/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+    "title": "Blocks should be synchronized on \"private readonly\" fields"
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: I like the current title more. It's to the point and reasonably short.

rules/S2445/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,47 @@
+Locking on a class field synchronizes not on the field itself, but on the object assigned to it. So locking on a non-``++readonly++`` field makes it possible for the field's value to change while a thread is in a block locked on the old value. That would allow a second thread, locked on the new value, to enter the block at the same time.
```
pavel-mikula-sonarsource, line 1:
```suggestion
Locking on a class field synchronizes not on the field itself, but on the object assigned to it. So locking on a non-`readonly` field makes it possible for the field's value to change while a thread is in a block locked on the old value. That would allow a second thread, locked on the new value, to enter the block at the same time.
```

rules/S2445/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+
+The story is very similar for locking on a local variable or on a new instance; two different threads running the method in parallel would lock on two different object instances, completely undermining the synchronization.
+
+
```
pavel-mikula-sonarsource, line 6: A paragraph about string interning is a possible message but is missing in the description.

rules/S2445/csharp/rule.adoc
```diff
@@ -1,12 +1,14 @@
-include::../rule.adoc[]
+include::rule-except-see.adoc[]
```
pavel-mikula-sonarsource, line 1: This doesn't need to say what it doesn't have. As long as it has description, we're fine.
```suggestion
include::rule.adoc[]
```

rules/S2445/csharp/rule-except-see.adoc
pavel-mikula-sonarsource, line 1: This content should be in `rule.adoc`

rules/S2445/rule.adoc
```diff
@@ -1,3 +1 @@
-include::rule-except-see.adoc[]
```
pavel-mikula-sonarsource, line 1: This file should be deleted

rules/S2445/csharp/rule.adoc
```diff
@@ -1,12 +1,14 @@
...
+include::../see.adoc[]
 
 ifdef::env-github,rspecator-view[]
 
 '''
 == Implementation Specification
 (visible only on this page)
 
-include::../message.adoc[]
+include::message.adoc[]
```
pavel-mikula-sonarsource, line 11: The ../highlighting.adoc is not C#. Either split it, or update it to be language agnostic

rules/S2445/java/rule.adoc
```diff
@@ -1,4 +1,4 @@
-include::../rule-except-see.adoc[]
+include::rule-except-see.adoc[]
```
pavel-mikula-sonarsource, line 1: This extraction doesn't make much sense. The content should be here

pavel-mikula-sonarsource (changes requested):
Cleanup

rules/S2445/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,47 @@
+Locking on a class field synchronizes not on the field itself, but on the object assigned to it. So locking on a non-``++readonly++`` field makes it possible for the field's value to change while a thread is in a block locked on the old value. That would allow a second thread, locked on the new value, to enter the block at the same time.
+
+
+The story is very similar for locking on a local variable or on a new instance; two different threads running the method in parallel would lock on two different object instances, completely undermining the synchronization.
+
```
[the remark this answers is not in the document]
antonioaversa: I have rewritten the description entirely, to also take into account special cases of new instances such as string interning.

A small difference with the proposal above is that I haven't mentioned the "access to a shared resource": one may lock on an object just to impose serializability, not necessarily to access a shared resource. 

Sometimes the shared resource is there but hidden behind an abstraction (`Console` with multiple `AppDomain`, db, ...). Sometimes there may be multiple shared resources for a single lock. 
In general, I don't see a 1-to-1 mapping between locking and resource access.

rules/S2445/csharp/rule-except-see.adoc
antonioaversa: Changed into `Synchronized access via "color"`, to mark the difference between the critical section being synchronized and the object which helps to do so and avoid confusion with synchronization in other languages, where those are syntactically in 1-to-1 relation (namely Java).

rules/S2445/csharp/rule-except-see.adoc
antonioaversa: Added.

rules/S2445/csharp/rule.adoc
antonioaversa: Recap of the restructure made:
- content of `csharp\rule-except-see.adoc` moved at the beginning of `csharp\rule.adoc`
- `csharp\rule-except-see.adoc` removed
- content of `java\rule-except-see.adoc` moved at the beginning of `java\rule.adoc`
- `java\rule-except-see.adoc` removed
- highlighting split for Java and C#
- I have left the `rule.adoc` at root level, including `see.adoc`, although it is not used anymore by C# nor Java
- the name of the rule at root level is still the Java one

rules/S2445/csharp/rule-except-see.adoc
antonioaversa: Check https://github.com/SonarSource/rspec/pull/1570#discussion_r1114263403

rules/S2445/rule.adoc
antonioaversa: Check https://github.com/SonarSource/rspec/pull/1570#discussion_r1114263403

rules/S2445/csharp/rule.adoc
antonioaversa: Check https://github.com/SonarSource/rspec/pull/1570#discussion_r1114263403

rules/S2445/java/rule.adoc
antonioaversa: Check https://github.com/SonarSource/rspec/pull/1570#discussion_r1114263403

antonioaversa:
@pavel-mikula-sonarsource @mary-georgiou-sonarsource I addressed your comments and I am reassigning this PR back to you, although GitHub Checks are currently failing.

The issue with the GitHub Checks seems to come from https://next.sonarqube.com/sonarqube/batch/index returning 503, rather than by actual issues with the adoc content itself.

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,72 @@
...
+Locking on a class field synchronizes not on the field itself, but on the object assigned to it. So locking on a non-``readonly`` field makes it possible for the field's value to change while a thread is in a block locked on the old value. That would allow a second thread, locked on the new value, to enter the block at the same time.
+
+
+Locking on a ``readonly`` field of a class which is not ``private`` allows external code to lock the field, potentially interfering with synchronization of methods in that class. 
+
+
+Locking on a local variable or on a new instance undermines the synchronization: two different threads running the same method in parallel would lock on two different object instances. That would allow a second thread, locked on that new value, to enter the same block concurrently.
+
+
+Locking on a string literal is even more dangerous: depending on whether the string is interned or not, different threads may or may not synchronize on the same object instance.
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: What is not clear to me is in point 3 - why when locking on a local variable we have the same issues with locking on a new object. This might need to be clarified a bit as from this text is not clear to me.

rules/S2445/java/rule.adoc
```diff
@@ -1,4 +1,43 @@
-include::../rule-except-see.adoc[]
+Synchronizing on a class field synchronizes not on the field itself, but on the object assigned to it. So synchronizing on a non-``++final++`` field makes it possible for the field's value to change while a thread is in a block synchronized on the old value. That would allow a second thread, synchronized on the new value, to enter the block at the same time.
```
mary-georgiou-sonarsource, line 1: why this file is changed? 
It seems unrelated to our PR - could you please revert?

mary-georgiou-sonarsource (changes requested):
- Refining a bit more the description 
- One file probably needs reverting
- Please add the educational links

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,72 @@
...
+{
+  lock (lockObj) 
+  {
+    //...
+    color = new Color("green");
+    // ...
+  }
+}
+----
+
```
mary-georgiou-sonarsource, line 58: I think we can add some educational links at the end of the description like:
 https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock and https://learn.microsoft.com/en-us/dotnet/api/system.string.intern?view=net-7.0


See [this rule](https://sonarsource.github.io/rspec/#/rspec/S6419) on where and how to add them/

rules/S2445/java/rule.adoc
antonioaversa: Compare https://github.com/SonarSource/rspec/pull/1570#discussion_r1112639616 and 
https://github.com/SonarSource/rspec/pull/1570#discussion_r1114263403

The original description at the root level was specific to Java, so I have moved it into the Java directory.

mary-georgiou-sonarsource: Missed that -thanks for the expalanation.

rules/S2445/csharp/rule.adoc
antonioaversa: Links added

rules/S2445/csharp/rule.adoc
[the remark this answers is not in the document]
antonioaversa: Description updated

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
-include::../rule.adoc[]
+Locking on a class field synchronizes not on the field itself, but on the object assigned to it. Thus, there are some good practices to follow to avoid bugs related to thread synchronization.
+
+1. Locking on a non-``readonly`` field makes it possible for the field's value to change while a thread is in the code block locked on the old value. This allows another thread to lock on the new value and access the same block concurrently.
+
+2. Locking on a ``readonly`` non-``private`` field allows, for example, a method from another class to lock on the same field, potentially interfering with the synchronization. 
+
+3. Locking on a local variable or a new instance of an object undermines the synchronization because two different threads running the same method in parallel will lock on two different instances of the same object, allowing them to access the synchronized block at the same time.
```
mary-georgiou-sonarsource, line 7: Still here it's not clear to me why locking on a local variable creates an issue.
The explanation after `or a new instance of an object undermines` seems to be related only for locking on `new Object()`.

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
-include::../rule.adoc[]
+Locking on a class field synchronizes not on the field itself, but on the object assigned to it. Thus, there are some good practices to follow to avoid bugs related to thread synchronization.
+
+1. Locking on a non-``readonly`` field makes it possible for the field's value to change while a thread is in the code block locked on the old value. This allows another thread to lock on the new value and access the same block concurrently.
+
+2. Locking on a ``readonly`` non-``private`` field allows, for example, a method from another class to lock on the same field, potentially interfering with the synchronization.
```
mary-georgiou-sonarsource, line 5:
```suggestion
2. Locking on a ``readonly`` non-``private`` field allows external code to be able to lock on the same field, potentially interfering with the synchronization.  For example a method from `Class A` can use a public object that is used for locking in another method in `Class B` and as a result we can have two unrelated methods waiting for each other.
```


IMO, synchronization is a difficult topic, so bringing also a more concrete example might help the reader to better understand the rule. WDYT?

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
...
+
+[source,csharp]
+----
+private Color colorObject = new Color("red");
+private readonly colorString = "red";
+
+private void DoSomething()
+{
+  // Synchronizing access via "colorObject"
+  lock (colorObject) // Noncompliant; lock is actually on object instance "red" referred to by the color field
```
mary-georgiou-sonarsource, line 22:
```suggestion
  lock (colorObject) // Noncompliant; lock is actually on object instance "red" referred to by the `colorObject` field
```

mary-georgiou-sonarsource (changes requested):
Just a bit more decription refinement - looks good otherwise.

rules/S2445/csharp/message.adoc
```diff
@@ -0,0 +1,12 @@
+=== Message
+
+"xxx" is not "private readonly", and should not be used for locking.
+
+"xxx" is a local variable, and should not be used for locking.
+
+Use fields from "yyy" for locking.
```
mary-georgiou-sonarsource, line 7: I see here that there's also this best practice, about choosing fields that belong to the same type that the locking is in, but we don't have anything in the description for it.

rules/S2445/csharp/rule.adoc
antonioaversa: I would say that it basically depends on how the local variable is used.

Taking a step back, the idea underlying the use of `private readonly` fields, is that their reference is guaranteed to be constant throughout the entire lifespan of the object. If, for any reason, the Garbage Collector needs to relocate such `private readonly` reference (for performance, for example), the runtime itself will guarantee that all references will be updated accordingly.

Bottom-line is: if the runtime does its job correctly, you never read twice the same field and get an inconsistent reference.

If you use a local variable your reference is on the stack, and if you are using such a reference, it's in some levels of CPU cache and/or in a register. In either case, you can alter that data (see the case of ref locals).

So indeed, while new object creation (almost, see string interning) renders synchronization useless, using a local variable _can_ render synchronization useless. 

Would it be better if I would change `undermines` into `can undermine`?

rules/S2445/csharp/highlighting.adoc
```diff
@@ -0,0 +1,4 @@
+=== Highlighting
+
+``lock (xxx)``
```
pavel-mikula-sonarsource, line 3:
```suggestion
`lock (xxx)` statement
```

rules/S2445/csharp/message.adoc
pavel-mikula-sonarsource: That sounds like a different rule

rules/S2445/csharp/message.adoc
```diff
@@ -0,0 +1,12 @@
+=== Message
+
+"xxx" is not "private readonly", and should not be used for locking.
+
+"xxx" is a local variable, and should not be used for locking.
```
pavel-mikula-sonarsource, line 5: See `Messages` section of the guideline: 
https://docs.sonarqube.org/latest/extension-guide/adding-coding-rules/
It should say what's wrong and how to fix it.

This applies to all.
```suggestion
Do not lock on a local variable "xxx", use a private readonly field instead.
```
This applies to all of them.

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
-include::../rule.adoc[]
+Locking on a class field synchronizes not on the field itself, but on the object assigned to it. Thus, there are some good practices to follow to avoid bugs related to thread synchronization.
+
+1. Locking on a non-``readonly`` field makes it possible for the field's value to change while a thread is in the code block locked on the old value. This allows another thread to lock on the new value and access the same block concurrently.
```
pavel-mikula-sonarsource, line 3: There should be a single quite - this applies everywhere
```suggestion
1. Locking on a non-`readonly` field makes it possible for the field's value to change while a thread is in the code block locked on the old value. This allows another thread to lock on the new value and access the same block concurrently.
```

rules/S2445/rule.adoc
pavel-mikula-sonarsource: If it is not used, there's no reason for this to exist. 
Currently, it's only a rename any, making it confusing.

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
...
+  lock (lockObj) 
+  {
+    //...
+    color = new Color("green");
+    // ...
+  }
+}
+----
+
+== See
```
pavel-mikula-sonarsource, line 58: There's `see.adoc` in the root of the rule. Where is it used? If nowhere, delete it.

pavel-mikula-sonarsource (changes requested):
We'll need a PR to revisit the messages on the rule. And that can also update the rspec with it.

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
...
+
+4. Locking on a string literal is also dangerous since, depending on whether the string is interned or not, different threads may or may not synchronize on the same object instance.
+
+
+== Noncompliant Code Example
+
+[source,csharp]
+----
+private Color colorObject = new Color("red");
+private readonly colorString = "red";
```
pavel-mikula-sonarsource, line 17: Invalid syntax
```suggestion
private readonly string colorString = "red";
```

rules/S2445/csharp/rule.adoc
```diff
@@ -1,14 +1,76 @@
...
+  }
+}
+----
+
+
+== Compliant Solution
+
+[source,csharp]
+----
+private Color colorObject = new Color("red");
```
pavel-mikula-sonarsource, line 44: Not aligned with it's usage below
```suggestion
private Color color = new Color("red");
```

rules/S2445/csharp/rule.adoc
mary-georgiou-sonarsource: Sounds good for `can undermine`

rules/S2445/csharp/rule.adoc
mary-georgiou-sonarsource: @antonioaversa  what do you thing about this?

rules/S2445/csharp/rule.adoc
mary-georgiou-sonarsource: resolving as Pavel suggested to rename to `color`

rules/S2445/csharp/message.adoc
mary-georgiou-sonarsource: @pavel-mikula-sonarsource 
Yes and maybe no too.
To lock on a field of a different type you are probably locking on a non-private field.
IMO it's still related.

Maybe adding something to the description though - rethinking about it - might be an overkill.

pavel-mikula-sonarsource: I would not extend the scope, as public static fields, and nested classes will make it a rabbit hole.

rules/S2445/rule.adoc
antonioaversa: I don't understand what "rename any" means.

Anyway, I have removed both `rule.adoc` and `see.adoc` at root-level, which, after changes  https://github.com/SonarSource/rspec/pull/1570#discussion_r1114263403, are indeed not used anymore.

rules/S2445/csharp/rule.adoc
antonioaversa: Changed. Used `"` instead of ````, since it is inside a code block.

rules/S2445/csharp/rule.adoc
antonioaversa: Ok for `can undermine`.

As for "different instances of the same local variable": you lock on an object in the heap by its reference, you don't lock on the reference itself. There wouldn't be space on the stack to allocate the header which is used by the `Monitor.Enter` and `Monitor.Exit`, and there would be problems when the method returns.

Maybe there's a way to lock on a local variable via a `stackalloc` or a pinned reference, but I never seen such a scenario and, if exists, it's probably very very niche.

antonioaversa: As discussed offline, changed to:
```cs
3. Locking on a local variable or a new instance of an object can undermine synchronization because two different threads running the same method in parallel will potentially lock on different instances of the same object, allowing them to access the synchronized block at the same time.
```

rules/S2445/csharp/message.adoc
mary-georgiou-sonarsource: @pavel-mikula-sonarsource I think in the same spirit also the case of interned string should be removed.
And then maybe create a new rule proposal about "Good practices with lock".

rules/S2445/csharp/rule.adoc
antonioaversa: Good for me, changed to conform to the proposal above.

rules/S2445/csharp/rule.adoc
antonioaversa: Fixed

rules/S2445/csharp/rule.adoc
antonioaversa: Fixed

rules/S2445/csharp/rule.adoc
antonioaversa: Removed.
It was referenced by `rule.adoc` at top-level. 
Now that `rule.adoc` has been removed, there is no reason for this `see.adoc` to be kept.

rules/S2445/csharp/message.adoc
antonioaversa: The format of messages was borrowed from the Java counterpart.

I have changed them in
```
Do not lock on not "private readonly" field "xxx", use a private readonly field instead.
Do not lock on local variable "xxx", use a private readonly field instead.
Use fields from "yyy" for locking.
Locking on a new instance is a no-op, use a private readonly field instead.
Strings can be interned, and should not be used for locking. Use a private readonly field instead.
```

Although seems to me at least debatable, to direct users to use `private readonly` fields, rather than private get-only property backed by a field, for cases like a local variable, a new instance or a new string instance, since we don't know and can't extrapolate about user intent.

rules/S2445/csharp/highlighting.adoc
antonioaversa: I have realized that the code puts the squiggle on the target of the `lock`, not on the `lock` statement itself.

So I have changed this to `locked object in `lock (xxx)` statement`. Makes sense?

antonioaversa:
> We'll need a PR to revisit the messages on the rule. And that can also update the rspec with it.

@pavel-mikula-sonarsource We have the already existing https://github.com/SonarSource/sonar-dotnet/pull/6750. 
Can't we use that one?

rules/S2445/csharp/message.adoc
antonioaversa: A bit of background.

The message `Use fields from "yyy" for locking.`  has been introduced during the 1st round of review of the implementation for the rule, in a discussion with @zsolt-kolbay-sonarsource : https://github.com/SonarSource/sonar-dotnet/pull/6750#discussion_r1108453102

The reasoning for introducing a custom message was the following (Zsolt: correct me if I am wrong):
- if you lock on a field of another class (e.g. `lock (AnotherClass.staticReadonlyField) { }`), the rule would tell you that `staticReadonlyField` is not `private readonly`, and to use a `private readonly` field instead
- but you can't use a `private readonly` field from `AnotherClass`, unless `AnotherClass` is your outer class, because as soon as you make `staticReadonlyField` private, you won't see anymore from your current scope
- therefore, instead of suggesting to the user something he can't possibly do, we suggest something he can do: use fields from its own class

If, in this case, we consider adherence to the rule as more important than feasibility of the suggestion, I can revert this one to the original message.
Option 2: don't trigger the rule if the `private readonly` field is in a class different than the one that contains the method with the `lock` statement. Very conservative: we would probably disable the rule where it's most needed. We would, however, not have false positive or wrong messages.

antonioaversa: Outcome of the discussion had offline on the subject reported [here](https://github.com/SonarSource/rspec/pull/1570#issue-1582097418).

rules/S2445/rule.adoc
pavel-mikula-sonarsource: typo 😒 It should have been "rename only"

rules/S2445/csharp/highlighting.adoc
pavel-mikula-sonarsource: I don't mind

rules/S2445/csharp/message.adoc
```diff
@@ -0,0 +1,10 @@
+=== Message
+
+Do not lock on non-"readonly" field "xxx", use a "readonly" field instead.
```
pavel-mikula-sonarsource, line 3: I would remove all double quotes around `readonly` word. We can live without them. The "xxx" is OK

pavel-mikula-sonarsource:
> @pavel-mikula-sonarsource We have the already existing [SonarSource/sonar-dotnet#6750](https://github.com/SonarSource/sonar-dotnet/pull/6750).
> Can't we use that one?

Yes, sure. I wasn't aware in what state the actual implementation PR is.

rules/S2445/csharp/message.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+Do not lock on local variable "xxx", use a "readonly" field instead.
+
+Locking on a new instance is a no-op, use a "readonly" field instead.
```
pavel-mikula-sonarsource, line 7: To align the messages.

This will enable us to use a single format string with varying reasons in the implementation.
```suggestion
Do not lock on a new instance because is a no-op, use a "readonly" field instead.
```

rules/S2445/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+    "title": "Blocks should be synchronized on \"readonly\" fields"
```
pavel-mikula-sonarsource, line 2: The rule was renamed in the root. As the name from the root is valid for C# too, this should not be here. Both variants are good enough

rules/S2445/csharp/message.adoc
antonioaversa: The list of messages listed here became stale after [this](https://github.com/SonarSource/rspec/pull/1570#discussion_r1118576245) and [this](https://github.com/SonarSource/rspec/pull/1570#discussion_r1118693193) discussions.

New messages are:
```
Do not lock on writable field "xxx", use a readonly field instead.
Do not lock on local variable "xxx", use a readonly field instead.
Do not lock on a new instance because is a no-op, use a readonly field instead.
Do not lock on strings as they can be interned, use a readonly field instead.
```

pavel-mikula-sonarsource (approved):
LGTM

pavel-mikula-sonarsource:
Is there a ticket created for the part that was left out? It should be in sonar-dotnet,labeled with the new rule idea

antonioaversa:
> Is there a ticket created for the part that was left out? It should be in sonar-dotnet,labeled with the new rule idea

Created https://github.com/SonarSource/sonar-dotnet/issues/6804

PR #1571, opened 2023-02-13
Modify rule S4545: Add VB.NET specifics, Move C# specifics

antonioaversa:
Implementation https://github.com/SonarSource/sonar-dotnet/issues/6703

Kept as draft as it is to be merged after https://github.com/SonarSource/sonar-dotnet/pull/6728.

Remark: implementation merged before corresponding RSPEC is ready. Changes to the sonar-dotnet repo are to be expected, when running rspec update upon release.


rules/S4545/csharp/rule-except-see.adoc
```diff
@@ -0,0 +1,35 @@
+The ``++DebuggerDisplayAttribute++`` is used to determine how an object is displayed in the debugger window.
```
zsolt-kolbay-sonarsource, line 1: AFAIK we no longer need to surround these snippets with `++`.

rules/S4545/rule.adoc
```diff
@@ -4,32 +4,7 @@ The ``++DebuggerDisplayAttribute++`` is used to determine how an object is displ
 The ``++DebuggerDisplayAttribute++`` constructor takes a single argument: the string to be displayed in the value column for instances of the type. Any text within curly braces is evaluated as the name of a field, property, or method.
```
zsolt-kolbay-sonarsource, line 4: Is the original rule.adoc file still needed?

zsolt-kolbay-sonarsource (changes requested):
Some minor comments. I'm not a fan of the name _rule-expect-see.adoc_, but I've seen it used in other rules as well, so it's just my preference.

rules/S4545/csharp/rule-except-see.adoc
antonioaversa: I've learnt that yesterday from your PR :-)

Removed.

rules/S4545/rule.adoc
antonioaversa: It's not, removed.

antonioaversa:
> Some minor comments. I'm not a fan of the name _rule-expect-see.adoc_, but I've seen it used in other rules as well, so it's just my preference.

I have restructured a bit the content:
- removed the `rule-except-see.adoc` files
- moved the content to the `rule.adoc` of each language
- fixed the `highlighting.adoc`, which was inconsistent with the rule implementation
- refined the description and align it with the implementation of the rule, which has evolved in the last few days

zsolt-kolbay-sonarsource (approved):
LGTM

rules/S4545/csharp/rule.adoc
```diff
@@ -1,4 +1,37 @@
-include::../rule.adoc[]
+The ``DebuggerDisplayAttribute`` is used to determine how an object is displayed in the debugger window.
+
+
+The ``DebuggerDisplayAttribute`` constructor takes a single mandatory argument: the string to be displayed in the value column for instances of the type. Any text within curly braces is evaluated as the name of a member, or any complex expression converted to a string value, containing method calls and operators.
```
mary-georgiou-sonarsource, line 4:
```suggestion
The `DebuggerDisplayAttribute` constructor takes a single mandatory argument: the string to be displayed in the value column for instances of the type. Any text within curly braces is evaluated as the name of a field, property, or method or as a complex expression containing method calls and operators that is evaluated to a string.
```

rules/S4545/vbnet/rule.adoc
```diff
@@ -1,4 +1,39 @@
-include::../rule.adoc[]
+The ``DebuggerDisplayAttribute`` is used to determine how an object is displayed in the debugger window.
+
+
+The ``DebuggerDisplayAttribute`` constructor takes a single mandatory argument: the string to be displayed in the value column for instances of the type. Any text within curly braces is evaluated as the name of member, or any complex expression converted to a string value, containing method calls and operators.
```
mary-georgiou-sonarsource, line 4: same proposal as above.

mary-georgiou-sonarsource (changes requested):
Generally looks good - only a minor change to the description to make it IMO clearer.

rules/S4545/message.adoc
```diff
@@ -1,4 +1,4 @@
 === Message
 
-Fix the name between the curly braces so that it matches a field, property or method of this object
+Fix the name between the curly braces so that it matches a field or property of this object
```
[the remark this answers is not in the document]
antonioaversa: Placeholder

rules/S4545/csharp/rule.adoc
antonioaversa: - Double backticks to single backticks => OK
- Member to "field, property" => OK (no "method", see [here](https://github.com/SonarSource/rspec/pull/1571#discussion_r1119027741))
- "converted to a string value" to "That is evaluated to a string" => The second seems incorrect to me since expressions such as `DebuggerDisplay("{42}")` are correctly evaluated, despite `42` not being evaluated to a string. It's rather that the expression is evaluated to whatever `object` it evaluates to, and then a `ToString()` is called on that object. In any case, I have removed the subordinate altogether.

rules/S4545/vbnet/rule.adoc
antonioaversa: Fixed in the same way it is described in https://github.com/SonarSource/rspec/pull/1571#discussion_r1119031588

rules/S4545/message.adoc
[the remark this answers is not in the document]
antonioaversa: Method re-added: a method group has a `ToString`, so it indeed works:
image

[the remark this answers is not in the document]
mary-georgiou-sonarsource: To be sure I understood: We just don't want to advice people on using something that brings probably no value while debugging?

mary-georgiou-sonarsource (approved):
LGTM

rules/S4545/message.adoc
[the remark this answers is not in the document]
antonioaversa: I realize now I have misinterpreted the screenshot above. I thought it broke the evaluation of the `DebuggerDisplayAttribute`, but it doesn't. Instead, it prints the `ToString` of the method group, which happens to have braces in it.

Anyway, yes: I don't think anybody would add the name of a method in the format string of the attribute, just to print the signature of a method. Moreover, the signature of the method is the same for all instances of an object, so I don't see any valid user scenario when the user would need it at debug time, to assess the state specific to an object.

It's much more likely that somebody wanted to write `DebuggerDisplay("{SomeMethod()}")`, instead of `DebuggerDisplay("{SomeMethod}")`, so a "complex expression containing method calls", rather than the name of a method.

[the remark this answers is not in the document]
mary-georgiou-sonarsource: Thanks for the explanation.
That's what I had in mind after your first reply in this thread.

PR #1572, opened 2023-02-14
S3398: Add C# description

zsolt-kolbay-sonarsource:
https://github.com/SonarSource/sonar-dotnet/issues/6706


rules/S3398/description.adoc
```diff
@@ -0,0 +1 @@
+When a ``++private++`` method is only invoked by an inner class, there's no reason not to move it into that class. It will still have the same access to the outer class' members, but the outer class will be clearer and less cluttered.
```
antonioaversa, line 1: Up to you actually, but I would not use the word `inner class`, but `nested class`. 

Nested is the standard way "types within types" are defined (c.p. https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/nested-types).

While calling them nested or inner doesn't make any difference in C#, it does in Java (where the rule was first implemented), where inner classes have a reference to the outer class `this` and nested classes (which are called `static class`) don't.

The example in the Java code shows that they actually meant inner, since `Outie.this.increment();` is used. So using `nested` in C# would in my opinion avoid some confusion.

antonioaversa (approved):
LGTM, just a non-blocking remark on the use of the word `inner`.

rules/S3398/description.adoc
zsolt-kolbay-sonarsource: Couldn't agree more. Renamed it.

rules/S3398/csharp/rule.adoc
```diff
@@ -1,4 +1,57 @@
-include::../rule.adoc[]
+When a ``++private static++`` method is only invoked by a nested class, there's no reason not to move it into that class. It will still have the same access to the outer class' members, but the outer class will be clearer and less cluttered.
```
andrei-epure-sonarsource, line 1: no need for `++` - this is only needed for special characters.

see https://github.com/SonarSource/rspec/pull/1506#discussion_r1086269791 for details

andrei-epure-sonarsource: should you also say 
```suggestion
When a ``++private static++`` method is only invoked by a nested class, there's no reason not to move it into that class. It will still have the same access to the outer class' static members, but the outer class will be clearer and less cluttered.
```

rules/S3398/csharp/rule.adoc
```diff
@@ -1,4 +1,57 @@
...
+
+== Noncompliant Code Example
+
+[source,csharp]
+----
+public class Outer
+{
+    public void OuterMethod()
```
andrei-epure-sonarsource, line 9: I don't get this : where is this used? shouldn't it be removed?

rules/S3398/csharp/rule.adoc
```diff
@@ -1,4 +1,57 @@
...
+public class Outer
+{
+    public void OuterMethod() 
+    {
+        Console.WriteLine("Called from an outer method.");
+    }
+
+    private static void Print(int num) // Noncompliant - method is only used by the nested class, should be moved there
+    {
+        Console.WriteLine(num);
```
andrei-epure-sonarsource, line 16: you could also have a private static field inside `Outer` and have it here `num + staticNum` (that's what "It will still have the same access to the outer class' members" is mentioning)

[like this](https://sharplab.io/#v2:C4LghgzsA+ACDMACWAmRA5A9gOwMaYFsAHAGwEsxthEBvAWAChFnEiAnMgNzGAFNkAjADZEZKogBmZXiQAmAbkYtkSMdQBi0uYgC8APklaFSlieYB6c4nyz+sAAyIAktmy82Z1h259BI2AKOALK8wAAWmLIAFACUugYARAmIANSGMsYMnuxcPHZoLm4eTCz0JcrMCILBoRGyArHxiCHhkbGK5YgAvow9Wf1VqIgAwoSkFOJlyjk+dsKi4lIZHcpVaoiaGU1Lch2eltaRdo6F7ozZ3nnIBa5nnVMVzDNXAf6BzbVtcfqISanpu3OnVWSACNVa9UaPxadXaZk8fT6QA===)

rules/S3398/java/rule.adoc
```diff
@@ -1,3 +1,4 @@
+include::../description.adoc[]
```
andrei-epure-sonarsource, line 1: what about `swift`?

https://sonarsource.github.io/rspec/#/rspec/S3398/swift

rules/S3398/rule.adoc
```diff
@@ -1,6 +1,3 @@
-When a ``++private++`` method is only invoked by an inner class, there's no reason not to move it into that class. It will still have the same access to the outer class' members, but the outer class will be clearer and less cluttered.
```
andrei-epure-sonarsource, line 1: why extract this in a separate description for the rest of the languages, if anyway we don't use it in C#?

Wouldn't it be cleaner just to keep the C# changes and let everything else as it is? (as it doesn't change anything)?

andrei-epure-sonarsource (changes requested):

andrei-epure-sonarsource (approved):
LGTM

PR #1578, opened 2023-02-16
Modify rule S1444(C#): Add info on decision to not implement this rule for C# 

mary-georgiou-sonarsource:


rules/S1444/comments-and-links.adoc
```diff
@@ -4,5 +4,5 @@
 Is implemented by \http://jira.codehaus.org/browse/SONARPLUGINS-3223 for Flex
```
mary-georgiou-sonarsource, line 4: this link is dead and no replacement exists - but it's in a comment so I guess I can leave as is?

rules/S1444/csharp/rule.adoc
```diff
@@ -1,4 +1,4 @@
-include::../rule.adoc[]
+This rule won't be implemented for C# as it's already covered by rule S2223.
```
mary-georgiou-sonarsource, line 1: Is this ok?
I did not find something in metadata to communicate the decision.

arseniy-sonar: this is ok to communicate the decision, indeed. But it is important to mark the rule as "Closed" in the metadata, see https://github.com/SonarSource/rspec/#to-delete-a-rule

rules/S1444/comments-and-links.adoc
arseniy-sonar: Yes, links in the comments are not checked. Also, I think the `\` prevents asciidoctor from making it an active link

arseniy-sonar (changes requested):

rules/S1444/csharp/metadata.json
```diff
@@ -1,3 +1,33 @@
...
+  "scope": "Main",
+  "securityStandards": {
+    "CWE": [
+      500
+    ]
+  },
+  "defaultQualityProfiles": [
+    "Sonar way"
+  ],
+  "quickfix": "unknown"
```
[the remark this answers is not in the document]
arseniy-sonar: Just to clarify. All these fields are automatically inherited by the language-specific version of the rule, so it is not necessary to copy them over. You can keep only the metadata fields that are different.
Note however, that this applies only to the top-level metadata fields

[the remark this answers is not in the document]
mary-georgiou-sonarsource: Thanks for taking the time to explain this!

PR #1583, opened 2023-02-17
Modify rule S2223: Update description to include also the case of global state

mary-georgiou-sonarsource:


rules/S2223/csharp/rule.adoc
```diff
@@ -1,4 +1,12 @@
-A ``++static++`` field that is neither constant nor read-only is not thread-safe. Correctly accessing these fields from different threads needs synchronization with ``++lock++``s. Improper synchronization may lead to unexpected results, thus publicly visible static fields are best suited for storing non-changing data shared by many consumers. To enforce this intent, these fields should be marked ``++readonly++`` or converted to constants.
+Non-private `static` fields that are neither `const` nor `readonly` can lead to errors and unpredictable behavior.
+
+This can happen because:
+* Any object can modify these fields and alter the shared state between several other objects.
+* Correctly accessing these fields from different threads needs synchronization with `lock`. Improper synchronization may lead to unexpected results.
+* They make code more difficult to read, debug and test.
+
+Publicly visible static fields are best suited for storing non-changing data shared by objects/consumers.
```
andrei-epure-sonarsource, line 8: for some reason my brain needs to think twice when reading "non-changing data" 

I suggest to make it a bit more wordy as in [the docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2211)

Also, the objects/consumers is weird - wouldn't a consumer be always an object in OOP?

Also, let's be more opinionated than "best suited" and simply use "should"

```suggestion
Publicly visible static fields should only be used to store shared data that does not change.
```

rules/S2223/csharp/rule.adoc
```diff
@@ -1,4 +1,12 @@
-A ``++static++`` field that is neither constant nor read-only is not thread-safe. Correctly accessing these fields from different threads needs synchronization with ``++lock++``s. Improper synchronization may lead to unexpected results, thus publicly visible static fields are best suited for storing non-changing data shared by many consumers. To enforce this intent, these fields should be marked ``++readonly++`` or converted to constants.
+Non-private `static` fields that are neither `const` nor `readonly` can lead to errors and unpredictable behavior.
+
+This can happen because:
+* Any object can modify these fields and alter the shared state between several other objects.
+* Correctly accessing these fields from different threads needs synchronization with `lock`. Improper synchronization may lead to unexpected results.
+* They make code more difficult to read, debug and test.
```
andrei-epure-sonarsource, line 6: the often used terminology is "global state" , "global variable", "global scope" https://en.wikipedia.org/wiki/Global_variable

and we can merge some lines

```suggestion
* Any object can modify these fields and alter the global state. This makes the code more difficult to read, debug and test.
* Correctly accessing these fields from different threads needs synchronization with `lock`. Improper synchronization may lead to unexpected results.
```

andrei-epure-sonarsource (changes requested):

antonioaversa (approved):
LGTM! Very well phrased.

mary-georgiou-sonarsource:
@andrei-epure-sonarsource pipeline is failing due to next being down.

andrei-epure-sonarsource (approved):
lgtm!

PR #1591, opened 2023-02-28
AsciiDoc Dos & Don'ts

ilia-kebets-sonarsource:
As discussed with @arseniy-sonar, this initiates a document for AsciiDoc format "gotchas".

### DIFF
--- a/docs/asciidoc_faq.adoc
+++ b/docs/asciidoc_faq.adoc
@@ -1,0 +1,10 @@
+= AsciiDoc FAQ
+
+=== Multiple include
+
+* When using multiple `include` statements, make sure that there is at least an empty line between them like:
+
+```adoc
+include::how-to-fix-it/core.adoc[]
+include::how-to-fix-it/symfony.adoc[]
+```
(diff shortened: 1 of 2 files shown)


docs/asciidoc_faq.adoc
```diff
@@ -0,0 +1,10 @@
+= AsciiDoc FAQ
+
+=== Multiple include
+
+* When using multiple `include` statements, make sure that there is at least an empty line between them like:
+
+```adoc
+include::how-to-fix-it/core.adoc[]
+include::how-to-fix-it/symfony.adoc[]
```
arseniy-sonar, line 9: This looks like a counter example to the rule above.
I think the best is to show the wrong and the right examples, as is done in https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc

arseniy-sonar (changes requested):
Thank you for starting off this list. I am sure there will be many more tricky points as we continue using asciidoc.
I am not sure the FAQ is a good name here, as this is not really a question, it is rather a pitfall. So I would go for something like "AsciiDoc Pitfalls", or "AsciiDoc Dos&Don'ts" or "AsciiDoc Tips&Tricks"

arseniy-sonar (approved):
Perfect!

PR #1597, opened 2023-03-01
S3900: Update rule description and message

zsolt-kolbay-sonarsource:
https://github.com/SonarSource/sonar-dotnet/issues/2643

### DIFF
--- a/rules/S3900/csharp/rule.adoc
+++ b/rules/S3900/csharp/rule.adoc
@@ -70,9 +70,9 @@ public class MyClass
 
 
 == Exceptions
-Arguments validated for `null` via helper methods should be annotated with the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull[`[NotNull]`] attribute.
-Method parameters marked with the `[NotNull]` https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ItemNotNullAttribute[Resharper code annotation attribute] are supported as well.
-To create a custom null validation method declare an attribute with name `ValidatedNotNullAttribute` and mark the parameter that is validated for null in your method declaration with it:
+* Arguments validated for `null` via helper methods should be annotated with the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull[`[NotNull]`] attribute.
+* Method parameters marked with the `[NotNull]` https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ItemNotNullAttribute[Resharper code annotation attribute] are supported as well.
+* To create a custom null validation method declare an attribute with name `ValidatedNotNullAttribute` and mark the parameter that is validated for null in your method declaration with it:
 
 [source,csharp]
 ----
--- a/rules/S3900/csharp/rule.adoc
+++ b/rules/S3900/csharp/rule.adoc
@@ -1,7 +1,7 @@
-A publicly accessible method can be called from anywhere, which means you should validate parameters to be within the expected constraints. In general, checking against ``++null++`` is recommended defensive programming.
+Methods declared as public, protected, or protected internal can be accessed from other assemblies, which means you should validate parameters to be within the expected constraints. In general, checking against `null` is recommended in defensive programming.
 
 
-This rule raises an issue when a parameter of a publicly accessible method is not validated against ``++null++`` before being dereferenced.
+This rule raises an issue when a parameter of a publicly accessible method is not validated against `null` before being dereferenced.
 
 
 == Noncompliant Code Example
@@ -85,10 +85,7 @@ public static class Utils
     public static string ToUpper(string value)
     {
         Guard.NotNull(value, nameof(value));
-        if (value == null)
-        {
-            return value.ToString();
-        }
+
         return value.ToUpper(); // Compliant
     }
 }
--- a/rules/S3900/csharp/rule.adoc
+++ b/rules/S3900/csharp/rule.adoc
@@ -12,12 +12,12 @@ public class MyClass
 {
     private MyOtherClass other;
 
-    public void Bar(MyOtherClass other) 
+    public void Foo(MyOtherClass other) 
     {
         this.other = other.Clone(); // Noncompliant
     }
 
-    protected void FooBar(MyOtherClass other)
+    protected void Bar(MyOtherClass other)
     {
         this.other = other.Clone(); // Noncompliant
     }
@@ -33,7 +33,7 @@ public class MyClass
 {
     private MyOtherClass other;
 
-    public void Bar(MyOtherClass other) 
+    public void Foo(MyOtherClass other) 
     {
         if (other != null) 
         {
@@ -41,14 +41,14 @@ public class MyClass
         }
     }
 
-    public void Baz(MyOtherClass other) 
+    public void Bar(MyOtherClass other) 
     {
         ArgumentNullException.ThrowIfNull(other);
 
         this.other = other.Clone();
     }
 
-    protected void FooBar(MyOtherClass other) 
+    protected void Baz(MyOtherClass other) 
     {
         if (other != null) 
         {
@@ -56,7 +56,7 @@ public class MyClass
         }
     }
 
-    public void Foo(MyOtherClass other) 
+    public void Qux(MyOtherClass other) 
     {
         this.other = other; // Compliant: "other" is not being dereferenced
     }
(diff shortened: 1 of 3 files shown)


rules/S3900/csharp/rule.adoc
```diff
@@ -1,7 +1,7 @@
-A publicly accessible method can be called from anywhere, which means you should validate parameters to be within the expected constraints. In general, checking against ``++null++`` is recommended defensive programming.
+Methods declared as public, protected, or protected internal can be accessed from other assemblies, which means you should validate parameters to be within the expected constraints. In general, checking against `null` is recommended in defensive programming.
 
 
-This rule raises an issue when a parameter of a publicly accessible method is not validated against ``++null++`` before being dereferenced.
+This rule raises an issue when a parameter of a publicly accessible method is not validated against `null` before being dereferenced.
```
martin-strecker-sonarsource, line 5: Below, in the example:
* Beginning the *Noncompliant Code Example* section with a compliant example is confusing. Maybe we should put the `Foo` example last?
* Grammar: `// Compliant: other not being dereferenced` -> `// Compliant: "other" is not dereferenced`
* Simplify ``++ValidatedNotNullAttribute++`` 
* Consider adding an AttributeUsage attribute to the user-defined Attribute as it is good practice to do so `[AttributeUsage(AttributeTargets.Parameter, Inherited=false)]`

martin-strecker-sonarsource: Have you discussed the exceptions already? I would add
* [ArgumentNullException.ThrowIfNull Method (System) | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.throwifnull?view=net-7.0) (.Net 7+)
* [NotNullAttribute](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull) (C#8)
* [ReSharper Code annotation attributes: NotNullAttribute](https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#NotNullAttribute)

martin-strecker-sonarsource: We should consider to provide a correct implementation for the Guard class that corresponds to 
https://github.com/dotnet/runtime/blob/6a218ab799e56d4ebaffda550de06a9a73a88b14/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs#L59
Downside is, that CallerArgumentExpression is .Net Core 3+ only.

I would also remove the `this` as the method isn't called as an extension method, and providing extension methods for `object` is not recommended.

```cs
public static void NotNull<T>([ValidatedNotNullAttribute] T value, [CallerArgumentExpression("value")] string name = "") where T : class`

Guard.NotNull(value);
```

martin-strecker-sonarsource (changes requested):

rules/S3900/csharp/rule.adoc
zsolt-kolbay-sonarsource: Thank you for the thorough review!
- I moved the `Foo()` compliant method to the Compliant section and addressed the grammar/formatting issues.
- I've added the `AttributeUsage` attribute.
- I've added a sample for `ArgumentNullException.ThrowIfNull()` to the Compliant section, and the Microsoft/Resharper annotations to the list of exceptions.
- `NotNull()` is no longer an extension method, and I've added the `CallerArgumentExpression` attribute to the second parameter.

rules/S3900/csharp/rule.adoc
```diff
@@ -51,29 +41,43 @@ public class MyClass
...
+        this.other = other; // Compliant: "other" is not being dereferenced
+    }
 }
 ----
 
 
 == Exceptions
-
-To create a custom null validation method declare an attribute with name ``++ValidatedNotNullAttribute++`` and mark the parameter that is validated for null in your method declaration with it:
+Parameters marked with https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis[attributes for null-state static analysis] - `[NotNull]` or `[DisallowNull]` - are ignored.
```
martin-strecker-sonarsource, line 68: I think `[DisallowNull]` is a pre-condition and does not fit here. It is only useful if applied to ref parameter as far as I can tell. See e.g. this [example](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzguATEBqAPgAQCYBGAWACh8AGAAn2IDoARKCAcxgHs0oBjFBgGFOGAKYBBGBAA2ATxRQUAbgoUAxDACu06RGDTRNUVP2jVlAMx1CNQTQDeFGs5oB6VzQDaLFDOmcAdwA5bWkAXRoAWgA+GgAVAAtDACdRADMaAAcIZIgwUQRRZJowKDYEhBpgQy0dGgg0wuSAnIwUOCrNSsUaLkq/QNEMLNQUGlhe0KcXfCt8ABYaAFkACm9FAeDQsNSMzmAAK1FeBAB+Gk4ASgdplxdOGgBeSZ0VcjuAX1uab9m6RaWxBWl2+jned3uh2OZwuT16ogCFyhJ2BbwhLncsOetWkSjcHgAPJEaC1kjAxlA9uMxjjvndVrsLpc0S4vuQPkA==).

martin-strecker-sonarsource:
```suggestion
Arguments validated for `null` via helper methods should be annotated with the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull[`[NotNull]`] attribute.
```

rules/S3900/csharp/rule.adoc
```diff
@@ -51,29 +41,43 @@ public class MyClass
...
+Parameters marked with the `[NotNull]` https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ItemNotNullAttribute[Resharper code annotation attribute] are ignored as well.
```
martin-strecker-sonarsource, line 69:
```suggestion
Parameters marked with the `[NotNull]` https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ItemNotNullAttribute[Resharper code annotation attribute] are supported as well.
```

rules/S3900/csharp/rule.adoc
```diff
@@ -51,29 +41,43 @@ public class MyClass
...
 ----
 using System;
 
+[AttributeUsage(AttributeTargets.Parameter, Inherited=false)]
 public sealed class ValidatedNotNullAttribute : Attribute { }
 
 public static class Guard
 {
-    public static void NotNull<T>([ValidatedNotNullAttribute] this T value, string name) where T : class
+    public static void NotNull<T>([ValidatedNotNullAttribute] T value,  [CallerArgumentExpression("value")] string name) where T : class
```
martin-strecker-sonarsource, line 80:
```suggestion
    public static void NotNull<T>([ValidatedNotNullAttribute] T value,  [CallerArgumentExpression("value")] string name = "") where T : class
```

rules/S3900/csharp/rule.adoc
```diff
@@ -12,11 +12,6 @@ public class MyClass
 {
     private MyOtherClass other;
```
martin-strecker-sonarsource, line 14: I think it would make sense to name the methods
* Foo
* Bar
* Baz
* Qux

in that order or something related to https://en.wikipedia.org/wiki/Metasyntactic_variable

martin-strecker-sonarsource (changes requested):

rules/S3900/csharp/rule.adoc
zsolt-kolbay-sonarsource: Changed it.

rules/S3900/csharp/rule.adoc
zsolt-kolbay-sonarsource: Done.

rules/S3900/csharp/rule.adoc
```diff
@@ -39,41 +34,50 @@ public class MyClass
...
-    public void Bar(MyOtherClass other) 
     {
         if (other != null) 
         {
             this.other = other.Clone();
         }
     }
 
-    protected void FooBar(MyOtherClass other) 
+    public void Bar(MyOtherClass other)
```
martin-strecker-sonarsource, line 44: The `Bar` in the compliant section is the one with the `protected` modifier. You exchange the code of the Bar and Baz methods.

rules/S3900/csharp/rule.adoc
```diff
@@ -39,41 +34,50 @@ public class MyClass
...
 
 
 == Exceptions
-
-To create a custom null validation method declare an attribute with name ``++ValidatedNotNullAttribute++`` and mark the parameter that is validated for null in your method declaration with it:
+Arguments validated for `null` via helper methods should be annotated with the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull[`[NotNull]`] attribute.
+Method parameters marked with the `[NotNull]` https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ItemNotNullAttribute[Resharper code annotation attribute] are supported as well.
+To create a custom null validation method declare an attribute with name `ValidatedNotNullAttribute` and mark the parameter that is validated for null in your method declaration with it:
 
 ----
```
martin-strecker-sonarsource, line 72:
```suggestion
[source,csharp]
----
```

rules/S3900/csharp/rule.adoc
```diff
@@ -39,41 +34,50 @@ public class MyClass
...
 using System;
 
+[AttributeUsage(AttributeTargets.Parameter, Inherited=false)]
 public sealed class ValidatedNotNullAttribute : Attribute { }
 
 public static class Guard
 {
-    public static void NotNull<T>([ValidatedNotNullAttribute] this T value, string name) where T : class
+    public static void NotNull<T>([ValidatedNotNullAttribute] T value,  [CallerArgumentExpression("value")] string name = "") where T : class
```
martin-strecker-sonarsource, line 80:
```suggestion
    public static void NotNull<T>([ValidatedNotNullAttribute] T value, [CallerArgumentExpression("value")] string name = "") where T : class
```

martin-strecker-sonarsource (changes requested):

martin-strecker-sonarsource (approved):
LGTM

PR #1599, opened 2023-03-02
SONARPY-1296 Update quickfix metadata for Python rules

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
I believe S5717 still doesn't have the information on quick fix availability.
It was added before the recent sprints, but it seems that it wasn't updated at the time.
Can you take the opportunity to update metadata for this one as well?
LGTM otherwise.

PR #1606, opened 2023-03-03
Frontend deployment ignores the 'common' utility folder in rules directories

christophe-zurn-sonarsource:


frontend/src/deployment/utils.ts
```diff
@@ -2,6 +2,9 @@
 import fs from 'fs';
 import path from 'path';
 
+// List of allowed non-language directories used for utility purposes in a rule folder.
+const ALLOWED_UTILITY_DIRECTORY_NAMES = ['common'];
```
arseniy-sonar, line 6: Should this be a set?
```suggestion
const ALLOWED_UTILITY_DIRECTORY_NAMES = {'common'};
```

arseniy-sonar (approved):
LGTM, only one minor suggestion, but no need for another round of reveiw

frontend/src/deployment/utils.ts
christophe-zurn-sonarsource: Unfortunately, there are no set literals in JS/TS. We would need to use `const ALLOWED_UTILITY_DIRECTORY_NAMES = new Set<string>(['common']);`, I'm not sure it's better than simply using arrays.
I don't have a strong opinion on the matter, so if you prefer the alternative with Set, I'll update it accordingly. WDYT ?

arseniy-sonar: Ah, that's a pity. I think simplicity is more important here, so let's keep an array

PR #1587, opened 2023-02-23
Add checks for education rule format

ilia-kebets-sonarsource:
Fixes https://sonarsource.atlassian.net/browse/RULEAPI-776

- Add checks for the new "How to fix it?" section
- Add docs for fixtures creation
- Add checks for progressive education format root sections
- Remove one check that we don't rely anymore as the checks implementation is more robust
- Add checks for the new "Resources" section

PR #1607, opened 2023-03-03
Add checks for education format

ilia-kebets-sonarsource:
Original work on https://github.com/SonarSource/rspec/pull/1587
I created a new PR because the target branch was rebased, breaking the PR

- Add checks for the new "How to fix it?" section
- Add docs for fixtures creation
- Add checks for progressive education format root sections
- Remove one check that we don't rely anymore as the implementation of the checks is more robust
- Add checks for the new "Resources" section
- Add checks for the "Code examples" subsections

I merged the https://github.com/SonarSource/rspec/tree/czurn/update-rules-headers branch inside here so the new checks pass on these rules in the CI.
I had to fix a few other rules: https://github.com/SonarSource/rspec/pull/1607/commits/e217c9280c26527cae92e7a56bf35cd062f0b62e, https://github.com/SonarSource/rspec/pull/1607/commits/794a067cdca8c649a23a20e6603d868076013d7d


rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,166 @@
...
+def read_file(path):
+  SECTION_NAMES_PATH = Path(__file__).parent.parent.parent.parent.joinpath(path)
+  return SECTION_NAMES_PATH.read_text(encoding='utf-8').split('\n')
+
+def parse_names(path):
+  SECTION_NAMES_FILE = read_file(path)
+  return [s.replace('* ', '').strip() for s in SECTION_NAMES_FILE if s.strip()]
+
+HOW_TO_FIX_IT = 'How to fix it'
+HOW_TO_FIX_IT_REGEX = re.compile(HOW_TO_FIX_IT)
```
arseniy-sonar, line 21: Why do you use two ways to match this header?
`section.startswith(HOW_TO_FIX_IT)` and `HOW_TO_FIX_IT_REGEX.match(s)`

arseniy-sonar (changes requested):
Thank you for converging the multiple lists of permitted sections into an (almost) single source of truth.
I think it can be improved further in two aspects:
1. "all_section_names.adoc" contains some of the sections from the "education_format_example.adoc", so this makes it harder to keep consistent (if I want to rename or add an education section, I need to remember to change it in these two files), and makes it less precise because now the education sections are allowed for non-education rules, which wasn't the intention, I believe. (BTW, why is it called '*_example"? Is it not the definitive list?).
To address this, I propose to keep two files - one for education, and one for non-education rules: "legacy_section_names.adoc", and "education_format_section_names.adoc", for example. One will be used for non-education rules, and will fail on education rule, and vice-versa. This should also reduce the total validation code for section names.
2. The handling of the "education_format_example.adoc" looks really narrow, and might break if we decide to add a new `== ` section, rename, or reorder the sections. the parsing code is tightly coupled with this particular document. I like how you collect the optional sections. I propose to not make any assumptions about the `== `-level sections, and transform the 2-level nested list into an array of `sections` as you do now + an map of arrays of subsections, i.e.:
`{"Resources" -> ["Articles", "Standards", ...], "How to fix it" -> ["Pitfalls", "Going extra mile"], ...}`
(you can skip the `{framework display name}` in the section name)

I understand that you will need special handling of the `How to fix it in ...` section, but at least all the remaining can be handled in a generic way.

ci/validate_asciidoc.sh
```diff
@@ -97,8 +97,8 @@ done
 ADOC_COUNT=$(find rules -name "tmp*.adoc" | wc -l)
 
 if (( ADOC_COUNT > 0 )); then
-  if asciidoctor --failure-level=WARNING -o /dev/null rules/*/*/tmp*.adoc; then
-      if asciidoctor -a rspecator-view --failure-level=WARNING -o /dev/null rules/*/*/tmp*.adoc; then
+  if asciidoctor -o /dev/null rules/*/*/tmp*.adoc; then
+      if asciidoctor -a rspecator-view -o /dev/null rules/*/*/tmp*.adoc; then
```
arseniy-sonar, line 101: Is there a way to keep failing on other `asciidoctor` warnings?
Would it be possible to skip print the header-level skip warnings for thousands of rules during validation?
Finally, you can remove the `echo "== Description" > "$TMP_ADOC"` workaround on l.76 if you are disabling this check.

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,159 @@
...
+  current_section_name = ''
+  for line in education_format_file:
+    if line.startswith('== '):
+      section = line.replace('== ', '').strip()
+      if section.endswith('(optional)'):
+        section = section.replace(' (optional)', '')
+        current_map = optional_sections
+      else:
+        current_map = sections
+      if HOW_TO_FIX_IT_REGEX.match(section):
```
arseniy-sonar, line 37:
```suggestion
      if section.starts_with(HOW_TO_FIX_IT):
```
This might be slightly faster, but more importantly it makes it more correlated with the body of the `if`

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,159 @@
...
+        current_map[HOW_TO_FIX_IT] = set()
+        current_section_name = HOW_TO_FIX_IT
+      else:
+        current_section_name = section
+        current_map[section] = set()
```
arseniy-sonar:
```suggestion
        section = HOW_TO_FIX_IT # overwrite previous instances of "How to fix it *" section specifications
      current_section_name = section
      current_map[section] = set()
```

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,159 @@
...
+[
+  SECTIONS,
+  OPTIONAL_SECTIONS
+  ] = parse_education_section_names('docs/header_names/education_format_names.adoc')
+
+
+
+def intersection(lst1, lst2):
+  lst3 = [value for value in lst1 if value in lst2]
+  return lst3
```
arseniy-sonar, line 70:
```suggestion
  return list(set(lst1).intersection(lst2))
```
In harmony with the `difference` function you have below

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,160 @@
...
+        current_map = optional_sections
+      else:
+        current_map = sections
+      if HOW_TO_FIX_IT_REGEX.match(section):
+        # we store "How to fix it" and "How to fix it in ..." together
+        section = HOW_TO_FIX_IT
+      current_section_name = section
+      if section not in current_map:
+        # If a section appears multiple times in the doc, we don't overwrite its subsections
+        current_map[section] = set()
```
arseniy-sonar, line 43:
```suggestion
        section = HOW_TO_FIX_IT
      else:
        if section in current_map:
          raise RuleValidationError(f'Section {section} is mentioned more than once in the specification file')
      current_section_name = section
      current_map[section] = set()
```
If we don't overwrite its subsections, they are gonna accumulate and be duplicated. 
I think the correct way of handling this is to assert that the section wasn't mentioned before, except if it is a HOW_TO_FIX_IT subsection. See my suggestion there

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,160 @@
...
+  """Validates all h2-level section names"""
+
   descr = rule_language.description
-  for h2 in descr.find_all('h2'):
-    name = h2.text.strip()
-    if name not in ACCEPTED_SECTION_NAMES:
-      raise RuleValidationError(f'Rule {rule_language.id} has unconventional header "{name}"')
+  h2_titles = list(map(lambda x: x.text.strip(), descr.find_all('h2')))
+
+  education_titles = intersection(h2_titles, list(SECTIONS.keys()) + list(OPTIONAL_SECTIONS.keys()))
```
arseniy-sonar, line 80: If you change the `intersection` second parameter type to a set, you can write this shorter:
```suggestion
  education_titles = intersection(h2_titles, SECTIONS.keys() + OPTIONAL_SECTIONS.keys())
```

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,160 @@
...
+    raise RuleValidationError(f'Rule {rule_language.id} has more than 6 "{HOW_TO_FIX_IT}" sections. Please ensure this limit can be increased with PM/UX teams')
+  if not how_to_fix_it_sections:
+    raise RuleValidationError(f'Rule {rule_language.id} is missing a "{HOW_TO_FIX_IT}" section')
+  if HOW_TO_FIX_IT in how_to_fix_it_sections and len(how_to_fix_it_sections) > 1:
+    raise RuleValidationError(f'Rule {rule_language.id} is mixing "{HOW_TO_FIX_IT}" with "How to fix it in FRAMEWORK NAME" sections. Either use a single "{HOW_TO_FIX_IT}" or one or more "How to fix it in FRAMEWORK"')
+  framework_sections_seen = set()
+  for section_name in how_to_fix_it_sections:
+    validate_how_to_fix_it_framework(section_name, rule_language, framework_sections_seen)
+
+def validate_how_to_fix_it_framework(section_name, rule_language, framework_sections_seen):
```
arseniy-sonar, line 108:
```suggestion
  for section_name in how_to_fix_it_sections:
    validate_how_to_fix_it_framework(section_name, rule_language)

def validate_how_to_fix_it_framework(section_name, rule_language):
    framework_sections_seen = set()
```
Why do you keep a single set for all "How to fix it " sections?

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -98,3 +203,18 @@ def validate_source_language(rule_language: LanguageSpecificRule):
...
+    if resources_section is not None:
+      titles = collect_titles(resources_section, 3)
+      subsections_seen = set()
+      for title in titles:
+        name = title.text.strip()
+        if name not in OPTIONAL_SECTIONS[optional_section]:
+          raise RuleValidationError(f'Rule {rule_language.id} has a "{optional_section}" subsection with an unallowed name: "{name}"')
+        if name in subsections_seen:
+          raise RuleValidationError(f'Rule {rule_language.id} has duplicate "{optional_section}" subsections. There are 2 occurences of "{name}"')
+        subsections_seen.add(name)
```
arseniy-sonar:
```suggestion
        if name not in OPTIONAL_SECTIONS[optional_section]:
          raise RuleValidationError(f'Rule {rule_language.id} has a "{optional_section}" subsection with an unallowed name: "{name}"')
        if name in subsections_seen:
          raise RuleValidationError(f'Rule {rule_language.id} has duplicate "{optional_section}" subsections. There are 2 occurences of "{name}"')
        subsections_seen.add(name)
```
This code is duplicated in `validate_how_to_fix_it_subsections` and `validate_how_to_fix_it_framework`

arseniy-sonar (changes requested):
Looks good. Only one important issue remains: the asciidoctor built-in validation, see my inline comment

rspec-tools/tests/resources/invalid-rules/S200/scala/rule.adoc
```diff
@@ -0,0 +1,6 @@
+== Resources
+=== Documentation
+=== Documentation
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: Yes, I'll add a comment

rspec-tools/tests/resources/invalid-rules/S200/javascript/rule.adoc
```diff
@@ -0,0 +1,5 @@
+== Why is this an issue?
+== How to fix it in Razor
+== How to fix it in Razor
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: duplicate frameworks, will add a comment

rspec-tools/tests/resources/invalid-rules/S101/javascript/rule.adoc
```diff
@@ -1,8 +1,9 @@
...
-=== How to fix it in Express.js
+// the maximum amount of frameworks is 6
+== Why is this an issue?
+== How to fix it in Express.js
+== How to fix it in Express.js
+== How to fix it in Express.js
+== How to fix it in Express.js
+== How to fix it in Express.js
+== How to fix it in Express.js
+== How to fix it in Express.js
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: I believe the number gets triggered before in the implementation. I'll refactor the titles to trigger that only.

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: I've reworked this one now.

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: Yes, it has been refactored as `validate_subsections`

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: I'm getting the following error:

```
>     education_titles = intersection(h2_titles, SECTIONS.keys() + OPTIONAL_SECTIONS.keys())
E     TypeError: unsupported operand type(s) for +: 'dict_keys' and 'dict_keys'
```

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: `validate_how_to_fix_it_framework` checks a single `How to fix it in...` section, so I must collect the framework names one level above.

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: I used sets instead of the initial lists to handle the duplicate subsections.

ilia-kebets-sonarsource: I believe that a check for the tooling itself is a bit overkill, no?

ci/validate_asciidoc.sh
ilia-kebets-sonarsource: > Is there a way to keep failing on other asciidoctor warnings?
not sure, but I've changed to fail on [errors](https://docs.asciidoctor.org/asciidoctor/latest/errors-and-warnings/#error) 
> Would it be possible to skip print the header-level skip warnings for thousands of rules during validation?
I guess with a regex on stdout?
> Finally, you can remove the echo "== Description" > "$TMP_ADOC" workaround on l.76 if you are disabling this check.
Since we're keeping errors, we still need this, right?

rspec-tools/rspec_tools/validation/description.py
arseniy-sonar: Nevermind then, sorry. I thought `dict.keys()` returns a set, but it returns `dict_keys`

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -98,3 +186,28 @@ def validate_source_language(rule_language: LanguageSpecificRule):
...
+    else:
+      validate_subsections_for_map(rule_language, mandatory_section, SECTIONS)
+
+def validate_subsections_for_map(rule_language: LanguageSpecificRule, section_name: str, map, section_regex = None):
+  if not section_regex:
+    section_regex = section_name
+  descr = rule_language.description
+  resources_section = descr.find('h2', string=section_regex)
+  if resources_section is not None:
+    titles = collect_titles(resources_section, 3)
```
arseniy-sonar, line 205:
```suggestion
  top_level_section = descr.find('h2', string=section_regex)
  if top_level_section is not None:
    titles = collect_titles(top_level_section, 3)
```

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -98,3 +186,28 @@ def validate_source_language(rule_language: LanguageSpecificRule):
...
+def validate_subsections(rule_language: LanguageSpecificRule):
+  for optional_section in list(OPTIONAL_SECTIONS.keys()):
+    validate_subsections_for_map(rule_language, optional_section, OPTIONAL_SECTIONS)
+  for mandatory_section in list(SECTIONS.keys()):
+    if mandatory_section == HOW_TO_FIX_IT:
+      validate_subsections_for_map(rule_language, mandatory_section, SECTIONS, HOW_TO_FIX_IT_REGEX)
+    else:
+      validate_subsections_for_map(rule_language, mandatory_section, SECTIONS)
+
+def validate_subsections_for_map(rule_language: LanguageSpecificRule, section_name: str, map, section_regex = None):
```
arseniy-sonar, line 199:
```suggestion
def validate_subsections_for_map(rule_language: LanguageSpecificRule, section_name: str, allowed_subsections: set[str], section_regex = None):
```
this makes it more explicit, and reduces the number of lookups to the `SECTIONS` map

rspec-tools/rspec_tools/validation/description.py
arseniy-sonar: I see now. I find this pretty hard to follow. Moreover, you are essentially passing validation of the "how to fix it" section name to the `validate_how_to_fix_it_framework` callee.
Since you already have the list of all "how to fix it in" sections: `how_to_fix_it_sections`, why not check duplication of section names in `validate_how_to_fix_it_sections_names`? For example:
```
  duplicate_names = [x for x in how_to_fix_it_sections if how_to_fix_it_sections.count(x) > 1] # O(n*n) is fine, given n <= 6
  if 0 < duplicate_names.len():
    raise RuleValidationError(f'Rule {rule_language.id} has duplicate "{HOW_TO_FIX_IT}" sections "{duplicate_names}"')
```

arseniy-sonar (changes requested):
The validation code looks good, good job refactoring!
The only point remaining is the [spam of the warnings](https://cirrus-ci.com/task/5020001413890048?logs=asciidoc_tests#L86) coming from level-skipping of the code-example subsections. Even now, with few rules implementing this format these warning constitute 484 lines of the full 573-line long validation log. This makes the log hard to read, and it is important to see when some other rule error happens, because that has direct effect on the rule descriptions that are displayed to the clients.
It is not an option to send the asciidoctor output to /dev/null as it has other useful diagnostics.
What do you think can be done here?
Shall we raise the examples to the h3 level (the same level as the "What is the potential impact?" subsection)?
This would also allow RSPEC tooling to check the names of these sections, as it does it for the legacy format. And is an important point since, I think products might rely on the names of these sections, for example, to show examples side-by-side.

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: done in e45c99c72

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: Thanks!

ci/validate_asciidoc.sh
arseniy-sonar: Right, I did not realise it was an error and warning-level

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: done

rspec-tools/rspec_tools/validation/description.py
```diff
@@ -1,55 +1,118 @@
...
-SECTION_NAMES_FILE = SECTION_NAMES_PATH.read_text(encoding='utf-8').split('\n')
-ACCEPTED_SECTION_NAMES: Final[list[str]] = [s.replace('* ', '').strip() for s in SECTION_NAMES_FILE if s.strip()]
+LEGACY_SECTION_NAMES: Final[list[str]] = parse_names('docs/header_names/legacy_section_names.adoc')
 # The list of all the framework names currently accepted by the script.
-FRAMEWORK_NAMES_PATH = Path(__file__).parent.parent.parent.parent.joinpath('docs/allowed_framework_names.adoc')
-FRAMEWORK_NAMES_FILE = FRAMEWORK_NAMES_PATH.read_text(encoding='utf-8').split('\n')
-ACCEPTED_FRAMEWORK_NAMES: Final[list[str]] = [s.replace('* ', '').strip() for s in FRAMEWORK_NAMES_FILE if s.strip()]
+ACCEPTED_FRAMEWORK_NAMES: Final[list[str]] = parse_names('docs/header_names/allowed_framework_names.adoc')
+
+SECTIONS = {
```
arseniy-sonar, line 32: Just like in the doc, please, add a comment here warning the developer that they should change the list in the documentation if these maps are updated

docs/description.adoc
```diff
@@ -65,25 +65,37 @@ so as to be as comprehensible as possible for the users.
...
+** Pitfalls
+** Going the extra mile
+* Resources (optional)
+** Documentation
+** Articles & blog posts
+** Conference presentations
+** Standards
+
+
+Content of the section "_3. How to fix it_" (i.e. subsections 3.a, 3.b, 3.c and 3.d) may be repeated multiple times, as
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: I didn't touch this because it is heavily modified by the other PR: https://github.com/SonarSource/rspec/pull/1610/files

rspec-tools/rspec_tools/validation/description.py
ilia-kebets-sonarsource: done https://github.com/SonarSource/rspec/pull/1607/commits/f01bbbb92b07348717aa21cf5b668dbadef7a6d1

PR #1615, opened 2023-03-07
Modify rule S2137: Change text to include JavaScript standard builtins

francoismora:
The rule now checks for JavaScript standard built-in identifiers instead of the list previously defined in the rule. The code samples have been updated to use `const`/`let` instead of `var`. The rule actually doesn't check function constructor calls, so the corresponding code examples have been removed: `new Function("arguments", "return 17;")`.

Fixes SonarSource/SonarJS#3708

### DIFF
--- a/rules/S2137/javascript/rule.adoc
+++ b/rules/S2137/javascript/rule.adoc
@@ -1,10 +1,10 @@
 JavaScript has special identifiers that, while not reserved, still should not be used as identifiers. They form the JavaScript standard built-in objects and are available in all environments. They include identifiers like:
 
-* ``++eval++`` - evaluates a string as JavaScript code
-* ``++arguments++`` - used to access function arguments through indexed properties. It exists only inside function declarations and function expressions
-* ``++undefined++`` - returned for values and properties that have not yet been assigned
+* ``++eval++`` - Evaluates a string as JavaScript code.
+* ``++arguments++`` - Used to access function arguments through indexed properties. It exists only inside function declarations and function expressions.
+* ``++undefined++`` - Returned for values and properties that have not yet been assigned.
 * ``++NaN++`` - Not a Number; returned when math functions fail.
-* ``++Infinity++`` - when a number exceeds the upper limit of the floating point numbers
+* ``++Infinity++`` - When a number exceeds the upper limit of the floating point numbers.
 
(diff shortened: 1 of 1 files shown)


rules/S2137/javascript/rule.adoc
```diff
@@ -1,10 +1,9 @@
...
-
+JavaScript has special identifiers that, while not reserved, still should not be used as identifiers. They form the JavaScript standard built-in objects and are available in all environments. They include identifiers like:
 
 * ``++eval++`` - evaluates a string as JavaScript code
-* ``++arguments++`` - used to access function arguments through indexed properties. 
+* ``++arguments++`` - used to access function arguments through indexed properties. It exists only inside function declarations and function expressions
 * ``++undefined++`` - returned for values and properties that have not yet been assigned
-* ``++NaN++`` - Not a Number; returned when math functions fail. 
+* ``++NaN++`` - Not a Number; returned when math functions fail.
 * ``++Infinity++`` - when a number exceeds the upper limit of the floating point numbers
```
yassin-kammoun-sonarsource, line 7: Could you make consistent each bullet point by either adding an ending point or removing it from all of them, please? Also, some of these bullet points start with an uppercase while others don't.

rules/S2137/javascript/rule.adoc
```diff
@@ -17,13 +16,12 @@ These words should not be bound or assigned, because doing so would overwrite th
...
 ++eval; // Noncompliant
-var obj = { set p(arguments) { } }; // Noncompliant
-var eval; // Noncompliant
-try { } catch (arguments) { } // Noncompliant
-function x(eval) { } // Noncompliant
-function arguments() { } // Noncompliant
-var y = function eval() { }; // Noncompliant
-var f = new Function("arguments", "return 17;"); // Noncompliant
+const obj = { set p(arguments) { } }; // Noncompliant
+let eval; // Noncompliant
```
yassin-kammoun-sonarsource, line 20: 👍

yassin-kammoun-sonarsource (approved):
LGTM!

rules/S2137/javascript/rule.adoc
francoismora: Updated

PR #1616, opened 2023-03-07
Modify rule S1472: Add tagged template literals

francoismora:
The change adds a paragraph about tagged template literals.

Fixes SonarSource/SonarJS#3705

### DIFF
--- a/rules/S1472/javascript/rule.adoc
+++ b/rules/S1472/javascript/rule.adoc
@@ -12,7 +12,7 @@ The first option will be the one chosen by the JavaScript interpreter.
 By extension, and to improve readability, any kind of function call argument should not start on new line.
 
 Similarly, tagged templates allow for advanced forms of string interpolation by evaluating the tag as a function to call, passing the
-template literal elements as arguments. Therefore the rule also verifies that template literals don't start on a separate line.
+template literal elements as arguments. Therefore, the rule also verifies that template literals don't start on a separate line.
 
 == Noncompliant Code Example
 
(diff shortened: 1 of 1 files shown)


yassin-kammoun-sonarsource (changes requested):
I don't think the current RSPEC format supports multiple pairs of noncompliant/compliant sections. Maybe it's technically possible, but I don't remember seeing any rule doing that. We will probably need to define a `rule.adoc` file specific for JavaScript that includes both genera and JavaScript-specific descriptions as well as move the noncompliant/compliant snippets for tagged templates in the already existing noncompliant/compliant sections.

You can still prove me wrong and find an RSPEC that includes multiple sections of noncompliant and compliant snippets.

rules/S1472/javascript/rule.adoc
```diff
@@ -1,4 +1,76 @@
...
+. Define a function and then execute some unrelated code inside a closure ?
+. Pass the second function as a parameter to the first one ?
+
+The first option will be the one chosen by the JavaScript interpreter.
+
+
+By extension, and to improve readability, any kind of function call argument should not start on new line.
+
+Similarly, tagged templates allow for advanced forms of string interpolation by evaluating the tag as a function to call, passing the
+template literal elements as arguments. Therefore the rule also verifies that template literals don't start on a separate line.
```
yassin-kammoun-sonarsource, line 15:
```suggestion
template literal elements as arguments. Therefore, the rule also verifies that template literals don't start on a separate line.
```

yassin-kammoun-sonarsource (approved):

PR #1620, opened 2023-03-08
S3655: Include cast scenario for C#

antonioaversa:


rules/S3655/noncompliant.adoc
```diff
@@ -1,8 +1,10 @@
 == Noncompliant Code Example
 
-[source,text]
+[source,csharp]
 ----
 int? nullable = null;
 ...
-UseValue(nullable.Value); // Noncompliant
+Console.WriteLine(nullable.Value); // Noncompliant
```
zsolt-kolbay-sonarsource, line 7: I'd put this in a method, just to show that the rule is showing issues where the variable might not have a value (instead of showing an example where it never does).

```
void Print(int? number)
{
    Console.WriteLine(nullable.Value); // Noncompliant
    // or
    Console.WriteLine((int)nullable); // Noncompliant
}
```

rules/S3655/description.adoc
```diff
@@ -1 +1 @@
-Nullable value types can hold either a value or ``++null++``. The value held in the nullable type can be accessed with the ``++Value++`` property, but ``++.Value++`` throws an ``++InvalidOperationException++`` when the value is ``++null++``. To avoid the exception, a nullable type should always be tested before ``++.Value++`` is accessed.
+Nullable value types can hold either a value or `null`. The value held in the nullable type can be accessed with the `Value` property or casting to the underlying type, but both operations throw an `InvalidOperationException` when the value is `null`. To avoid the exception, a nullable type should always be tested before the value is accessed.
```
zsolt-kolbay-sonarsource, line 1: or **by** casting **it** to the underlying type

zsolt-kolbay-sonarsource (changes requested):
Just some minor comments.

rules/S3655/noncompliant.adoc
antonioaversa: My understanding is that the use of `number` in the method you show above should not trigger rule S3655.
`number` would have an unknown value, and, while there is no path where the variable is definitely non-`null`, there is no path where the variable is definitely `null` either.

Asking confirmation of the above to @pavel-mikula-sonarsource 

Incidentally, when nullable types are enabled, CS8629 is triggered: `Nullable value type may be null`.

zsolt-kolbay-sonarsource (approved):
lgtm

rules/S3655/noncompliant.adoc
```diff
@@ -1,8 +1,10 @@
...
+// or
+Console.WriteLine((int)nullable);  // Noncompliant
```
pavel-mikula-sonarsource, line 9: This is problematic, because copy/pasting the example into IDE will not show the `Noncompliant` here. Because the previous line has learned that it has `NotNull`.

We need to split it. See the proposal above.

pavel-mikula-sonarsource (changes requested):
We'll need to improve the old examples.
Users should be able to copy/paste them somewhere to see it in action.
And both `Noncompliant` cases should just raise.

rules/S3655/noncompliant.adoc
antonioaversa: Changed to match the `Sample` method above.

rules/S3655/compliant.adoc
```diff
@@ -1,22 +1,38 @@
...
-int? nullable = null;
-...
-if (nullable.HasValue)
+public void Sample(bool condition)
 {
-  UseValue(nullable.Value);
+    int? nullableValue = condition ? 42 : null;
+    if (nullableValue.HasValue)
+    {
+      Console.WriteLine(nullableValue.Value); // Noncompliant
```
pavel-mikula-sonarsource, line 10: Copy/Paste issue: This is not Noncompliant anymore :D
```suggestion
      Console.WriteLine(nullableValue.Value);
```

rules/S3655/compliant.adoc
```diff
@@ -1,22 +1,38 @@
...
+    }
+
+    int? nullableCast = condition ? 42 : null;
+    if (nullableCast.HasValue)
+    {
+      Console.WriteLine((int)nullableCast);   // Noncompliant
```
pavel-mikula-sonarsource, line 16: Also here and below

rules/S3655/compliant.adoc
```diff
@@ -1,22 +1,38 @@
...
+    }
 }
 ----
 or
 
-[source,text]
+[source,csharp]
```
pavel-mikula-sonarsource, line 22: I don't think we need this second example.

rules/S3655/compliant.adoc
```diff
@@ -1,22 +1,38 @@
...
 ----
-int? nullable = null;
-...
-if (nullable != null)
+public void Sample(bool condition)
 {
-  UseValue(nullable.Value);
+    int? nullableValue = condition ? 42 : null;
+    if (nullableValue != null)
```
pavel-mikula-sonarsource, line 27: I'd use later C#
```suggestion
    if (nullableValue is not null)
```

rules/S3655/compliant.adoc
```diff
@@ -1,22 +1,38 @@
...
+    {
+      Console.WriteLine(nullableValue.Value); // Noncompliant
+    }
+
+    int? nullableCast = condition ? 42 : null;
+    if (nullableCast != null)
```
pavel-mikula-sonarsource, line 33: Also here
```suggestion
    if (nullableCast is not null)
```

pavel-mikula-sonarsource (changes requested):
The `Noncompliant` needs to go away

rules/S3655/compliant.adoc
antonioaversa: The compliant examples show two different ways of making the code compliant: using `.Value` and `!= null`.

1. Should I merge them in the following way:
```cs
public void Sample(bool condition)
{
    int? nullableValue = condition ? 42 : null;
    if (nullableValue.HasValue)
    {
      Console.WriteLine(nullableValue.Value);
    }

    int? nullableCast = condition ? 42 : null;
    if (nullableCast is not null)
    {
      Console.WriteLine((int)nullableCast);
    }
}
```

2. or just don't propose `is not null` as a possible fix;
3. or don't propose `.HasValue` as a possible fix.

I couldn't find any recommendation for using one rather than the other in general literature (apart from not-so-backed up opinions on the subject [here](https://stackoverflow.com/questions/676078/what-is-the-difference-between-nullablet-hasvalue-or-nullablet-null)).

pavel-mikula-sonarsource: 1) Merge them.

The merged version provides the same information on half of the code.

PR #1634, opened 2023-03-13
Modify rule S1862: Consider identical cases of switch statements

yassin-kammoun-sonarsource:


ilia-kebets-sonarsource (approved):
See suggestion otherwise LGTM

PR #1643, opened 2023-03-14
Modify rule S4275: Update code samples

francoismora:
Fixes SonarSource/SonarJS#3759

### DIFF
--- a/rules/S4275/javascript/rule.adoc
+++ b/rules/S4275/javascript/rule.adoc
@@ -5,21 +5,42 @@ include::../description.adoc[]
 [source,javascript]
 ----
 class A {
-  private _x: number = 0;
-  private y: number = 0;
+  #x: number = 0;
+  #y: number = 0;
 
-  public get x() {  // Noncompliant: field 'x' is not used in the return value
-    return this.y;
+  get x() {  // Noncompliant: field 'x' is not used in the return value
+    return this.#y;
   }
 
-  public setX(val: number) { // Noncompliant: field 'x' is not updated
-    this.y = val;
+  setX(val: number) { // Noncompliant: field 'x' is not updated
+    this.#y = val;
   }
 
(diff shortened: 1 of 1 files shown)


rules/S4275/javascript/rule.adoc
```diff
@@ -5,41 +5,66 @@ include::../description.adoc[]
...
 
-  public setX(val: number) {
-    this.x = val;
+  getY() {
+    return this.#y;
   }
 
-  public getY() {
-    return this.y;
+  set y(val: number) {
```
yassin-kammoun-sonarsource, line 66: I am a bit confused here. The non compliant snippet for this setter defines `setY` while here we use the syntax `set y`. Is that on purpose?

rules/S4275/javascript/rule.adoc
```diff
@@ -5,41 +5,66 @@ include::../description.adoc[]
...
 }
+
+let x = 0;
+let y = 0;
+const obj2 = {};
+Object.defineProperty(obj2, 'x', {
+  get() {
+    return y;
+  }
+});
```
yassin-kammoun-sonarsource, line 43: Can we add `// Noncompliant` annotations to highlight the issue?

yassin-kammoun-sonarsource: We are missing the compliant counterpart of these.

yassin-kammoun-sonarsource (changes requested):

rules/S4275/javascript/rule.adoc
francoismora: I guess I pasted the wrong code snippet. Fixed.

rules/S4275/javascript/rule.adoc
francoismora: Fixed.

yassin-kammoun-sonarsource (approved):

PR #1651, opened 2023-03-15
Modify rule S3649: Fix typo in code samples

hendrik-buchwald-sonarsource:


jamie-anderson-sonarsource (approved):
LGTM. I've seen that the fix is detected as expected.

PR #1673, opened 2023-03-21
SONARPY-1301 rule S5994 update for python

maksim-grebeniuk-sonarsource:


rules/S5994/python/rule.adoc
```diff
@@ -0,0 +1,22 @@
+include::../description.adoc[]
+
+
+== Noncompliant Code Example
+
+[source,python]
+----
+import re
+2match1 = re.match(r"a++abc", "aaaabc", re.DOTALL) # Noncompliant, the second 'a' never matches
+3match2 = re.match(r"\d*+[02468]", "1234", re.DOTALL) # Noncompliant, the sub-pattern "[02468]" never matches
```
guillaume-dequenne-sonarsource, line 10: I think the leading number is a bit odd, can we remove it to have `match1` and `match2` instead?
Same for the compliant examples below.
It also looks a bit odd that we are trying to match a literal immediately. Maybe we could have `pattern1 = re.compile` calls, consistently with what we do for Java?

guillaume-dequenne-sonarsource (changes requested):
Left a minor comment.

rules/S5994/python/rule.adoc
maksim-grebeniuk-sonarsource: done

PR #1674, opened 2023-03-21
Modify rule S5950: Fix typos + rewording (CPP-4142)

amelie-renard-sonarsource:


rules/S5950/cfamily/rule.adoc
```diff
@@ -1,39 +1,31 @@
...
+Note: This scenario can only happen before {cpp}17. From {cpp}17, the standard states that each argument needs to be thoroughly evaluated before evaluating the other arguments. So the explicit construction of `unique_ptr` and `shared_ptr` become exception-safe.
 
 *Performance*
 
-
-While ``++make_uniqe()++`` doesn't have an impact on performance, ``++make_shared()++`` does.
-
-``++make_shared()++`` performs one heap-allocation. While constructing ``++shared_ptr()++`` explicitly will require two: one for the object being managed and the other for the control block that stores data about the ref-counts and the ``++shared_ptr()++`` deleter.
+Constructing `shared_ptr()` explicitly requires two heap allocations: one for the managed object and the other for the control block that stores data about the ref-counts and the `shared_ptr()` deleter. `make_shared()` still impacts performances, but it performs only one heap allocation. +
+`make_unique()`, on the other hand, doesn't impact performances.
```
phil-nash-sonarsource, line 28: Replace the two uses of "performances" with "performance"

phil-nash-sonarsource: "`make_shared()` still impacts performance" - relative to what?

phil-nash-sonarsource (changes requested):
See inline comments

rules/S5950/cfamily/rule.adoc
amelie-renard-sonarsource: Mmm, I misunderstood the paragraph. I've rewritten it, I hope it's clearer

rules/S5950/cfamily/rule.adoc
```diff
@@ -1,39 +1,31 @@
...
-. The constructed ``++Lhs++`` object is leaked since the ``++unique_ptr++`` isn't constructed yet
-
-Note: This scenario can only happen before ``{cpp}17``. the new standard states that each argument needs to be fully evaluated before the evaluation of the other arguments. In this case, the explicit construction of ``++unique_ptr++`` and ``++shared_ptr++`` is exception-safe.
+. Memory allocation for `Lhs`
+. Construction of the `Lhs` object
+. Call to `throwingFunction` (before the `unique_ptr` construction)
+. `throwingFunction` throws an exception
+. The constructed `Lhs` object is leaked since the `unique_ptr` isn't constructed yet
 
+Note: This scenario can only happen before {cpp}17. From {cpp}17, the standard states that each argument needs to be thoroughly evaluated before evaluating the other arguments. So the explicit construction of `unique_ptr` and `shared_ptr` become exception-safe.
```
loic-joly-sonarsource, line 23: I'm not sure, but this wording sounds slightly strange to me...
```suggestion
Note: This scenario can only happen before {cpp}17. Starting with {cpp}17, the standard states that each argument needs to be thoroughly evaluated before evaluating the other arguments. So the explicit construction of `unique_ptr` and `shared_ptr` become exception-safe.
```

loic-joly-sonarsource: I'm not sure either about the word "thoroughly".
```suggestion
Note: This scenario can only happen before {cpp}17. From {cpp}17, the standard states that even though the order of evaluation of each argument is still unspecified, interleaving the evaluation of different arguments is no longer allowed. This makes the direct construction of `unique_ptr` and `shared_ptr` exception-safe.
```

loic-joly-sonarsource (commented):
I'd be happy if a native speaker (yes, looking at you @phil-nash-sonarsource) could check my comments.

rules/S5950/cfamily/rule.adoc
amelie-renard-sonarsource: I don't know about the first comment, but I like your second suggestion.

philsquared: I think "From" was ok. "Starting with" is better. Or perhaps, "Since".

philsquared: I agree that Loïc's second comment is a good improvement. LGTM. Ship it!

rules/S5950/cfamily/rule.adoc
```diff
@@ -53,18 +45,21 @@ auto uniqueP = std::make_unique<MyClass>(42);
...
 
 == Exceptions
 
-This rule ignores code that uses features not supported by `make_shared` or `make_unique`:
+This rule ignores code that uses features not supported by `make_shared` and `make_unique`:
+
+* custom deleters
+* calling placement-new, i.e., version of `new` with arguments, like `new(std::nothrow)`
+
+In addition, `make_shared` does not support the following:
```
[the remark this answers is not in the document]
philsquared: That is the case I had in mind, yes. I think a mention in the rspec that this is a valid exception should be enough. I don't think it is that common, in practice.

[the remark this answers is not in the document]
amelie-renard-sonarsource: Okay, I'll add something about that

rules/S5950/cfamily/rule.adoc
```diff
@@ -1,40 +1,33 @@
...
+Indeed, constructing explicitly a `shared_ptr()` requires two heap allocations: one for the managed object and the other for the control block that stores data about the ref-counts and the `shared_ptr()` deleter. `make_shared()` on the other hand, performs only one heap allocation.
 
-While ``++make_uniqe()++`` doesn't have an impact on performance, ``++make_shared()++`` does.
-
-``++make_shared()++`` performs one heap-allocation. While constructing ``++shared_ptr()++`` explicitly will require two: one for the object being managed and the other for the control block that stores data about the ref-counts and the ``++shared_ptr()++`` deleter.
-
+Note: Because `make_shared` performs only one allocation for both the object and the control block, the memory occupied by the object will be deallocated when no `shared_ptr` or `weak_ptr` points to it. If the object is large, a `weak_ptr` is used, and memory is a concern, explicitly calling the constructor of `shared_ptr` may be preferred. This way, the object's memory will be deallocated when there are no more shared owners, never mind the weak ones.
```
phil-nash-sonarsource, line 30: I think "never mind the weak ones" sounds a little too informal for this setting. How about something like, "independently of any `weak_ptr`s" ?

phil-nash-sonarsource (changes requested):
Just one more minor change (inline) and this is good to go.

rules/S5950/cfamily/rule.adoc
amelie-renard-sonarsource: Sounds good to me

PR #1681, opened 2023-03-23
Modify rule S2222: Update description

pavel-mikula-sonarsource:
Update description as proposed here by @ganncamp :
https://discuss.sonarsource.com/t/c-lock-release-rule-description-seems-to-skip-some-steps/10639/14

### DIFF
--- a/rules/S2222/description-dotnet.adoc
+++ b/rules/S2222/description-dotnet.adoc
@@ -2,6 +2,6 @@ If a lock is held or acquired and then released within a method, it should be re
 
 Failing to do so could make an application deadlock-prone because there is a risk of the lock not being released.
 
-The types tracked by the rule are: `Monitor`, `Mutex`, `ReaderWriterLock`, `ReaderWriterLockSlim` and `SpinLock` from the `System.Threading` namespace.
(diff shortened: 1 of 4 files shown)


rules/S2222/description-dotnet.adoc
```diff
@@ -3,3 +3,5 @@ If a lock is held or acquired and then released within a method, it should be re
 Failing to do so could make an application deadlock-prone because there is a risk of the lock not being released.
 
 The types tracked by the rule are: `Monitor`, `Mutex`, `ReaderWriterLock`, `ReaderWriterLockSlim` and `SpinLock` from the `System.Threading` namespace.
+
+This rule raises an issue when a lock is acquired in a method but only released on some paths. No issue is raised when the lock is never released in the method. The assumption is that callers will release it.
```
martin-strecker-sonarsource, line 7: NIT:
```suggestion
This rule raises an issue when a lock is acquired in a method but only released on some paths. No issue is raised when the lock is never released in the method because the assumption is that callers will release it.
```

martin-strecker-sonarsource (approved):
LGTM. An oxford comma is missing and one NIT suggestion

PR #1705, opened 2023-03-30
Update S2871 (`no-alphabetical-sort`): Cover cases from `require-array-sort-compare`

francoismora:
…y-sort-compare`


ilia-kebets-sonarsource (approved):
LGTM!

PR #1716, opened 2023-04-03
Modify rule S2757: Cover confusing non-null assertions for TS

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
we're going for a separate rule: https://github.com/SonarSource/rspec/pull/1719

PR #1742, opened 2023-04-12
Modify rule S6500: Adjust title and wording

hendrik-buchwald-sonarsource:
Implementation ticket: https://sonarsource.atlassian.net/browse/SONARIAC-731
AppSec ticket: https://sonarsource.atlassian.net/browse/APPSEC-624

### DIFF
--- a/rules/S6500/description.adoc
+++ b/rules/S6500/description.adoc
@@ -1,5 +1,5 @@
-Installing recommended packages automatically can lead to additional
-vulnerabilities in the Docker image.
+Installing recommended packages automatically can lead to vulnerabilities in the
+Docker image.
 
 Potentially unnecessary packages are installed via a known Debian package
 manager. These packages might contain vulnerabilities, thus weakening the
--- a/rules/S6500/description.adoc
+++ b/rules/S6500/description.adoc
@@ -1,13 +1,13 @@
-Installing recommended dependencies automatically can lead to additional
+Installing recommended packages automatically can lead to additional
 vulnerabilities in the Docker image.
 
-Potentially unnecessary dependencies are installed via a known Debian package
-manager. These dependencies might contain vulnerabilities, thus weakening the
+Potentially unnecessary packages are installed via a known Debian package
+manager. These packages might contain vulnerabilities, thus weakening the
 overall security posture of the created containers. +
 Depending on the introduced vulnerabilities, a malicious actor accessing such a
 container could use these for privilege escalation. +
-Removing the unused dependencies can also significantly reduce your Docker image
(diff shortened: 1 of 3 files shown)


rules/S6500/description.adoc
```diff
@@ -1,13 +1,12 @@
...
+Installing recommended packages automatically can lead to vulnerabilities in the
+Docker image.
 
-Unnecessary dependencies are installed via a known Debian package manager.
-These unused dependencies might contain
-vulnerabilities, thus weakening the overall security posture of the created
-containers. +
+Potentially unnecessary packages are installed via a known Debian package
+manager. These packages might contain vulnerabilities, thus weakening the
+overall security posture of the created containers. +
```
gaetan-ferry-sonarsource, line 6: It might not be clear why packages from a trusted source can contain vulnerabilities. I would add some details here.
```suggestion
manager. These packages will increase the attack surface of the created container. Indeed, they might contain unidentified vulnerabilities or malicious code. Those packages could be used as part of a broader supply chain attack. In general, the more packages are installed on a container, the weaker its security posture is. +
```

rules/S6500/implementation.adoc
```diff
@@ -6,7 +6,7 @@ ifdef::env-github,rspecator-view[]
 
 == Message
 
-* Make sure that installing unnecessary dependencies is safe here.
+* Make sure automatically installed packages are really used.
```
gaetan-ferry-sonarsource, line 9: This does not fit the hotspots messaging guideline I would say.

```suggestion
* Make sure automatically installing recommended packages is safe here.
```

rules/S6500/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Installing unnecessary packages is security-sensitive",
+  "title": "Installing recommended packages automatically is security-sensitive",
```
gaetan-ferry-sonarsource, line 2: I would have moved the "automatically" at the beginning. But I am not really sure.

```suggestion
  "title": "Automatically installing recommended packages is security-sensitive",
```

gaetan-ferry-sonarsource (changes requested):
I think we could add some details about why having more packages is bad in general. Also some slight wording suggestions.

rules/S6500/metadata.json
hendrik-buchwald-sonarsource: Yea, that looks nicer.

rules/S6500/description.adoc
hendrik-buchwald-sonarsource: I applied it, just slightly rewritten.

PR #1759, opened 2023-04-14
Modify rule S3655: Streamlining for C# and VB 

Tim-Pohlmann:
Update the message for C# and the code samples for VB.
Implementation Ticket: https://github.com/SonarSource/sonar-dotnet/pull/7084


antonioaversa (approved):
LGTM

PR #1775, opened 2023-04-18
Modify rule S2492: Fix the noncompliant example (PLSQL-S2492)

gregory-paidis-sonarsource:
This is a fix for [this issue](https://discuss.sonarsource.com/t/parse-error-on-noncompliant-example-of-plsql-namingtypescheck/12627)

### DIFF
--- a/rules/S2492/plsql/rule.adoc
+++ b/rules/S2492/plsql/rule.adoc
@@ -8,7 +8,7 @@ With the default regular expression ``++[a-zA-Z]([a-zA-Z0-9_]*[a-zA-Z0-9])?++``:
 [source,sql]
 ----
 DECLARE
-  TYPE Collection-type_ IS VARRAY(42) OF PLS_INTEGER; -- Noncompliant
+  TYPE Collection_type_ IS VARRAY(42) OF PLS_INTEGER; -- Noncompliant
 BEGIN
   NULL;
 END;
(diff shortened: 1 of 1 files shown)


rules/S2492/plsql/rule.adoc
```diff
@@ -8,7 +8,7 @@ With the default regular expression ``++[a-zA-Z]([a-zA-Z0-9_]*[a-zA-Z0-9])?++``:
 [source,sql]
 ----
 DECLARE
-  TYPE Collection-type_ IS VARRAY(42) OF PLS_INTEGER; -- Noncompliant
+  TYPE Collection_type_ IS VARRAY(42) OF PLS_INTEGER; -- Noncompliant
 BEGIN
```
csaba-sagi-sonarsource, line 12: Why didn't you address the "_Compliant Solution has unneeded {code}_" comment from Michal?

csaba-sagi-sonarsource (changes requested):
I think we can improve the examples further.

rules/S2492/plsql/rule.adoc
gregory-paidis-sonarsource: you are right, thanks for noticing it :)

csaba-sagi-sonarsource (approved):
LGTM!

PR #1777, opened 2023-04-19
Modify rule S6506: Add examples for wget

jamie-anderson-sonarsource:
Implementation ticket: [SONARIAC-750](https://sonarsource.atlassian.net/browse/SONARIAC-750)

[SONARIAC-750]: https://sonarsource.atlassian.net/browse/SONARIAC-750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rules/S6506/docker/highlighting.adoc
+++ b/rules/S6506/docker/highlighting.adoc
@@ -1,3 +1,9 @@
 === Highlighting
 
-Highlight the curl command, the `-L` option, and the URL.
+For `curl`:
+
+* Highlight the `curl` command, the `-L` option, and the URL.
+
+For `wget`:
+
+* Highlight the `wget` command and the URL.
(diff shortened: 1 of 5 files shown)


hendrik-buchwald-sonarsource (approved):
LGTM!

The other RSPEC PR we can merge ourselves. After that, we should remove the warning in the first post and maybe rebase the PR.
I would also recommend starting the first post with "Implementation ticket:".

jonas-wielage-sonarsource:
@jamie-anderson-sonarsource It would be nice if you could also adapt the Implementation Specification for the rule in case of wget.
Should the message remain the same?

jamie-anderson-sonarsource:
I've added a new message for `wget` given that you cannot enforce HTTPS. The message for `curl` has been slightly modified for consistency.

I've also added a note for how `wget` should be highlighted.

PR #1793, opened 2023-04-21
Modify rule S5693: Add syntax highlighting and remove Java specific tags

dorian-burihabwa-sonarsource:


leonardo-pilastri-sonarsource (approved):
Looks good, maybe you could just change the last part of the PR name to `and remove tags`, since we're getting rid of all of them?

PR #1798, opened 2023-04-24
Modify rule S3260: Update rule to include file access modifier

sebastien-marichal:


rules/S3260/csharp/impact.adoc
```diff
@@ -0,0 +1,67 @@
+=== What is the potential impact?
+
+We measured ~2x improvement in execution time by running the following snippet with https://github.com/dotnet/BenchmarkDotNet[BenchmarkDotNet].
```
costin-zaharia-sonarsource, line 3: It looks to me that the code is at least 3 times faster.
```suggestion
We measured at least 3x improvement in the execution time by running the following snippet with https://github.com/dotnet/BenchmarkDotNet[BenchmarkDotNet].
```

rules/S3260/csharp/how.adoc
```diff
@@ -0,0 +1,3 @@
+== How to fix it
+
+The issue can be fixed by adding the `sealed` keyword in front of the `class` or `record`.
```
costin-zaharia-sonarsource, line 3: What do you think about?

> The code can be improved by adding the `sealed` keyword in front of the `class` or `record` types that have no inheritors.

rules/S3260/csharp/why.adoc
```diff
@@ -0,0 +1,3 @@
+== Why is this an issue?
+
+Classes and records with either `private` or `file` access modifier aren't visible outside of their assemblies, so if they're not extended inside the assemblies or file, they should be made explicitly non-extensible with the addition of the `sealed` keyword.
```
costin-zaharia-sonarsource, line 3:
```suggestion
Classes and records with either `private` or `file` access modifiers aren't visible outside of their assemblies or files, so if they're not extended inside their scope, they should be made explicitly non-extensible with the addition of the `sealed` keyword.
```

costin-zaharia-sonarsource (changes requested):

PR #1800, opened 2023-04-24
S1155, S3216, S3260, S3353: Add the performance tag

costin-zaharia-sonarsource:


rules/S1155/csharp/metadata.json
sebastien-marichal, line 1: The tag could be also applied to the `vbnet` version of the rule

sebastien-marichal (changes requested):

rules/S1155/csharp/metadata.json
costin-zaharia-sonarsource: Done. From what I've seen only S1155 has a VB spec.

PR #1811, opened 2023-04-26
Modify rule S2094: improve JS rule definition

ilia-kebets-sonarsource:


yassin-kammoun-sonarsource (approved):
LGTM! Let's maybe reword the commit/PR message:

- `Modify rule S2094: Split the non/compliant code samples into two pairs`

PR #1815, opened 2023-04-26
Modify rule S6590: Add code example and documentation header

ilia-kebets-sonarsource:


rules/S6590/javascript/rule.adoc
```diff
@@ -19,16 +19,42 @@ Replace the explicit literal type assertion with `as const`.
...
 ----
-let foo = { bar: 'baz' as 'baz' };
+const foo: 42 = 42;
 ----
 
 ==== Compliant solution
 
 [source,typescript,diff-id=1,diff-type=compliant]
 ----
+const foo = 42 as const;
```
yassin-kammoun-sonarsource, line 29: We need to use `let` here and in the noncompliant snippet because `foo` in `const foo = 42` has type `42` because of `const`, which makes the compliant snippet not ideal since `as const` would be redundant.

yassin-kammoun-sonarsource (changes requested):

rules/S6590/javascript/rule.adoc
ilia-kebets-sonarsource: indeed, the examples I ran into on Peach were either `public static` fields or keyword less ones. Maybe use a class with static fields since `let ... as const` sounds stupid?

yassin-kammoun-sonarsource: Yes, I guess.

PR #1820, opened 2023-04-27
Modify rule S4043: suggest ES2023 `Array.toReversed()` and `Array.toSorted()` functions

philnash:
Also updates `var` to `const`.


ilia-kebets-sonarsource (approved):
LGTM!

PR #1822, opened 2023-04-27
RULEAPI-785 RSPEC: education format "How to fix it" section should be optional

christophe-zurn-sonarsource:


docs/description.adoc
```diff
@@ -71,14 +71,14 @@ This new structure is defined as follows:
 // This needs to be kept in sync with the [maps in the validation script](https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_tools/validation/description.py#L33-L40).
```
victor-diez-sonarsource, line 71: Should we keep this link in sync with the changes?

christophe-zurn-sonarsource: Good catch :D

PR #1825, opened 2023-05-01
Modify rule S3949: Remove CBDE from C#, add VB.NET

pavel-mikula-sonarsource:
Fixes https://github.com/SonarSource/sonar-dotnet/issues/7139

### DIFF
--- a/rules/S3949/csharp/rule.adoc
+++ b/rules/S3949/csharp/rule.adoc
@@ -1,19 +1,32 @@
-== Why is this an issue?
+include::../description.adoc[]
 
-Numbers are infinite, but the types that hold them are not. Each numeric type has hard upper and lower bounds. Try to calculate or assign numbers beyond those bounds, and the result will be a value that has silently wrapped around from the expected positive value to a negative one, or vice versa.
-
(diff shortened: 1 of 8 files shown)


rules/S3949/csharp/rule.adoc
```diff
@@ -1,33 +1,33 @@
...
+----
+public long Transform(int value)
+{
+    if (value <= 0)
+    {
+        return value;
+    }
+    long number = int.MaxValue;
+    return number + value;  // Noncompliant
+}
```
Tim-Pohlmann, line 30:
```suggestion
public long Transform(int value)
{
    if (value <= 0)
    {
        return value;
    }
    long number = int.MaxValue;
    return number + value;
}
```

Tim-Pohlmann (changes requested):

rules/S3949/csharp/rule.adoc
pavel-mikula-sonarsource: Oops, and it was missing on the VB version

Tim-Pohlmann (approved):
LGTM

pavel-mikula-sonarsource:
There are 2 new commits to review

Tim-Pohlmann (approved):
LGTM

PR #1828, opened 2023-05-04
Modify rule S1912: CPP-4028 Take namespace into consideration

LoghinVladDev:


LoghinVladDev:
To be merged together with https://github.com/SonarSource/sonar-cpp/pull/2736

LoghinVladDev:
@necto Please Review details of name lookup, I'm trying to think of a more direct explanation. Input appreciated
Also, would this not be a candidate for the new format? this 'name lookup' could be in the last level of the 'learning' process

rules/S1912/cfamily/rule.adoc
```diff
@@ -7,6 +7,16 @@ It is especially important that multi-threaded applications do not call the same
 This rule will trigger an issue each time a function in the configurable list is invoked.
 
 
+A call will be matched differently depending on the presence of the scope resolution operator ``::`` in the function name from the configurable list.
+Consider the call ``a::b::f()``, and the following cases for the name specified to the list:
```
necto, line 11: This does not say this isn't nested in a different namespace (i.e. your call in this context `namespace c {...; a::b::f();}` won't fit the bullet points below). Moreover, this might confuse users who have a call without full qualification, e.g. `namespace a { ...; using namespace b; ...; f(); }` should still adhere to the same rules described below.

I propose to focus on the function declaration here instead of the function call site. I.e., speak about `namespace a { namespace b { void f(); } }`.

necto (changes requested):

rules/S1912/cfamily/rule.adoc
```diff
@@ -7,6 +7,17 @@ It is especially important that multi-threaded applications do not call the same
 This rule will trigger an issue each time a function in the configurable list is invoked.
 
 
+A call will be matched differently depending on the presence of the scope resolution operator ``::`` in the function name from the configurable list.
+Consider the following declaration ``namespace a { namespace b { void f(); } }``, and the following cases for how this function could be added to specified to the list:
+
+- If the function name is not qualified (list contains ``f``), any call to any function ``f`` will be trigger an issue, regardless of declaration scope.
+- If the function name is partially qualified (list contains ``b::f``), any call to a function ``f`` declared inside a scope named ``b`` will trigger an issue
+(calling the function ``namespace c { namespace b { void f(); } }`` will trigger the issue).
+- If the function name is fully qualified (list contains ``::a::b::f``), only calls to the function ``f``, from the named scope ``b``, from the global named scope ``a`` will trigger an issue
```
necto, line 16: I find the description in ASTMatcher easier to grasp. Something along the lines:
```suggestion
For example, `namespace a { namespace b { void f(); } }` can be matched with "f", "b::f", "a::b::f", "::a::b::f" (fully qualified name yielding most precise results).
```
Note, that this remark should be concise enough: it matters only when you configure the rule, so it shouldn't take too much space when a developer on your team wants to read about the meaning of the rule.

rules/S1912/cfamily/rule.adoc
```diff
@@ -7,6 +7,17 @@ It is especially important that multi-threaded applications do not call the same
 This rule will trigger an issue each time a function in the configurable list is invoked.
 
 
+A call will be matched differently depending on the presence of the scope resolution operator ``::`` in the function name from the configurable list.
```
necto, line 10:
```suggestion
Depending on the use of the scope resolution operator ``::`` in a function name in `nonReentrantFunctionList`, the rule might flag calls of functions declared in different name scopes.
```

rules/S1912/cfamily/rule.adoc
```diff
@@ -7,6 +7,17 @@ It is especially important that multi-threaded applications do not call the same
...
+Consider the following declaration ``namespace a { namespace b { void f(); } }``, and the following cases for how this function could be added to specified to the list:
+
+- If the function name is not qualified (list contains ``f``), any call to any function ``f`` will be trigger an issue, regardless of declaration scope.
+- If the function name is partially qualified (list contains ``b::f``), any call to a function ``f`` declared inside a scope named ``b`` will trigger an issue
+(calling the function ``namespace c { namespace b { void f(); } }`` will trigger the issue).
+- If the function name is fully qualified (list contains ``::a::b::f``), only calls to the function ``f``, from the named scope ``b``, from the global named scope ``a`` will trigger an issue
+
+It is recommended to provide fully qualified names to the configurable list. This also applies to functions in the global scope.
```
necto, line 18:
```suggestion
It is recommended to provide fully qualified names to the configurable list (i.e., start each name with `::`), even for the functions in the global namespace.
```
To be clarified, but I think "global namespace" is more common term than "global name scope"

LoghinVladDev: I think this is correct for C++. I though from the perspective of C programming. I assume people using POSIX non-reentrant functions would be in majority C programmers. 
Although the topic at hand only happens in C++, so, it would make sense to adopt full C++ terminology

rules/S1912/cfamily/rule.adoc
```diff
@@ -7,6 +7,15 @@ It is especially important that multi-threaded applications do not call the same
 This rule will trigger an issue each time a function in the configurable list is invoked.
 
 
+A call will be matched differently depending on the presence of the scope resolution operator ``::`` in the function name from the configurable list.
+
+
```
necto, line 12:
```suggestion
A call will be matched differently depending on the presence of the scope resolution operator ``::`` in the function name from the configurable list.
```

The second sentence clarifies the first one, so it would make sense for them to go into the same paragraph.

necto (approved):

PR #1831, opened 2023-05-05
Modify rule S2871: Add Array.toSorted()

philnash:
This goes with https://github.com/SonarSource/SonarJS/pull/3873. `toSorted()` is a new, copying method for arrays that otherwise behaves the same as `sort()`.


ilia-kebets-sonarsource (changes requested):
The description looks great, could you also update the rule title in `rules/S2871/javascript/metadata.json`?

ilia-kebets-sonarsource (approved):

PR #1834, opened 2023-05-05
Modify rule S6544: remove mention of error handling, which we will not cover anymore

ilia-kebets-sonarsource:
Fixes [#3891](https://github.com/SonarSource/SonarJS/issues/3891)

### DIFF
--- a/rules/S6544/javascript/rule.adoc
+++ b/rules/S6544/javascript/rule.adoc
@@ -12,10 +12,6 @@ This rule forbids returning promises where another type is expected such as in:
 * void returns
 * spread operators
 
-[.underline]#*Missing error handling:*#
-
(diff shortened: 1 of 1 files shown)


rules/S6544/javascript/rule.adoc
```diff
@@ -12,10 +12,6 @@ This rule forbids returning promises where another type is expected such as in:
 * void returns
 * spread operators
 
-[.underline]#*Missing error handling:*#
```
yassin-kammoun-sonarsource, line 15: Maybe we can remove the above "separator" `[.underline]#*Unresolved promises:*#` since we consider a single use case now.

yassin-kammoun-sonarsource (approved):

PR #1847, opened 2023-05-09
Modify rule S6344: Update for educational format

jonas-wielage-sonarsource:


rules/S6344/php/rule.adoc
```diff
@@ -1,14 +1,47 @@
 == Why is this an issue?
 
-PHP doesn't fail if two `define` statements assign a value to the same constant name: in such a case, PHP only issues a warning and ignores the second `define`. However, that code should be fixed. In the best case, the second `define` tries to assign the same value to the constant: it's just useless code which can be removed. In the worst case, the second `define` attempts to assign a different value and is followed by code which wrongly assumes the value of the constant: it can be a nasty bug.
+PHP doesn't fail if two or more `define` statements assign a value to the same constant name.
+In such a case, PHP only issues a warning and ignores the second and further `define`.
```
nils-werner-sonarsource, line 4: Even when this is the same wording as before, I would use the opportunity to add 1 or two sentences here to start with more educational wording. Starting with a negotiation may be not the right entry in a rule description. Like, it may not affect you, but we still raise the issue.

rules/S6344/php/rule.adoc
```diff
@@ -1,14 +1,47 @@
...
 
-=== Noncompliant code example
+=== What is the potential impact?
 
-[source,php]
+In the best case, the additional `define` statements try to assign the same value to the constant.
```
nils-werner-sonarsource, line 8: Maybe `additional` is not the right wording. Isn't it more like a duplicated constant assignment?

nils-werner-sonarsource (approved):
LG! Please see my suggestion

PR #1849, opened 2023-05-09
Modify rule S6347: Update for educational format

jonas-wielage-sonarsource:


rules/S6347/php/rule.adoc
```diff
@@ -1,10 +1,18 @@
 == Why is this an issue?
 
-WordPress makes it possible to define options using `define` statements inside a configuration file named `wp-config.php`. However, if `define` statements are located at the end of this file, they are ignored by WordPress. This rule raises an issue when a `define` statement appears after `wp-settings.php` is loaded.
+WordPress makes it possible to define options using `define` statements inside a configuration file named `wp-config.php`.
+However, if `define` statements are located at the end of this file, they are ignored by WordPress.
```
nils-werner-sonarsource, line 4:
```suggestion
However, if the statements are located after the settings are loaded at the end of this file, they are not taken into account by WordPress.
```

rules/S6347/php/rule.adoc
```diff
@@ -1,10 +1,18 @@
...
+This rule raises an issue when a `define` statement appears after `wp-settings.php` is loaded.
```
nils-werner-sonarsource, line 5: You could add a few lines about "=== What is the potential impact?" Something like: security-relevant configuration could be ignored and not considered without any warning.

nils-werner-sonarsource (approved):
LGTM! Please see my comments

PR #1854, opened 2023-05-10
Modify rule S1780: Migrate to new educational format

nils-werner-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM. I saw that the rule describes adhering to PSR-2. As it is deprecated it would be best to create a ticket in PHP to update this to PSR-12

PR #1859, opened 2023-05-10
Modify rule S1599: Migration to new educational format

jonas-wielage-sonarsource:


jonas-wielage-sonarsource:
Close this PR because the branch was not generated from master.
See [PR-1860](https://github.com/SonarSource/rspec/pull/1860) for the actual PR.

PR #1860, opened 2023-05-10
Modify rule S1599: Migration to new educational format

jonas-wielage-sonarsource:


rules/S1599/php/rule.adoc
```diff
@@ -1,11 +1,10 @@
 == Why is this an issue?
 
-PHP's "variable variables" feature (dynamically-named variables) is temptingly powerful, but can lead to unmaintainable code. 
+PHP's "variable variables" feature (dynamically-named variables) is temptingly powerful, but can lead to unmaintainable code.
```
nils-werner-sonarsource, line 3:
```suggestion
Variable variables in PHP allow you to use the value of a variable as the name of another variable. This feature can be useful in dynamic programming scenarios where variable names need to be dynamically determined and manipulated.

=== What is the potential impact?
The use of variable variables in PHP can make code harder to read and understand, as it introduces a level of indirection and can lead to confusion. It can also increase the risk of security vulnerabilities, such as allowing user input to directly manipulate variable names, potentially leading to injection attacks or unintended behavior.
```

nils-werner-sonarsource (approved):
LGTM! Please see my suggestion

PR #1861, opened 2023-05-10
Modify rule S1808: Migration to new educational format

jonas-wielage-sonarsource:


nils-werner-sonarsource (approved):
LGTM! Please remove the rule from the SonarWay

PR #1888, opened 2023-05-12
Modify rule S2430: add to default profile, change quickfix to infeasible

alexander-kamushkin-sonarsource:
The rule was not part of the default quality profile (Sonar way), so we want to fix that


yassin-kammoun-sonarsource (approved):
LGTM

PR #1896, opened 2023-05-12
Modify rule S2014: Migration to new educational format

jonas-wielage-sonarsource:


rules/S2014/php/rule.adoc
```diff
@@ -1,50 +1,66 @@
 == Why is this an issue?
 
-``++$this++`` refers to the current class instance. But static methods can be accessed without instantiating the class, and ``++$this++`` is not available to them. Using ``++$this++`` in a static context will result in a fatal error at runtime.
+In PHP, `$this` is a keyword used inside a class, which refers to the current class instance.
+You have access to class properties and methods through the use of this keyword.
```
nils-werner-sonarsource, line 4:
```suggestion
To reference the current class instance the keyword $this can be used.
Through the use of this keyword you have access to class properties and methods.
```

nils-werner-sonarsource, line 4:
```suggestion
To reference the current class instance the keyword `$this` can be used.
Through the use of this keyword, you have access to class properties and methods.
```

rules/S2014/php/rule.adoc
```diff
@@ -1,50 +1,66 @@
...
 
+Static methods can be accessed without instantiating the class, so ``++$this++`` is not available for them.
+Using ``++$this++`` in a static context will result in a runtime error.
```
nils-werner-sonarsource, line 7:
```suggestion
Static methods can be accessed without instantiating the class, so `$this` is not available for them.
Using `$this` in a static context will result in a runtime error.
```
So far I know the `++` are left overs from a migration. We used to have this data stored in Jira before RSPEC was moved to GitHub.

nils-werner-sonarsource (approved):
LGTM! See my suggestions

nils-werner-sonarsource (approved):
LGTM! See my suggestions

PR #1899, opened 2023-05-12
Modify rule S1605: Migrate to new educational format

nils-werner-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM, see my comments

PR #1911, opened 2023-05-16
Modify rule S1155: Migrate to new educational format

nils-werner-sonarsource:
Related discuss thread https://community.sonarsource.com/t/unclear-scope-of-rspec-1155/90909


jonas-wielage-sonarsource (approved):
LGTM, but broken link should be fixed

PR #1924, opened 2023-05-22
Modify rule S6535: Adapt to LaYC

ilia-kebets-sonarsource:


vilchik-elena (changes requested):
Let's move most of introduction to the first section

rules/S6535/javascript/rule.adoc
```diff
@@ -1,24 +1,24 @@
-In JavaScript, we can use the ``++/++`` (backslash) character to escape special characters inside a string, a template literal, or a regular expression.
+== Why is this an issue?
+
+In JavaScript, the ``++/++`` (backslash) character is used to escape special characters inside a string, a template literal, or a regular expression.
```
vilchik-elena, line 3: I am not sure why we need a new line for each sentence. It does not look very pretty. I would merge all of them into one paragraph

rules/S6535/javascript/rule.adoc
```diff
@@ -28,7 +28,7 @@ const path  = `\/${some}\/${dir}`;
 
 ==== Compliant solution
 
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
 const regex = /[[]/;
```
vilchik-elena, line 33: I think some link to the docs about which characters are special and could be escaped would not hurt

rules/S6535/javascript/rule.adoc
```diff
@@ -1,24 +1,24 @@
...
+Useless character escapes can safely be removed without changing the original behavior.
 
 === Code examples
 
 ==== Noncompliant code example
 
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
 const regex = /[\[]/;
```
vilchik-elena, line 23: could you add a comment here explaining why it's not compliant? `[` seems to be a special char in regex

rules/S6535/javascript/rule.adoc
```diff
@@ -1,24 +1,24 @@
...
 const octal = '\8';
```
vilchik-elena, line 24: and here :D

rules/S6535/javascript/rule.adoc
```diff
@@ -1,24 +1,24 @@
-In JavaScript, we can use the ``++/++`` (backslash) character to escape special characters inside a string, a template literal, or a regular expression.
+== Why is this an issue?
+
+In JavaScript, the ``++/++`` (backslash) character is used to escape special characters inside a string, a template literal, or a regular expression.
 
 It indicates that the next character should be treated as a literal character rather than as a special character or string delimiter.
 
 For instance, it is common to escape single quotes inside a string literal using the single quote delimiter like ``++'It\'s a beautiful day'++``.
 
 Escaping is only meaningful for special characters.
```
vilchik-elena, line 9: I think this phrase can be dropped, it does not bring much

rules/S6535/javascript/rule.adoc
```diff
@@ -1,24 +1,24 @@
...
 
-== Why is this an issue?
-
 Escaping non-special characters in strings, template literals, and regular expressions doesn't affect their value. Therefore, useless escapes impact code readability and could even denote a bug in the code if the developer left it by mistake or intended to escape another special character instead.
 
 == How to fix it
 
-We can safely remove a useless character escape without changing the original behaviour.
+Useless character escapes can safely be removed without changing the original behavior.
```
vilchik-elena, line 15: how about adding a mention that the developer should first make sure that the intention was to simply use the escaped char and that there is no hidden bug.

vilchik-elena (changes requested):

rules/S6535/javascript/rule.adoc
ilia-kebets-sonarsource: There is no exhaustive link, I have added 2.

PR #1926, opened 2023-05-22
Modify rule S4663: Adapt to LaYC

frederic-tingaud-sonarsource:


rules/S4663/rule.adoc
```diff
@@ -1,15 +1,5 @@
...
-----
-/* */
-
-/*
-
- */
-----
+An empty multi-line comment doesn't help to improve the readability of the code and might indicate an oversight.
 
+A meaningful comment should be added or the comment markers should be removed.
```
vilchik-elena, line 5:
```suggestion
A meaningful text should be added to the comment or the comment markers should be removed.
```

rules/S4663/rule.adoc
```diff
@@ -1,15 +1,5 @@
...
-
-[source,text]
-----
-/* */
-
-/*
-
- */
-----
+An empty multi-line comment doesn't help to improve the readability of the code and might indicate an oversight.
```
vilchik-elena, line 3: I think `doesn't help` sounds passive aggressive and also I am not sure why we are mentioning readability for this rule.
```suggestion
An empty multi-line comment might indicate an oversight.
```

vilchik-elena (changes requested):
I suggest to add code examples for each language

rules/S4663/common/code-example.adoc
```diff
@@ -0,0 +1,8 @@
+[source,java]
```
vilchik-elena, line 1: why did you decide to put this fine in the subdir?

vilchik-elena (approved):

rules/S4663/common/code-example.adoc
frederic-tingaud-sonarsource: I _felt_ like it would be better in that case, because the include::{variable}[] pattern can be a little confusing. The variable is declared in a subfolder, but interpreted by the include at root. So for example you are inside the java subfolder and your variable says `:example: java/code-example.adoc`. If somebody writes `:example: code-example.adoc` and there is a code-example.adoc at the root, we don't know if they wanted to include the common file and they succeeded or if they didn't understand they need to repeat the folder name. If the common file is also in a subfolder, there can be no doubt it's on purpose.

PR #1928, opened 2023-05-22
Change rule S4487: Adapt to LaYC

frederic-tingaud-sonarsource:


vilchik-elena:
hey @frederic-tingaud-sonarsource could you split this into 2 PRs? to have one PR per rule

PR #1931, opened 2023-05-22
Modify rule S4143: Adapt to LaYC

frederic-tingaud-sonarsource:


rules/S4143/javascript/rule.adoc
```diff
@@ -1,19 +1,31 @@
 == Why is this an issue?
 
-It is highly suspicious when a value is saved for a key or index and then unconditionally overwritten. Such replacements are likely in error.
+include::../description.adoc[]
```
vilchik-elena, line 4: let's add a phase here that rule also reports when the same value is added to the set.

rules/S4143/python/rule.adoc
```diff
@@ -2,14 +2,15 @@
 
 :link-with-uscores1: https://docs.python.org/3/reference/datamodel.html#object.__setitem__
 
-It is highly suspicious when a value is saved in a collection for a given key or index and then unconditionally overwritten. Such replacements are likely errors.
+Storing a value inside a collection at a given key or index and then unconditionally overwriting it without reading the initial value is a case of "dead store".
```
vilchik-elena, line 5: I would reference the top level description to avoid duplication

rules/S4143/python/rule.adoc
```diff
@@ -2,14 +2,15 @@
...
 
+At best it is a useless operation that will confuse the reader, most often it is an error and not what the developer intended to do.
 
 This rule raises an issue when the {link-with-uscores1}[``++__setitem__++``] method of the same object is called multiple times with the same index, slice or key without any other action done between the calls.
```
vilchik-elena, line 9: IMHO it's confusing to reference here this internal method.
I would drop this paragraph, general description seems enough.
WDYT?

rules/S4143/rule.adoc
```diff
@@ -2,4 +2,24 @@
 
 include::description.adoc[]
```
vilchik-elena, line 3: I seems only java is using this file. I suggest we move it into java subdir

vilchik-elena (changes requested):

rules/S4143/python/rule.adoc
frederic-tingaud-sonarsource: I'm not confident enough in how necessary this explanation is, so I'd let @guillaume-dequenne-sonarsource decide.

rules/S4143/javascript/rule.adoc
frederic-tingaud-sonarsource: Done

rules/S4143/rule.adoc
frederic-tingaud-sonarsource: I see a lot more languages using it (apex, cobol, flex, ruby and multiple others), although they might not actually implement the rule.

rules/S4143/python/rule.adoc
frederic-tingaud-sonarsource: done

rules/S4143/python/rule.adoc
guillaume-dequenne-sonarsource: I think it's fair to drop this: it's not even actually true, since direct calls to `__setitem__` are not covered in the current implementation of the rule. I think this was an intention from the initial specification that was never implemented.

rules/S4143/description.adoc
```diff
@@ -1 +1,3 @@
-It is highly suspicious when a value is saved for a key or index and then unconditionally overwritten. Such replacements are likely errors.
+Storing a value inside a collection at a given key or index and then unconditionally overwriting it without reading the initial value is a case of "dead store".
+
+At best it is a useless operation that will confuse the reader, most often it is an error and not what the developer intended to do.
```
guillaume-dequenne-sonarsource, line 3: By the way, we sometimes received the feedback that the word `useless` is generally considered to be quite strong/aggressive (and can sound really bad in case of false positives) and that rephrasing to something like `redundant` is generally better.
Not sure if our stance has changed, but you might want to consider this.

frederic-tingaud-sonarsource: I like redundant in this context. It is clearer. Thanks!

rules/S4143/python/rule.adoc
frederic-tingaud-sonarsource: done

vilchik-elena:
@frederic-tingaud-sonarsource merged and update the spreadsheet

PR #1933, opened 2023-05-23
Modify rule S1528: add quickfix

alexander-kamushkin-sonarsource:
https://github.com/SonarSource/SonarJS/issues/3887


yassin-kammoun-sonarsource (approved):
LGTM

PR #1935, opened 2023-05-23
Modify rule S4144: Adapt to LaYC format

frederic-tingaud-sonarsource:


frederic-tingaud-sonarsource:
I don't understand this sentence in the description:
"Numerical and string literals are not taken into account". Does it mean that two methods with identical code except for a numerical value will be considered as identical?

rules/S4144/python/rule.adoc
```diff
@@ -39,6 +35,8 @@ class MyClass:
         pass  # on purpose
 ----
 
+Numerical and string literals are not taken into account.
```
frederic-tingaud-sonarsource, line 38: @guillaume-dequenne-sonarsource do you understand this sentence? I moved it here, but I don't understand what it tries to say. IMO, we should reformulate it, remove it, or put some code example to make it clear.

rules/S4144/cfamily/rule.adoc
```diff
@@ -63,6 +56,8 @@ public:
 };
 ----
 
+This rule raises an exception when two methods implemented inside a class definition share the same implementation.
```
vilchik-elena, line 59:
```suggestion
This rule raises an issue when two methods implemented inside a class definition have the same implementation.
```

rules/S4144/description.adoc
```diff
@@ -1 +1 @@
-When two methods have the same implementation, either it was a mistake - something else was intended - or the duplication was intentional, but may be confusing to maintainers. In the latter case, one implementation should invoke the other. Numerical and string literals are not taken into account.
```
vilchik-elena, line 1: I see that you moved this sentence to the end, but for "go" for example it's not done

rules/S4144/python/rule.adoc
vilchik-elena: @frederic-tingaud-sonarsource it means that if 2 functions differ only by string literal value they are considered identical

rules/S4144/apex/rule.adoc
```diff
@@ -19,9 +18,7 @@ public String getName() {  // Noncompliant
 }
 ----
 
-=== Compliant solution
-
-[source,apex]
+[source,apex,diff-id=1,diff-type=compliant]
```
vilchik-elena, line 21: IMHO we should avoid 2 separate code blocks without any text between then.
What about `To avoid duplication one implementation can invoke the other.`

rules/S4144/apex/rule.adoc
```diff
@@ -19,9 +18,7 @@ public String getName() {  // Noncompliant
...
 ----
 private String code = 'bounteous';
```
vilchik-elena, line 24: the root `exceptions.adoc` file which is included here, does it make sense? is it used a lot? I wonder if it's better to inline it

Also I am not sure what the content of that file means:

```
Methods that are not accessors (getters and setters), with fewer than 2 statements are ignored.
```

Are accessors always ignored? or never?

vilchik-elena (changes requested):

rules/S4144/description.adoc
frederic-tingaud-sonarsource: The sentence is not here in the other languages unless I missed something. And I wonder if it's because they don't ignore numerical and literal values.

rules/S4144/apex/rule.adoc
frederic-tingaud-sonarsource: The file is used by the root rules.adoc and by PHP. I understand it as "getter and setter are never ignored, but it might be interesting to have the opinion of somebody from a squad that implemented this rule. Perhaps @leonardo-pilastri-sonarsource you can confirm?

rules/S4144/description.adoc
frederic-tingaud-sonarsource: Removing the sentence altogether because no language actually implements this.-

rules/S4144/apex/rule.adoc
leonardo-pilastri-sonarsource: I did not work on the rule but by having a look at least at the way the Java check works, it looks like all methods with less than 2 statements are ignored, **except** for accessors (getters and setters are considered even if they are 1-liners). So I think the exception text is trying to say that all methods are evaluated, except for the ones that are not getters/setters **AND** are less than 2 lines (IMO not super clear by reading it). As for Apex I was told the biztech team actually use the language, maybe you can ping them?

frederic-tingaud-sonarsource:
Adding @sebastian-hungerecker-sonarsource to review the Ruby code example: The idea is to change the example so that it explains how one can refactor to remove duplication. I am not even convinced that this code compiles, let alone looks like good Ruby.

rules/S4144/ruby/rule.adoc
sebastian-hungerecker-sonarsource, line 1: A couple of comments on the code:

* I don't understand the variable name `ray`. Does French use the same word for "ray" and "radius" by any chance? Anyway, I think "radius" is what's meant here.
* Ruby uses snake_case for method and variable names, not camelCase.
* Ruby does not use the `get_foo` / `set_foo` naming convention for getters and setters. Instead getters should just use the name of the variable and setters should use assignment syntax. So for an instance variable `@foo`, the getter should be named `foo` and the setter `foo=`.
* Conventionally, the parameter name of a setter should be the name of the variable being set.

With that in mind, the code should be:

```ruby
class Circle
  def initialize(radius)
    @radius = radius
  end

  def width=(width)
    self.diameter = width
  end

  def height=(height)
    self.diameter = height
  end

  def diameter=(diameter)  # Implementation is shared
    @radius = diameter / 2
    update_shape()
  end

  def update_shape()
    # ...
  end
end
```

frederic-tingaud-sonarsource:
Adding @guillaume-dequenne-sonarsource to review the Scala code example: The idea is to change the example so that it explains how one can refactor to remove duplication. I am not even convinced that this code compiles, let alone looks like good Scala.

rules/S4144/scala/rule.adoc
```diff
@@ -1,37 +1,45 @@
...
 
-  def area: Int = {
-    val s = length * width
-    s * height
+  def setHeight(size: Int) = { // Noncompliant: duplicates setWidth
+    radius = size / 2
+    updateShape()
   }
+
+  def updateShape() {...}
```
quentin-jaquier-sonarsource, line 20: _nitpick_
```suggestion
  def updateShape() = {...}
```

rules/S4144/scala/rule.adoc
```diff
@@ -1,37 +1,45 @@
...
   }
 
-  def area: Int = {
-    length * width
+  def setDiameter(size: Int) = { // Implementation is shared
+    radius = size / 2
+    updateShape()
   }
+
+  def updateShape() {...}
```
quentin-jaquier-sonarsource, line 42: _nitpick_
```suggestion
  def updateShape() = {...}
```

rules/S4144/scala/rule.adoc
```diff
@@ -1,37 +1,45 @@
...
+include::../description.adoc[]
 
-[source,scala]
+[source,scala,diff-id=1,diff-type=noncompliant]
 ----
-class Box(length: Int, width: Int, height: Int) {
-  def volume: Int = {
-    val s = length * width
-    s * height
+class Circle(radius: Int) {
```
quentin-jaquier-sonarsource, line 9: Fields are immutable by default in scala.
```suggestion
class Circle(var radius: Int) {
```

rules/S4144/scala/rule.adoc
```diff
@@ -1,37 +1,45 @@
...
+include::../inter.adoc[]
 
-[source,scala]
+[source,scala,diff-id=1,diff-type=compliant]
 ----
-class Box(length: Int, width: Int, height: Int) {
-  def volume: Int = {
-    val s = length * width
-    s * height
+class Circle(radius: Int) {
```
quentin-jaquier-sonarsource, line 28:
```suggestion
class Circle(var radius: Int) {
```

rules/S4144/scala/rule.adoc
```diff
@@ -1,37 +1,45 @@
...
+  def setWidth(size: Int) = {
+    setDiameter(size)
+  }
+
+  def setHeight(size: Int) = {
+    setDiameter(size)
   }
```
quentin-jaquier-sonarsource, line 35: _Optional:_ the previous code works fine, but I would say it is common practise to remove the block and directly go for:
```suggestion
  def setWidth(size: Int) = setDiameter(size)

  def setHeight(size: Int) = setDiameter(size)
```
(You could even go for `def setWidth = setDiameter`, but I don't think I would recommend this code...)

quentin-jaquier-sonarsource (approved):

rules/S4144/scala/rule.adoc
sebastian-hungerecker-sonarsource: I'd go with assignment syntax over `setFoo` (same as in Ruby):

```scala
def width_=(width: Int) {
  diameter = width
}

def diameter_=(diameter: Int) {
  radius = diameter / 2
  updateShape()
}
```

Also I don't think it's idiomatic to use `=` when there is no return value.

quentin-jaquier-sonarsource: Fair point, I guess my Scala skills are a bit rusty!

vilchik-elena (approved):

PR #1937, opened 2023-05-23
Modify rule S3981: Adapt to LaYC

frederic-tingaud-sonarsource:


rules/S3981/csharp/rule.adoc
```diff
@@ -2,29 +2,16 @@
...
 ----
-if(collection.Count >= 0){...}
-
-if(enumerable.Count() < 0){...}
-
-if(array.Length >= 0){...}
-
-bool result = array.Length >=0;
-----
+if(collection.Count >= 0){...} // Noncompliant always true
```
vilchik-elena, line 7: I suggest to add some punctuation after `// Noncompliant`, colon or comma

rules/S3981/csharp/rule.adoc
```diff
@@ -2,29 +2,16 @@
...
-[source,csharp]
+if(array.Length >= 0){...} // Noncompliant always true
 ----
-if (list.Any()) { ... }
 
-if (list.Count > 0) { ... }
-
-if (array.Length >= 42) { ... }
-----
+Either remove the redundant code if the current behavior is correct or fix the code to properly check for emptiness if it was the intent.
```
vilchik-elena, line 14: I don't think that there is ever the case when the current behavior is correct. In 99.999999% cases the comparison operator was chosen incorrectly.

rules/S3981/csharp/rule.adoc
```diff
@@ -2,29 +2,16 @@
```
vilchik-elena, line 2: How about something like this (splitting true and false cases)

```
The size of a collection and the length of an array are always greater than or equal to zero. So testing that size or length is greater than or equal to zero doesn't make sense, since the result is always ``++true++``.
[source,csharp]
----
if(collection.Count >= 0){...} // Noncompliant, always true
if(array.Length >= 0){...} // Noncompliant, always true
----
In this example, the condition is always true, and probably `>` should have been used. 

Similarly testing that it is less than zero will always return ``++false++``.
[source,csharp]
----
if(enumerable.Count() < 0){...} // Noncompliant, always false
----

Most often the intent is to check the (non)emptiness of the collection or array, and ` > 0` or ` == 0` operations should be used.

```

rules/S3981/kotlin/rule.adoc
```diff
@@ -1,10 +1,22 @@
 == Why is this an issue?
 
-include::description.adoc[]
+The size of a collection or an array is always greater than or equal to zero. So testing that a size is greater than or equal to zero doesn't make sense, since the result is always `true`. Similarly testing that it is less than zero will always return `false`.
 
-include::noncompliant.adoc[]
+[source,text]
+----
+if (myList.size >= 0) { // Noncompliant always true
```
vilchik-elena, line 7: I think it was nice and concise to use `{...}`
Still I wonder if it's better to comment ellipsis so that the code is valid (in that case multiline block will do better)

vilchik-elena (changes requested):

rules/S3981/csharp/rule.adoc
frederic-tingaud-sonarsource: Done

rules/S3981/csharp/rule.adoc
frederic-tingaud-sonarsource: Changed

rules/S3981/kotlin/rule.adoc
frederic-tingaud-sonarsource: Reverted

rules/S3981/csharp/rule.adoc
```diff
@@ -1,31 +1,24 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+The size of a collection and the length of an array are always greater than or equal to zero. So testing that a size or length is greater than or equal to zero doesn't make sense, since the result is always ``++true++``.
```
vilchik-elena, line 3:
```suggestion
The size of a collection and the length of an array are always greater than or equal to zero. So testing that doesn't make sense, since the result is always ``++true++``.
```

vilchik-elena (approved):
LGTM, but I suggested removing the repetition

PR #1943, opened 2023-05-25
Updated S2384 rule metadata to explicitly state that it applies only to private mutable members

ADarko22:


necto:
@ADarko22, please make sure the PR title and labels comply with [our policy](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule)

alban-auzeill (approved):

PR #1955, opened 2023-05-29
Modify rule S3433: Migrate to new educational format

cristian-ambrosini-sonarsource:
Removing `swift` rule since it's not implemented yet.
I'll leave a comment in the rule as soon as this PR is merged.


rules/S3433/csharp/rule.adoc
```diff
@@ -1,4 +1,61 @@
...
+[TestMethod]
+public void MyIgnoredTestMethod()
+{ /* ... */ }
+
+[TestMethod]
+public void MyIgnoredGenericTestMethod(int foo)
+{ /* ... */ }
+----
+
+=== Exceptions
```
sebastien-marichal, line 46: As per [documentation](https://github.com/SonarSource/rspec/blob/master/docs/description.adoc#2-education-format), the `Exceptions` should be under the `Why is this an issue?` section.

rules/S3433/csharp/rule.adoc
```diff
@@ -1,4 +1,61 @@
...
+=== Compliant solution
+
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+[TestMethod]
+public void TestNullArg()
+{  /* ... */  }
+
+[TestMethod]
+public void MyIgnoredTestMethod()
```
sebastien-marichal, line 38: Maybe not really important: an `async Task` compliant example could be nice.

rules/S3433/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+  "quickfix": "targeted"
```
sebastien-marichal, line 2: Following my comment on the `async Task` test method, not sure if the quick fix is feasible as there are at least two possibilities in this case. 
If we are not required to cover all cases with the quick fix, you can ignore and resolve this comment.

sebastien-marichal (changes requested):

rules/S3433/csharp/rule.adoc
cristian-ambrosini-sonarsource: Yes, and it should be under the proper section, right? :worried:
Maybe what's confusing you is that we don't have a "How to fix it" section in this rule so the compliant/noncompliant examples are inside the "Why" section. Either that or I'm missing something, let me know

rules/S3433/csharp/rule.adoc
cristian-ambrosini-sonarsource: Yeah that would be nice, my only concern is that having a 4/3 match on compliant/noncompliant examples can be confusing for users. WDYT?

rules/S3433/csharp/metadata.json
cristian-ambrosini-sonarsource: That's a good point, I'm assuming that `targeted` can result in either `partial` or `covered` once the quick fix is implemented so we should be fine.

rules/S3433/csharp/rule.adoc
sebastien-marichal: I was thinking of replacing this compliant solution with the `async Task` one, keeping the 3/3 match.

sebastien-marichal (approved):
LGTM!

PR #1956, opened 2023-05-29
Modify rule S2699: Migrate to new educational format

cristian-ambrosini-sonarsource:


rules/S2699/csharp/rule.adoc
```diff
@@ -13,57 +13,80 @@ This rule raises an exception when no assertions from any of the following frame
...
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 [TestMethod]
-public void MyMethod_WhenSomething_ExpectsSomething()
+public void TestGetFoo()
```
antonioaversa, line 34: I agree on changing the name of the test method to match the name of the tested method. 
However, why removing the `_WhenSomething_ExpectsSomething`?
I would keep it and make it into a real scenario. E.g. (compliant solution):

```cs
[TestMethod]
public void ByScalar_WhenMatrixInvertibleAndScalarNonZero_ExpectsInvertibleResult()
{
    var matrix = new Matrix(1, 2, 1, 3);
    MatrixAssert.IsInvertible(matrix.ByScalar(3));
}
```

rules/S2699/csharp/rule.adoc
```diff
@@ -1,9 +1,9 @@
 == Why is this an issue?
 
-A test case without assertions ensures only that no exceptions are thrown. Beyond basic runnability, it ensures nothing about the behavior of the code under test.
+The rule is targetting test methods that lack an assertion and consist solely of an action and optionally a setup. Such tests are designed to verify that the process runs without throwing an exception.
```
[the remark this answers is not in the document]
cristian-ambrosini-sonarsource: I would not consider it an exception, it's more a FP. I'll try to improve a bit the first part, let me know what you think 👍

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
+public class CustomTestExample
+{
+    [TestMethod]
+    public void Add_SingleNumber_ReturnsSameNumber() => Validator.Caclulator(0); // Compliant
+}
+
+public static class Validator
+{
+    [AssertionMethod]
+    public static void Caclulator(int value) =>
```
antonioaversa, line 46: `Caclulator` -> `Calculator`

Besides the name, the example doesn't compile (`StringCalculator` is not defined), and even if it was, the assertion method would not compile because it returns `void`, whereas the body is a boolean expression.

Moreover, `Calculator` isn't a typical name for an assertion method, which typically is something like `AssertX` or `BeX` as in `Should.BeX` etc. It also doesn't return a `bool`, but it raises exception when it fails, so that the test runner can report the test failure.

For this example to have value and not confuse the user, I think it should better capture a real scenario where users would use a custom assertion method, which is when `Assert.X` or `CollectionAssert.X` etc. are not enough.

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
+		new StringCalculator().Add(value.ToString()) == value;
+}
+
+public class AssertionMethodAttribute : Attribute { }
+----
+
+== Hot to fix it in MSTest
```
antonioaversa, line 53: I think [this list](https://github.com/SonarSource/rspec/blob/master/docs/header_names/allowed_framework_names.adoc) should also be extended, with the list of testing frameworks for .NET.

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
+public void Add_SingleNumber_ReturnsSameNumber()
 {
-    var myClass = new Class();
-    var result = myClass.GetFoo();
-    Assert.IsTrue(result);
+    var stringCalculator = new StringCalculator();
+
+    var actual = stringCalculator.Add("0");
+
+    Assert.AreEqual(0, actual);
```
antonioaversa, line 81: Too many empty lines:

```cs
var stringCalculator = new StringCalculator();
var actual = stringCalculator.Add("0");
Assert.AreEqual(0, actual);
```

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
 [TestMethod]
-public void MyMethod_WhenSomething_ExpectsSomething()
+public void Add_SingleNumber_ReturnsSameNumber()
```
antonioaversa, line 75: `Add_Zero_ReturnsSameNumber`?

If you add a single (digit) number, you don't (always) get the same number.

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
+[TestMethod]
+public void Add_SingleNumber_ReturnsSameNumber()
 {
-    [TestMethod]
-    public void TestMethod1() => Validator.CustomMethod(42);    // Compliant
+    var stringCalculator = new StringCalculator();
+
+    var actual = stringCalculator.Add("0");
+	
+	Assert.That(0, Is.EqualTo(actual));
```
antonioaversa, line 113: Indenting and empty lines

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
 [source,csharp]
 ----
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[TestClass]
+public class CustomTestExample
+{
+    [TestMethod]
+    public void Add_SingleNumber_ReturnsSameNumber() => Validator.Caclulator(0); // Compliant
```
antonioaversa, line 40: As in comment below, should the test be called `Add_Zero_ReturnsSameNumber`?

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
+=== Code examples
 
+==== Noncompliant code example
 
-[source,csharp]
+[source,csharp,diff-id=2,diff-type=noncompliant]
 ----
-using System;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+[TestMethod]
```
antonioaversa, line 93: The NUnit attribute is `Test`, defined in a `TestFixture`.
`TestMethod` is MSTest specific.
Check [here ](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-nunit)for some examples.

rules/S2699/csharp/rule.adoc
```diff
@@ -13,56 +14,149 @@ This rule raises an exception when no assertions from any of the following frame
...
-public static class Validator
+== Hot to fix it in xUnit
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=3,diff-type=noncompliant]
+----
+[TestMethod]
```
antonioaversa, line 125: xUnit uses `Fact` and `Theory`. `TestMethod` is MSTest specific.

Check [here](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test) for examples.

rules/S2699/csharp/rule.adoc
```diff
@@ -1,9 +1,10 @@
 == Why is this an issue?
 
-A test case without assertions ensures only that no exceptions are thrown. Beyond basic runnability, it ensures nothing about the behavior of the code under test.
+The rule is targeting test methods that lack an assertion and consist solely of an action and optionally a setup. 
+Such tests are only verifying that no exception is thrown, without providing any guarantees regarding the behavior of the code under test.
```
antonioaversa, line 4: It would be better in my opinion, to have a single noncompliant example, shown after this first paragraph.
That way, users would understand right away what we are talking about.

It would also avoid having non-compliant blocks for each testing framework, which look very similar to each other. It also renders less useful the fourth-level sections under code examples.

antonioaversa (changes requested):
More changes related to code examples and frameworks

antonioaversa (changes requested):
More changes related to code examples and frameworks

rules/S2699/csharp/rule.adoc
cristian-ambrosini-sonarsource: Good catch! I wonder why there are no tests that verify this (and thus what's the purpose of the list in the first place)

rules/S2699/csharp/rule.adoc
cristian-ambrosini-sonarsource: Yeah that was my fault. I just changed the assertion invocation based on the framework and not the attributes, thanks for catching it!

rules/S2699/csharp/rule.adoc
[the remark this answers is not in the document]
antonioaversa: Good!

rules/S2699/csharp/rule.adoc
cristian-ambrosini-sonarsource: You should see the diff visually, isn't it mandatory to have noncompliant/compliant pair inside a `Code Examples` subsection?

rules/S2699/csharp/rule.adoc
antonioaversa: I don't know exactly but I suspect it may just be for verification by the rspec tooling on CI, as I would infer from [here](https://github.com/SonarSource/rspec/blob/65008d6b56be390dd88703092d02473dda30ebdb/rspec-tools/tests/validation/test_description_validation.py#L93) and [here](https://github.com/SonarSource/rspec/blob/65008d6b56be390dd88703092d02473dda30ebdb/docs/description.adoc#L206).
Nothing user-facing on SQ or SC I think.

@arseniy-sonar : am I right?

rules/S2699/csharp/rule.adoc
antonioaversa: After discussion, adding an example in the why section after the first paragraph is enough for the user to grasp what we are talking about. No changes to How to fix sections are required.=

rules/S2699/csharp/rule.adoc
```diff
@@ -13,57 +14,147 @@ This rule raises an exception when no assertions from any of the following frame
...
+{
+    [TestMethod]
+    public void Add_SingleNumber_ReturnsSameNumber() => Validator.IsSameNumber(0); // Compliant
+}
+
+public static class Validator
+{
+    [AssertionMethod]
+    public static void IsSameNumber(int value)
+	{
```
antonioaversa, line 47: `{` should be vertically aligned to `public static void IsSameNumber(int value)`.

rules/S2699/csharp/rule.adoc
necto: Yes. This list is maintained to make sure we use consistent spelling for the frameworks. If the framework used in the title of this section is missing from the list, it should fail the rule-description validation in the CI check. Let me know if that does not happen

rules/S2699/csharp/rule.adoc
```diff
@@ -13,57 +14,147 @@ This rule raises an exception when no assertions from any of the following frame
...
+
+In addition to using built-in assertion methods, you also have the option to create custom assertion methods. To do this, declare an attribute named `[AssertionMethodAttribute]` and apply it to the respective method. This allows you to encapsulate specific validation logic within your custom assertion methods without raising the issue.
+Here's an example:
 
 [source,csharp]
 ----
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[TestClass]
```
antonioaversa, line 36: Alternative:
```cs
    [TestClass]
    public class CustomTestExample
    {
        [TestMethod]
        public void Add_SingleNumber_ReturnsSameNumber()
        {
            var stringCalculator = new StringCalculator();
            var actual = stringCalculator.Add("0");
            Validator.AssertCustomEquality(0, actual); // Compliant
        }
    }

    public static class Validator
    {
        [AssertionMethod]
        public static void AssertCustomEquality(int expected, int actual)
        {
            // ...
        }
    }
```

antonioaversa (approved):
LGTM
Open discussions to be addressed first.

rules/S2699/csharp/rule.adoc
cristian-ambrosini-sonarsource: If I remember correctly the CI was green even tho those 3 test frameworks were missing from the list. I can do a quick test later to verify that

PR #1961, opened 2023-05-30
Modify rule S1111: more context to Why is this an issue

leonardo-pilastri-sonarsource:


rules/S1111/java/rule.adoc
```diff
@@ -1,6 +1,10 @@
 == Why is this an issue?
 
-According to the official javadoc documentation, this Object.finalize() is called by the garbage collector on an object when garbage collection determines that there are no more references to the object. Calling this method explicitly breaks this contract and so is misleading. 
+According to the official javadoc documentation, `Object.finalize()` is called by the garbage collector on an object when garbage collection determines that there are no more references to the object. The general contract is that the `finalize` method should only be called once per object, thus calling this method explicitly is misleading and does not respect this contract.
```
dorian-burihabwa-sonarsource, line 3: Let's pick a more active voice and split the sentences over their own respective line
```suggestion
Before it reclaims storage from an object that is no longer referenced, the garbage collector calls `finalize()` on the object.
This is a good time to release resources held by the object.
Because the general contract is that the `finalize` method should only be called once per object, calling this method explicitly is misleading and does not respect this contract. 
```

rules/S1111/java/rule.adoc
```diff
@@ -1,6 +1,10 @@
 == Why is this an issue?
 
-According to the official javadoc documentation, this Object.finalize() is called by the garbage collector on an object when garbage collection determines that there are no more references to the object. Calling this method explicitly breaks this contract and so is misleading. 
+According to the official javadoc documentation, `Object.finalize()` is called by the garbage collector on an object when garbage collection determines that there are no more references to the object. The general contract is that the `finalize` method should only be called once per object, thus calling this method explicitly is misleading and does not respect this contract. 
+
+=== What is the potential impact?
+
+An explicit call to an Object's finalize method will perform operations that most likely were supposed to be performed only when the Object was not referenced anymore by any thread. Since it is an acceptable practice to override the finalize method in any subclass of Object, by invoking it explicitly, we will run code that was designed to only be ran at a different time.
```
dorian-burihabwa-sonarsource, line 7: Split over 2 lines
```suggestion
An explicit call to an Object's finalize method will perform operations that most likely were supposed to be performed only when the Object was not referenced anymore by any thread.
Since it is an acceptable practice to override the finalize method in any subclass of Object, by invoking it explicitly, we will run code that was designed to only be ran at a different time.
```

dorian-burihabwa-sonarsource (changes requested):
Let's make sure that we apply the writing principles discussed in the workshop and reference from first hand resource like the JLS[0]

[0] https://docs.oracle.com/javase/specs/jls/se17/html/jls-12.html#jls-12.6

PR #1962, opened 2023-05-30
Modify rule S2190: LaYC format

gregory-paidis-sonarsource:
Also remove unused languages.


rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+Non-termination occurs when control enters a loop that lacks an exit condition. 
+It can arise from one of the following scenarios:
 
-=== Noncompliant code example
+1. a method invoking itself
```
vilchik-elena, line 6: If you want to be so well-structured and split the possible cases in the list I would recommend putting a code example after each. That way reader will quickly grasp what you mean.

rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
...
+2. a group of methods mutually invoking each other
+3. `while` and `for` loops
+4. `goto` statements 
 
-[source,csharp]
+=== What is the potential impact?
+
+The program will either be looping indefinitely or the stack will reach its limit and the program will crash due to a stack overflow error.
```
vilchik-elena, line 13: again, if the rule about recursion program will not be looping indefinitely.

rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
 == Why is this an issue?
```
vilchik-elena, line 1: I believe this rule deserves a proper explanation of what is recursion and its base case
Mention kinds of recursion (direct and indirect), why recursion is good, and why dangerous.

rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
...
 
-include::../description.adoc[]
+Non-termination occurs when control enters a loop that lacks an exit condition.
```
vilchik-elena, line 3: I am afraid the word "loop" might be confusing in the context of recursion

rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
...
+=== What is the potential impact?
+
+The program will either be looping indefinitely or the stack will reach its limit and the program will crash due to a stack overflow error.
+
+== How to fix it
+
+The program's logic should incorporate a mechanism to break out of the loop.
+Here are some ways to achieve that:
+
+1. self-invoking should happen under a condition that eventually evaluates to `false`
```
vilchik-elena, line 20: You should not assume that reader remembers the order of scenarios from another tab. This tab should be understandable on its own

rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
...
-  end:
-    goto start; // Noncompliant; there's no way to break out of this method
-}
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
```
vilchik-elena, line 52: Similar here add a paragraph about the fix

rules/S2190/csharp/rule.adoc
```diff
@@ -52,6 +72,21 @@ void WhileLoop()
   }
 }
 ----
+
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/iteration-statements#the-for-statement["for" statement]
```
vilchik-elena, line 80: I would add `the` like in the documentation 
`The "for" statement`

vilchik-elena (changes requested):

rules/S2190/csharp/rule.adoc
```diff
@@ -1,48 +1,68 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+Non-termination occurs when control enters a loop that lacks an exit condition. 
+It can arise from one of the following scenarios:
 
-=== Noncompliant code example
+1. a method invoking itself
+2. a group of methods mutually invoking each other
+3. `while` and `for` loops
```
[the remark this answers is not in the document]
gregory-paidis-sonarsource: So, the rule applies to both, because it checks endless execution via the CFG.
Do you think we should rename it to something like `Loops and recursions should not be infinite`?

rules/S2190/csharp/rule.adoc
gregory-paidis-sonarsource: Same as the title, if applies to loops as well.

rules/S2190/csharp/rule.adoc
gregory-paidis-sonarsource: I rewrote it in a way that it starts talking about looping and then describes recursion as well, so it should be a bit more clear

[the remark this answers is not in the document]
gregory-paidis-sonarsource: Removed non-termination from the entire text

rules/S2190/csharp/rule.adoc
```diff
@@ -1,58 +1,186 @@
...
   {  
-    num = num * Pow(num, exponent-1);
+    num = num * Pow(num, exponent - 1);
   }
   return num;
 }
+----
 
-void WhileLoop()
+In this example, `Pow` will call `Pow` only if the exponent is greater than one, while passing `exponent-1` to the method, a smaller value than the one it received, `exponent`. This means that the recursion will run `exponent-1` times, until `exponent` is equal to one, and then return the result.
```
vilchik-elena, line 145: I would drop this paragraph. You try to literally describe what is happening in the code, and you are not doing it very precise. E.g. `recursion will run `exponent-1` times`: not recursion, but method, and which value of exponent you imply in "`exponent-1` times"

The first sentence is too complicated

rules/S2190/csharp/rule.adoc
```diff
@@ -1,58 +1,186 @@
...
+
+int Bar(int n) 
+{
+  Console.WriteLine("bar");
+  Foo(n-1);
+}
 ----
-ifdef::env-github,rspecator-view[]
+
+In this example, `Foo` is invoking `Bar` conditionally, only if `n` is a positive number. Also, both `Foo` and `Bar` are decrementing the value of `n`, which means it will always decrease, until it reaches zero and exit `Foo`, without calling `Bar` again.
```
vilchik-elena, line 166: This explanation is more confusing than helping. Do we need the both methods to decrement the value to avoid the loop?

rules/S2190/csharp/rule.adoc
```diff
@@ -1,58 +1,186 @@
...
+* **Performance overhead**: Recursive function calls can lead to poor performance due to the need to push and pop https://en.citizendium.org/wiki/Stack_frame#:~:text=In%20computer%20science%2C%20a%20stack,only%20exist%20at%20run%2Dtime[stack frames], making them potentially slower than iterative solutions.
+* **Difficulty in debugging**: Debugging recursive code can be challenging, as multiple recursive calls can make it harder to track the flow of execution and identify logical errors.
+* **Space complexity**: Recursive algorithms may require more memory compared to iterative approaches, as each recursive call adds a new frame to the call stack.
+* **Lack of control**: Recursion can sometimes lead to infinite loops or unexpected behavior if not properly implemented or terminated, making it crucial to have proper base cases and exit conditions.
+
+
+== How to fix it
+
+The program's logic should incorporate a mechanism to break out of the loop.
+Here are some ways to achieve that:
```
vilchik-elena, line 96: I am not sure it's appropriate to say so, when each way corresponds to a different original problem

rules/S2190/csharp/rule.adoc
```diff
@@ -1,58 +1,186 @@
...
+  Console.WriteLine("bar");
+  Foo(n-1); // always calls Foo
+}
+----
+
+In this example, `Foo` will always call `Bar`, which will always call `Foo`, etc. So the program will never terminate.
+
+=== Recursion in Depth
+
+Recursion can have a lot of benefits:
```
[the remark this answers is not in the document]
gregory-paidis-sonarsource: rewrote it as `some` instead of `a lot` :)

rules/S2190/csharp/rule.adoc
gregory-paidis-sonarsource: Dropped

rules/S2190/csharp/rule.adoc
gregory-paidis-sonarsource: I refactored the examples to only have `n-1` inside `Foo`, and removed this sentence, as I do not believe it adds much value

vilchik-elena (approved):

PR #1965, opened 2023-05-30
S1909: Remove C# since the rule is not applicable 

gregory-paidis-sonarsource:
See [this conversation](https://github.com/SonarSource/sonar-dotnet/issues/6661) for more details.


csaba-sagi-sonarsource (changes requested):
According to the guidelines, we should not remove the content, but mark the status as closed. See: https://github.com/SonarSource/rspec#to-delete-a-rule

csaba-sagi-sonarsource (approved):
LGTM!

PR #1966, opened 2023-05-30
Modify rule S3464: LaYC format

antonioaversa:
RSpec: https://sonarsource.github.io/rspec/#/rspec/S3464

### DIFF
--- a/rules/S3464/csharp/rule.adoc
+++ b/rules/S3464/csharp/rule.adoc
@@ -1,22 +1,94 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+Recursion is a technique used in computer science to define a problem in terms of the problem itself, usually in terms of a simpler version of the problem itself.
 
-=== Noncompliant code example
+For example the implementation of the generator for the n-th value of the Fibonacci sequence comes naturally from its mathematical definition, when recursion is used:
+
+[source,csharp]
+----
+static int NthFibonacciNumber(int n)
+{
+    if (n <= 1)
+    {
+        return 1;
+    }
+    else
+    {
+        return NthFibonacciNumber(n - 1) + NthFibonacciNumber(n - 2);
+    }
+}
+----
+
+As opposed to:
+
+[source,csharp]
+----
+static int NthFibonacciNumber(int n)
+{
+    int previous = 0;
+	int last = 1;
+	for (var i = 0; i < n; i++)
+	{
+        (previous, last) = (last, last + previous);
+	}
+	return last;
+}
+----
+
+Use of recursion is acceptable in methods, like the one above, where you can break out of it. 
+
+[source,csharp]
+----
+static int NthFibonacciNumber(int n)
+{
+    if (n <= 1)
+    {
+        return 1; // Base case => stop recurring
+    }
+    // ...
+}
+----
+
+It is also acceptable and makes sense in some type definitions:
+
+[source,csharp]
+----
+class Box<T> : IComparable<Box<T>>
+{
+    public int CompareTo(Box<T>? other)
+    {
+        // Compare the two Box<T> instances...
+    }
+}
+----
+
+With types, some invalid recursive definitions are catched by the compiler:
+
+[source,csharp]
+----
+class C2<T> : C2<T>     // Error CS0146: Circular base type dependency
+{
+}
+
+class C2<T> : C2<C2<T>> // Error CS0146: Circular base type dependency
+{
+}
+----
+
+In more complex scenarios, however, the code will compile but execution will result into a `TypeLoadException` if you try to instantiate the class.
 
 [source,csharp]
 ----
 class C1<T>
 {
 }
+
 class C2<T> : C1<C2<C2<T>>> // Noncompliant
 {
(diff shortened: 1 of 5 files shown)


rules/S3464/csharp/rule.adoc
```diff
@@ -1,23 +1,95 @@
...
 
-include::../description.adoc[]
+Recursion is a technique used in computer science to define a problem in terms of the problem itself, usually in terms of a simpler version of the problem itself.
 
-=== Noncompliant code example
+For example the implementation of the generator for the n-th value of the Fibonacci sequence comes naturally from its mathematical definition, when recursion is used:
+
+[source,csharp]
+----
+static int NthFibonacciNumber(int n)
```
gregory-paidis-sonarsource, line 9: I think you can remove `static` to simplify the examples.
Applies to all code snippets.

rules/S3464/csharp/rule.adoc
```diff
@@ -1,23 +1,95 @@
...
+
+Use of recursion is acceptable in methods, like the one above, where you can break out of it. 
+
+[source,csharp]
+----
+static int NthFibonacciNumber(int n)
+{
+    if (n <= 1)
+    {
+        return 1; // Base case => stop recurring
```
gregory-paidis-sonarsource, line 46: (Nitpick, optional)
The arrow seems like a lambda to me, maybe use a colon:

`Base case: ...`

Also, `recurring` feels slightly unnatural, maybe something like:

`(stop|exit) the recursive calls` or `(stop|exit) the recursion`

rules/S3464/csharp/rule.adoc
```diff
@@ -1,23 +1,95 @@
...
+    }
+    // ...
+}
+----
+
+It is also acceptable and makes sense in some type definitions:
+
+[source,csharp]
+----
+class Box<T> : IComparable<Box<T>>
```
gregory-paidis-sonarsource, line 56: Do you need the CRTP _target_ to be generic? Is it semantically different than this, as far as the pattern is concerned:

```csharp
class Box: IComparable<Box> { }
```

I feel like if you try to convey CRTP, having one less generic makes it a bit more readable/understandable.

rules/S3464/csharp/rule.adoc
```diff
@@ -1,23 +1,95 @@
...
+{
+    public int CompareTo(Box<T>? other)
+    {
+        // Compare the two Box<T> instances...
+    }
+}
+----
+
+With types, some invalid recursive definitions are catched by the compiler:
```
gregory-paidis-sonarsource, line 65: "catched" -> "caught"

rules/S3464/csharp/rule.adoc
```diff
@@ -1,23 +1,95 @@
...
+class C2<T> : C2<T>     // Error CS0146: Circular base type dependency
+{
+}
+
+class C2<T> : C2<C2<T>> // Error CS0146: Circular base type dependency
+{
+}
+----
+
+In more complex scenarios, however, the code will compile but execution will result into a `TypeLoadException` if you try to instantiate the class.
```
gregory-paidis-sonarsource, line 78: Either here or in "Resources" tab, you can add this:

- [TypeLoadException](https://learn.microsoft.com/en-us/dotnet/api/system.typeloadexception?view=net-7.0)

Some more suggestions:
- [Recursion - wiki](https://en.wikipedia.org/wiki/Recursion_(computer_science))
- [Clippert's "Curiouser and curiouser"](https://ericlippert.com/2011/02/02/curiouser-and-curiouser/)
- [Cleary's CRGP](https://blog.stephencleary.com/2022/09/modern-csharp-techniques-1-curiously-recurring-generic-pattern.html)

You can also add [wiki's CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern), but it's basically all about C++

gregory-paidis-sonarsource (changes requested):
Looks very readable, left a couple of suggestions. Consider most stuff for VB as well, since it's very similar.

rules/S3464/csharp/rule.adoc
antonioaversa: Yes, I don't think it is actually needed and it doesn't add value. Removed.

rules/S3464/csharp/rule.adoc
antonioaversa: Nice catch :-)

rules/S3464/csharp/rule.adoc
```diff
@@ -1,22 +1,96 @@
...
```
gregory-paidis-sonarsource, line 78: Consider adding the link for `TypeLoadException` here as well, as we discussed.

gregory-paidis-sonarsource (changes requested):
Some more minor changes suggested.

rules/S3464/csharp/rule.adoc
antonioaversa: whoops, fixed!

gregory-paidis-sonarsource (approved):
LGTM!

PR #1969, opened 2023-05-31
Modify rule S4622: Add exception for TypeScript Utility Types

yassin-kammoun-sonarsource:


ilia-kebets-sonarsource (approved):
That's great, could you also add a link to utility types?

PR #1970, opened 2023-05-31
Modify rule S1114: Expand `Why is this an issue` and add resource

dorian-burihabwa-sonarsource:


dorian-burihabwa-sonarsource:
Closing because the rule will likely be deprecated in the future

PR #1971, opened 2023-05-31
Modify rule S3443: LaYC format

sebastien-marichal:


rules/S3443/csharp/rule.adoc
```diff
@@ -1,12 +1,28 @@
...
 
-If you call ``++GetType()++`` on a ``++Type++`` variable, the return value will always be ``++typeof(System.Type)++``. So there's no real point in making that call. The same applies to passing a type argument to ``++IsInstanceOfType++``. In both cases the results are entirely predictable.
+Calling `GetType` on a variable of type `Type` will always return the `Type` representation of `System.Type` (aka. `typeof(System.Type)`). This also applies to passing a `Type` argument to `IsInstanceOfType`.
 
+In both cases, the results are entirely predictable and should be avoided.
 
-=== Noncompliant code example
+=== Exceptions
+
+Calling `GetType` on `System.Type` is considered compliant to get an instance of `System.RuntimeType`, like in the following example:
```
cristian-ambrosini-sonarsource, line 9: Educational: so if I understood correctly:
```cs
string str = string.Empty;

Type strType = str.GetType(); // Compliant (System.String)
Type strTypeType = strType.GetType(); // Noncompliant (System.RuntimeType)
typeof(Type).GetType(); // Compliant (System.RuntimeType)
```

So the exception of this rule is specifically the line `typeof(Type).GetType();` to allow one way to retrieve System.RuntimeType without raising the issue. Am I right?

rules/S3443/csharp/rule.adoc
```diff
@@ -1,12 +1,28 @@
...
+Make sure the usage of `GetType` or `IsInstanceOfType` is invoked with correct variable or argument type.
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
 var intType = typeof(int);
 var runtimeType = intType.GetType(); // Noncompliant, always typeof(System.RuntimeType)
```
cristian-ambrosini-sonarsource, line 27: This first example is not fixed in the compliant example, right?
WDYT about a test like the following?

```cs
void ExamineSystemType(string str, Type type) 
{
    type.GetType(); // Noncompliant
    
    if (type.IsInstanceOfType(typeof(string))) // Noncompliant
    { /* ... */ }
}
```
and then the compliant version:
```cs
void ExamineSystemType(string str, Type type) 
{
    str.GetType(); // Noncompliant
    
    if (str.IsInstanceOfType(typeof(string))) // Noncompliant
    { /* ... */ }
}
```

cristian-ambrosini-sonarsource (changes requested):
Looks good! I suggested a few changes

rules/S3443/csharp/rule.adoc
```diff
@@ -1,12 +1,28 @@
 == Why is this an issue?
 
-If you call ``++GetType()++`` on a ``++Type++`` variable, the return value will always be ``++typeof(System.Type)++``. So there's no real point in making that call. The same applies to passing a type argument to ``++IsInstanceOfType++``. In both cases the results are entirely predictable.
+Calling `GetType` on a variable of type `Type` will always return the `Type` representation of `System.Type` (aka. `typeof(System.Type)`). This also applies to passing a `Type` argument to `IsInstanceOfType`.
```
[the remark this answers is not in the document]
sebastien-marichal: I wanted to have a clearer sentence, I think the old version is not easy to understand for junior developers. And I agree, `type` is used too many times.

cristian-ambrosini-sonarsource (approved):
LGTM!

PR #1975, opened 2023-05-31
Modify rule S2275: LaYC format

cristian-ambrosini-sonarsource:


rules/S2275/csharp/rule.adoc
```diff
@@ -27,24 +40,42 @@ s = string.Format("no format"); // Compliant
...
 
 [source,csharp]
 ----
 var pattern = "{0} {1} {2}";
-var res = string.Format(pattern, 1, 2); // Compliant, not const string are not recognized
+var res = string.Format(pattern, 1, 2); // Compliant, non-constant string are not recognized
 ----
 
-* No issue is raised if the argument is not an inline creation array.
+* the argument is not an inline creation array.
```
gregory-paidis-sonarsource, line 54: I think it's better to say "array created inline" than "inline creation array"

gregory-paidis-sonarsource (approved):
LGTM, left a minor comment, feel free to decide if you agree or not.

PR #1976, opened 2023-05-31
Modify rule S6232: CPP-3254 Well defined alternatives to union type-punning

LoghinVladDev:
Also using LaYC format for specialized examples

### DIFF
--- a/rules/S6232/cfamily/metadata.json
+++ b/rules/S6232/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "\"std::bit_cast\" should be used instead of union type-punning",
+  "title": "Well defined type-punning should be used instead of union type-punning",
   "type": "BUG",
   "status": "ready",
   "remediation": {
@@ -7,7 +7,6 @@
     "constantCost": "5min"
   },
   "tags": [
-    "since-c++20",
     "symbolic-execution",
     "bad-practice",
     "pitfall"
--- a/rules/S6232/cfamily/metadata.json
+++ b/rules/S6232/cfamily/metadata.json
@@ -26,5 +26,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "targeted"
 }
--- a/rules/S6232/cfamily/metadata.json
+++ b/rules/S6232/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "Well defined type-punning should be used instead of union type-punning",
+  "title": "Well-defined type-punning method should be used instead of a union-based one",
   "type": "BUG",
   "status": "ready",
   "remediation": {
--- a/rules/S6232/cfamily/rule.adoc
+++ b/rules/S6232/cfamily/rule.adoc
@@ -1,19 +1,18 @@
 == Why is this an issue?
 
-// {cpp}20 introduced the ``++std::bit_cast++`` function. It standardizes the diverse and sub-optimal approaches of reinterpreting a value as a different type of the same length preserving its binary representation.
 
-In some performance-oriented algorithms, a solution to certain slow operations is reinterpreting a value as a different type of the same length, while preserving its binary representation.
+In some performance-oriented algorithms, a solution to certain slow operations is reinterpreting a value as a different type of the same length while preserving its binary representation.
 
 
-One of the superseded solutions, know as "union type-punning", is to use a `union` with two members with types corresponding to the source and the target types of the cast.
-The operation is performed by saving the value in the member with source type and then reading the value of the target type.
+One of the superseded solutions, known as "union type-punning", is to use a `union` with two members with types corresponding to the source and the target types of the cast.
+The operation is performed by saving the value in the member with the source type and then reading the value of the target type.
 Despite being allowed in C, this operation has undefined behavior according to {cpp} standard and should be replaced by `std::memcpy` (or `std::bit_cast` in {cpp}20).
 
 
-Union type-punning can also be used to remove `const`. This can create readability issues, and should be replaced with `const_cast`.
+Union type-punning can also be used to remove `const`. This can create readability issues and should be replaced with `const_cast`.
 
 
-This rule raises an issue on any use of a `union` that should be replaced with `std::memcpy`, `std::bit_cast` or `const_cast`.
+This rule raises an issue on any use of a `union` that should be replaced with `std::memcpy`, `std::bit_cast`, or `const_cast`.
 
 
 == How to fix it
--- a/rules/S6232/cfamily/rule.adoc
+++ b/rules/S6232/cfamily/rule.adoc
@@ -1,15 +1,28 @@
 == Why is this an issue?
 
-{cpp}20 introduced the ``++std::bit_cast++`` function. It standardizes the diverse and sub-optimal approaches of reinterpreting a value as a different type of the same length preserving its binary representation.
+// {cpp}20 introduced the ``++std::bit_cast++`` function. It standardizes the diverse and sub-optimal approaches of reinterpreting a value as a different type of the same length preserving its binary representation.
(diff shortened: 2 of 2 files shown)


rules/S6232/cfamily/rule.adoc
```diff
@@ -48,7 +80,7 @@ float fastInvSqrt(float number) {
 
 == Resources
 
-* S6181 - replacing ``++std::memcpy++`` with ``++std::bit_cast++``.
+* S6181 - replacing `std::memcpy` with `std::bit_cast++`.
```
necto, line 83: Looks like you've unintentionally left the extra `++` remain
```suggestion
* S6181 - replacing `std::memcpy` with `std::bit_cast`.
```

rules/S6232/cfamily/metadata.json
```diff
@@ -1,13 +1,12 @@
 {
-  "title": "\"std::bit_cast\" should be used instead of union type-punning",
+  "title": "Well defined type-punning should be used instead of union type-punning",
```
necto, line 2: "type-punning" is quite a heavy word to repeat in a single sentence.
What do you think of the following options?
```suggestion
  "title": "Well-defined alternative should be used instead of union type-punning",
```
```suggestion
  "title": "Well-defined type-punning method should be used instead of a union-based one",
```

rules/S6232/cfamily/rule.adoc
```diff
@@ -1,15 +1,27 @@
...
 
+In some performance-oriented algorithms, a solution to certain slow operations is reinterpreting a value as a different type of the same length while preserving its binary representation.
 
-One of the superseded solutions, know as "union type-punning", is to use a ``++union++`` with two members with types corresponding to the source and the target types of the cast. The operation is performed by saving the value in the member with source type and then reading the value of the target type. Despite being allowed in C, this operation has undefined behavior according to {cpp} standard and should be replaced by either ``++std::bit_cast++`` (or ``++std::memcpy++``).
 
+One of the superseded solutions, known as "union type-punning", is to use a `union` with two members with types corresponding to the source and the target types of the cast.
+The operation is performed by saving the value in the member with the source type and then reading the value of the target type.
+Despite being allowed in C, this operation has undefined behavior according to {cpp} standard and should be replaced by `std::memcpy` (or `std::bit_cast` in {cpp}20).
```
necto, line 9: Perhaps it is worth mentioning that `memcpy` should not have any performance impact, and will likely be removed during compilation (to be verified)

rules/S6232/cfamily/rule.adoc
```diff
@@ -28,8 +40,28 @@ float fastInvSqrt(float number)
 }
 ----
 
+==== Compliant solution
+
+Solution until {cpp}20:
```
necto, line 45:
```suggestion
Solution before {cpp}20:
```

rules/S6232/cfamily/rule.adoc
```diff
@@ -1,15 +1,27 @@
...
+Union type-punning can also be used to remove `const`. This can create readability issues and should be replaced with `const_cast`.
 
-=== Noncompliant code example
+
+This rule raises an issue on any use of a `union` that should be replaced with `std::memcpy`, `std::bit_cast`, or `const_cast`.
+
+
+== How to fix it
+
+Replace any "union type-punning" solution with a standard compliant solution.
```
necto, line 20:
```suggestion
Replace any "union type-punning" solution with a standard-compliant solution.
```

rules/S6232/cfamily/rule.adoc
```diff
@@ -1,15 +1,27 @@
...
 
-{cpp}20 introduced the ``++std::bit_cast++`` function. It standardizes the diverse and sub-optimal approaches of reinterpreting a value as a different type of the same length preserving its binary representation.
 
+In some performance-oriented algorithms, a solution to certain slow operations is reinterpreting a value as a different type of the same length while preserving its binary representation.
 
-One of the superseded solutions, know as "union type-punning", is to use a ``++union++`` with two members with types corresponding to the source and the target types of the cast. The operation is performed by saving the value in the member with source type and then reading the value of the target type. Despite being allowed in C, this operation has undefined behavior according to {cpp} standard and should be replaced by either ``++std::bit_cast++`` (or ``++std::memcpy++``).
 
+One of the superseded solutions, known as "union type-punning", is to use a `union` with two members with types corresponding to the source and the target types of the cast.
+The operation is performed by saving the value in the member with the source type and then reading the value of the target type.
```
necto, line 8:
```suggestion
The operation is performed by saving the value in the member of the source type and then reading the value from the member of the target type.
```

necto (changes requested):

necto:
Also, I would seize the opportunity and marked the "quickfix" as "targeted" because it looks feasible for some cases (but I'm not suggesting implementing it in the current sprint!)

rules/S6232/cfamily/rule.adoc
```diff
@@ -48,7 +80,7 @@ float fastInvSqrt(float number) {
 
 == Resources
```
necto, line 81: Perhaps, include the video Balazs mentioned in the ticket comment
Do you know other good resources on the subject?

necto:
It might be also worth adding the "What is the potential impact?" section citing one or two horrors of the non-conformance to the standard

rules/S6232/cfamily/rule.adoc
LoghinVladDev: Clang/Apple Clang/GCC/MSVC optimizes it to no performance impact. It's good to mention that most modern compilers optimize it

rules/S6232/cfamily/rule.adoc
```diff
@@ -28,8 +40,28 @@ float fastInvSqrt(float number)
...
+
+Solution before {cpp}20:
+
+[source,cpp]
+----
+float fastInvSqrt(float number) {
+  constexpr float threehalfs = 1.5F;
+  const float x2 = number * 0.5F;
+
+  std::uint32_t i;
```
LoghinVladDev, line 53: Given that in 'what is the potential impact' I want to specify size difference between architectures, would it be worth to change this to also include a size assert before cpp 20, since bit cast does this implicitly

```diff
+ static_assert(sizeof(i) == sizeof(number), "Safe type-punning not possible on this architecture");
std::memcpy(&i, &number, sizeof(float));
```

rules/S6232/cfamily/rule.adoc
```diff
@@ -1,15 +1,45 @@
...
+
+
+=== What is the potential impact?
+
+
+The {cpp} standard states that only one union member can be "active" at any time.
+A member becomes active once assigned a value, making the other union members "inactive".
+The standard also states that reading from an "inactive" member is undefined behavior.
+
+Since union type-punning relies on reading the "inactive" member, code using it is undefined behavior.
```
necto, line 28:
```suggestion
Since union type-punning relies on reading the "inactive" member, code using it exercises undefined behavior.
```
Grammatically, "code" cannot be "behavior"

rules/S6232/cfamily/rule.adoc
```diff
@@ -1,15 +1,45 @@
...
+Such code can be unintentionally removed in aggressive levels of optimization.
+
+Further problems could also arise from using union-type punning in cross-platform solutions.
+Since this method is mainly used with Fundamental Types, which vary in size depending on the underlying architecture, it could also create undefined behavior.
+
+
+
+== How to fix it
+
+Replace any "union type-punning" solution with a standard-compliant solution.
```
necto, line 38:
```suggestion
Replace any "union-based type punning" solution with a standard-compliant solution.
```

necto (approved):
I couple of nits, and looking great otherwise!

Note that you use different spellings: "union-type punning" and "union type-punning". I propose (in the change suggestions below) to stick to the term "union-based type punning"

PR #1981, opened 2023-05-31
Modify rule S2368: LaYC format

sebastien-marichal:


antonioaversa (changes requested):
A bit more education on what multidimensional and jagged arrays are, and some very minor cosmetic suggestions.

rules/S2368/why-dotnet.adoc
```diff
@@ -0,0 +1,4 @@
+Using multidimensional and jagged arrays as function parameters in C# can be challenging for developers. When these methods are exposed to external users, it requires advanced knowledge of the language for effective usage. Additionally, determining the appropriate data to pass to these parameters may not be intuitive. As a result, it is recommended to avoid exposing such methods to external users.
```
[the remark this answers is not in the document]
sebastien-marichal: Having a paragraph about what is multi-dimensional and jagged array seems too much to be in the `why` in my opinion. For me, it should be added as documentation or article resources.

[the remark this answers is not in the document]
sebastien-marichal: I improve the `why` section to include the code example. Didn't want to add a too complex example since we could lose junior devs. Hopefully, the one I added is good enough.
Happy to have your feedback on those.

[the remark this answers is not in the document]
antonioaversa: I think it's good!

rules/S2368/csharp/exceptionCode.adoc
```diff
@@ -0,0 +1,10 @@
+[source,csharp]
+----
+public class Program
```
antonioaversa, line 3: Is the `public class Program` class necessary? 
The why section of the rule mentions "they are not publicly exposed".
If the rule also checks for type visibility, in addition to method visibility (effective visibility), then it makes sense to keep the class and clearly show its visibility, since it would have an effect on whether the rule would be reported or not. And maybe make a bit clearer what `publicly visible` means. 
If the rule only checks for method visibility, then I would remove the class and only show the method, with its visibility.

rules/S2368/csharp/rule.adoc
```diff
@@ -1,41 +1,9 @@
-== Why is this an issue?
+:unintuitiveCode: csharp/unintuitiveCode.adoc
```
antonioaversa, line 1: I like the breakdown!

rules/S2368/vbnet/exceptionCode.adoc
```diff
@@ -0,0 +1,8 @@
+[source,vbnet,diff-id=1,diff-type=compliant]
+----
+Public Class Program
```
antonioaversa, line 3: Same remark as in the C# example

rules/S2368/vbnet/exceptionCode.adoc
```diff
@@ -0,0 +1,8 @@
+[source,vbnet,diff-id=1,diff-type=compliant]
```
antonioaversa, line 1: Is this actually compliant with the same diff id as the intuitive and unintuitive examples?
You seem to have two compliant for one non-compliant: do you know what happens in the UI?

rules/S2368/vbnet/intuitiveCode.adoc
```diff
@@ -0,0 +1,13 @@
+[source,vbnet,diff-id=1,diff-type=compliant]
+----
+Public Class Matrix2x2
+    ' ...
+End Class
+
+Public Class Program
+    Public Sub WriteMatrix(matrix As Matrix2x2) ' Compliant
```
antonioaversa, line 8: In the "exception" example, `ByVal` was used, but here is not. 
I would align the two uses, and either make `ByVal` implicit in both or explicit in both.

My personal preference is to make implicit, because `ByVal` is the default anyway and doesn't add any valuable information in this context: the rule would trigger all the same if the parameter was `ByRef`. So I would remove it, so that the user would not wonder why it is there in the first place.

rules/S2368/why-dotnet.adoc
```diff
@@ -0,0 +1,20 @@
+== Why is this an issue?
+
+Using https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/multidimensional-arrays[multidimentsional] and https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/jagged-arrays[jagged] arrays as method parameters in C# can be challenging for developers. When these methods are exposed to external users, it requires advanced language knowledge for effective usage. Determining the appropriate data to pass to these parameters may not be intuitive.
```
antonioaversa, line 3: multidimentsional -> multidimensional

I would also split this paragraph into three: one per period.

antonioaversa (changes requested):
The why explanation is good! Some small remarks related to method visibility scope, orthography and cosmetic.

antonioaversa:
@sebastien-marichal The `validate_asciidoc` check is failing.

rules/S2368/csharp/rule.adoc
sebastien-marichal: Unfortunately, that's the reason why the `validate_asciidoc` check is falling 😞

rules/S2368/csharp/exceptionCode.adoc
sebastien-marichal: Nice catch, I will add a second example with a `private` or `internal` visibility for the class. Will check before if it is implemented that way.

rules/S2368/vbnet/exceptionCode.adoc
sebastien-marichal: Copy-paste mistake

rules/S2368/csharp/rule.adoc
sebastien-marichal: This will be fixed soon https://github.com/SonarSource/rspec/pull/2028! 🎉

antonioaversa:
@sebastien-marichal : the PR currently shows 38 commits - I think you may want to rebase on the current HEAD of `dotnet-layc`.

frederic-tingaud-sonarsource:
My bad. I rebased the branch to make sure the fix that landed on master was fixing this branch, I didn't realize it wasn't based on master.

PR #1987, opened 2023-05-31
SONARJAVA-4482 Modify S1130 rewrite Why is this an issue in LaYC form, refactor exceptions

leonardo-pilastri-sonarsource:


rules/S1130/java/rule.adoc
```diff
@@ -22,49 +26,65 @@ void bar() throws Throwable, Exception {}  // Noncompliant; Exception is a subcl
...
    * @throws IOException Overriding classes may throw this exception if they print values into a file
    */
-  protected void print() throws IOException { // no issue, method is overridable and the exception has proper javadoc
+  protected void overridable() throws IOException { // no issue, method is overridable and the exception has proper javadoc
     System.out.println("foo");
   }
 }
 ----
 
+Also, the rule won't raise issues on `RuntimeException`, or one of its descendants, because explicating runtime exceptions which could be thrown can ultimately help the method's users, and can even be considered as good practice.
```
dorian-burihabwa-sonarsource, line 78:
```suggestion
Also, the rule will not raise issues on `RuntimeException`, or one of its sub-classes, because documenting runtime exceptions which could be thrown can ultimately help users of the method understand its behavior.
```

dorian-burihabwa-sonarsource (approved):

PR #1988, opened 2023-05-31
Modify rule S3875: LaYC format

gregory-paidis-sonarsource:


rules/S3875/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+
```
sebastien-marichal, line 2: I guess a forgotten space appeared here 😄

sebastien-marichal (approved):
LGTM! 
Just a lonely space needs to be cleaned

PR #1992, opened 2023-05-31
Modify S6290(secrets): Migrate to LayC

loris-s-sonarsource:


loris-s-sonarsource:
Closed because abandonned for now

PR #1997, opened 2023-06-01
Modify rule S3889: LAYC format

gregory-paidis-sonarsource:


rules/S3889/csharp/rule.adoc
```diff
@@ -2,24 +2,15 @@
 
 include::../description.adoc[]
```
antonioaversa, line 3: I think `description.adoc` should be renamed to `why-dotnet.adoc`, or `why.adoc` (as the rule only applies to csharp and vbnet, so there is no name conflict with other languages). I would call it `why-dotnet.adoc` though, for consistency.

rules/S3889/csharp/rule.adoc
```diff
@@ -2,24 +2,15 @@
...
 
-=== Noncompliant code example
+include::../resources-dotnet.adoc[]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock[lock statement]
```
antonioaversa, line 6: Examples of resources shown [here](https://github.com/SonarSource/rspec/tree/a51217c6d91abfa5e1d77d0ae0843e3903adf2d0/rules/S5131/common/resources) capitalize the first letter, whereas the [styling guide for adoc](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc#lists) prescribes not avoid capitalizing on list items.
In this case, I think that we don't have an introducing sentence terminated by `:`, and that the first letter for each element of the list should be capitalized, as in the examples.

rules/S3889/description.adoc
```diff
@@ -1 +1,8 @@
-``++Thread.Suspend++`` and ``++Thread.Resume++`` can give unpredictable results, and both methods have been deprecated. Indeed, if ``++Thread.Suspend++`` is not used very carefully, a thread can be suspended while holding a lock, thus leading to a deadlock. Other safer synchronization mechanisms should be used, such as ``++Monitor++``, ``++Mutex++``, and ``++Semaphore++``.
+`Thread.Suspend` and `Thread.Resume` can give unpredictable results, and both methods have been deprecated. Indeed, if `Thread.Suspend` is not used very carefully, a https://en.wikipedia.org/wiki/Thread_(computing)[thread] can be suspended while holding a https://en.wikipedia.org/wiki/Lock_(computer_science)[lock], thus leading to a https://en.wikipedia.org/wiki/Deadlock[deadlock].
+
+There are other synchronization mechanisms that are safer and should be used, such as:
```
antonioaversa, line 3: suggestion: should be used instead

rules/S3889/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Neither \"Thread.Resume\" nor \"Thread.Suspend\" should be used",
+  "title": "\"Thread.Resume\" and \"Thread.Suspend\" should be used",
```
antonioaversa, line 2: "title": "\"Thread.Resume\" and \"Thread.Suspend\" should **not** be used",

antonioaversa (changes requested):
Title to be fixed, the rest is minor

rules/S3889/csharp/rule.adoc
gregory-paidis-sonarsource: I agree in general, but `lock` is written in lowercase, both in the documentation and while writing code. 
Based on that, do you still think it should be capitalized?

rules/S3889/metadata.json
gregory-paidis-sonarsource: 🤦‍♂️

rules/S3889/csharp/rule.adoc
antonioaversa: No OK, makes sense.

antonioaversa (approved):
LGTM!

PR #2009, opened 2023-06-01
Modify rule S6522: Adapt to LaYC

victor-diez-sonarsource:


rules/S6522/javascript/rule.adoc
```diff
@@ -1,30 +1,24 @@
...
-==== Compliant solution
+More importantly, it will cause a runtime error when using ES modules. With TypeScript, it will raise a compilation error.
+
+[source,javascript]
+----
+import { readFile } from 'fs';
+readFile = 'hello world!';   //Noncompliant: TypeError: Assignment to constant variable.
+----
+
+Instead, use constants for `import` and `require`, so that their value is always the same in the scope where they are defined.
```
yassin-kammoun-sonarsource, line 21:
```suggestion
Instead, use constants for `import` and `require` statements, so that their value is always the same in the scope where they are defined.
```

rules/S6522/javascript/rule.adoc
```diff
@@ -1,30 +1,24 @@
...
 [source,javascript]
 ----
 let { readFile } = require('fs');
 readFile = 'hello world!';
 ----
 
-==== Compliant solution
+More importantly, it will cause a runtime error when using ES modules. With TypeScript, it will raise a compilation error.
+
+[source,javascript]
```
yassin-kammoun-sonarsource, line 15: Do we still need to use the noncompliant tag?

rules/S6522/javascript/rule.adoc
```diff
@@ -1,30 +1,24 @@
...
+----
+import { readFile } from 'fs';
+readFile = 'hello world!';   //Noncompliant: TypeError: Assignment to constant variable.
+----
+
+Instead, use constants for `import` and `require`, so that their value is always the same in the scope where they are defined.
 
 [source,javascript]
```
yassin-kammoun-sonarsource, line 23: Should we use the compliant tag?

yassin-kammoun-sonarsource (approved):
Left minor suggestions, otherwise looks good to me.

vilchik-elena (commented):
Clarify the scope of the rule first

rules/S6522/javascript/rule.adoc
victor-diez-sonarsource: yes now we need to add an explanation why it is non-compliant

rules/S6522/javascript/rule.adoc
```diff
@@ -1,45 +1,45 @@
...
-//=== How does this work?
-
-//=== Pitfalls
+This rule will not raise an issue when a module mutates the imported object. Please be aware the mutated value can be observed by all other modules importing the same value.
 
-//=== Going the extra mile
+[source,javascript]
+----
+import fs from 'node:fs';
+fs.readPath = fs.readFile; // fs.default now contains readPath
```
vilchik-elena, line 35: not sure I understand the comment. Could you elaborate?

vilchik-elena (approved):
LGTM! nice work!

rules/S6522/javascript/rule.adoc
victor-diez-sonarsource: when doing a default import, the import variable will point to the exports.default, which is an object that can be modified externally. So this rule does not raise an issue because it's valid js, but that's why I mention: `Please be aware the mutated value can be observed by all other modules importing the same value.`

victor-diez-sonarsource: probably it should say `Please be aware the mutated value can be observed by all other modules importing the same object.`

vilchik-elena: check here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#default_import

the way you import is already default. 
I think you should replace import with `import * as fs from ...`

vilchik-elena: @victor-diez-sonarsource and remove `Please` from that phrase

vilchik-elena (changes requested):
As discussed please group the cases by the type of imported object (sealed vs live binding)

vilchik-elena (approved):

PR #2016, opened 2023-06-01
S2925: Add c# and vbnet specifications

gregory-paidis-sonarsource:


rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,110 @@
...
+}
+----
+
+An alternative is a task-based asynchronous approach, using https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/[async and await].
+
+More specifically the https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-7.0[Task.Delay] method should be used, because of the following advantages:
+
+* It is asynchronous. This means that the thread will not be blocked, but instead will be reused by other operations.
+* It is more precise than `Thread.Sleep`.
+* It can be cancelled and continued, which gives more flexibility and control in the timing of your code.
```
antonioaversa, line 21: What about having `asynchronous`, `precise`, `cancelled and continued` in bold?

rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,110 @@
...
+----
+[TestMethod]
+public async Task SomeTest()
+{
+    await Task.Delay(500);
+    // assertions...
+}
+----
+
+Another scenario is when some data might need to be mocked using https://github.com/moq/moq4[Moq], and a delay needs to be introduced.
```
antonioaversa, line 34: Another scenario where `Thread.Sleep` may appear is when some data might need to be mocked using https://github.com/moq/moq4[Moq] or a similar framework, and a delay needs to be introduced.

If I understand well, this is not just about Moq right? It can be any mocking library I guess, such as https://www.telerik.com/products/mocking.aspx.

rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,110 @@
...
+        {
+            Thread.Sleep(500); // Noncompliant
+            return Task.FromResult(expected);
+        });
+
+    // assertions...
+}
+----
+
+An alternative is to use `ReturnsAsync` and pass the delay amount there:
```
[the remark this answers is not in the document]
antonioaversa: `delay amount` -> `amount of time to delay` or `time to delay`

rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+ifdef::env-github,rspecator-view[]
+
+'''
+== Implementation Specification
+(visible only on this page)
+
+=== Message
+
+Remove this use of "Thread.sleep()".
```
antonioaversa, line 84: sleep with capital S

rules/S2925/java/metadata.json
```diff
@@ -1,29 +1,2 @@
...
-  "defaultSeverity": "Major",
-  "ruleSpecification": "RSPEC-2925",
-  "sqKey": "S2925",
-  "scope": "Tests",
-  "defaultQualityProfiles": [
-    "Sonar way"
-  ],
-  "quickfix": "unknown"
-}
+}
```
antonioaversa, line 2: The title is with small s in java and capital s in dotnet, so I guess title has to be overridden either here or for both csharp and vbnet.

antonioaversa (changes requested):
Title to be fixed for java, comments section to be shared and other more cosmetic remarks.

rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,110 @@
...
+    // assertions...
+}
+----
+
+An alternative is a task-based asynchronous approach, using https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/[async and await].
+
+More specifically the https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-7.0[Task.Delay] method should be used, because of the following advantages:
+
+* It is asynchronous. This means that the thread will not be blocked, but instead will be reused by other operations.
+* It is more precise than `Thread.Sleep`.
```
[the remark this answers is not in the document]
gregory-paidis-sonarsource: I meant that it has better precision in milliseconds.
I will clarify with "more precise in timing the delay", if that's fine with you.

rules/S2925/csharp/rule.adoc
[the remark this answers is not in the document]
gregory-paidis-sonarsource: made it more specific

rules/S2925/java/metadata.json
gregory-paidis-sonarsource: overriden in java, for brevity

rules/S2925/vbnet/rule.adoc
```diff
@@ -0,0 +1,98 @@
+== Why is this an issue?
```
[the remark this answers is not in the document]
antonioaversa: There are some differences between the csharp version and the vbnet one:
- `It is **asynchronous**: This means that the thread will not be blocked, but instead will be reused by other operations` vs `It is **asynchronous**: The thread will not be blocked, but instead will be reused by other operations`
- `Another scenario is when some data might need to be mocked and a delay needs to be introduced.
One example of this is when using https://github.com/moq/moq4[Moq].` vs `Another scenario is when some data might need to be mocked using https://github.com/moq/moq4[Moq], and a delay needs to be introduced.`
- [Optional and only personal preference] the non-compliant version of the code in csharp goes to new line at every `.`, whereas the vb.net doesn't. I think non inlining in VB.NET is more cumbersome, since you need to prepend each line with `_`. Still, I would align the style used in VB.NET to the one used in the C# version.
- [Optional and only personal preference] there is an empty line in the compliant version in C#, which is not there in VB.NET

rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,110 @@
...
+    Thread.Sleep(500); // Noncompliant
+    // assertions...
+}
+----
+
+An alternative is a task-based asynchronous approach, using https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/[async and await].
+
+More specifically the https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-7.0[Task.Delay] method should be used, because of the following advantages:
+
+* It is asynchronous. This means that the thread will not be blocked, but instead will be reused by other operations.
```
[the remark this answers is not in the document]
antonioaversa: First words are still capitalized

antonioaversa (approved):
LGTM! Two conversations still open and to be checked, before merging.

rules/S2925/csharp/rule.adoc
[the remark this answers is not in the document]
antonioaversa: Makes sense as it is the start of a sentence.

rules/S2925/csharp/rule.adoc
```diff
@@ -0,0 +1,87 @@
...
+== Implementation Specification
+(visible only on this page)
+
+=== Message
+
+Remove this use of "Thread.Sleep()".
+
+include::../comments-and-links.adoc[]
+
+endif::env-github,rspecator-view[]
```
pavel-mikula-sonarsource, line 87: This ugly (hard to maintain) `ifdef` section can be extracted into `rspecator.adoc` or `rspecator-dotnet.adoc` in the root

rules/S2925/resources-dotnet.adoc
```diff
@@ -0,0 +1,9 @@
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.sleep[Thread.Sleep method]
+* https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay[Task.Delay method]
+* https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/[Asynchronous programming with async and await]
+* https://github.com/moq/moq4[Moq mocking library]
+* https://www.telerik.com/products/mocking.aspx[Telerik mocking library]
```
pavel-mikula-sonarsource, line 9: I'm not sure why the Telerik is here, when the rule doesn't speak about it.

rules/S2925/vbnet/rule.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+Using `Thread.Sleep` in a test might introduce unpredictable and incosistent results depending on the environment. Furthermore, it will block the https://en.wikipedia.org/wiki/Thread_(computing)[thread], which means the system resources are not being fully used.
+
+[source,vbnet,diff-id=1,diff-type=noncompliant]
+----
+<TestMethod>
+Public Sub SomeTest()
+    Threading.Thread.Sleep(500) ' Noncompliant
+
```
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Same below

rules/S2925/vbnet/rule.adoc
```diff
@@ -0,0 +1,79 @@
...
+End Function
+----
+
+Another scenario is when some data might need to be mocked using https://github.com/moq/moq4[Moq], and a delay needs to be introduced:
+
+[source,vbnet,diff-id=2,diff-type=noncompliant]
+----
+<TestMethod>
+Public Sub UserService_Test()
+    Dim userService = New Mock(Of UserService)()
```
pavel-mikula-sonarsource, line 39:
```suggestion
    Dim UserService As New Mock(Of UserService)
```

rules/S2925/vbnet/rule.adoc
```diff
@@ -0,0 +1,79 @@
...
+    Dim expected = New User()
+    userService.Setup(Function(m) m.GetUserById(42)).Returns(
```
pavel-mikula-sonarsource, line 41:
```suggestion
    UserService.Setup(Function(X) X.GetUserById(42)).Returns(
```

rules/S2925/vbnet/rule.adoc
```diff
@@ -0,0 +1,79 @@
...
+    ' assertions...
+End Sub
+----
+
+An alternative to `Thread.Sleep` while mocking with `Moq` is to use `ReturnsAsync` and pass the amount of time to delay there:
+
+[source,vbnet,diff-id=2,diff-type=compliant]
+----
+<TestMethod>
+Public Sub UserService_Test()
```
pavel-mikula-sonarsource, line 56: Same nitpicks as above

rules/S2925/vbnet/rule.adoc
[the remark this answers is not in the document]
pavel-mikula-sonarsource: The inline version is fine. Leave it as it is.

`_` is not used in VB except some extreme cases.
The VB formatting would be like this with auto line continuation
```
        UserService.
            Setup(Function(m) m.GetUserById(42)).
            ReturnsAsync(expected, TimeSpan.FromMilliseconds(500))
```

pavel-mikula-sonarsource (approved):
LGTM, left some nitpicks above

PR #2019, opened 2023-06-02
Modify rule S2953: LaYC format

antonioaversa:
Adapts the rule to LaYC format level 2, for csharp


rules/S2953/csharp/metadata.json
```diff
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
```
sebastien-marichal, line 27: It was suggested to me to move the metadata content to the rule root folder keeping here only the properties that need to override.
WDYT?

rules/S2953/csharp/rule.adoc
```diff
@@ -1,14 +1,22 @@
...
+The interface declares a https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose[Dispose] method, which the implementer has to define.
 
+The method name `Dispose` should be used exclusively to implement `IDisposable.Dispose` to prevent any confusion.
 
-=== Noncompliant code example
+It may be tempting to create a `Dispose` method for other purposes, but doing so will result in confusion and likely lead to problems in production.
 
-[source,csharp]
+== How to fix it
+
```
sebastien-marichal, line 14: I think we should have here some paragraphs to explain a bit more about how to fix it.
But it might be redundant from what is already in the `why`.

rules/S2953/csharp/rule.adoc
```diff
@@ -48,11 +55,11 @@ public class GarbageDisposal
 
 === Exceptions
```
sebastien-marichal, line 56: I think the `Exceptions` should go under the `Why this is an issue?` section. As mentioned in the [documentation](https://github.com/SonarSource/rspec/blob/master/docs/description.adoc#2-education-format)

sebastien-marichal (changes requested):
I suggested to move on section and left some open-question comments.

rules/S2953/csharp/metadata.json
antonioaversa: Yes, makes perfect sense to me. Changed.

rules/S2953/csharp/rule.adoc
antonioaversa: I agree it makes sense since they now appear in two tabs in SQ, and when the user will be looking at the "How to fix" section, he won't have the why section visible, so he may lose some context.
I have added a paragraph of a few of sentences, to explain the code shown below.
Let me know what you think.

rules/S2953/csharp/rule.adoc
antonioaversa: Yes, also reported by a failing check of the CI. Fixed!

rules/S2953/csharp/rule.adoc
sebastien-marichal: Nice!

rules/S2953/csharp/metadata.json
antonioaversa: It wouldn't make any difference: the rule is .NET specific and regardless, automation is infeasible since we don't know user intention.
Anyway, changed.

sebastien-marichal (approved):
LGTM!

sebastien-marichal (approved):
LGTM!

PR #2029, opened 2023-06-02
Modify rule S6617: Add Exception section

sebastien-marichal:


antonioaversa (approved):
LGTM

rules/S6617/csharp/rule.adoc
```diff
@@ -1,5 +1,14 @@
 include::../why-dotnet.adoc[]
 
+[source,csharp]
```
andrei-epure-sonarsource, line 3: Do I understand well that this code snippet has been inserted between the "Why" and the "Impact"?

andrei-epure-sonarsource: why is this code snippet necessary in the rule description?

PR #2030, opened 2023-06-02
SONARJAVA-4488 Modify rule S2142: rewrite in new education format

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #2036, opened 2023-06-02
Modify rule S4159: LaYC format

antonioaversa:
Adapts the rule to LaYC format level 2, for dotnet languages (csharp, vbnet).


gregory-paidis-sonarsource (changes requested):
Looks good, left a comment about a typo

gregory-paidis-sonarsource (approved):
LGTM

PR #2037, opened 2023-06-02
Modify rule S2930: LaYC format

sebastien-marichal:


rules/S2930/csharp/rule.adoc
```diff
@@ -1,45 +1,71 @@
 == Why is this an issue?
 
-When writing managed code, you don't need to worry about allocating or freeing memory: The garbage collector takes care of it. For efficiency reasons, some objects such as ``++Bitmap++`` use unmanaged memory, enabling for example the use of pointer arithmetic. Such objects have potentially huge unmanaged memory footprints, but will have tiny managed ones. Unfortunately, the garbage collector only sees the tiny managed footprint, and fails to reclaim the unmanaged memory (by calling ``++Bitmap++``'s finalizer method) in a timely fashion.
```
gregory-paidis-sonarsource, line 3: I think you need to link the following concepts (make them clickable):
- [managed code](https://learn.microsoft.com/en-us/dotnet/standard/managed-code)
- [garbage collector](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/)
- [pointer arithmetic](https://stackoverflow.com/questions/394767/pointer-arithmetic)
- [finalizer](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/finalizers)

rules/S2930/csharp/rule.adoc
```diff
@@ -1,45 +1,71 @@
...
+When writing managed code, you don't need to worry about allocating or freeing memory: The garbage collector takes care of it. For efficiency reasons, some objects such as `Bitmap` use unmanaged memory, enabling, for example, the use of pointer arithmetic. Such objects have potentially huge unmanaged memory footprints but will have tiny managed ones. Unfortunately, the garbage collector only sees the tiny managed footprint and fails to reclaim the unmanaged memory (by calling `Bitmap`'s finalizer method) promptly. 
 
+Moreover, memory is not the only system resource that needs to be managed on time: The operating system can only handle having so many file descriptors (e.g. `FileStream`) or sockets (e.g. `WebClient`) open at any given time. Therefore, it is important to `Dispose` of them as soon as they are no longer needed, rather than relying on the garbage collector to call these objects' finalizers at some nondeterministic point in the future.
```
gregory-paidis-sonarsource, line 5: maybe link to what an [fd](https://en.wikipedia.org/wiki/File_descriptor) is

gregory-paidis-sonarsource: also [web socket](https://en.wikipedia.org/wiki/WebSocket)

rules/S2930/csharp/rule.adoc
```diff
@@ -1,45 +1,71 @@
...
 
-Moreover, memory is not the only system resource which needs to be managed in a timely fashion: The operating system can only handle having so many file descriptors (e.g. ``++FileStream++``) or sockets (e.g. ``++WebClient++``) open at any given time. Therefore, it is important to ``++Dispose++`` of them as soon as they are no longer needed, rather than relying on the garbage collector to call these objects' finalizers at some nondeterministic point in the future.
+This rule tracks `private` fields and local variables of the following `IDisposable` / `IAsyncDisposable` types, which are never disposed, closed, aliased, returned, or passed to other methods.
 
+* `System.IO` namespace
```
gregory-paidis-sonarsource, line 9: do we need all these specifics?
Personally I would delete the entire list, as it feels too "busy", but I will leave it up to you. :)

rules/S2930/csharp/rule.adoc
```diff
@@ -1,45 +1,71 @@
...
 
-* ``++System.IO.File.Create()++``
-* ``++System.IO.File.Open()++``
-* ``++System.Drawing.Image.FromFile()++``
-* ``++System.Drawing.Image.FromStream()++``
+[source,csharp]
+----
+public Stream WriteToFile(string path, string text)
+{
+  var fs = new FileStream(path, FileMode.Open); // Compliant, because it is returned
```
gregory-paidis-sonarsource, line 43: I think I would use the same convention everywhere, meaning:

`[Compliant|Noncompliant]: reason`

For example here:

`Compliant: it is returned`

same applies to all the other snippets

gregory-paidis-sonarsource (changes requested):
Looks good, left a few suggestions

gregory-paidis-sonarsource (approved):
LGTM!

PR #2040, opened 2023-06-05
Modify rule S3972: LaYC Conditional on new line

frederic-tingaud-sonarsource:


rules/S3972/php/rule.adoc
```diff
@@ -1,8 +1,6 @@
 == Why is this an issue?
 
-Code is clearest when each statement has its own line. Nonetheless, it is a common pattern to combine on the same line an ``++if++`` and its resulting _then_ statement. However, when an ``++if++`` is placed on the same line as the closing ``++}++`` from a preceding ``++else++`` or ``++elseif++``, it is either an error - ``++else++`` is missing - or the invitation to a future error as maintainers fail to understand that the two statements are unconnected.
-
-=== Noncompliant code example
+When an `if` is placed on the same line as the closing ``++}++`` from a preceding  `then`, `else` or `elseif` part, it is either an error - `else` is missing - or the invitation to a future error as maintainers fail to understand that the two statements are unconnected.
```
amelie-renard-sonarsource, line 3: `then` does not seem to exist in PHP, does it?

rules/S3972/rule.adoc
```diff
@@ -1,11 +1,8 @@
 == Why is this an issue?
 
-Code is clearest when each statement has its own line. Nonetheless, it is a common pattern to combine on the same line an ``++if++`` and its resulting _then_ statement. However, when an ``++if++`` is placed on the same line as the closing ``++}++`` from a preceding  _then_, _else_ or _else if_ part, it is either an error - ``++else++`` is missing - or the invitation to a future error as maintainers fail to understand that the two statements are unconnected.
+When an `if` is placed on the same line as the closing ``++}++`` from a preceding  `then`, `else` or `else if` part, it is either an error - `else` is missing - or the invitation to a future error as maintainers fail to understand that the two statements are unconnected.
```
amelie-renard-sonarsource, line 3: I don't know about the other languages, but shouldn't we remove the mention of `then` in the CFamily description?

rules/S3972/rule.adoc
```diff
@@ -1,11 +1,8 @@
...
 
-
-=== Noncompliant code example
```
amelie-renard-sonarsource, line 6: I understand the point of removing the title but I would still put a sentence. Maybe something like:
"Check the following code as an example:"

amelie-renard-sonarsource (changes requested):

rules/S3972/php/rule.adoc
frederic-tingaud-sonarsource: I understood `then` as "not else", but I agree it's confusing. fixed

rules/S3972/rule.adoc
frederic-tingaud-sonarsource: fixed

rules/S3972/rule.adoc
amelie-renard-sonarsource: Yes, far better, thanks!

PR #2041, opened 2023-06-05
Modify rule S3464: LaYC format

antonioaversa:
Adapts the rule to LaYC format level 2, for csharp.


rules/S2223/csharp/rule.adoc
```diff
@@ -1,62 +1,117 @@
...
+    {
+    }
+}
+----
 
-=== Noncompliant code example
+Non-private `static` fields that are neither `const` nor `readonly`, like the ones in the examples above, can lead to errors and unpredictable behavior.
+
+This can happen because:
+* Any object can modify these fields and alter the global state. This makes the code more difficult to read, debug and test.
```
sebastien-marichal, line 34: I am concerned about how this will be rendered; having the examples within bullet points might have a strange result.

Asciidoctor provides [list continuation](https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#list-continuation), but I don't know if we support it.

sebastien-marichal (approved):
LGTM!
Left an open comment about rendering, can probably be ignored since it does not depend on us.

rules/S2223/csharp/rule.adoc
antonioaversa: Ok thanks for the info! I didn't know about this list continuation.

It looks better on the rendering by VSCode!

I have applied the change. Let's see if everything is rendered correctly.

PR #2042, opened 2023-06-05
Modify rule S2187: LaYC format

sebastien-marichal:


rules/S2187/csharp/rule.adoc
```diff
@@ -1,50 +1,24 @@
...
-    [TestMethod]
-    public void SomeMethodShouldReturnTrue() { }
-}
-----
+It does not apply to `xUnit` since `xUnit` does not provide a https://xunit.net/docs/comparisons#attributes[test class attribute].
 
 === Exceptions
 
-* abstract classes
+* `abstract` classes
```
antonioaversa, line 14: Could you explain this scenario a bit more in detail?

It is not clear to me why one would need to mark an `abstract` class as `TestClass` or similar.

It's enough to mark the concrete derivation as `TestClass`.

rules/S2187/csharp/rule.adoc
```diff
@@ -1,50 +1,24 @@
...
 * derived classes that inherit from a base class that does have test methods
```
antonioaversa, line 15: This scenario is more clear. It would still be good to have a code example, to make clearer what we are talking about:
```cs
public abstract class CollectionInterfaceTests<T>
{
    private Func<ICollection<T?>> CollectionProvider { get; }

    protected CollectionInterfaceTests(Func<ICollection<T?>> listProvider)
    {
        CollectionProvider = listProvider;
    }

    [TestMethod]
    public void Count_IsZero_AfterClean()
    {
        var collection = CollectionProvider();
        collection.Add(default);
        Assert.AreNotEqual(0, collection.Count);
        collection.Clear();
        Assert.AreEqual(0, collection.Count);
    }
}

[TestClass]
public class IntListTests : CollectionInterfaceTests<int>
{
    public IntListTests() : base(() => new List<int>()) { }
}

[TestClass]
public class StringListTests : CollectionInterfaceTests<string>
{
    public StringListTests() : base(() => new List<string?>()) { }
}

[TestClass]
public class StringLinkedListTests : CollectionInterfaceTests<string>
{
    public StringLinkedListTests() : base(() => new LinkedList<string?>()) { }
}

```

rules/S2187/csharp/rule.adoc
```diff
@@ -1,50 +1,24 @@
...
-* in *MSTest*, classes that contain methods marked with either ``++AssemblyInitialize++`` or ``++AssemblyCleanup++``.
+* in `MSTest`, classes that contain methods marked with either `AssemblyInitialize` or `AssemblyCleanup`.
```
antonioaversa, line 16: Example (as far as I understand the scenario):
```cs
[TestClass]
public class TestsInstrumentation
{
    [AssemblyInitialize]
    public static void AssemblyInit(TestContext context)
    {
        // ...
    }

    [AssemblyCleanup]
    public static void AssemblyCleanup()
    {
        // ...
    }
}
```

I would also mention and refer the following:
```
The test engine runs a method that is marked with the AssemblyInitialize attribute only if that method is a member of a class that is marked with the TestClass attribute.
```
from [here](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2008/ms245278(v=vs.90)).

Otherwise, there would be no reason to add `TestClass` to a class which doesn't contain `TestMethod` and only contains setup or tear-down methods, such as `[AssemblyInitializeAttribute]` or `[AssemblyCleanupAttribute]`.

Another thing that would be good to explain is why an exception is made for `[AssemblyInitializeAttribute]` or `[AssemblyCleanupAttribute]`, but not for `[ClassInitializeAttribute]` or `[ClassCleanupAttribute]` or `[TestInitializeAttribute]` or `[TestCleanupAttribute]`.

To be confirmed, but the reason I think it's that, while to have setup/tear-down methods at assembly level it makes sense to define a dedicated class for those methods, without actual test methods, having setup or tear-down methods at class level only makes sense for abstract classes (so exception mentioned above).

That is because, if the `TestClass` is concrete and also marked as `ClassInitialize` or `MethodInitialize`, it would instrument only setup/tear-down methods of the test class itself... but there are no methods on the test class itself!

rules/S2187/csharp/how-mstest.adoc
```diff
@@ -0,0 +1,31 @@
...
+* `DataTestMethod`
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
+[TestClass]
+public class SomeOtherClassTest { } // Noncompliant - no test
```
antonioaversa, line 18: Optional: // Noncompliant: no test**s**

rules/S2187/csharp/how-nunit.adoc
```diff
@@ -0,0 +1,33 @@
...
+* `Theory`
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
+[TestFixture]
+public class SomeClassTest { } // Noncompliant - no test
```
antonioaversa, line 20: Optional: // Noncompliant: no test**s**

rules/S2187/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
```
antonioaversa, line 1: Optional proposal

Long time ago Nicolas made the following remark:
```
=== on 24 Nov 2014, 19:28:35 Nicolas Peru wrote:
\[~ann.campbell.2]What does "without any tests" means ? is that for classes without any test methods ? or is it tests without asserts ? This rule seems to lack some example.
```

I had the same doubt until I have read the "why" section: "tests" can be interpreted as "test methods" or "test assertions".
Could we change the title of the rule, at least for .NET, to "TestCases should contain test methods"?
WDYT?

antonioaversa (changes requested):
Some more examples required, mainly on the exception section.
Proposal to change the title into something more precise.

antonioaversa (changes requested):
Some more examples required, mainly on the exception section.
Proposal to change the title into something more precise.

rules/S2187/csharp/metadata.json
sebastien-marichal: I think we should rename it anyway, from `.NET` point of view, `TestCases` does not mean the same thing as `Java` (the rule has been implemented in Java first).

A more adequate title for `.NET` could be "Test classes should contain at least one test method".
I don't know if we can have different titles for different languages.
WDYT?

sebastien-marichal: An even more accurate title could be: "Test classes should contain at least one test **case**"

rules/S2187/csharp/rule.adoc
sebastien-marichal: Looked for a while but found when this exception has been included and why: https://github.com/SonarSource/sonar-dotnet/issues/929

rules/S2187/csharp/rule.adoc
sebastien-marichal: I went for explanation paragraphs to avoid to have a way bigger `Exception` then the `Why` section.
Hopefully, it is clear enough.
Let me know what you think.

rules/S2187/csharp/metadata.json
sebastien-marichal: I update the rule title, don't know if it works this way though.

rules/S2187/csharp/how.adoc
```diff
@@ -0,0 +1,3 @@
+When working with `{framework}`, to fix this issue, it is important that all test classes annotated with the `[{classAnnotation}]` attribute contains at least one test case.
```
antonioaversa, line 1: Suggestion: `When working with `{framework}`, to fix this issue` -> `To fix this issue in ``{framework}```

rules/S2187/csharp/rule.adoc
```diff
@@ -1,50 +1,54 @@
...
+To ensure proper testing, it is important to include test cases in a test class. If a test class does not have any test cases, it can give the wrong impression that the class being tested has been thoroughly tested, when in reality, it has not.
 
+This rule will raise an issue when any of these conditions are met:
 
-Supported test frameworks are ``++NUnit++`` and ``++MSTest++`` (not applicable to ``++xUnit++``).
+* For `NUnit`, a class is marked with `TestFixture` but does not contain any method marked with `Test`, `TestCase`, `TestCaseSource`, or `Theory`.
+* For `MSTest`, a class is marked with `TestClass` but does not contain any method marked with `TestMethod` or `DataTestMethod`.
 
-This rule will raise an issue when any of these conditions are met:
+It does not apply to `xUnit` since it does not require a https://xunit.net/docs/comparisons#attributes[test class attribute].
```
antonioaversa, line 10: It = the rule? I would change it.

antonioaversa (approved):

antonioaversa (approved):
LGTM
Just some minor grammatical mistakes and some optional, non-blocking suggestions.

rules/S2187/csharp/rule.adoc
antonioaversa: Good for me!

PR #2045, opened 2023-06-05
Modify rule S2346: LaYC format

antonioaversa:
Adapts the [rule](https://sonarsource.github.io/rspec/#/rspec/S2346) to LaYC format level 2, for dotnet languages (csharp, vbnet).


rules/S2346/csharp/rule.adoc
```diff
@@ -1,12 +1,87 @@
...
+enum Days
+{
+    // ...
+    None = 0,                                                    // 0b00000000
+    Weekdays = Monday | Tuesday | Wednesday | Thursday | Friday, // 0b00011111
+    Weekend = Saturday | Sunday,                                 // 0b01100000
+    All = Weekdays | Weekend                                     // 0b01111111
+}
+
+var someDays = Days.Wednesday | Days.Weekend;                    // 0b01100100
```
sebastien-marichal, line 36: I would include this line in the next example as, to me, it is not as relevant in this example as it could be in the next one.

While reading documentation or tutorials, I will sometimes skip to the example that seems to be the most relevant one which could be the next one, then some context will be missing.

rules/S2346/resources.adoc
```diff
@@ -0,0 +1,13 @@
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute[FlagsAttribute]
+* https://en.wikipedia.org/wiki/Bit_field[Bit field]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators[Bitwise and shift operators (C# reference)]
```
sebastien-marichal, line 7: I think we should separate resources when we language specific links.

antonioaversa: I think there has been the opposite recommendation in a previous PR.
I will ask the question on Slack.

antonioaversa: As discussed on Slack: separating resources when lang-specific content is present, from now on.

rules/S2346/csharp/rule.adoc
```diff
@@ -29,6 +104,20 @@ enum FruitType
...
+* https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute[FlagsAttribute]
+* https://en.wikipedia.org/wiki/Bit_field[Bit field]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators[Bitwise and shift operators (C# reference)]
+* https://learn.microsoft.com/en-us/dotnet/api/system.enum.hasflag[Enum.HasFlag(Enum) Method]
+
+=== Articles & blog posts
+
+* https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229062(v=vs.100)[Designing Flags Enumerations]
+
 ifdef::env-github,rspecator-view[]
```
sebastien-marichal, line 121: You can extract this `ifdef/endif` section to a dedicated `rspecator.adoc` file and include it in both languages.

sebastien-marichal (changes requested):
One last small change. Otherwise all good!

PR #2046, opened 2023-06-05
Modify rule S2222: LaYC format

sebastien-marichal:


rules/S2222/why-dotnet.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+* `Monitor`
+* `Mutex`
+* `ReaderWriterLock`
+* `ReaderWriterLockSlim`
+* `SpinLock` 
+
+An issue is reported when a lock is acquired within a method but not released on all paths.
+
+=== Exceptions
```
sebastien-marichal, line 17: This is a bit overkill, might remove it.
@gregory-paidis-sonarsource WDYT?

rules/S2222/why-dotnet.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+Failing to release locks properly can lead to potential deadlocks, where the lock might not be released, causing issues in the application.
+
+This rule specifically focuses on tracking the following types from the `System.Threading` namespace:
+
+* `Monitor`
+* `Mutex`
+* `ReaderWriterLock`
+* `ReaderWriterLockSlim`
+* `SpinLock`
```
gregory-paidis-sonarsource, line 13: I think it would be nice if you linked all of these (Monitor, Mutex, etc)

gregory-paidis-sonarsource (changes requested):
Looks good, left a couple suggestions

rules/S2222/csharp/rule.adoc
```diff
@@ -1,30 +1,43 @@
...
       Monitor.Exit(obj);
     }
   }
+}
+----
 
+[source,csharp,diff-id=2,diff-type=noncompliant]
+----
+class MyClass 
+{
```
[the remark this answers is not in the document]
sebastien-marichal: The logic is identical, but since the solution is mentioned in the `How to fix it` section, I will keep it. I think it is better to show multiple solutions to an issue.

For instance, I don't think I would have thought of the `try-finally` solution to release a lock object before reading this rule.

gregory-paidis-sonarsource (approved):
LGTM!

PR #2047, opened 2023-06-05
Modify rule S5856: LaYC format

gregory-paidis-sonarsource:


rules/S5856/why-dotnet.adoc
```diff
@@ -0,0 +1,5 @@
+https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expressions[Regular expressions] have their own syntax that is understood by regular expression engines. Those engines will throw an exception at runtime if they are given a regular expression that does not conform to that syntax.
+
+To avoid syntax errors, special characters should be https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-escapes-in-regular-expressions[escaped with backslashes] when they are intended to be matched literally and references to capturing groups should use the correctly spelled name or number of the group.
+
+Negative https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference#lookarounds-at-a-glance[lookahead] and negative lookbehind groups cannot be combined with https://learn.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions[RegexOptions.NonBacktracking]. Such combination would throw an exception during runtime.
```
sebastien-marichal, line 5: `lookbehind` could be included in the link.

```suggestion
Negative https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference#lookarounds-at-a-glance[lookahead and negative lookbehind] groups cannot be combined with https://learn.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions[RegexOptions.NonBacktracking]. Such combination would throw an exception during runtime.
```

rules/S5856/csharp/rule.adoc
```diff
@@ -1,45 +1,36 @@
 == Why is this an issue?
```
sebastien-marichal, line 1: Pedantic: The `why` could be included in the `why-dotnet.adoc` file

rules/S5856/why-dotnet.adoc
gregory-paidis-sonarsource: good point, restated it as "Negative _lookaround_ groups" instead, as that is what the documentation calls it

sebastien-marichal (approved):
LGTM!

PR #2052, opened 2023-06-05
Modify rule S1598: Reworked into new educational format

kaufco:


kaufco:
Wrong branch name -> closed

PR #2055, opened 2023-06-05
SONARJAVA-4488 Modify rule S2153: reformulate description and code examples

leonardo-pilastri-sonarsource:


rules/S2153/java/rule.adoc
```diff
@@ -1,37 +1,36 @@
 == Why is this an issue?
 
-Boxing is the process of putting a primitive value into an analogous object, such as creating an ``++Integer++`` to hold an ``++int++`` value. Unboxing is the process of retrieving the primitive value from such an object.
+Boxing is the process of putting a primitive value into a wrapper object, such as creating an `Integer` to hold an `int` value.
+Unboxing is the process of retrieving the primitive value from such an object.
+Since the original value is unchanged during boxing and unboxing, there's no point in doing either when not needed.
```
dorian-burihabwa-sonarsource, line 5:
```suggestion
Since the original value is unchanged during boxing and unboxing, there is no point in doing either when not needed.
```

rules/S2153/java/rule.adoc
```diff
@@ -1,37 +1,36 @@
...
 
-
-Since the original value is unchanged during boxing and unboxing, there's no point in doing either when not needed. This also applies to autoboxing and auto-unboxing (when Java implicitly handles the primitive/object transition for you).
+Instead you should rely on Java implicit boxing/unboxing to convert from the primitive type to the wrapper type and vice versa, for better readability.
```
dorian-burihabwa-sonarsource, line 7:
```suggestion
Instead you should rely on Java's implicit boxing/unboxing to convert from the primitive type to the wrapper type and vice versa, for better readability.
```

dorian-burihabwa-sonarsource (approved):

PR #2067, opened 2023-06-06
Modify rule S3449: LAYC format

gregory-paidis-sonarsource:


rules/S3449/csharp/rule.adoc
```diff
@@ -1,31 +1,19 @@
 == Why is this an issue?
 
-include::../description.adoc[]
-
-=== Noncompliant code example
+Numbers can be shifted with the `<<` and `>>` https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#left-shift-operator-[operators], but the right operand of the operation needs to be an `int` or a type that has an https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/conversions#102-implicit-conversions[implicit conversion] to `int`. However, with `dynamic`, the compiler's type checking is turned off, so you can pass anything to a shift operator and have it compile. And if  the argument can't be implicitly converted to `int` at runtime, then a https://learn.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.runtimebinderexception[RuntimeBinderException] will be raised.
```
gregory-paidis-sonarsource, line 3: There is some duplication both here and in `Resources`, but I think there are enough differences to justify the simplicity of this approach.

rules/S3449/csharp/rule.adoc
```diff
@@ -1,31 +1,21 @@
...
```
sebastien-marichal, line 3: There is an additional space:
```suggestion
Numbers can be shifted with the `<<` and `>>` https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#left-shift-operator-[operators], but the right operand of the operation needs to be an `int` or a type that has an https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/conversions#102-implicit-conversions[implicit conversion] to `int`. However, with `dynamic`, the compiler's type checking is turned off, so you can pass anything to a shift operator and have it compile. And if the argument can't be implicitly converted to `int` at runtime, then a https://learn.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.runtimebinderexception[RuntimeBinderException] will be raised.
```

rules/S3449/csharp/rule.adoc
sebastien-marichal: This could be resolved with variables, but I agree, it could be too much for such a small paragraph.

rules/S3449/csharp/rule.adoc
sebastien-marichal, line 3: The `dynamic` section was not clear to me, at first I thought we were talking about when the right operand is `dynamic`, but looking at the example made me realize it was about the left operand. I find it confusing.

Also, a link to `dynamic` documentation would be great.

rules/S3449/vbnet/rule.adoc
```diff
@@ -1,8 +1,6 @@
 == Why is this an issue?
 
-include::../description.adoc[]
-
-=== Noncompliant code example
+Numbers can be shifted with the `<<` and `>>` https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/bit-shift-operators[operators], but the right operand of the operation needs to be an `int` or a type that has an https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/data-types/implicit-and-explicit-conversions[implicit conversion] to `int`. However, with `object`, the compiler's type checking is turned off, so you can pass anything to a shift operator and have it compile. And if  the argument can't be implicitly converted to `int` at runtime, then a https://learn.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.runtimebinderexception[RuntimeBinderException] will be raised.
 
 [source,vbnet]
 ----
```
sebastien-marichal, line 6: Pedantic: an example using the left shift operator (`<<`) could be added for parity.

rules/S3449/vbnet/rule.adoc
```diff
@@ -1,8 +1,6 @@
 == Why is this an issue?
 
-include::../description.adoc[]
-
-=== Noncompliant code example
+Numbers can be shifted with the `<<` and `>>` https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/bit-shift-operators[operators], but the right operand of the operation needs to be an `int` or a type that has an https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/data-types/implicit-and-explicit-conversions[implicit conversion] to `int`. However, with `object`, the compiler's type checking is turned off, so you can pass anything to a shift operator and have it compile. And if  the argument can't be implicitly converted to `int` at runtime, then a https://learn.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.runtimebinderexception[RuntimeBinderException] will be raised.
```
sebastien-marichal, line 3: Same here, there is an additional space:
```suggestion
Numbers can be shifted with the `<<` and `>>` https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/bit-shift-operators[operators], but the right operand of the operation needs to be an `int` or a type that has an https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/data-types/implicit-and-explicit-conversions[implicit conversion] to `int`. However, with `object`, the compiler's type checking is turned off, so you can pass anything to a shift operator and have it compile. And if the argument can't be implicitly converted to `int` at runtime, then a https://learn.microsoft.com/en-us/dotnet/api/microsoft.csharp.runtimebinder.runtimebinderexception[RuntimeBinderException] will be raised.
```

sebastien-marichal (changes requested):
The description is a bit confusing in my opinion, it could be rewritten for better clarity.

rules/S3449/vbnet/rule.adoc
gregory-paidis-sonarsource: Changed one of the examples so that we have one of each

sebastien-marichal (approved):
LGTM!

PR #2072, opened 2023-06-06
SONARJAVA-4482 Review LaYC format migration changes

leonardo-pilastri-sonarsource:


rules/S1130/java/rule.adoc
```diff
@@ -5,7 +5,7 @@ An exception in a `throws` clause in Java is superfluous if it is:
```
dorian-burihabwa-sonarsource, line 5: In the sentence above, we can remove the "in Java" part

rules/S1149/java/rule.adoc
```diff
@@ -1,17 +1,17 @@
 == Why is this an issue?
 
 Early classes of the Java API, such as `Vector`, `Hashtable` and `StringBuffer`, were synchronized to make them thread-safe.
-However, synchronization has a big negative impact on performance, even when using these collections from a single thread.
+However, synchronization has a big negative impact on performance, even when accessing these collections from a single thread.
 
 
-It is often best to use their new non-synchronized counterparts:
+It is often best to use their more recent non-synchronized counterparts:
```
dorian-burihabwa-sonarsource, line 7: Because the issue is not related to the modernity of the types maybe we should remove all notions of new/modern/more recent here. WDYT?

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):

PR #2073, opened 2023-06-06
Modify rule S3447: LayC format

csaba-sagi-sonarsource:


gregory-paidis-sonarsource (approved):
Looks good.

I would only rewrite the noncompliant message, from this:

```csharp
// This doesn't compile, CS7036 shows
```

to this:

```csharp
// Compilation Error [CS7036]
```

Leaving it up to you.

PR #2077, opened 2023-06-06
Modify rule S105: LaYC format

amelie-renard-sonarsource:


vilchik-elena:
@amelie-renard-sonarsource will you update the spreadsheet or should I?

amelie-renard-sonarsource:
I'll do it, thanks @vilchik-elena

PR #2078, opened 2023-06-06
Modify rule S4635: LaYC format

sebastien-marichal:


rules/S4635/csharp/rule.adoc
csaba-sagi-sonarsource, line 1: I suggest also adding a link to the SubString method`s doc on MSDN.

PR #2080, opened 2023-06-06
Modify rule S2692: LaYC format

antonioaversa:
Adapts the [rule](https://sonarsource.github.io/rspec/#/rspec/S2692) to LaYC format level 2, for dotnet languages (csharp, vbnet).
Also, removes the directories for the following languages (rule not implemented):
- cfamily
- flex
- go
- php
- python
- vb6

Changes have been made for the following languages:
- java - rspecator section
- javascript - rspecator section


rules/S2692/csharp/metadata.json
```diff
@@ -1,3 +1,4 @@
 {
-  "title": "\"IndexOf\" checks should not be for positive numbers"
+  "title": "\"IndexOf\" checks should not be for positive numbers",
+  "quickfix": "infeasible"
```
csaba-sagi-sonarsource, line 3: Why do you think that there cannot be a quick fix for the rule?

rules/S2692/csharp/rule.adoc
```diff
@@ -1,16 +1,38 @@
...
+Any checks which look for values `> 0` ignore the first element, which is likely a bug. If the intent is merely to check inclusion of a value in a `string`, `List`, or an array, consider using the https://learn.microsoft.com/en-us/dotnet/api/system.string.contains[Contains] method instead.
 
-This rule also raises an issue when ``++IndexOfAny++``, ``++LastIndexOf++`` or ``++LastIndexOfAny++`` from a ``++string++`` is tested against ``++> 0++``
+[source,csharp]
+----
+strings.Contains(someString) // bool result
+----
 
-=== Noncompliant code example
+This rule raises an issue when an `IndexOf` value retrieved from a `string`, `List` or array is tested against `> 0`.
```
csaba-sagi-sonarsource, line 19: Why did you drop the other methods?
I think it would be nice to have them and maybe also give links to them on MSDN?

csaba-sagi-sonarsource, line 19: This and the part with the other methods the rule covers could be merged.

rules/S2692/rspecator.adoc
```diff
@@ -1,16 +1,15 @@
-include::../rule.adoc[]
-
 ifdef::env-github,rspecator-view[]
 
 '''
 == Implementation Specification
 (visible only on this page)
 
-include::../message.adoc[]
+include::message.adoc[]
```
csaba-sagi-sonarsource, line 7: I am not missing something, the message is only referenced here. I think we can directly put the content here and get rid of the file.

rules/S2692/csharp/metadata.json
antonioaversa: Makes sense to change it to `targeted`.

rules/S2692/csharp/rule.adoc
antonioaversa: I have merged them. I restructured a bit because in actuality not all the methods apply to all types.

csaba-sagi-sonarsource (approved):
LGTM!

PR #2084, opened 2023-06-06
Modify rule S2055(Java): Update rule according to the LayC

alban-auzeill:
image
image
image

### DIFF
--- a/rules/S2055/java/rule.adoc
+++ b/rules/S2055/java/rule.adoc
@@ -3,7 +3,8 @@
 Java serialization is the conversion from objects to byte streams for storage or transmission.
 And later, java deserialization is the reverse conversion, it reconstructs objects from byte streams.
 
-Subtypes of the `java.io.Serializable` interface are called serializable classes.
+To make a java class serializable, this class should implement the `java.io.Serializable` interface directly or
+through its inheritance.
 
 [source,java]
 ----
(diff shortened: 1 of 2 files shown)


rules/S2055/java/rule.adoc
```diff
@@ -1,61 +1,163 @@
 == Why is this an issue?
```
leveretka, line 1: I think in this section you still haven't explained, why this is an issue. I think it's worth mentioning, what happens if you don't follow this rule. Like that there will be an exception at runtime.

rules/S2055/java/rule.adoc
```diff
@@ -1,61 +1,163 @@
...
 
+public class OtherSerializableClass extends SerializableClass {
+  // is also serializable because it is a subtype of Serializable
+}
+----
 
-=== Noncompliant code example
+Serialization stores the fields of serializable classes and ignores the fields of non-serializable classes.
+Deserialization instantiates serializable objects and fields without using any declared constructors. And for the
+non-serializable objects, it calls the implicit or explicit no-argument constructor.
```
leveretka, line 25: I find this section more confusing than the previous description if you don't know what it is about. How about:

"Serializable classes can have non-serializable ancestors, and to initialize fields inherited from such ancestor the no-argument constructor will be invoked. If there's no implicit or explicit no-argument constructor, the error occurs at runtime".

leveretka (changes requested):

rules/S2055/java/rule.adoc
alban-auzeill: @leveretka I have reworked the rule and updated the above screen capture.

leveretka (approved):
LGTM!

PR #2085, opened 2023-06-06
Modify rule S2365: Migrate to new educational format

gregory-paidis-sonarsource:
https://sonarsource.github.io/rspec/#/rspec/S2365/csharp


rules/S2365/vbnet/rule.adoc
```diff
@@ -1,55 +1,47 @@
...
 
-[source,vbnet]
+==== Noncompliant code example
+
+[source,vbnet,diff-id=1,diff-type=noncompliant]
 ----
-Module Module1
-    ' Internal state
-    Dim array = {"apple", "banana", "orange", "pineapple", "strawberry"}
+Module SomeModule
```
csaba-sagi-sonarsource, line 13: NITPICK:
The C# and VB.NET examples are not the same. Here we have a module, while for C# the property is not nested inside a class.

rules/S2365/vbnet/rule.adoc
```diff
@@ -1,55 +1,47 @@
 == Why is this an issue?
 
-Most developers expect property access to be as efficient as field access. However, if a property returns a copy of an array or collection, it will be much slower than a simple field access, contrary to the caller's likely expectations. Therefore, such properties should be refactored into methods so that callers are not surprised by unexpectedly poor performance.
+Most developers expect https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/properties[property] access to be as efficient as https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/objects-and-classes/#fields-and-properties[field] access. However, if a property returns a copy of an array or https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/concepts/collections[collection], it will be much slower than simple field access, contrary to the caller's likely expectations. Therefore, such properties should be refactored into methods so that callers are not surprised by the unexpectedly slower performance.
```
csaba-sagi-sonarsource, line 3: Just thinking out loud:
This is why I am not sure it is a good idea to put the links here. We have duplicated the whole text because of the language-specific links + anyway, the links will be in the resources part (some more duplication) + if the links break we will need to fix it in more places (higher maintenance) + do we really need a link here for it? Advanced users will not click them because they know what it is while people with less experience should anyway look at all the resources in the documentation part.

rules/S2365/rspecator-dotnet.adoc
csaba-sagi-sonarsource, line 1: If I understand it correctly, there is nothing dotnet specific in this file, we can name it just rspecator and if ever a third language comes it can also include it.

rules/S2365/vbnet/rule.adoc
gregory-paidis-sonarsource: That's a good point, I removed the links and moved it to `why-dotnet.adoc`

rules/S2365/csharp/rule.adoc
```diff
@@ -1,16 +1,21 @@
-== Why is this an issue?
+include::../why-dotnet.adoc[]
 
-Most developers expect property access to be as efficient as field access. However, if a property returns a copy of an array or collection, it will be much slower than simple field access, contrary to the caller's likely expectations. Therefore, such properties should be refactored into methods so that callers are not surprised by the unexpectedly poor performance.
+This rule tracks calls to the following methods inside properties:
 
+* https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tolist[Enumerable.ToList]
+* https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.toarray[Enumerable.ToArray]
+* https://learn.microsoft.com/en-us/dotnet/api/system.array.clone[Array.Clone]
```
csaba-sagi-sonarsource, line 7: This should also be part of why-dotnet.adoc

rules/S2365/vbnet/rule.adoc
```diff
@@ -1,55 +1,41 @@
...
-Most developers expect property access to be as efficient as field access. However, if a property returns a copy of an array or collection, it will be much slower than a simple field access, contrary to the caller's likely expectations. Therefore, such properties should be refactored into methods so that callers are not surprised by unexpectedly poor performance.
+== How to fix it
 
+=== Code examples
 
-=== Noncompliant code example
+==== Noncompliant code example
 
-[source,vbnet]
+[source,vbnet,diff-id=1,diff-type=noncompliant]
```
csaba-sagi-sonarsource, line 9: The code snippets are still different, could you please make them identical for c# and vb.net?

csaba-sagi-sonarsource (changes requested):
I left two small notes, please address. After that it is good to go!

PR #2088, opened 2023-06-07
Create tone_guide.adoc

John-Clifton-SonarSource:
Created a tone guide to help rules authors make their content as compelling for their users as possible.

## Review

- [x] text style and tone

### DIFF
--- a/docs/tone_guide.adoc
+++ b/docs/tone_guide.adoc
@@ -1,0 +1,37 @@
+= Tone Guide
+
+When writing a rule, we must consider who we are talking to, what they are trying to achieve, and what their mindset is likely to be.
+
+== Who are we talking to?
+
+We will be talking to a range of people; experienced engineers fixing something for the 10th time, people very new to the language, working on code they don't know well, and everyone in between. It is also important to realize that many of our users are not native English speakers.
+
+What might they be doing, and how do they feel?
+
+* They want to merge a PR and have a failing quality gate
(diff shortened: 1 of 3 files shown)


docs/tone_guide.adoc
```diff
@@ -0,0 +1,37 @@
+= Tone Guide
+
+When writing a rule, we must consider who we are talking to, what they are trying to achieve, and what their mindset is likely to be.
+
+== Who are we talking to?
+
+We will be talking to a range of people; experienced engineers fixing something for the 10th time, people very new to the language, working on code they don't know well, and everyone in between. It is also important to realize that many of our users are not native English speakers.
```
antonioaversa, line 7: I would say not only engineers. 
I would consider their managers too, who may occasionally go to SQ or SC and may click on the rule description to better understand how serious reported issues are.
Should we talk to them too?

John-Clifton-SonarSource: I think that is a fair point. For the moment, I think we want to focus on the people who actively make code changes based on the rules since they are our main consumers. Also, if we follow the guidance in terms of simplicity of language, I think we should end up with rules that are consumable by managers as well as engineers.

hendrik-buchwald-sonarsource (commented):
LGTM :ok_hand: As background information, this addition is related to the following initiative: https://discuss.sonarsource.com/t/tone-guide-for-rule-updates/14380

PR #2089, opened 2023-06-07
Modify rule S3218: LaYC format

cristian-ambrosini-sonarsource:


rules/S3218/csharp/rule.adoc
```diff
@@ -1,4 +1,51 @@
-include::../rule.adoc[]
+== Why is this an issue?
+
+Naming the members of an inner class the same as the static members of its enclosing class is possible but generally considered a bad practice. That's because maintainers may be confused about which members are being used in a given context. Instead the inner class member should be given distinct and descriptive name, and all references to it should be updated accordingly.
```
csaba-sagi-sonarsource, line 3: Do we need the `and all references to it should be updated accordingly` part? I mean if one does not update it, the project will not build.

rules/S3218/csharp/rule.adoc
```diff
@@ -1,4 +1,51 @@
...
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
```
csaba-sagi-sonarsource, line 5: The previous state contained 3 examples, where the second was showcasing why the problem exists. I think we lost this educational information by removing it.
I suggest going back to the previous state with 3 snippets.

rules/S3218/csharp/rule.adoc
csaba-sagi-sonarsource, line 3: Thinking out loud, should we do the changes for java too? As the rule was completely common before?

rules/S3218/csharp/rule.adoc
```diff
@@ -1,4 +1,51 @@
...
+----
+
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions[Common Coding Conventions]
+* https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/nested-types[Nested Types]
 
 ifdef::env-github,rspecator-view[]
```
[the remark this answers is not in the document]
cristian-ambrosini-sonarsource: Umm.. I think I didn't merge rspecator anywhere except where we had both csharp and vbnet (using `rspecator-dotnet`). That's a good point, I will change it in this PR, but I'm afraid we should go through the rules already merged to check this change. This might also be scripted, it seems common for every rules/languages

rules/S3218/csharp/rule.adoc
cristian-ambrosini-sonarsource: I kept the first two examples, removing the last one because it was redundant with the second snippet. I think it's pretty straightforward and we haven't lost any information. What do you think it's missing?

rules/S3218/csharp/rule.adoc
cristian-ambrosini-sonarsource: Yep I thought the same; let's see what the team thinks about that (see Slack conversation)

cristian-ambrosini-sonarsource: The why is now shared with Java.
Using "_inner class_" wording should be valid for each of those languages.
In CSharp _nested_ would be more specific but in Java it means a different thing:
Nested classes are divided into two categories: `non-static` and `static`. `Non-static` nested classes are called **inner classes**. Nested classes that are declared static are called static **nested classes**. Given what the rule is targetting we should be good with a common description using the _inner_ terminology.

rules/S3218/csharp/rule.adoc
csaba-sagi-sonarsource: What I am afraid is that based on these samples, people will only rename the member and from that point the functionality will change as it will return the outer and not the inner. Maybe indeed the intent was to return the inner one. That is why I liked the 3 snippet example that we had.

rules/S3218/java/rule.adoc
```diff
@@ -1,20 +1,46 @@
-include::../rule.adoc[]
+include::../why.adoc[]
 
-== Resources
+[source,text]
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: And also I would not change the snippets for java at all. Let them do that part. That way we improved the rule in a non invasive way.

csaba-sagi-sonarsource (changes requested):
One more round of comments.

rules/S3218/java/rule.adoc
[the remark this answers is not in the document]
csaba-sagi-sonarsource: You cannot have diff-id=1,diff-type=compliant for two snippets, it will not work like that.
https://sonarsource.slack.com/archives/C04P0SB2KJP/p1686150697710199
Please fix it.

csaba-sagi-sonarsource (changes requested):
One last round. I promise!

rules/S3218/rspecator.adoc
```diff
@@ -0,0 +1,14 @@
+ifdef::env-github,rspecator-view[]
+
+'''
+== Implementation Specification
+(visible only on this page)
+
+include::./message.adoc[]
```
csaba-sagi-sonarsource, line 7: NITPICK:
The message and the comments and links can be embedded directly and the files removed.

PR #2094, opened 2023-06-07
Modify rule S2127: Update rule according to the LayC

irina-batinic-sonarsource:


rules/S2127/java/rule.adoc
```diff
@@ -1,16 +1,28 @@
...
+
+Double.longBitsToDouble(i);   // Noncompliant
+Double.longBitsToDouble(b);   // Noncompliant
+Double.longBitsToDouble(s);   // Noncompliant
+Double.longBitsToDouble('c'); // Noncompliant
+Double.longBitsToDouble(42);  // Noncompliant
 ----
 
+== Resources
+* https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Double.html#doubleToLongBits(double)[Oracle Java SE - Double.doubleToLongBits]
```
kaufco, line 25: I would add the following two resources here, because they describe the two different bit patterns for IEEE-754 32 bit and 64 bit floating point values:

- https://en.wikipedia.org/wiki/Double-precision_floating-point_format
- https://en.wikipedia.org/wiki/Single-precision_floating-point_format

rules/S2127/java/rule.adoc
```diff
@@ -1,16 +1,28 @@
...
 int i = 42;
-double d = Double.longBitsToDouble(i);  // Noncompliant
+byte b = 42;
+short s = 42;
+
+Double.longBitsToDouble(i);   // Noncompliant
+Double.longBitsToDouble(b);   // Noncompliant
+Double.longBitsToDouble(s);   // Noncompliant
+Double.longBitsToDouble('c'); // Noncompliant
+Double.longBitsToDouble(42);  // Noncompliant
```
kaufco, line 21: I also would add an example with `int` here (and put this into a "How to fix it" section), because that might also be what the developer really intended:

```
Float.intBitsToFloat(i); // Compliant
```

kaufco (changes requested):

PR #2100, opened 2023-06-07
SONARJAVA-4492 Modify rule S2438: rewrite title and description, simplified code samples

leonardo-pilastri-sonarsource:


rules/S2438/java/rule.adoc
```diff
@@ -1,42 +1,38 @@
...
+The semantics of `Thread` and `Runnable` are different, and while it is technically correct to use `Thread` where a `Runnable` is expected, it is a bad practice to do so.
 
-While it is technically correct to use a ``++Thread++`` where a ``++Runnable++`` is called for, the semantics of the two objects are different, and mixing them is a bad practice that will likely lead to headaches in the future.
-
-
-The crux of the issue is that ``++Thread++`` is a larger concept than ``++Runnable++``. A ``++Runnable++`` is an object whose running should be managed. A ``++Thread++`` expects to manage the running of itself or other ``++Runnables++``. 
+The crux of the issue is that `Thread` is a larger concept than `Runnable`.
+A `Thread` represents a single thread in the OS, and implements the functionalities needed to handle its lifecycle (start, stop, pause).
+It is both a task and a lifecycle management.
+A `Runnable` only represents a task without the lifecycle management.
```
dorian-burihabwa-sonarsource, line 8: Let's reshuffle, the text so that we add knowledge progressively and remove the OS thread mention since it is specifically not mentioned in the JavaDoc
```suggestion
The crux of the issue is that `Thread` is a larger concept than `Runnable`.
A `Runnable` represents a task.
A `Thread` represents a task and its execution management (ie: how it should behave when started, stopped, resumed, ...).
It is both a task and a lifecycle management.
```

dorian-burihabwa-sonarsource (changes requested):

PR #2103, opened 2023-06-07
SONARJAVA-4492 Modify rule S2441: rewrite title and description

leonardo-pilastri-sonarsource:


rules/S2441/java/rule.adoc
```diff
@@ -1,24 +1,33 @@
 == Why is this an issue?
 
-If you have no intention of writting an ``++HttpSession++`` object to file, then storing non-``++serializable++`` objects in it may not seem like a big deal. But whether or not you explicitly serialize the session, it may be written to disk anyway, as the server manages its memory use in a process called "passivation". Further, some servers automatically write their active sessions out to file at shutdown & deserialize any such sessions at startup.
+`HttpSession` is managed by web servers and can be serialized and stored on disk as the server manages its memory use in a process called "passivation".
```
dorian-burihabwa-sonarsource, line 3:
```suggestion
`HttpSession`s are managed by web servers and can be serialized and stored on disk as the server manages its memory use in a process called "passivation" (and later restored during "activation").
```

rules/S2441/java/rule.adoc
```diff
@@ -1,24 +1,33 @@
...
 
-
-The point is, that even though ``++HttpSession++`` does not ``++extend Serializable++``, you must nonetheless assume that it will be serialized, and understand that if you've stored non-serializable objects in the session, errors will result. 
+Even though `HttpSession` does not extend `Serializable`, you must nonetheless assume that it will be serialized.
+If non-serializable objects were stored in the session, its serialization might fail.
```
dorian-burihabwa-sonarsource, line 6:
```suggestion
If non-serializable objects are stored in the session, serialization might fail.
```

dorian-burihabwa-sonarsource (changes requested):
I think we can make a few tweaks in the phrasing here

PR #2105, opened 2023-06-07
Modify rule S4275: Adapt to LaYC

victor-diez-sonarsource:


rules/S4275/javascript/rule.adoc
```diff
@@ -1,38 +1,62 @@
...
-  setY(val: number) { // Noncompliant: field 'y' is not updated
+  get x() { // Noncompliant: field 'x' is not used in the return value
+    return this.#y;
   }
 }
+----
+[source,javascript,diff-id=2,diff-type=compliant]
+----
+class A {
+  _x: number = 0;
```
vilchik-elena, line 48: you renamed this variable compared to the previous code sample

rules/S4275/javascript/rule.adoc
```diff
@@ -83,6 +75,15 @@ Object.defineProperty(o, 'x', {
   }
 });
 ----
+
+== Resources
+
+=== Documentation
+
+- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get[MDN get]
```
vilchik-elena, line 83: Add `-` after MDN

rules/S4275/javascript/rule.adoc
```diff
@@ -83,6 +75,15 @@ Object.defineProperty(o, 'x', {
   }
 });
 ----
+
+== Resources
+
+=== Documentation
```
vilchik-elena, line 81: I suggest to add this page as well
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

rules/S4275/javascript/rule.adoc
```diff
@@ -1,38 +1,62 @@
 == Why is this an issue?
+Getters and setters provide a way to enforce encapsulation by providing methods that give controlled access to class fields. However, in classes with multiple fields, it is not unusual that copy and paste is used to quickly create the needed getters and setters, which can result in the wrong field being accessed by a getter or setter.
 
-include::../description.adoc[]
+This rule raises an issue in any of these cases:
 
-=== Noncompliant code example
+* A setter does not update the field with the corresponding name (if it exists).
+* A getter does not return any value or does not access the field with the corresponding name (if it exists).
```
vilchik-elena, line 7: Let's add code example where getter does not return anything (may be even conditionally)

rules/S4275/javascript/rule.adoc
```diff
@@ -1,38 +1,62 @@
 == Why is this an issue?
+Getters and setters provide a way to enforce encapsulation by providing methods that give controlled access to class fields. However, in classes with multiple fields, it is not unusual that copy and paste is used to quickly create the needed getters and setters, which can result in the wrong field being accessed by a getter or setter.
 
-include::../description.adoc[]
+This rule raises an issue in any of these cases:
```
vilchik-elena, line 4:
```suggestion
This rule raises an issue in the following cases.
```

rules/S4275/javascript/rule.adoc
vilchik-elena, line 7: Did you consider splitting this point into two? 
* getter does not return anything
* getter returns the wrong thing

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #2106, opened 2023-06-07
Modify rule S3358: LaYC for Nested ternaries

frederic-tingaud-sonarsource:


rules/S3358/cfamily/rule.adoc
```diff
@@ -1,28 +1,10 @@
...
-  if (p1 > p2) {
-    return p1 > p3 ? p1 : p3;
-  } else {
-    return p2 > p3 ? p2 : p3;
-  }
-}
-----
+include::../description.adoc[]
 
 === Exceptions
```
amelie-renard-sonarsource, line 9: Maybe rework this paragraph as well with a concrete example? For example:
"
In C++11, nested ternaries inside constexpr functions are compliant. Indeed, in C++11, such functions are limited to a single return statement, so no other solution is possible. The later standards lifted this restriction, so this exception is no longer valid.
+ code example
"

rules/S3358/php/rule.adoc
```diff
@@ -1,27 +1,11 @@
...
-----
-function get_readable_status($is_running, $has_errors) {
-  if ($is_running) {
-    return "Running";
-  }
-  return $has_errors ? "Failure. " : "Succeeded ";
-}
-----
 
 === Exceptions
```
amelie-renard-sonarsource, line 10: A like adding
` // Compliant by exception`
in the code example, but that's up to you

rules/S3358/python/rule.adoc
```diff
@@ -15,7 +10,7 @@ class Job:
         return "Running" if job.is_running else "Failed" if job.errors else "Succeeded"  # Noncompliant
 ----
 
-=== Compliant solution
+Instead, use another line to express the nested operation as a separate statement.
 
 [source,python]
 ----
```
amelie-renard-sonarsource, line 16: In the exception: is "comprehensions" a word that every Python developer, even a beginner, knows? Should we add a code example?

rules/S3358/description.adoc
```diff
@@ -1,4 +1,7 @@
-Just because you _can_ do something, doesn't mean you should, and that's the case with nested ternary operations. Nesting ternary operators results in the kind of code that may seem clear as day when you write it, but six months later will leave maintainers (or worse - future you) scratching their heads and cursing.
+Nested ternaries are hard to read and the operator prevalence rules might not be obvious to maintainers.
 
+include::{noncompliant}[]
 
-Instead, err on the side of clarity, and use another line to express the nested operation as a separate statement.
+Instead, use another line to express the nested operation as a separate statement.
```
amelie-renard-sonarsource, line 5:
```suggestion
Instead, use another line to express the nested operation in a separate statement.
```

rules/S3358/description.adoc
```diff
@@ -1,4 +1,7 @@
-Just because you _can_ do something, doesn't mean you should, and that's the case with nested ternary operations. Nesting ternary operators results in the kind of code that may seem clear as day when you write it, but six months later will leave maintainers (or worse - future you) scratching their heads and cursing.
+Nested ternaries are hard to read and the operator prevalence rules might not be obvious to maintainers.
```
amelie-renard-sonarsource, line 1: I'm afraid "Operator prevalence rules" might be hard to understand. What about:
```suggestion
Nested ternaries are hard to read and can make the order of operations complex to understand.
```

amelie-renard-sonarsource (changes requested):

frederic-tingaud-sonarsource:
@guillaume-dequenne-sonarsource could you check that the new wording of the Python exception and the code example that comes with it make sense?

rules/S3358/description.adoc
frederic-tingaud-sonarsource: Way better, thanks!

rules/S3358/description.adoc
frederic-tingaud-sonarsource: Changed

rules/S3358/python/rule.adoc
frederic-tingaud-sonarsource: Proposed a change, to check with the Python squad

rules/S3358/php/rule.adoc
frederic-tingaud-sonarsource: done

rules/S3358/cfamily/rule.adoc
frederic-tingaud-sonarsource: Done

rules/S3358/cfamily/rule.adoc
```diff
@@ -1,36 +1,25 @@
...
-=== Noncompliant code example
+=== Exceptions
 
-[source,cpp]
-----
-int max(int p1, int p2, int p3) {
-  return p1 > p2 ? (p1 > p3 ? p1 : p3) : (p2 > p3 ? p2 : p3); // Noncompliant
-}
-----
+In {cpp}11 mode the issue is not raised for ternary operators inside `constexpr` functions.
```
amelie-renard-sonarsource, line 11:
```suggestion
In {cpp}11, the rule ignores ternary operators inside `constexpr` functions.
```

amelie-renard-sonarsource (approved):

rules/S3358/python/rule.adoc
guillaume-dequenne-sonarsource: Currently the rule doesn't report on comprehensions, whatever they are (list, set dictionary and generators). It would make sense to keep the "comprehension" part without specifying it has to be a list comprehension.

I'd tend to think it's a common concept in Python, but the added example certainly helps.

rules/S3358/python/rule.adoc
```diff
@@ -29,7 +24,12 @@ class Job:
 
 === Exceptions
 
-No issue is raised on conditional expressions in comprehensions.
+No issue is raised on conditional expressions in list comprehensions.
```
frederic-tingaud-sonarsource, line 27:
```suggestion
No issue is raised on conditional expressions in comprehensions.
```

PR #2109, opened 2023-06-08
Modify rule S1699: LaYC format

cristian-ambrosini-sonarsource:
## Review

- [ ] text style and tone


rules/S1699/java/rule.adoc
```diff
@@ -2,7 +2,36 @@
 
 include::../description.adoc[]
```
csaba-sagi-sonarsource, line 3: You could move the descirption here directly just like with the noncompliant example. I think it is not used anywhere else.

rules/S1699/csharp/rule.adoc
```diff
@@ -64,4 +73,13 @@ The main difference between C# and Java in this case is that you need to opt in
 === on 23 Jun 2015, 11:51:47 Ann Campbell wrote:
 looks good to me [~tamas.vajk]
 
+=== on 8 Jun 2023, 11:15:00 Cristian Ambrosini wrote:
```
csaba-sagi-sonarsource, line 76: If I understood correctly the rspecator view should be common for all the languages if possible. So, you might move this and merge with other languages rspecator part.

csaba-sagi-sonarsource (changes requested):
Left a few small comments.

rules/S1699/java/rule.adoc
```diff
@@ -1,19 +1,48 @@
...
-include::../description.adoc[]
+Calling an overridable method from a constructor could result in failures or strange behaviors when instantiating a subclass which overrides the method.
 
-include::../noncompliant.adoc[]
+
+For example:
+
+* The subclass class constructor starts by contract by calling the parent class constructor.
+* The parent class constructor calls the method, which has been overridden in the child class.
+* If the behavior of the child class method depends on fields that are initialized in the child class constructor, unexpected behavior (like a ``++NullPointerException++``) can result, because the fields aren't initialized yet.
```
csaba-sagi-sonarsource, line 10:
```suggestion
* If the behavior of the child class method depends on fields that are initialized in the child class constructor, unexpected behavior (like a `NullPointerException`) can result, because the fields aren't initialized yet.
```

csaba-sagi-sonarsource (approved):
LGTM!

PR #2118, opened 2023-06-08
Modify rule S3973: Remove unsupported languages

sebastien-marichal:
## Review

- [ ] text style and tone


rules/S3973/csharp/rule.adoc
```diff
@@ -47,7 +47,7 @@ include::../message.adoc[]
 
 === Highlighting
 
-* Primary: ``++if (condition)++``
+* Primary: `if (condition)`
```
csaba-sagi-sonarsource, line 50: Looking at the implementation, it covers a lot more than just simple Ifs. See [here](https://github.com/SonarSource/sonar-dotnet/blob/351706c6ebe74cf2dbe9e47447aac7a569451c73/analyzers/src/SonarAnalyzer.CSharp/Rules/IndentSingleLineFollowingConditional.cs#L34). Could you please update the C# description to reflect this.

csaba-sagi-sonarsource (changes requested):
One small comment.

csaba-sagi-sonarsource (approved):
LGTM!

PR #2121, opened 2023-06-09
Modify s6426 layc

cynthiabethea:
Reworded the text slightly. I left the noncompliant/compliant titles in because I didn't think it was worthwhile to move them to the second tab. 

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6426/javascript/rule.adoc
+++ b/rules/S6426/javascript/rule.adoc
@@ -1,6 +1,7 @@
 == Why is this an issue?
 
-Testing frameworks like Mocha and Jest provide a way to run a single test case for a file by appending `.only()` to the test function. This is useful when debugging a specific use case. However, it should not be used in production or development, as it is likely a leftover from debugging and serves no purpose in those contexts. 
+
+When using testing frameworks like Mocha and Jest, appending `.only()` to the test function allows running a single test case for a file. This is useful when debugging a specific use case. However, it should not be used in production or development, as it is likely a leftover from debugging and serves no purpose in those contexts. 
 
 It is strongly recommended not to include `.only()` usages in version control.
 
--- a/rules/S6426/javascript/rule.adoc
+++ b/rules/S6426/javascript/rule.adoc
@@ -1,7 +1,8 @@
 == Why is this an issue?
 
-Testing frameworks like Mocha and Jest provide a way to run a single test case for a file by appending `.only()` to the test function. This is convenient when debugging a specific use case. It makes, however, no sense to have it in production or development, and it was likely left by mistake after such a debugging session.
-Therefore, it is strongly recommended not to commit usages of `.only()` to version control.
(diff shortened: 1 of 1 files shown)


rules/S6426/javascript/rule.adoc
```diff
@@ -1,7 +1,9 @@
 == Why is this an issue?
 
-Testing frameworks like Mocha and Jest provide a way to run a single test case for a file by appending `.only()` to the test function. This is convenient when debugging a specific use case. It makes, however, no sense to have it in production or development, and it was likely left by mistake after such a debugging session.
-Therefore, it is strongly recommended not to commit usages of `.only()` to version control.
+
+When using testing frameworks like Mocha and Jest, appending `.only()` to the test function allows running a single test case for a file. This is useful when debugging a specific use case. However, it should not be used in production or development, as it is likely a leftover from debugging and serves no purpose in those contexts. 
+
+It is strongly recommended not to include `.only()` usages in version control.
 
 === Noncompliant code example
```
vilchik-elena, line 8: Add `// Noncompliant` in the code sample

rules/S6426/javascript/rule.adoc
```diff
@@ -1,7 +1,9 @@
 == Why is this an issue?
```
vilchik-elena, line 1: Can we add some resources?

rules/S6426/javascript/rule.adoc
vilchik-elena, line 8: I suggest to drop these titles, and move last paragraph in between code samples to glue things.

vilchik-elena (changes requested):

rules/S6426/javascript/rule.adoc
```diff
@@ -1,7 +1,9 @@
...
 
-Testing frameworks like Mocha and Jest provide a way to run a single test case for a file by appending `.only()` to the test function. This is convenient when debugging a specific use case. It makes, however, no sense to have it in production or development, and it was likely left by mistake after such a debugging session.
-Therefore, it is strongly recommended not to commit usages of `.only()` to version control.
+
+When using testing frameworks like Mocha and Jest, appending `.only()` to the test function allows running a single test case for a file. This is useful when debugging a specific use case. However, it should not be used in production or development, as it is likely a leftover from debugging and serves no purpose in those contexts.
```
vilchik-elena, line 4: Can we also mention that having it in the code means that no other test from this file is executed.

rules/S6426/javascript/rule.adoc
cynthiabethea: There should be resources at the end, do you think we need more?

cynthiabethea: @vilchik-elena the other changes have been made

vilchik-elena: @cynthiabethea Indeed, missed it.

Could you add subtitles like the format requires?
(current section is result of automatic migration)

vilchik-elena (approved):

PR #2130, opened 2023-06-09
Modify rule S107: LaYC format

amelie-renard-sonarsource:


rules/S107/php/split-example.adoc
```diff
@@ -0,0 +1,11 @@
+[source,cpp]
```
hendrik-buchwald-sonarsource, line 1:
```suggestion
[source,php]
```

rules/S107/php/noncompliant.adoc
```diff
@@ -0,0 +1,7 @@
+[source,php]
+----
+function setCoordinates($x1, $y1, $z1, $x2, $y2, $z2) // Noncompliant
+{
```
hendrik-buchwald-sonarsource, line 4:
```suggestion
function setCoordinates($x1, $y1, $z1, $x2, $y2, $z2) { // Noncompliant
```

Not required but this keeps it in the same style.

rules/S107/php/struct-example.adoc
```diff
@@ -0,0 +1,12 @@
+[source,php]
+----
+class Point { // In geometry, Point is a logical structure to group data
+    public $x;
+    public $y;
+    public $z;
+}
+
+function setCoordinates($p1, $p2) {
```
hendrik-buchwald-sonarsource, line 9:
```suggestion
function setCoordinates(Point $p1, Point $p2) {
```

This way it is clear that the arguments represent an object of that type.

rules/S107/php/struct-example.adoc
```diff
@@ -0,0 +1,12 @@
+[source,php]
+----
+class Point { // In geometry, Point is a logical structure to group data
+    public $x;
+    public $y;
+    public $z;
+}
```
hendrik-buchwald-sonarsource, line 7:
```suggestion
class Point
{
    // In geometry, Point is a logical structure to group data
    public function __construct(
        public int $x,
        public int $y,
        public int $z
    ) {}
}
```

The code is correct but not recommendable. I have included a short PHP 8 version.

rules/S107/php/split-example.adoc
```diff
@@ -0,0 +1,11 @@
...
+----
+// Each function does a part of what the original setCoordinates function was doing, so confusion risks are lower
+function setOrigin($x, $y, $z) {
```
hendrik-buchwald-sonarsource, line 4:
```suggestion
function setOrigin(int $x, int $y, int $z) {
```

Not required but in my opinion we should lead with good example in the samples.

rules/S107/php/split-example.adoc
```diff
@@ -0,0 +1,11 @@
...
+   // ...
+}
+
+function setSize($width, $height, $depth) {
```
hendrik-buchwald-sonarsource, line 8:
```suggestion
function setSize(int $width, int $height, float $depth) {
```

Not required but in my opinion we should lead with good example in the samples.

hendrik-buchwald-sonarsource (commented):
As requested, some comments about PHP.

hendrik-buchwald-sonarsource (commented):
As requested, some comments about PHP.

amelie-renard-sonarsource:
Perfect, thanks @hendrik-buchwald-sonarsource !

rules/S107/python/rule.adoc
```diff
@@ -1,28 +1,46 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+Functions, methods, or lambdas with a long parameter list are difficult to use, as maintainers must figure out the role of each parameter and keep track of their position.
 
-=== Noncompliant code example
+[source,python]
+----
+def setCoordinates(x1, y1, z1, x2, y2, z2): # Noncompliant
```
guillaume-dequenne-sonarsource, line 7: It would be better to use snake case to respect Python conventions (`set_coordinates`)

rules/S107/python/rule.adoc
```diff
@@ -1,28 +1,46 @@
...
 
 [source,python]
 ----
-def do_something(param1, param2, param3, param4, param5):
-	...
+# Each function does a part of what the original setCoordinates function was doing, so confusion risks are lower
+def setOrigin(x, y, z):
+   # ...
+
+def setSize(width, height, depth):
```
guillaume-dequenne-sonarsource, line 21: Same here.

rules/S107/python/rule.adoc
```diff
@@ -1,28 +1,46 @@
...
 ----
-def do_something(param1, param2, param3, param4):
-	...
+@dataclass
+class Point: # In geometry, Point is a logical structure to group data
+    x: int
+    y: int
+    z: int
+
+def setCoordinates(p1, p2):
```
guillaume-dequenne-sonarsource, line 35: Same here.
Also, I assume `p1` and `p2` are meant to be `Point` objects? In which case you can type hint them accordingly, to make the example clearer: 
```
def set_coordinates(p1: Point, p2: Point): 
  ...
```

rules/S107/python/rule.adoc
```diff
@@ -1,28 +1,46 @@
...
+    # ...
 ----
 
+This rule raises an issue when a function, a method, or a lambda has more parameters than the given number.
```
guillaume-dequenne-sonarsource, line 39: This sentence seems odd (not only for Python): the given number of what?
I understand we're referring to the rule property `max_parameter` that can vary depending on the language, but it's not very clear in isolation.

guillaume-dequenne-sonarsource (commented):
Left a few comments related to the Python version of the rule.

rules/S107/python/rule.adoc
amelie-renard-sonarsource: Yes, it's correct. Fixed, thanks!

rules/S107/python/rule.adoc
amelie-renard-sonarsource: Thanks for the feedback. I have changed the end of the sentence, is it clearer now?

rules/S107/go/split-example.adoc
```diff
@@ -0,0 +1,11 @@
+[source,go]
+----
+// Each function does a part of what the original setCoordinates function was doing, so confusion risks are lower
+void setOrigin(x int, y int, z int) {
```
dorian-burihabwa-sonarsource, line 4:
```suggestion
func setOrigin(x int, y int, z int) {
```

rules/S107/go/split-example.adoc
```diff
@@ -0,0 +1,11 @@
...
+   // ...
+}
+
+void setSize(width int, height int, depth int) {
```
dorian-burihabwa-sonarsource, line 8:
```suggestion
func setSize(width int, height int, depth int) {
```

rules/S107/ruby/noncompliant.adoc
```diff
@@ -0,0 +1,6 @@
+[source,ruby]
+----
+def setCoordinates(x1, y1, z1, x2, y2, z2) # Noncompliant
```
sebastian-hungerecker-sonarsource, line 3: Ruby uses snake_case, not camelCase. Also setters would usually be defined using the `=` operator, not the `set_` prefix, so `def coordinates=`, not `def set_coordinates`, but that doesn't really work when there's more than one argument.

rules/S107/ruby/split-example.adoc
```diff
@@ -0,0 +1,11 @@
+[source,ruby]
+----
+# Each function does a part of what the original setCoordinates function was doing, so confusion risks are lower
+def setOrigin(int x, int y, int z)
```
sebastian-hungerecker-sonarsource, line 4: Given my earlier point about the `set_` prefix, I'd rename these to `move` and `resize`.

Also Ruby doesn't have parameter types (which I'm sure you know, but added in this example out of habit).

rules/S107/ruby/struct-example.adoc
```diff
@@ -0,0 +1,8 @@
+[source,ruby]
+----
+Struct.new("Point", :x, :y, :z) # In geometry, Point is a logical structure to group data
```
sebastian-hungerecker-sonarsource, line 3: This should be `Point = Struct.new(:x, :y, :z)`

rules/S107/ruby/struct-example.adoc
```diff
@@ -0,0 +1,8 @@
...
+
+def setCoordinates(p1, p2)
```
sebastian-hungerecker-sonarsource, line 5: I'd be inclined to want to change this to `def coordinates=(points)` where `points` is expected to be an array of two points, but that may make the example unclear. Plus putting the arguments into an array kinda circumvents this rule, so we probably shouldn't do that in the examples.

Maybe we can use an example that's not about setters?

sebastian-hungerecker-sonarsource (changes requested):

rules/S107/ruby/struct-example.adoc
sebastian-hungerecker-sonarsource: Maybe instead of a setter, we can define the constructor? Something like this:

Noncompliant:

```ruby
class Cube
  def initialize(x1, y1, z1, x2, y2, z2) # Noncompliant
    # ...
  end
end
```

Compliant:

```ruby
Point = Struct.new(:x, :y, :z)

class Cube
  def initialize(p1, p2) # Compliant
    # ...
  end
end
```

Though in that case, I guess we'd have to drop the example with `move` and `resize` because it wouldn't really be good practice to remove the constructor and instead require the user to call two methods to initialize the object instead of having a constructor.

rules/S107/scala/struct-example.adoc
```diff
@@ -0,0 +1,12 @@
+[source,scala]
+----
+class Point: // In geometry, Point is a logical structure to group data
+  public var _x = 0
+  public var _y = 0
+  public var _z = 0
+end Point
```
dorian-burihabwa-sonarsource, line 7: Because most of our examples use Scala2, I would suggest this version
```suggestion
class Point(var x: Int, var y: Int, var z: Int) { // In geometry, Point is a logical structure to group data
}

```

dorian-burihabwa-sonarsource (changes requested):
Thank you for the great work,  I reviewed the examples for Apex, Go, Kotlin, Java and Scala (I saw that Ruby was commented on as I was reviewing so I did not cover these samples).

The examples are pretty good but some might be tweaked to follow conventions (eg: implicit or explicit function types) or language versions.

Ping me if you want another review on the same languages

rules/S107/javascript/noncompliant.adoc
```diff
@@ -0,0 +1,6 @@
+[source,javascript]
+----
+void setCoordinates(int x1, int y1, int z1, int x2, int y2, int z2) { // Noncompliant
```
vilchik-elena, line 3:
```suggestion
function setCoordinates(x1, y1, z1, x2, y2, z2) { // Noncompliant
```

rules/S107/javascript/split-example.adoc
```diff
@@ -0,0 +1,11 @@
+[source,javascript]
+----
+// Each function does a part of what the original setCoordinates function was doing, so confusion risks are lower
+void setOrigin(int x, int y, int z) {
+   // ...
+}
+
+void setSize(int width, int height, int depth) {
+   // ...
+}
```
vilchik-elena, line 10:
```suggestion
function setOrigin(x, y, z) {
   // ...
}

function setSize(width, height, depth) {
   // ...
}
```

rules/S107/javascript/struct-example.adoc
```diff
@@ -0,0 +1,12 @@
...
+----
+struct Point { // In geometry, Point is a logical structure to group data
+    int x;
+    int y;
+    int z;
+};
+
+void setCoordinates(Point p1, Point p2) {
+    // ...
+}
```
vilchik-elena, line 11:
```suggestion
// In geometry, Point is a logical structure to group data
let point1 = { x: x1, y: y1, z: z1};
let point2 = { x: x1, y: y1, z: z1};
setCoordinates(point1, point2);

function setCoordinates(p1, p2) {
   // ...
}
```

vilchik-elena (approved):

sebastian-hungerecker-sonarsource (approved):
LGTM for Ruby.

rules/S107/csharp/rule.adoc
```diff
@@ -1,31 +1,55 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+Methods with a long parameter list are difficult to use, as maintainers must figure out the role of each parameter and keep track of their position.
```
martin-strecker-sonarsource, line 3: Consider making "function" an attribute and use it in rules/S107/rule.adoc:
`{function}s with a long parameter list are difficult to use, as maintainers must figure out the role of each parameter and keep track of their position.`

You could then define it here `:function: method` and reuse the existing documents as you do with most of the other rules.

Hint: We had some discussion in our squad to create a common dictionary for all languages for such cases (concepts and keywords that have different names/casing but refer to the same thing). This would allow us to resolve such minor differences.

You can find the discussion here: https://sonarsource.slack.com/archives/C059FQ512EA/p1686562366420019

One major problem with this is also affecting you here: You need two variables because you need the capitalized and the lowercase variant as asciidoc does not have any built-in functions to solve this.

amelie-renard-sonarsource: Thanks, @martin-strecker-sonarsource for your insight. TBH, I hesitated to introduce these two variables (one upper case, one lower case). So I thought I'd wait for the rules to be approved before deciding.
But indeed, I'll add them.

The common dictionary is a nice idea!

martin-strecker-sonarsource: Rewording the sentence might be another option.

rules/S107/apex/rule.adoc
```diff
@@ -1,26 +1,8 @@
-== Why is this an issue?
+:noncompliant: apex/noncompliant.adoc
+:split-example: apex/split-example.adoc
+:struct-example: apex/struct-example.adoc
```
martin-strecker-sonarsource, line 3: You may consider shortening the definition to the only part that ever changes:
```suggestion
:language: apex
```

amelie-renard-sonarsource: Indeed, I didn't think of that. Thanks!

rules/S107/plsql/rule.adoc
```diff
@@ -1,8 +1,12 @@
...
+Functions and procedures with a long parameter list are difficult to use, as maintainers must figure out the role of each parameter and keep track of their position.
 
-=== Noncompliant code example
+The solution can be to:
+
+* Split the function or the procedure into smaller ones
+
+* Find a better data structure for the parameters
 
 With the default threshold of 10:
```
martin-strecker-sonarsource, line 11:
```suggestion
With a customized threshold of 5 (the default is 10):
```

Noncompliant
```sql
SET SERVEROUTPUT ON

CREATE PROCEDURE storeCube( -- Noncompliant, too many parameters
  x1 PLS_INTEGER,
  y1 PLS_INTEGER,
  z1 PLS_INTEGER,
  x2 PLS_INTEGER,
  y2 PLS_INTEGER,
  z2 PLS_INTEGER)
BEGIN
  -- store the qube values
END;
/
BEGIN
  storeCube(0, 0, 0, 1, 1, 1);
END;
/

DROP PROCEDURE storeCube;
```

Split
```sql
SET SERVEROUTPUT ON

CREATE PROCEDURE storeVector(
  x PLS_INTEGER,
  y PLS_INTEGER,
  z PLS_INTEGER)
BEGIN
  -- store the vectors values
END;
/
BEGIN
storeVector(0, 0, 0);
storeVector(1, 1, 1);
END;
/

DROP PROCEDURE storeVector;
```
struct
```sql
SET SERVEROUTPUT ON

CREATE TYPE point AS OBJECT(
  x PLS_INTEGER,
  y PLS_INTEGER,
  z PLS_INTEGER);

CREATE PROCEDURE storeQube(
  p1 point,
  p2 point)
BEGIN
END;
/
BEGIN
storeQube(point(0, 0, 0), point(1, 1, 1))
END;
/

DROP PROCEDURE storeQube;
DROP TYPE point;
```

rules/S107/vb6/rule.adoc
```diff
@@ -1,34 +1,13 @@
...
 
-A long parameter list can indicate that a new structure should be created to wrap the numerous parameters or that the procedure is doing too many things.
+Procedures with a long parameter list are difficult to use, as one must figure out the role of each parameter.
 
-=== Noncompliant code example
+The solution can be to:
 
-With a maximum number of 3 parameters:
+* Split the procedure into smaller ones
+* Find a better data structure for the parameters that group data in a way that makes sense for the specific application domain
```
martin-strecker-sonarsource, line 8: === Noncompliant code example

```vb6
Public Sub SetCoordinates(ByVal x1 As Integer, ByVal y1 As Integer, ByVal z1 As Integer, ByVal x2 As Integer, ByVal y2 As Integer, ByVal z2 As Integer) ' Noncompliant
  ' ...
End Sub
```

=== Compliant solution

Split

```vb6
Public Sub SetOrigin(ByVal x As Integer, ByVal y As Integer, ByVal z As Integer)
  ' ...
End Sub

Public Sub SetSize(ByVal width As Integer, ByVal height As Integer, ByVal depth As Integer)
  ' ...
End Sub
```

struct
```vb6
Public Type Point
    X As Integer
    Y As Integer
    Z As Integer
End Type

Public Sub SetCoordinates(p1 As Point, p2 As Point)
  ' ...
End Sub
```

amelie-renard-sonarsource: Thanks!

rules/S107/plsql/rule.adoc
amelie-renard-sonarsource: Thanks!

rules/S107/cfamily/rule.adoc
```diff
@@ -1,27 +1,7 @@
-== Why is this an issue?
+:language: cfamily
```
frederic-tingaud-sonarsource, line 2: :upper_function: Function

:function: function

rules/S107/plsql/rule.adoc
```diff
@@ -1,65 +1,20 @@
...
 
-BEGIN
-  DBMS_OUTPUT.PUT_LINE('Sum is ' || sumCorrected(11));
-END;
-/
+* Find a better data structure for the parameters that group data in a way that makes sense for the specific application domain
 
-DROP FUNCTION sumCorrected;
-----
+include::struct-example.adoc[]
```
frederic-tingaud-sonarsource, line 15: It's not necessary to have these as includes, they could be inlined. But it's not a big deal, if you prefer like that I'm ok with it.

frederic-tingaud-sonarsource (changes requested):

rules/S107/scala/noncompliant.adoc
```diff
@@ -0,0 +1,23 @@
+[source,scala]
+----
+def setCoordinates(x1:Int, y1:Int, z1:Int, x2:Int, y2:Int, z2:Int): Unit = { // Noncompliant
+    // ...
+}
+----
+
+ifdef::env-github,rspecator-view[]
```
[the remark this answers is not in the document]
amelie-renard-sonarsource: Indeed, removed

frederic-tingaud-sonarsource (approved):

rules/S107/vbnet/rule.adoc
```diff
@@ -1,50 +1,31 @@
...
+
 End Class
 
-Public Class DerivedClass
+Class DerivedClass
     Inherits BaseClass
 
     Public Sub New(Param1 As Integer, Param2 As Integer, Param3 As Integer, Param4 As Integer, Param5 As Long)
-        MyBase.New(Param1) ' Compliant, the parameters intended for the base class constructor do not count in the sum of the parameter list.
+        MyBase.New(Param1) ' Compliant by exception
```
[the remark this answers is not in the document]
amelie-renard-sonarsource: Fixed in https://github.com/SonarSource/rspec/pull/2130/commits/5368290adb82c87176ced10c7639c244d7ff1a34

rules/S107/ruby/noncompliant.adoc
```diff
@@ -0,0 +1,11 @@
...
+----
+def modify(x1, y1, z1, x2, y2, z2) # Noncompliant
+ # ...
+end
+----
+
+
+
+
+
```
dorian-burihabwa-sonarsource, line 11: I guess we can remove the extra blank lines
```suggestion
```

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Checked for Apex, Go, Java, Kotlin, Ruby, and Scala.

I left a suggestion on the ruby non-compliant example that contains some extra blank lines

PR #2133, opened 2023-06-09
Modify rule S108: LaYC format

amelie-renard-sonarsource:


rules/S108/exceptions.adoc
```diff
@@ -1,3 +1,3 @@
 === Exceptions
 
-When a block contains a comment, this block is not considered to be empty.
\ No newline at end of file
+A code block is not empty when it contains a comment.
```
frederic-tingaud-sonarsource, line 3:
```suggestion
A code block is not considered empty when it contains a comment.
```
I think the "considered" word is important to keep because from an AST point of view, it _is_ empty.

rules/S108/noncompliant.adoc
```diff
@@ -1,6 +1,4 @@
-=== Noncompliant code example
-
 [source,text]
```
frederic-tingaud-sonarsource, line 1:
```suggestion
[source,java]
```

rules/S108/description.adoc
```diff
@@ -1 +1,8 @@
-Most of the time a block of code is empty when a piece of code is really missing. So such empty block must be either filled or removed.
+== Why is this an issue?
+
+An empty code block may indicate that a piece of code is missing. 
+
+include::{example}[]
+
+Removing or filling the empty code blocks takes away ambiguity.
```
frederic-tingaud-sonarsource, line 7:
```suggestion
Removing or filling the empty code blocks takes away ambiguity and will generally result in a simpler and less surprising code.
```

frederic-tingaud-sonarsource (changes requested):

rules/S108/description.adoc
amelie-renard-sonarsource: Modified

rules/S108/noncompliant.adoc
amelie-renard-sonarsource: Fixed

rules/S108/exceptions.adoc
amelie-renard-sonarsource: I rewrote it differently. WDYT?

frederic-tingaud-sonarsource: Great, thanks!

rules/S108/description.adoc
```diff
@@ -1 +1,7 @@
-Most of the time a block of code is empty when a piece of code is really missing. So such empty block must be either filled or removed.
+== Why is this an issue?
+
+An empty code block is confusing and surprising for maintainers as it may indicate missing code.
```
frederic-tingaud-sonarsource, line 3: Not a fan of this wording, because it implies that it's because code might be missing that it is surprising, while the two are independent.
```cpp
for (int i = 0; i < 10; doSomething(), ++i) {}
```
has no missing code but is super surprising.

frederic-tingaud-sonarsource (changes requested):
Just one remaining point where I don't completely agree.

frederic-tingaud-sonarsource (approved):

PR #2134, opened 2023-06-09
S4015: Migrate to LAYC level 2

gregory-paidis-sonarsource:
## Review

- [ ] text style and tone

https://sonarsource.github.io/rspec/#/rspec/S4015/csharp

### DIFF
--- a/rules/S4015/csharp/rule.adoc
+++ b/rules/S4015/csharp/rule.adoc
@@ -1,32 +1,65 @@
 == Why is this an issue?
 
-Changing an inherited member to ``++private++`` will not prevent access to the base class implementation.
-
-
-This rule raises an issue when a ``++private++`` method in an unsealed type has a signature that is identical to a ``++public++`` method declared in a base type.
-
-
-=== Noncompliant code example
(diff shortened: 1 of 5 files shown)


rules/S4015/csharp/rule.adoc
```diff
@@ -1,80 +1,106 @@
 == Why is this an issue?
 
-Changing an inherited member to ``++private++`` will not prevent access to the base class implementation.
+Changing an inherited method that is not an https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/override[override] to https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/private[private] will not prevent access to the base class implementation. Furthermore, having a derived class shadow the name of a base method can lead to confusion.
```
sebastien-marichal, line 3: I am not sure of the relevance of this:
> ...will not prevent access to the base class implementation.
I think it can be removed, as it might not be why they changed the access modifier.

My suggestions:
```suggestion
Decreasing the https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers[accessibility level] of an inherited method that is not https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/override[overridable] to https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/private[private] will shadow the name of the base method and can lead to confusion.
```

sebastien-marichal (changes requested):
Good job!
Only one comment :)

sebastien-marichal (approved):
LGTM!

PR #2135, opened 2023-06-09
Modify rule S3966: Migrate to LaYC

mary-georgiou-sonarsource:


rules/S3966/csharp/rule.adoc
```diff
@@ -1,55 +1,39 @@
...
+{
 
-[source,csharp]
-----
-using var d = new Disposable();
+}
 ----
 
 
-ifdef::env-github,rspecator-view[]
```
sebastien-marichal, line 37: I think we should keep this `ifdef`, otherwise it might appear in the product

rules/S3966/csharp/rule.adoc
```diff
@@ -1,55 +1,39 @@
...
+include::../see.adoc[]
 
-'''
 == Implementation Specification
```
sebastien-marichal, line 37: Why not keep the `rspecator.adoc` file?

We did add a rspecator file in those cases (several languages) where we would include adoc files such as `message` or `highlighting`. We would then include the rspecator file in each languages.

rules/S3966/description.adoc
```diff
@@ -0,0 +1,5 @@
+== Why is this an issue?
+
+Disposing an object twice in the same method, either with the `using` keyword or by calling `Dispose` directly, is confusing and error-prone. For example, another developer might try to use an already-disposed object, or there can be runtime errors for specific paths in the code.
+
+In addition, even if the documentation[https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?redirectedfrom=MSDN#System_IDisposable_Dispose] explicitly states that calling the `Dispose` method multiple times should not throw an exception, some implementations still do it. Thus it is safer to not dispose of an object twice when possible.
```
sebastien-marichal, line 5:
```suggestion
In addition, even if the https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?redirectedfrom=MSDN#System_IDisposable_Dispose[documentation] explicitly states that calling the `Dispose` method multiple times should not throw an exception, some implementations still do it. Thus it is safer to not dispose of an object twice when possible.
```

rules/S3966/description.adoc
```diff
@@ -0,0 +1,5 @@
+== Why is this an issue?
+
+Disposing an object twice in the same method, either with the `using` keyword or by calling `Dispose` directly, is confusing and error-prone. For example, another developer might try to use an already-disposed object, or there can be runtime errors for specific paths in the code.
```
sebastien-marichal, line 3: The user might familiar with `using`, so I suggest adding a link to documentation: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using

sebastien-marichal (changes requested):
Content is good.

rules/S3966/description.adoc
mary-georgiou-sonarsource: I added link to the resources and not to the description because the links are language specific.

Also based on this (that the using statement links are language specific) I added a resources.adoc to each rule.

rules/S3966/csharp/rule.adoc
mary-georgiou-sonarsource: Change reverted

rules/S3966/vbnet/resources.adoc
```diff
@@ -0,0 +1,9 @@
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?redirectedfrom=MSDN#System_IDisposable_Dispose[Dispose method]
```
sebastien-marichal, line 5: Miss that from the first round 🙊 
```suggestion
* https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose#System_IDisposable_Dispose[Dispose method]
```

rules/S3966/why.adoc
```diff
@@ -0,0 +1,5 @@
+== Why is this an issue?
+
+Disposing an object twice in the same method, either with the `using` keyword or by calling `Dispose` directly, is confusing and error-prone. For example, another developer might try to use an already-disposed object, or there can be runtime errors for specific paths in the code.
+
+In addition, even if the https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?redirectedfrom=MSDN#System_IDisposable_Dispose[documentation] explicitly states that calling the `Dispose` method multiple times should not throw an exception, some implementations still do it. Thus it is safer to not dispose of an object twice when possible.
```
sebastien-marichal, line 5:
```suggestion
In addition, even if the https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose#System_IDisposable_Dispose[documentation] explicitly states that calling the `Dispose` method multiple times should not throw an exception, some implementations still do it. Thus it is safer to not dispose of an object twice when possible.
```

sebastien-marichal (approved):
LGTM!
I left some minor comments that do not require another review.

PR #2140, opened 2023-06-09
SONARJAVA-4488 Second pass fixing typos and using diff'd code samples

dorian-burihabwa-sonarsource:


leonardo-pilastri-sonarsource (approved):
Looks good!

PR #2141, opened 2023-06-09
Modify rule S4019: LaYC format

cristian-ambrosini-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4019/csharp/metadata.json
+++ b/rules/S4019/csharp/metadata.json
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
 }
--- a/rules/S4019/csharp/rule.adoc
+++ b/rules/S4019/csharp/rule.adoc
@@ -1,48 +1,61 @@
 == Why is this an issue?
 
-When a method in a derived class has the same name as a method in the base class but with a signature that only differs by types that are weakly derived (e.g. ``++object++`` vs ``++string++``), the result is that the base method becomes hidden.
+When a method in a derived class has the same name as a method in the base class but with a signature that only differs by types that are weakly derived (for example `object` vs `string`), the result is that the base method becomes hidden.
 
+As shown in the following code snippet, when an instance of the derived class is used, invoking the method with an argument that matches the weakly derived parameter type will invoke the derived class method instead of the base class method:
 
-=== Noncompliant code example
-
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
-using System;
-
-namespace MyLibrary
+class BaseClass
 {
-  class Foo
-  {
-    internal void SomeMethod(string s1, string s2) { }
-  }
-
-  class Bar : Foo
-  {
-    internal void SomeMethod(string s1, object o2) { }  // Noncompliant
-  }
+  internal void MyMethod(string str) => Console.WriteLine($"BaseClass: Method(string)");
 }
-----
 
-
-=== Compliant solution
-
-[source,csharp]
-----
-using System;
-
-namespace MyLibrary
+class DerivedClass : BaseClass
 {
-  class Foo
-  {
-    internal void SomeMethod(string s1, string s2) { }
-  }
-
-  class Bar : Foo
-  {
-    internal void SomeOtherMethod(string s1, object o2) { }
-  }
+  internal void MyMethod(object str) => Console.WriteLine($"DerivedClass: Method(object)"); // Noncompliant
 }
+
+// ...
+BaseClass baseObj = new BaseClass();
+baseObj.MyMethod("Hello"); // Output: BaseClass: Method(string)
+
+DerivedClass derivedObj = new DerivedClass();
+derivedObj.MyMethod("Hello"); // Output: DerivedClass: Method(object) - DerivedClass method is hiding the BaseClass method
+
+BaseClass derivedAsBase = new DerivedClass();
+derivedAsBase.MyMethod("Hello"); // Output: BaseClass: Method(string)
 ----
+
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+class BaseClass
+{
+  internal void MyMethod(string str) => Console.WriteLine($"BaseClass: Method(string)");
+}
+
+class DerivedClass : BaseClass
+{
+  internal void MyOtherMethod(object str) => Console.WriteLine($"DerivedClass: Method(object)"); // Compliant
+}
+
+// ...
+BaseClass baseObj = new BaseClass();
+baseObj.MyMethod("Hello"); // Output: BaseClass: Method(string)
+
+DerivedClass derivedObj = new DerivedClass();
+derivedObj.MyMethod("Hello"); // Output: BaseClass: Method(string)
+
+BaseClass derivedAsBase = new DerivedClass();
+derivedAsBase.MyMethod("Hello"); // Output: BaseClass: Method(string)
+----
+
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading[Member overloading]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier['new' modifier]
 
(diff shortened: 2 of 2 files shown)


rules/S4019/csharp/rule.adoc
```diff
@@ -1,49 +1,62 @@
 == Why is this an issue?
 
-When a method in a derived class has the same name as a method in the base class but with a signature that only differs by types that are weakly derived (e.g. ``++object++`` vs ``++string++``), the result is that the base method becomes hidden.
+When a method in a derived class has the same name as a method in the base class but with a signature that only differs by types that are weakly derived (for example `object` vs `string`), the result is that the base method becomes hidden.
```
antonioaversa, line 3: Can you investigate and explain this "weakly derivation" concept?

While I think I understand the example, I am foreign to the formal concept, and I don't know how far it goes. 

Moreover, if by "strong derivation" we intend the Liskov Substitution Principle, `string` derives from `object`, because every time a declared parameter, variable or field of type `object` is defined, you can assign a `string` to it, and the string behaves as an `object` (`isA` relationship).

So, I wouldn't know how I would go defining this `weak inheritance` relationship.

I see a definition [here](https://www.sciencedirect.com/topics/computer-science/inheritance-relationship#:~:text=A%20strong%20is%2Da%20relationship,faculty%20member%20is%20a%20person.), but I don't think it applies to the type of derivation described in this rule.

I actually suspect that by "weak derivation", the author intends `generalization` or `abstraction`, and that there is actually no weakness in this inheritance relationship, only the direction in which the relationship is considered.

rules/S4019/csharp/rule.adoc
```diff
@@ -1,49 +1,62 @@
...
 
-=== Noncompliant code example
-
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
-using System;
+class BaseClass
+{
+  internal void MyMethod(string str) => Console.WriteLine($"BaseClass: Method(string)");
```
antonioaversa, line 11: No need for `$`, since string interpolation is not used.
The same applies to the following lines too.

antonioaversa, line 11: I agree with removing usings, namespace and the first parameter: they don't add anything to the example and just complexify it, making it harder for the reader to grasp the main concept.

rules/S4019/csharp/rule.adoc
```diff
@@ -1,49 +1,62 @@
...
-  class Bar : Foo
-  {
-    internal void SomeMethod(string s1, object o2) { }  // Noncompliant
-  }
+  internal void MyMethod(object str) => Console.WriteLine($"DerivedClass: Method(object)"); // Noncompliant
 }
-----
 
+// ...
+BaseClass baseObj = new BaseClass();
```
antonioaversa, line 20: I also like this example: it shows the impact.

rules/S4019/csharp/rule.adoc
```diff
@@ -1,49 +1,62 @@
...
+
+BaseClass derivedAsBase = new DerivedClass();
+derivedAsBase.MyMethod("Hello"); // Output: BaseClass: Method(string)
 ----
 
+== Resources
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading[Member overloading]
```
antonioaversa, line 57: This one can also be useful: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/knowing-when-to-use-override-and-new-keywords

Method hiding doesn't exist in Java, where everything is `virtual` by default, so a junior Java developer who hasn't experience with C++ or C# would wonder when to use `virtual`, when to use `new`, when to go with interface explicit implementation, or when to just rename the method in the derived class.

rules/S4019/csharp/rule.adoc
```diff
@@ -1,49 +1,62 @@
...
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier['new' modifier]
```
antonioaversa, line 58: Good that you mention the `new` modifier here. I would also add a section explaining why the `new` doesn't apply here.

I was surprise when, pasting your code into LINQPad, I didn't get the CS0108 ("Use the new keyword if hiding was intended"). It would be good to explain when CS0108 and when (like in the code example above) it doesn't.

rules/S4019/csharp/rule.adoc
```diff
@@ -1,49 +1,62 @@
...
+
 ifdef::env-github,rspecator-view[]
```
antonioaversa, line 60: I would extract this section into `rspecator.adoc`, since it contains a conditional macro (`ifdef`), which just clutters the main content of the rule.

antonioaversa (changes requested):
Changes to and extensions of the examples are good!
Overriding vs hiding is a tricky subject which deserves more context and explanation.
I would also add a `What is the potential impact? (optional)` section exemplifying the bad things that can happen, if this issue is not fixed.

rules/S4019/csharp/metadata.json
```diff
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
```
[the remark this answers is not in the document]
cristian-ambrosini-sonarsource: What should the automatic renaming be? I thought about it and I wasn't sure: `MyMethod` should be renamed to.. `MyMethod_2`?

rules/S4019/csharp/rule.adoc
cristian-ambrosini-sonarsource: Thanks, I was interpolating with the `str` parameter but then I removed it and I forgot to remove the `$` too

rules/S4019/csharp/rule.adoc
cristian-ambrosini-sonarsource: If I understood correctly the compiler warning CS0108 raises when the two methods are identical copies (same name, same number of parameters, same parameter types, but not necessarily same parameter name). That's also why `new` doesn't work in the first place (it's not considered hidden if param types are derived and do not match exactly)

rules/S4019/csharp/metadata.json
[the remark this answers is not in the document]
antonioaversa: I don't know, it could be that or something else. I would just not rule out the quickfix, but I am in two minds about it, so I would let you go about `unknown vs infeasible` with whatever decision makes the most sense for you.
It's probably worth asking in the `team-lang-dotnet` channel on Slack.

rules/S4019/csharp/rule.adoc
antonioaversa: "Less derived" is better, but still confusing to me: less derived with respect to what? The derivation root: `object`?
Moreover, the example is about `object` itself, which is "0-derived" from `object`.

My proposal would be to use `ancestors` and do the following:
```
When a method in a derived class has:
* the same name as a method in the base class 
* but types of parameters that are ancestors (for example `string` in the base class and `object` in the derived class)
the result is that the base method becomes hidden.
```

rules/S4019/csharp/rule.adoc
antonioaversa: Yes. In addition to the cases you mentioned, there is also the scenario when the return type is different, which triggers CS0108, despite signatures being different.
```cs
class BaseClass
{
  internal string MyMethod(object str) => "";
}

class DerivedClass : BaseClass
{
  internal object MyMethod(object str) => ""; // CS0108
}
``` 

This is due to the support of [covariant return types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns), introduced in C# 9.0.

So there is an asymmetry here:
- if the "less derived/ancestor" type is on the output side, CS0108
- if the "less derived/ancestor" type is on the input side, CS0108 won't

In any case, the new paragraph is good and doesn't need to go into such tiny details.

antonioaversa (approved):
LGTM!
Before merging:
- re-check with the team https://github.com/SonarSource/rspec/pull/2141#discussion_r1227542534
- check suggestion in https://github.com/SonarSource/rspec/pull/2141#discussion_r1227548288

PR #2143, opened 2023-06-09
Modify rule S4635: Improve code example

sebastien-marichal:
## Review

- [ ] text style and tone


zsolt-kolbay-sonarsource (approved):
~~LGTM~~

rules/S4635/csharp/rule.adoc
```diff
@@ -22,7 +22,7 @@ To mitigate the creation of new objects while searching for a character within a
 
 [source,csharp,diff-id=1,diff-type=compliant]
 ----
-str.IndexOf(char1, startIndex) - startIndex;
+str.IndexOf(char1, startIndex) - startIndex;    // Compliant: no new instance of string created
 ----
 
 Using these methods gives the same result while avoiding the creation of additional `string` objects.
```
zsolt-kolbay-sonarsource, line 28: This will not have the same result in every scenario as the non-compliant version of the code.
If the character/string cannot be found in the string, then `IndexOf` will return -1. I've seen several code bases, where the return value of `IndexOf` was compared to -1 instead of checking whether it's smaller than 0. Consider the following code:
```
if (str.SubString(42).IndexOf('X') == -1)
{
   ...
}
```
Replacing this with 
```
if (str.IndexOf('X', 42) - 42 == -1)
{
   ...
}
```
will yield a different outcome if `'X'` is not found.
I think it's worth adding this as a note to the description.

zsolt-kolbay-sonarsource (changes requested):
I'd add a short description on how to handle negative values (when the character/string cannot be found in the string).

zsolt-kolbay-sonarsource (commented):
I will not be able to follow up on my change request this week, so I change the status and leave it up to you if you want to change it.

rules/S4635/csharp/rule.adoc
```diff
@@ -18,15 +18,38 @@ A new `string` object is created with every call to the https://learn.microsoft.
...
 
+=== Pitfalls
+
+As mentioned in the `String.IndexOf` documentation, when the value to seek within the string is not found, `String.IndexOf` will return the `-1`.
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: I don't understand the problem. The code can be rewritten as follows:
```cs
if (str.IndexOf(char1, startIndex) == -1)
```
Consider removing the pitfall section.

martin-strecker-sonarsource (changes requested):

martin-strecker-sonarsource:
I also think the title is a bit confusing as well:

❓ **String offset-based methods should be preferred for finding substrings from offsets**

What about something like:

❗  **The startString parameter should be preferred for seeking strings from offsets**

rules/S4635/csharp/rule.adoc
```diff
@@ -18,15 +18,38 @@ A new `string` object is created with every call to the https://learn.microsoft.
 
 Creating a new object consumes significant https://en.wikipedia.org/wiki/System_resource[system resources], especially CPU and memory. When using `Substring` repeatedly, such as in a loop, it can greatly impact the overall performance of the application or program. As the size of the string increases, the slower it becomes.
 
-To mitigate the creation of new objects while searching for a character within a substring, it is possible to utilize the overload of the mentioned methods that accepts an offset as the second parameter. Here is an example:
+To mitigate the creation of new objects while searching for characters within a substring, it is possible to use the overload of the mentioned methods that accept an offset as the second parameter. Here is an example:
```
martin-strecker-sonarsource, line 21: The parameter is always called `startIndex` so we could use the name to refer to it instead of the position
```suggestion
To mitigate the creation of new objects while searching for characters within a substring, it is possible to use the overload of the mentioned methods with a `startIndex` parameter. Here is an example:
```

rules/S4635/csharp/rule.adoc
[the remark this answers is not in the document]
sebastien-marichal: I think the section is still relevant with your version of the fix.
The rule's compliant solution suggests subtracting `startIndex` from the result of `IndexOf`.
```cs
str.IndexOf(char1, startIndex) - startIndex
```

I think, if not careful enough, devs might apply the fix and end up with code like this:

```cs
if (str.IndexOf(char1, startIndex) - startIndex == -1)
{
   // ...
}
```

Which will change the behavior of the application. It will, in most cases if not all, evaluate to `false`.

sebastien-marichal:
> I also think the title is a bit confusing as well:
> 
> ❓ **String offset-based methods should be preferred for finding substrings from offsets**
> 
> What about something like:
> 
> ❗ **The startString parameter should be preferred for seeking strings from offsets**

I agree the title could be better. I must admit, your suggestion is not totally clear to me. Maybe we can find something better.

martin-strecker-sonarsource:
> your suggestion is not totally clear to me

I had a typo in it. It should read  "The 'startIndex' parameter should be preferred for seeking strings from offsets."

Which part of it?

The rule is highlighting the `IndexOf()` invocation, so it should be clear what "startIndex parameter" refers to. Maybe we can change it to "startIndex parameter of the IndexOf method" but this is a bit imprecise because there are IndexOf, IndexOfAny, LastIndexOf, and LastIndexOfAny.

rules/S4635/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-    "quickfix": "targeted"
-}
+    "quickfix": "infeasible"
```
martin-strecker-sonarsource, line 2: It is possible to provide a quick fix for this case.

rules/S4635/csharp/metadata.json
```diff
@@ -1,3 +1,4 @@
 {
+    "title": "Seeking characters from a substring should be avoided",
```
sebastien-marichal, line 2: @martin-strecker-sonarsource This is my try to improve the title. From my understanding, this rule is looking of useless object creation when seeking character. I think this fits well.

I dropped the `startIndex` info since I don't know where to put it and I am not sure if really relevant.

martin-strecker-sonarsource (changes requested):
Title is fine. We still need to do something about the example/pitfall.

rules/S4635/csharp/rule.adoc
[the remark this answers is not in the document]
sebastien-marichal: I have removed the `Pitfalls` section and updated the examples to be more realistic.
I believe now, it should be more clear.

rules/S4635/csharp/rule.adoc
```diff
@@ -16,13 +19,16 @@ A new `string` object is created with every call to the https://learn.microsoft.
 
 === What is the potential impact?
 
-Creating a new object consumes significant https://en.wikipedia.org/wiki/System_resource[system resources], especially CPU and memory. When using `Substring` repeatedly, such as in a loop, it can greatly impact the overall performance of the application or program. As the size of the string increases, the slower it becomes.
+Creating a new object consumes significant https://en.wikipedia.org/wiki/System_resource[system resources], especially CPU and memory. When using `Substring` repeatedly, such as in a loop, it can greatly impact the overall performance of the application or program.
 
-To mitigate the creation of new objects while searching for a character within a substring, it is possible to utilize the overload of the mentioned methods that accepts an offset as the second parameter. Here is an example:
+To mitigate the creation of new objects while searching for characters within a substring, it is possible to use the overload of the mentioned methods with a `startIndex` parameter. Here is an example:
```
martin-strecker-sonarsource, line 24: There are many overloads with a `startIndex` parameter per method group. "Here is an example:" doesn't fit any more as you show the compliant solution.
```suggestion
To mitigate the creation of new objects while searching for characters within a substring, it is possible to use an overload of the mentioned methods with a `startIndex` parameter:
```

rules/S4635/csharp/rule.adoc
```diff
@@ -16,13 +19,16 @@ A new `string` object is created with every call to the https://learn.microsoft.
...
 
 [source,csharp,diff-id=1,diff-type=compliant]
 ----
-str.IndexOf(char1, startIndex) - startIndex;
+if (str.IndexOf(char1, startIndex) == -1)           // Compliant: no new instance of string created
```
martin-strecker-sonarsource, line 28:
```suggestion
if (str.IndexOf(char1, startIndex) == -1)           // Compliant: no new instance of string is created
```

martin-strecker-sonarsource (approved):
LGTM. Some minor grammar issues are still to be resolved.

PR #2150, opened 2023-06-12
Modify rule S3265: LaYC format

sebastien-marichal:
https://sonarsource.github.io/rspec/#/rspec/S3265

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3265/csharp/rule.adoc
+++ b/rules/S3265/csharp/rule.adoc
@@ -4,6 +4,27 @@ https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types
 
 However, they can also serve as https://en.wikipedia.org/wiki/Bit_field[bit flags], enabling bitwise operations to combine multiple elements within a single value.
 
+[source,csharp]
+----
+enum Days
+{
+  None      = 0b00000000,                   // 0
+  Monday    = 0b00000001,                   // 1
+  Tuesday   = 0b00000010,                   // 2
+  Wednesday = 0b00000100,                   // 4
+  Thursday  = 0b00001000,                   // 8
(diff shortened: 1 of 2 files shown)


rules/S3265/csharp/rule.adoc
```diff
@@ -1,122 +1,52 @@
...
-  Write = 2, 
-  Execute = 4 
+  Read = 1,
+  Write = 2,
+  Execute = 4
 }
 // ...
 
-var x = Permissions.Read | Permissions.Write;  // Noncompliant; enum is not marked with [Flags]
+var x = Permissions.Read | Permissions.Write;  // Noncompliant: enum is not annotated with [Flags]
```
antonioaversa, line 18: The non-compliant comment comes before the explanation of non-compliance, which is done in the following paragraph.
I understand you may want to give an example of the sentence in the first paragraph ("they can also serve as... to combine..."), but I find confusing of having an example of non-compliance before the non-compliance is explained.

What I would probably do is:
- have the non-compliant code example after the next paragraph and before the compliant example
- have another example that shows the idea, that doesn't necessarily require to define the enum. Something like:

```
var orange = Colors.Red | Colors.Yellow; 
// Red = 0b00000001, Yellow = 0b00000010, orange = 0b00000011 
```
or
```
var weekend = Days.Saturday | Color.Sunday; 
// Saturday = 0b1000000, Sunday = 0b0100000, weekend = 0b1100000
```
Without definition of `Colors` or `Days`.

antonioaversa (approved):
LGTM!
Doesn't need to go through me again, but the comment about non-compliant code example being placed before non-compliance explanation should in my opinion be addressed.

PR #2151, opened 2023-06-12
Modify rule S3445: LaYC format

costin-zaharia-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3445/csharp/rule.adoc
+++ b/rules/S3445/csharp/rule.adoc
@@ -1,15 +1,13 @@
 == Why is this an issue?
 
-In C#, the `throw` statement can be used in two different ways:
+In C#, the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#13106-the-throw-statement[throw] statement can be used in two different ways:
 
 * by specifying an expression
 * without specifying an expression
 
 === By specifying an expression
 
-In the software development context, an expression is a value or anything that executes and ends up being a value. The expression shall be implicitly convertible to `System.Exception`, and the result of evaluating the expression is converted to `System.Exception` before being thrown. If the result of the conversion is `null`, a `System.NullReferenceException` is thrown instead.
-
-In this case, the stack trace will be cleared, losing the list of method calls between the original method that threw the exception and the current method.
+In the software development context, an expression is a value or anything that executes and ends up being a value. The expression shall be implicitly convertible to `System.Exception`, and the result of evaluating the expression is converted to `System.Exception` before being thrown.
 
 [source,csharp]
 ----
@@ -18,9 +16,12 @@ try
 }
 catch(Exception exception)
 {
+  // code that uses the exception
   throw exception; // The exception stack trace is cleared up to this point.
 }
 ----
+
+In this case, the https://en.wikipedia.org/wiki/Stack_trace[stack trace], will be cleared, losing the list of method calls between the original method that threw the exception and the current method.
 
 === Without specifying an expression
 
@@ -33,6 +34,7 @@ try
 }
 catch(Exception exception)
 {
+  // code that uses the exception
   throw; // The stack trace of the initial exception is preserved.
 }
 ----
--- a/rules/S3445/csharp/rule.adoc
+++ b/rules/S3445/csharp/rule.adoc
@@ -1,71 +1,94 @@
 == Why is this an issue?
 
-When rethrowing an exception, you should do it by simply calling ``++throw;++`` and not ``++throw exc;++``, because the stack trace is reset with the second syntax, making debugging a lot harder.
+In C#, the `throw` statement can be used in two different ways:
 
+* by specifying an expression
+* without specifying an expression
 
-=== Noncompliant code example
+=== By specifying an expression
 
-[source,csharp]
-----
-try
-{}
-catch(ExceptionType1 exc)
-{
-  Console.WriteLine(exc);
-  throw exc; // Noncompliant; stacktrace is reset
-}
-catch (ExceptionType2 exc) 
-{
-  throw new Exception("My custom message", exc);  // Compliant; stack trace preserved
-}
-----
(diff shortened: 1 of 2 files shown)


rules/S3445/csharp/rule.adoc
```diff
@@ -1,71 +1,94 @@
...
-When rethrowing an exception, you should do it by simply calling ``++throw;++`` and not ``++throw exc;++``, because the stack trace is reset with the second syntax, making debugging a lot harder.
+In C#, the `throw` statement can be used in two different ways:
 
+* by specifying an expression
+* without specifying an expression
 
-=== Noncompliant code example
+=== By specifying an expression
+
+In the software development context, an expression is a value or anything that executes and ends up being a value. The expression shall be implicitly convertible to `System.Exception`, and the result of evaluating the expression is converted to `System.Exception` before being thrown. If the result of the conversion is `null`, a `System.NullReferenceException` is thrown instead.
```
sebastien-marichal, line 10: I would expect more details about the rethrown mechanism and how the stack is preserved in those cases.

I think the explanations are going too far.

rules/S3445/csharp/rule.adoc
```diff
@@ -1,71 +1,94 @@
...
-== Implementation Specification
-(visible only on this page)
+[source,csharp]
+----
+try
+{
+}
+catch(Exception exception)
+{
+  throw new Exception("Additional information", exception);
```
sebastien-marichal, line 51: This code example is throwing a brand-new `exception`, I would not consider it an exception per se.
However, it is a good and interesting addition.

rules/S3445/csharp/rule.adoc
```diff
@@ -1,71 +1,94 @@
...
+}
+----
 
-=== Message
+== How to fix it
```
sebastien-marichal, line 55: The code example from the `why` and `How to fix it` sections are the same.
I think the `How to fix it` section could be merged into the `why`

sebastien-marichal (changes requested):
Left some suggestions, to summarize:
- focus more on the rethrow mechanism in the description
- reconsider the exception example
- merge `how to` section in the `why` section

Open to discussing it if needed.

rules/S3445/csharp/rule.adoc
costin-zaharia-sonarsource: > I would expect more details about the rethrown mechanism and how the stack is preserved in those cases.

Do you mean updating the next section (Without specifying an expression)? I've added "preserving the stack trace." to make this obvious.

> I think the explanations are going too far.

I've tried to explain all the concepts involved:
- what is an expression
- how it's evaluated and how the throw works when the expression is specified
- how the stack trace is cleared and what that means "losing the list of method calls between.."

What do you think it's superflu? the null handling?

rules/S3445/csharp/rule.adoc
costin-zaharia-sonarsource: My understanding is that it's perfectly fine to have snippets from the `Why` section replicated in `How to fix it` and that we should not expect the users to have the context from the other tab.

I think that for this rule, having two separate sections helps. The first one explains what the problem is and what's the result of using both approaches, while the second section has a short and clear explanation, straight to the point, while providing code examples with the compliant and noncompliant snippets.

I think that mixing the description of how the `throw` works and the rule expectations will cause more confusion and make the topic harder to understand.

sebastien-marichal: It is not always very clear to me, to be honest.
You convince me, this rule might be too complex to skip the `How to fix it` section.

sebastien-marichal (changes requested):
I added more details to my previous comment.

rules/S3445/csharp/rule.adoc
```diff
@@ -1,71 +1,96 @@
...
-catch (ExceptionType2 exc) 
+catch(Exception exception)
 {
-  throw new Exception("My custom message", exc);  // Compliant; stack trace preserved
+  // code that uses the exception
+  throw exception; // The exception stack trace is cleared up to this point.
 }
 ----
 
+In this case, the https://en.wikipedia.org/wiki/Stack_trace[stack trace], will be cleared, losing the list of method calls between the original method that threw the exception and the current method.
```
costin-zaharia-sonarsource, line 24: I've linked the stack trace but I would not go into more details about what a stack trace because the resto of the sentence indirectly explains that: "losing the list of method calls between the original method that threw the exception and the current method"

rules/S3445/csharp/rule.adoc
```diff
@@ -1,71 +1,96 @@
...
-\[~ann.campbell.2] LGTM
+== Resources
 
-=== on 26 Jul 2016, 17:50:15 Tamas Vajk wrote:
-\[~ann.campbell.2] I don't think this is a bug.
+=== Documentation
 
-=== on 27 Jul 2016, 13:30:00 Freddy Mallet wrote:
-Agreed [~tamas.vajk] and this is the limit of the usage of this "Bug" issue type because this is really a "Reliability" issue that might prevent someone in production to understand an unexpected behavior as the root cause of the issue is lost. That's why we're classifying this issue as a bug. cc [~ann.campbell.2]
+* https://learn.microsoft.com/en-us/dotnet/api/system.exception?view=net-7.0#re-throwing-an-exception[Re-throwing an exception]
```
sebastien-marichal, line 93:
```suggestion
* https://learn.microsoft.com/en-us/dotnet/api/system.exception#re-throwing-an-exception[Re-throwing an exception]
```

rules/S3445/csharp/rule.adoc
```diff
@@ -1,71 +1,96 @@
...
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#13106-the-throw-statement[The throw statement]
```
sebastien-marichal, line 94: You can also add here the link to the stack trace wiki page

sebastien-marichal (approved):
LGTM!
Left some comments that do not need another review round.

PR #2154, opened 2023-06-12
Modify Rule S3011: LaYC reflection

frederic-tingaud-sonarsource:
@sebastian-hungerecker-sonarsource: Could you have a look at the Ruby example? It is long and I think it could be simplified.

@nils-werner-sonarsource or @hendrik-buchwald-sonarsource : I don't know what happened with the PHP version of the rule. Its example seems like somebody accidentally copied a whole project in RSPEC. Do you think it's easily salvageable or should I create an independent JIRA ticket for you to rewrite it in the future?


sebastian-hungerecker-sonarsource:
Since this rule isn't implemented for Ruby and probably is never going to be, maybe we can just remove the Ruby RSPEC for it?

Anyway, I think the main reasons of why the example is this verbose are:

* It illustrates two different ways to circumvent access control for methods (while excluding one much more common way and not covering access control for variables at all - which raises the question what the scope of this rule is supposed to be if it does ever get implemented)
* It tries to explain what effects these methods have rather than just showing them.
* It doesn't just show the code to make the variable public, but also the initial class definition that made the method private. This is not done for the Java example for example.

If we separate the two examples and remove the explanatory bits and the class definition, we're left with these:


```ruby
# Reopening the eigenclass to make the method public for one object only
class << some_object
  public :some_private_method
end
```

and

```ruby
A.class_eval do
  public :some_private_method
end
```

But as I said, these aren't actually the most common ways I'd expect this rule to be violated, so the following examples may actually make more sense:

```ruby
# For methods
some_object.send(:some_private_method)

# For variables
some_object.instance_variable_get(:@some_private_variable)
```

But of course that assumes that those methods are actually intended to be supported by this rule, which is unclear without an implementation or a detailed (Ruby-specific) rule description.

So I'd rather go with just dropping the RSPEC.

hendrik-buchwald-sonarsource:
I leave the decision for PHP to @SonarSource/languages-team-cloud since it is a code quality rule. From my side, I just would slim it down a little and focus on one or two cases and not all of them.

nils-werner-sonarsource:
> I leave the decision for PHP to @SonarSource/languages-team-cloud since it is a code quality rule. From my side, I just would slim it down a little and focus on one or two cases and not all of them.

Yes indeed, someone seemed to be motivated and all UT test cases were copied to the RSPEC. We will clean it up by the end of tomorrow.

PR #2155, opened 2023-06-12
Modify rule S4586: LaYC format

martin-strecker-sonarsource:
https://sonarsource.github.io/rspec/#/rspec/S4586

## Review

- [ ] text style and tone


rules/S4586/csharp/rule.adoc
```diff
@@ -1,43 +1,50 @@
 == Why is this an issue?
```
csaba-sagi-sonarsource, line 1: The Why is this an issue section is the same for C# and VB.net. Please extract it into a file called why-dotnet.adoc and link it, to remove duplication.

csaba-sagi-sonarsource (changes requested):
I left two small comments. Also, it might have not been shared with you that we took a decision to take the clean-as-you code approach and update the quickfix field for each rule we migrate to lake, so please evaluate if a quickfix is possible or not and update the field accordingly.

rules/S4586/csharp/rule.adoc
csaba-sagi-sonarsource: I really don't have a strong opinion. Some people are using it and Pavel also suggested to use it in this exact case.

csaba-sagi-sonarsource (approved):
I am approving the PR, The unresolved conversation is optional in my opinion.
If you wish apply it, otherwise feel free to merge as is.

martin-strecker-sonarsource:
Re-requested a review because of https://github.com/SonarSource/rspec/pull/2155#discussion_r1230766502

rules/S4586/csharp/rule.adoc
csaba-sagi-sonarsource: If we ever plan to introduce a dictionary for the keywords, I think this state is a better one to move forwards, otherwise, I am not sure we will revisit the rule ever to add the variables/attributes.
I don't have a strong opinion so I will let you decide and merge the rule in whichever state you wish.

csaba-sagi-sonarsource (approved):
LGTM!

PR #2156, opened 2023-06-13
Modify rule S2198: LaYC format

antonioaversa:
## Review

- [ ] text style and tone


rules/S2198/csharp/rule.adoc
```diff
@@ -1,35 +1,22 @@
...
+Specifically, the following comparisons will return either always `true` or always `false` depending on the kind of comparison:
 
-* Comparing a `char` with a numeric constant that is outside of the range of `char`.
-* Comparing a `float` with a numeric constant that is outside of the range of `float`.
-* Comparing a `long` with a numeric constant that is outside of the range of `long`.
-* Comparing a `ulong` with a numeric constant that is outside of the range of `ulong`.
+* comparing a `char` with a numeric constant that is outside of the range of `char`
+* comparing a `float` with a numeric constant that is outside of the range of `float`
+* comparing a `long` with a numeric constant that is outside of the range of `long`
+* comparing a `ulong` with a numeric constant that is outside of the range of `ulong`
```
csaba-sagi-sonarsource, line 10: Do you think it would be worth it linking the msdn documentation to the types? Like
https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/data-types/ulong-data-type for ulong?
This would enable users to find the ranges more quickly.

rules/S2198/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+    "quickfix": "targeted"
```
csaba-sagi-sonarsource, line 2: What would be the quickfix in your opinion?
Is it just removing the comparison?
I am afraid that might change the behaviour of the code. Imagine for example:
```
If (f > double.Maxvalue)
{
Console.WriteLine("I am never printed");
}
```
If you remove the check, the code will actually execute. Would you in this case remove also the  body of the if as it was unreachable code?

csaba-sagi-sonarsource (approved):
Two nitpicky comments. Plus in addition, I would like to ask you to remove the `++ stuff in the base rule.adoc.
I am approving as all of these are optional.

rules/S2198/csharp/rule.adoc
antonioaversa: I have added two links:
- to the paragraph on the page about integral numeric types, that shows ranges for all numeric types
- to the page for the `char` data type, that shows the range for the type

rules/S2198/csharp/metadata.json
antonioaversa: The fix would either remove the entire `if` block, or just the `if` condition.
For example, if the code is `if (f > double.MaxValue) Body` the entire `if` would be removed.
If the code is `if (f <= double.MaxValue) Body` the `if` would be replaced with `Body`.
The same would apply to conditions like `f < double.MinValue` and `f >= double.MinValue`.

Indeed, there is actually the risk of changing the behavior of the code, for example in the following example:
```cs
if (SomeMethod() > int.MaxValue) Console.WriteLine(""Never being executed");

int SomeMethod() => throw new Exception();
```

If we would apply the code fix here, the entire `if` block would be removed, and the code would not raise an exception anymore. However, we can limit the code fix to fields only, if we want to avoid this corner case. 

Anyway, even if there are subtleties to be taken care of, I would keep the `targeted`, since there is a realm where in principle the code fix is possible. So I would keep the door open for now, and defer the deeper analysis to a possible future task to specify the code fix behavior.

antonioaversa:
> Two nitpicky comments. Plus in addition, I would like to ask you to remove the `++ stuff in the base rule.adoc. I am approving as all of these are optional.

I have removed the `rule.adoc` at top-level completely since all other languages have been removed as not implemented.

csaba-sagi-sonarsource (approved):
LGTM!

PR #2160, opened 2023-06-13
Modify rule S121: LaYC format

amelie-renard-sonarsource:


rules/S121/tsql/rule.adoc
```diff
@@ -1,25 +1,30 @@
+Control structures are code statements that impact the program's control flow (e.g., if statements, for loops, etc.)
+
 == Why is this an issue?
 
-While not technically incorrect, the omission of ``++BEGIN...END++`` can be misleading, and may lead to the introduction of errors during maintenance.
+While not technically incorrect, the omission of ``++BEGIN...END++`` can be misleading and may lead to the introduction of errors during maintenance.
 
-=== Noncompliant code example
+In the following example, the two statements seem to be attached to the `if` statement, but only the first one is, and `somethingElse` will always be executed:
```
frederic-tingaud-sonarsource, line 7:
```suggestion
In the following example, the two statements seem to be attached to the `IF` statement, but only the first one is, and `somethingElse` will always be executed:
```

rules/S121/rule.adoc
```diff
@@ -1,8 +1,4 @@
-== Why is this an issue?
+:compliant: compliant.adoc
```
frederic-tingaud-sonarsource, line 1: One thing I try to do, is have the common files included through variable in a `common` folder, or with a different name than the specialized ones.
The reason is that the variables include is a bit of a pitfall, because it includes from a different folder than the one where you create the variable. If you are in the java folder and you write `:compliant: compliant.adoc`, you think this will include `java/compliant.adoc`, but instead it includes `../compliant.adoc`

frederic-tingaud-sonarsource (changes requested):
A few details remaining.

rules/S121/pli/rule.adoc
```diff
@@ -1,34 +1,35 @@
 == Why is this an issue?
 
-Not using a ``++DO ... END++`` compound statement can be error-prone.
+While not technically incorrect, the omission of ``++DO ... END++`` can be misleading and may lead to the introduction of errors during maintenance.
```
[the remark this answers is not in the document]
amelie-renard-sonarsource: Mmm it's unclear what we should do. In the documentation, I mainly see upper cases. I'll ask an expert :)

rules/S121/rule.adoc
amelie-renard-sonarsource: Good point, I'll do that

frederic-tingaud-sonarsource (approved):

PR #2168, opened 2023-06-13
Modify rule S4275: LaYC format

martin-strecker-sonarsource:
[S4275: Getters and setters should access the expected fields](https://sonarsource.github.io/rspec/#/rspec/S4275)

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4275/csharp/rule.adoc
+++ b/rules/S4275/csharp/rule.adoc
@@ -1,6 +1,16 @@
 == Why is this an issue?
 
-Properties provide a way to enforce encapsulation by providing accessors that give controlled access to `private` fields. However, in classes with multiple fields, it is not unusual that cut and paste is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
+Properties provide a way to enforce https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)[encapsulation] by providing accessors that give controlled access to `private` fields. However, in classes with multiple fields, it is not unusual that https://en.wikipedia.org/wiki/Copy-and-paste_programming[copy-and-paste] is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
+
+[source,csharp]
+----
+class C
+{
+    private int x;
+    private int y;
+    public int Y => x; // Noncompliant: The returned field should be 'y'
+}
+----
 
 This rule raises an issue in any of these cases:
 
--- a/rules/S4275/csharp/rule.adoc
+++ b/rules/S4275/csharp/rule.adoc
@@ -4,8 +4,8 @@ Properties provide a way to enforce encapsulation by providing accessors that gi
 
 This rule raises an issue in any of these cases:
 
-* A setter does not update the field with the corresponding name.
 * A getter does not access the field with the corresponding name.
+* A setter does not update the field with the corresponding name.
 
 For simple properties, it is better to use https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties[auto-implemented properties] (C# 3.0 or later).
 
@@ -13,7 +13,9 @@ Field and property names are compared as case-insensitive. All underscore charac
 
 == How to fix it
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 [source,csharp,diff-id=1,diff-type=noncompliant]
 ----
@@ -36,7 +38,7 @@ class A
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
 [source,csharp,diff-id=1,diff-type=compliant]
 ----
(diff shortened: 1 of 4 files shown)


rules/S4275/csharp/rule.adoc
```diff
@@ -1,21 +1,23 @@
 == Why is this an issue?
 
-Properties provide a way to enforce encapsulation by providing ``++public++``, ``++protected++`` or ``++internal++`` methods that give controlled access to ``++private++`` fields. However in classes with multiple fields it is not unusual that cut and paste is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
-
+Properties provide a way to enforce encapsulation by providing accessors that give controlled access to `private` fields. However, in classes with multiple fields, it is not unusual that cut and paste is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
```
gregory-paidis-sonarsource, line 3: You can link [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)), some junior devs might not know it.

gregory-paidis-sonarsource, line 3: why does it say `cut` and paste instead of `copy` and paste? 🤔 
Technically, `cut and paste` will not result in this issue.

rules/S4275/vbnet/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+    "title": "Property procedures should access the expected fields"
```
gregory-paidis-sonarsource, line 2: you need to change the `quickfix` field as well, both for c# and vb.net in `metadata.json`.

(I think it should be `infeasible` for this rule)

rules/S4275/vbnet/rule.adoc
```diff
@@ -1,21 +1,23 @@
 == Why is this an issue?
 
-Properties provide a way to enforce encapsulation by providing ``++public++``, ``++protected++`` or ``++internal++`` methods that give controlled access to ``++private++`` fields. However in classes with multiple fields it is not unusual that cut and paste is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
-
+Properties provide a way to enforce encapsulation by providing property procedures that give controlled access to `Private` fields. However, in classes with multiple fields, it is not unusual that cut and paste is used to quickly create the needed properties, which can result in the wrong field being accessed by the property procedures.
```
gregory-paidis-sonarsource, line 3: Same as C#:

- Link encapsulation
- cut -> copy

gregory-paidis-sonarsource (changes requested):
Looks good, left some minor suggestions

rules/S4275/csharp/rule.adoc
antonioaversa: A small code example would also help. 
Especially because all code examples for this rule are now in the How to fix tab.

martin-strecker-sonarsource: I added an example, but I'm not certain it works: https://github.com/SonarSource/sonar-dotnet/issues/7420

rules/S4275/vbnet/metadata.json
martin-strecker-sonarsource: It is possible. The rule makes sure that there is a fitting field, so we could replace the tokens.

rules/S4275/csharp/rule.adoc
```diff
@@ -1,21 +1,33 @@
...
-Properties provide a way to enforce encapsulation by providing ``++public++``, ``++protected++`` or ``++internal++`` methods that give controlled access to ``++private++`` fields. However in classes with multiple fields it is not unusual that cut and paste is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
+Properties provide a way to enforce https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)[encapsulation] by providing accessors that give controlled access to `private` fields. However, in classes with multiple fields, it is not unusual that https://en.wikipedia.org/wiki/Copy-and-paste_programming[copy-and-paste] is used to quickly create the needed properties, which can result in the wrong field being accessed by a getter or setter.
 
+[source,csharp]
+----
+class C
+{
+    private int x;
+    private int y;
+    public int Y => x; // Noncompliant: The returned field should be 'y'
```
martin-strecker-sonarsource, line 11: May not work:
https://github.com/SonarSource/sonar-dotnet/issues/7420
The implementation works on the symbol-level in most parts, so it should be fine, but it is not covered by tests.

martin-strecker-sonarsource:
@gregory-paidis-sonarsource There are directories, where the rule.adoc isn't adapted for
* cfammily
* java
* php
* swift

Are we supposed to delete these directories? The rule.adoc looks like so
```adoc
include::../rule.adoc[]

'''
(visible only on this page)

include::../message.adoc[]

include::../highlighting.adoc[]

```

If we delete it, where should it be documented?

gregory-paidis-sonarsource:
> Are we supposed to delete these directories? 

@martin-strecker-sonarsource  I think you are fine ignoring them. Take a look if they are included in [this gigantic PR](https://github.com/SonarSource/rspec/pull/2145) for the languages you wanted. If they are, don't do anything about them :)

gregory-paidis-sonarsource (approved):
LGTM, take a look at the comment for the removal of unused languages. If nothing is needed, merge at will.

PR #2169, opened 2023-06-13
Modify rule S2291: LAYC format

gregory-paidis-sonarsource:


rules/S2291/csharp/metadata.json
```diff
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
```
martin-strecker-sonarsource, line 27: targeted: The fix is to remove the `unchecked` block.

rules/S2291/csharp/rule.adoc
```diff
@@ -1,50 +1,36 @@
 == Why is this an issue?
 
-``++Enumerable.Sum()++`` always executes addition in a ``++checked++`` context, so an ``++OverflowException++`` will be thrown if the value exceeds ``++MaxValue++`` even if an ``++unchecked++`` context was specified. Using an ``++unchecked++`` context anyway represents a misunderstanding of how ``++Sum++`` works.
+The checked and unchecked https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked[statements] and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12819-the-checked-and-unchecked-operators[operators] specify the https://en.wikipedia.org/wiki/Integer_overflow[overflow]-checking context for integral-type arithmetic operations and conversions.
```
martin-strecker-sonarsource, line 3: This is a nice explanation but I would move the

> Enumerable.Sum() always executes addition in a `checked` context

paragraph to the beginning and move this part into a subsection like `=== Background` or the `How does it work` section in `How to fix it`.

rules/S2291/csharp/rule.adoc
```diff
@@ -1,50 +1,36 @@
...
-{
-  int d = unchecked(list.Sum());  // Noncompliant
-
-  unchecked 
-  {
-    int e = list.Sum();  // Noncompliant
-  }
-}
+uint a = uint.MaxValue;
+unchecked(a = a + 1); // a+1 overflows, results in 0
```
martin-strecker-sonarsource, line 12:
```suggestion
unchecked(a = a + 1); // a + 1 overflows, results in 0
```

rules/S2291/csharp/rule.adoc
```diff
@@ -1,50 +1,36 @@
...
 ----
 
-
-=== Compliant solution
-
+* When `checked` is specified, an https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[OverflowException] is thrown for a run-time overflow, and https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0220?f1url=%3FappId%3Droslyn%26k%3Dk(CS0220)[CS0220] compile-time error occurs for a constant overflow.
```
martin-strecker-sonarsource, line 15: I don't think, we need to talk about compile time errors in our rules. In the case of this rule, there is no compile time error possible.
```suggestion
* When `checked` is specified, an https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[OverflowException] is thrown instead.
```

rules/S2291/csharp/rule.adoc
```diff
@@ -1,50 +1,36 @@
...
-  {
-    int e = list.Sum();
-  } 
-  catch (System.OverflowException e) 
-  {
-    // exception handling...
-  }
-}
+const uint b = uint.MaxValue;
+checked(b + 1) // compile-time: Compiler Error CS0220
```
martin-strecker-sonarsource, line 23: The examples and explanations are almost identical to what is written in https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked Do we really need this section? This is more like an explanation of a language feature than an education about a code smell / bug. I think we usually assume the user knows the language and link to "Learn Microsoft" as you already did. My personal opinion is, that the link is enough already. What do you think?

rules/S2291/csharp/rule.adoc
```diff
@@ -1,50 +1,36 @@
...
 ----
 
 
+https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum[Enumerable.Sum()] always executes addition in a `checked` context, so an OverflowException will be thrown if the value exceeds `MaxValue`, even if an `unchecked` context was specified. Therefore, using this method inside an `unchecked` context will only make the code more confusing, since the behavior will still be `checked`.
```
martin-strecker-sonarsource, line 27: It would be good to back that claim with a source (unfortunately we can not use footnotes 😢 ). You can either cite the exception section:
> [OverflowException](https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception?view=net-7.0)
The sum is larger than [Int32.MaxValue](https://learn.microsoft.com/en-us/dotnet/api/system.int32.maxvalue?view=net-7.0).

or the source code

https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Core/System/Linq/Enumerable.cs#L1408-L1415

This link should go into the Resources section.

rules/S2291/csharp/rule.adoc
```diff
@@ -54,46 +40,69 @@ void Add(List<int> list)
...
+=== Documentation
 
-=== on 6 Jul 2015, 09:33:41 Tamas Vajk wrote:
-\[~ann.campbell.2] Looks good.
+* https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum[Enumerable.Sum Method]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked[checked and unchecked statements]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12819-the-checked-and-unchecked-operators[checked and unchecked operators]
+* https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/[Exceptions and Exception Handling]
+* https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[OverflowException Class]
+* https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0220?f1url=%3FappId%3Droslyn%26k%3Dk(CS0220)[Compiler Error CS0220]
```
martin-strecker-sonarsource, line 105: Not needed.
```suggestion
```

rules/S2291/csharp/rule.adoc
```diff
@@ -54,46 +40,69 @@ void Add(List<int> list)
...
-Double-check my edits please [~tamas.vajk]
+=== Documentation
 
-=== on 6 Jul 2015, 09:33:41 Tamas Vajk wrote:
-\[~ann.campbell.2] Looks good.
+* https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum[Enumerable.Sum Method]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked[checked and unchecked statements]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12819-the-checked-and-unchecked-operators[checked and unchecked operators]
+* https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/[Exceptions and Exception Handling]
+* https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[OverflowException Class]
```
martin-strecker-sonarsource, line 104:
```suggestion
* https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum[`Enumerable.Sum` Method]
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked[`checked` and `unchecked` statements]
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12819-the-checked-and-unchecked-operators[`checked` and `unchecked` operators]
* https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/[Exceptions and Exception Handling]
* https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[`OverflowException` Class]
```

martin-strecker-sonarsource (changes requested):
I'm a bit concerned about the section about the "checked" and "unchecked" language feaure.

rules/S2291/csharp/rule.adoc
gregory-paidis-sonarsource: Yes, what you say makes sense. I still struggle to find the balance of how explanatory I should be. Dropped the entire section.

rules/S2291/csharp/rule.adoc
```diff
@@ -1,99 +1,83 @@
 == Why is this an issue?
 
-``++Enumerable.Sum()++`` always executes addition in a ``++checked++`` context, so an ``++OverflowException++`` will be thrown if the value exceeds ``++MaxValue++`` even if an ``++unchecked++`` context was specified. Using an ``++unchecked++`` context anyway represents a misunderstanding of how ``++Sum++`` works.
+https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum[Enumerable.Sum()] always executes addition in a `checked` context, so an https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[OverflowException] will be thrown if the value exceeds `MaxValue`, even if an `unchecked` context was specified. Therefore, using this method inside an `unchecked` context will only make the code more confusing, since the behavior will still be `checked`.
```
martin-strecker-sonarsource, line 3:
```suggestion
https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum[`Enumerable.Sum()`] always executes addition in a `checked` context, so an https://learn.microsoft.com/en-us/dotnet/api/system.overflowexception[OverflowException] will be thrown if the value exceeds `MaxValue`, even if an `unchecked` context was specified. Therefore, using this method inside an `unchecked` context will only make the code more confusing, since the behavior will still be `checked`.
```

martin-strecker-sonarsource (approved):
LGTM. One minor suggestion, though.

rules/S2291/csharp/rule.adoc
martin-strecker-sonarsource: I don't know: https://sonarsource.slack.com/archives/C04P0SB2KJP/p1686642571396659 There is a meeting today with the documentation team where I will bring this up as well.

PR #2173, opened 2023-06-13
Modify rule S2225: LaYC format

gregory-paidis-sonarsource:


rules/S2225/csharp/rule.adoc
```diff
@@ -19,9 +17,7 @@ public override string ToString ()
 }
 ----
 
-=== Compliant solution
-
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
```
sebastien-marichal, line 20: Pedantic: You could add a small sentence to introduce the use of `string.Empty` as a compliant solution.

PR #2176, opened 2023-06-13
Modify rule S4830: LaYC format

hendrik-buchwald-sonarsource:
## Review

- [x] text style and tone


rules/S4830/common/fix/validation.adoc
```diff
@@ -0,0 +1,23 @@
...
+trusted root CAs, and certificates issued by these authorities are validated
+automatically. This is the best practice, and it requires no additional code or
+configuration.
+
+==== Working with self-signed certificates or non-standard CAs
+
+In some cases, you might need to work with a server using a self-signed
+certificate, or a certificate issued by a CA not included in your trusted roots.
+Rather than disabling certificate validation in your code, you can add the
+necessary certificates to your trust store.
```
loris-s-sonarsource, line 22: This looks like "extra mile" or "pitfalls" materials no ? Since the fix is only to activate validation

rules/S4830/csharp/how-to-fix-it/dot-net.adoc
```diff
@@ -0,0 +1,24 @@
...
+----
+using System.Net;
+
+public void send()
+{
+    ServicePointManager.ServerCertificateValidationCallback +=
+        (sender, certificate, chain, errors) => {
+            return true; // Noncompliant
+        };
+}
```
loris-s-sonarsource, line 19: WDYT of adding a little more? And then adding the benign code as compliant ? 

```suggestion
using System.Net;
using System.Net.Http;

public static void connect()
{
    ServicePointManager.ServerCertificateValidationCallback +=
	 (sender, certificate, chain, errors) => {
	     return true; // Noncompliant
    };

    HttpClient httpClient = new HttpClient();
    HttpResponseMessage response = httpClient.GetAsync("https://example.com").Result;
}
```

rules/S4830/csharp/how-to-fix-it/dot-net.adoc
```diff
@@ -0,0 +1,24 @@
+== How to fix it in .NET
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
```
loris-s-sonarsource, line 5: because it's a global change. I suggest adding:

"In the following example, the callback change impacts the entirety of HTTP requests made by the application.",

rules/S4830/javascript/how-to-fix-it/node-js.adoc
```diff
@@ -0,0 +1,89 @@
...
+----
+const https = require('node:https');
+
+let options = {
+  hostname: 'www.example.com',
+  port: 443,
+  path: '/',
+  method: 'GET',
+  secureProtocol: 'TLSv1_2_method',
+  rejectUnauthorized: false // Secondary location
```
loris-s-sonarsource, line 19:
```suggestion
  rejectUnauthorized: false
```

I don't think the users understand this comment or care. However the code rationale should indicate that the problem comes from this line

rules/S4830/php/how-to-fix-it/curl.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+==== Noncompliant code example
+
+[source,php,diff-id=1,diff-type=noncompliant]
+----
+curl_setopt($curl, CURLOPT_SSL_VERIFYPEER, false); // Noncompliant
```
loris-s-sonarsource, line 11: Here I suggest adding the initialization of $curl and optionally a request made

rules/S4830/python/how-to-fix-it/openssl.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+include::../../common/fix/code-rationale.adoc[]
+
+==== Noncompliant code example
+
+[source,python,diff-id=1,diff-type=noncompliant]
+----
+from OpenSSL import SSL
+
+ctx1 = SSL.Context(SSL.TLSv1_2_METHOD) # Noncompliant: by default certificate validation is not done
```
loris-s-sonarsource, line 13:
```suggestion
ctx1 = SSL.Context(SSL.TLSv1_2_METHOD) # Noncompliant:
```

I think this part belongs in the code rationale instead

rules/S4830/python/how-to-fix-it/ssl.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+include::../../common/fix/code-rationale.adoc[]
+
+==== Noncompliant code example
+
+[source,python,diff-id=1,diff-type=noncompliant]
+----
+import ssl
+
+ctx1 = ssl._create_unverified_context() # Noncompliant: by default certificate validation is not done
```
loris-s-sonarsource, line 13:
```suggestion
ctx1 = ssl._create_unverified_context() # Noncompliant
```

Same train of thought, I think it belongs in the code rationale instead.

And I have mixed feelings about adding multiple functions.  But I'm not sure so I'm not suggesting anything

rules/S4830/python/how-to-fix-it/ssl.adoc
```diff
@@ -0,0 +1,35 @@
...
+ctx2 = ssl._create_stdlib_context() # Noncompliant: by default certificate validation is not done
+
+ctx3 = ssl.create_default_context()
+ctx3.verify_mode = ssl.CERT_NONE # Noncompliant
+----
+
+==== Compliant solution
+
+[source,python,diff-id=1,diff-type=noncompliant]
```
loris-s-sonarsource, line 22:
```suggestion
[source,python,diff-id=1,diff-type=compliant]
```

rules/S4830/vbnet/how-to-fix-it/dot-net.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+==== Noncompliant code example
+
+[source,vbnet]
+----
+Imports System.Net
```
loris-s-sonarsource, line 11: Same suggestion as .Net Here

rules/S4830/docker/how-to-fix-it/common.adoc
```diff
@@ -0,0 +1,28 @@
+== How to fix it
```
loris-s-sonarsource, line 1: It's actually Curl instead of common right ?
```suggestion
== How to fix it in cURL
```

loris-s-sonarsource (changes requested):
In general, I recommend adding a small web request to add a little realism and show the user that this impacts web requests. 


I mentioned it in the .Net samples and in general, adding them could add compliant code samples and show more colorfully to the user that they shouldn't do that, thanks to the diff views. It is only done on Docker and JS for now.

To follow what's written in the impact sections (which LGTM ! ✅ very nice), we could make requests to "bank.example.com". This could have even more impact.

docs/header_names/allowed_framework_names.adoc
```diff
@@ -50,3 +55,5 @@
 * SQLAlchemy
 * Amazon DynamoDB
 * python-ldap
+* Request
+* SSL
```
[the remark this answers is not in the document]
hendrik-buchwald-sonarsource: True, fixed.

rules/S4830/common/fix/validation.adoc
```diff
@@ -0,0 +1,23 @@
+Addressing the vulnerability of disabled TLS certificate validation primarily
+involves re-enabling the validation. However, it is important to be aware that
+there might be times when you need to work with self-signed certificates, or
+certificates issued by a Certificate Authority (CA) not included in the standard
+trusted root CAs of your environment. In these cases, the solution is not to
+disable certificate validation in your code, but to manage your trusted CAs
+appropriately.
```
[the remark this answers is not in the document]
hendrik-buchwald-sonarsource: True, it repeats itself. I have pushed a new version. It is very similar to your suggestion but also contains a transition to the subsections.

rules/S4830/common/fix/validation.adoc
hendrik-buchwald-sonarsource: I don't think it fits the pitfalls. It could fit the extra mile. On the other hand, it could be also considered part of the fix. After all, there was a reason somebody disabled certificate validation. So if they simply reenable it, that probably does not solve their problem.

rules/S4830/docker/how-to-fix-it/common.adoc
hendrik-buchwald-sonarsource: I am not convinced it makes sense to have a code sample for every command that the rule supports. curl is in this case also representing other commands like wget, thus I made this a common fix for now.

rules/S4830/common/fix/validation.adoc
loris-s-sonarsource: Oh, I see the point. Sounds good

rules/S4830/vbnet/how-to-fix-it/dot-net.adoc
loris-s-sonarsource: It looks like it could look like this (I added both compliant and non compliant in the same snippet and took the current compliant from rspec:

``` vbnet
Imports System
Imports System.Linq
Imports System.Net
Imports System.Net.Security
				
Public Module Example
	Public Sub Main()
		Dim Development As Boolean = True
		Dim ValidCerts As String() = {"cert-example-1", "cert-example-2"}
		
		ServicePointManager.ServerCertificateValidationCallback = Function(sender, certificate, chain, errors) True ' Noncompliant
			
		ServicePointManager.ServerCertificateValidationCallback =
		Function(sender, certificate, chain, errors)
			If Development Then Return True ' For development, trust all certificates
			Return Errors = SslPolicyErrors.None AndAlso ValidCerts.Contains(certificate.GetCertHashString()) ' Compliant: trust only some certificates
		End Function
				
		Dim request As System.Net.HttpWebRequest = System.Net.HttpWebRequest.Create(New System.Uri("https://example.com"))
		request.Method = System.Net.WebRequestMethods.Http.Get
		Dim response As System.Net.HttpWebResponse = request.GetResponse()
		response.Close()
	End Sub
End Module
```

loris-s-sonarsource: It looks like it could work, depending on the web. config file, out of scope here. In dotnetfiddle it looks like the webconfig file is configured to be safe so it does not work.

I think we can assume that users running into this rule will have it enabled and it will work for them

rules/S4830/docker/how-to-fix-it/common.adoc
loris-s-sonarsource: got it

rules/S4830/vbnet/how-to-fix-it/dot-net.adoc
hendrik-buchwald-sonarsource: Okay, thanks!

rules/S4830/cfamily/how-to-fix-it/botan.adoc
```diff
@@ -0,0 +1,87 @@
+== How to fix it in Botan
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
```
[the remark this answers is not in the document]
hendrik-buchwald-sonarsource: Done. I hope it works :sweat_smile:

rules/S4830/javascript/how-to-fix-it/node-js.adoc
```diff
@@ -0,0 +1,93 @@
...
+----
+const https = require('node:https');
+
+let options = {
+  hostname: 'www.example.com',
+  port: 443,
+  path: '/',
+  method: 'GET',
+  secureProtocol: 'TLSv1_2_method',
+  rejectUnauthorized: false
```
loris-s-sonarsource, line 25: just to avoid unnecessary diff colors:
```suggestion
  rejectUnauthorized: false,
  secureProtocol: 'TLSv1_2_method'
```

rules/S4830/javascript/how-to-fix-it/node-js.adoc
```diff
@@ -0,0 +1,93 @@
...
+  path: '/',
+  method: 'GET',
+  secureProtocol: 'TLSv1_2_method'
+};
+
+let req = https.request(options, (res) => {
+  res.on('data', (d) => {
+    process.stdout.write(d);
+  });
+});
```
loris-s-sonarsource, line 87:
```suggestion
const tls = require('node:tls');

let options = {
    secureProtocol: 'TLSv1_2_method'
};

let socket = tls.connect(443, "www.example.com", options, () => {
  process.stdin.pipe(socket);
  process.stdin.resume();
});
```

rules/S4830/javascript/how-to-fix-it/node-js.adoc
```diff
@@ -0,0 +1,93 @@
...
+}); // Noncompliant
+----
+
+[source,javascript,diff-id=2,diff-type=noncompliant]
+----
+const tls = require('node:tls');
+
+let options = {
+    secureProtocol: 'TLSv1_2_method',
+    rejectUnauthorized: false
```
loris-s-sonarsource, line 41:
```suggestion
    rejectUnauthorized: false,
    secureProtocol: 'TLSv1_2_method'
```

rules/S4830/javascript/how-to-fix-it/request.adoc
```diff
@@ -0,0 +1,40 @@
...
+==== Noncompliant code example
+
+[source,javascript,diff-id=1,diff-type=noncompliant]
+----
+const request = require('request');
+
+let socket = request.get({
+  url: 'www.example.com',
+  secureProtocol: 'TLSv1_2_method',
+  rejectUnauthorized: false // Noncompliant
```
loris-s-sonarsource, line 22:
```suggestion
  rejectUnauthorized: false, // Noncompliant
  secureProtocol: 'TLSv1_2_method'
```

loris-s-sonarsource (changes requested):
Minor Changes, else LGTM

loris-s-sonarsource (approved):
LGTM

PR #2177, opened 2023-06-13
Modify rule S4277: Update to LaYC

csaba-sagi-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4277/csharp/rule.adoc
+++ b/rules/S4277/csharp/rule.adoc
@@ -1,10 +1,12 @@
-== Why is this an issue?
+include::../why-dotnet.adoc[]
 
-include::../description.adoc[]
+== How to fix it
 
-=== Noncompliant code example
+=== Code examples
 
-[source,csharp]
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 [Export(typeof(IFooBar))]
 [PartCreationPolicy(CreationPolicy.Shared)]
@@ -21,10 +23,9 @@ public class Program
 }
 ----
 
+==== Compliant solution
 
-=== Compliant solution
-
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
 [Export(typeof(IFooBar))]
 [PartCreationPolicy(CreationPolicy.Shared)]
@@ -41,14 +42,4 @@ public class Program
 }
 ----
 
-ifdef::env-github,rspecator-view[]
-
-'''
-== Implementation Specification
-(visible only on this page)
-
-include::../message.adoc[]
-
-include::../highlighting.adoc[]
-
-endif::env-github,rspecator-view[]
+include::../rspecator.adoc[]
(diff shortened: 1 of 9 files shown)


rules/S4277/metadata.json
```diff
@@ -25,5 +25,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
```
cristian-ambrosini-sonarsource, line 28: I think it can be feasible actually, isn't 'just' a matter of replacing `new MyClass()` with `serviceProvider.GetService<MyClass>()`?

rules/S4277/csharp/rule.adoc
```diff
@@ -41,14 +42,6 @@ public class Program
 }
```
cristian-ambrosini-sonarsource, line 42: Educational: why is this example getting the interface `GetService<IFooBar>` and not the class itself `GetService<FooBar>`?

cristian-ambrosini-sonarsource (changes requested):

rules/S4277/metadata.json
csaba-sagi-sonarsource: There are multiple problems with a quickfix here.

- It might be that actually the attribute is not needed and not the constructor is problematic. (For example, you could have copied the class together with the attribute for whatever reason and forgot o remove it)
- The `serviceProvider` might not yet be available at the place where the constructor was called. So if we would just do that replacement, we could create a non-buildable code for the user.

rules/S4277/csharp/rule.adoc
csaba-sagi-sonarsource: The idea is to depend on the interface and not the implementation, so if tomorrow there will be a Foobar2 class that implements the interface the consumer code should not change at all.
More on dependency injection: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection

rules/S4277/metadata.json
cristian-ambrosini-sonarsource: I don't really see the first one being a problem for the quickfix: the same reasoning can be applied to other rules as well. Moreover, if that's something that we should be worried then we might need a bit of clarification in the `How to fix it` section telling the users that one fix could also be to remove the attribute because it's not needed.
Regarding the second point, yes I thought about it, I suggest to mark the quickfix as `partial` (not `targeted`). The quickfix will be available only if serviceProvider is found in the context. WDYT?

csaba-sagi-sonarsource: How will one know if the serviceProvider is available? To know that one should look for all the symbols in the context and see if it is an IServiceProvider, which I would definitely not do for a quickfix.

cristian-ambrosini-sonarsource: Yeah, that's also a fair point, ok then you convinced me :smiley:

cristian-ambrosini-sonarsource (approved):
LGTM

PR #2178, opened 2023-06-13
Modify rule S2290: Update to LaYC

csaba-sagi-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S2290/csharp/rule.adoc
+++ b/rules/S2290/csharp/rule.adoc
@@ -1,10 +1,13 @@
 == Why is this an issue?
 
-Field-like events are events that do not have explicit https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/add[`add`] and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/remove[`remove`] methods. The compiler generates a `private` `delegate` field to back the event, as well as generating the implicit `add` and `remove` methods.
+https://learn.microsoft.com/en-us/dotnet/csharp/event-pattern#define-and-raise-field-like-events[Field-like] events are events that do not have explicit https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/add[`add`] and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/remove[`remove`] accessors.
+[source,csharp]
+----
+public event EventHandler MyEvent; // No add and remove accessors
+----
+The compiler generates a `private` `delegate` field to back the event, as well as generating the implicit `add` and `remove` accessors.
 
 When a `virtual` field-like `event` is overridden by another field-like `event`, the behavior of the C# compiler is to generate a new `private` `delegate` field in the derived class, separate from the parent's field. This results in multiple and separate events being created, which is rarely what's actually intended.
-
-To prevent this, remove the `virtual` designation from the parent class event.
 
 === Noncompliant code example
 
@@ -12,21 +15,21 @@ To prevent this, remove the `virtual` designation from the parent class event.
 ----
 abstract class Car
 {
-  public virtual event EventHandler OnRefueled; // Noncompliant
+  public virtual event EventHandler OnRefuel; // Noncompliant
 
   public void Refuel()
   {
-    // This OnRefueled will always be null
-     if (OnRefueled != null)
+    // This OnRefuel will always be null
+     if (OnRefuel != null)
      {
-       OnRefueled(this, null);
+       OnRefuel(this, EventArgs.Empty);
      }
   }
 }
 
 class R2 : Car
 {
-  public override event EventHandler OnRefueled;
+  public override event EventHandler OnRefuel;
 }
 
 class Program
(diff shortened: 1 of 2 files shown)


rules/S2290/csharp/rule.adoc
```diff
@@ -1,17 +1,14 @@
...
-
-
-To prevent this, remove the ``++virtual++`` designation from the parent class event.
-
+To prevent this, remove the `virtual` designation from the parent class event.
 
 === Noncompliant code example
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
```
antonioaversa, line 11: Comment for line 20 (I can't comment on the line directly, since it was not modified).
We should avoid giving examples not respecting [MS Guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/event):
```
DO NOT pass null as the event data parameter when raising an event.

You should pass EventArgs.Empty if you don’t want to pass any data to the event handling method. Developers expect this parameter not to be null.
```

A junior dev may copy-paste or replicate the code we show.

rules/S2290/csharp/rule.adoc
```diff
@@ -64,7 +60,10 @@ abstract class Car
...
 }
 
-class R2 : Car {}
+class R2 : Car 
+{
+
+}
 
 class Program
 {
```
antonioaversa, line 69: The following is equivalent and simpler:
```
        r2.OnRefueled += (o, a) =>
        {
            Console.WriteLine("This event will be called");
        };
```

rules/S2290/csharp/rule.adoc
```diff
@@ -80,7 +79,14 @@ class Program
...
 ----
 
+== Resources
 
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/add[add keyword]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/remove[remove keyword]
+* https://learn.microsoft.com/en-us/dotnet/csharp/delegate-class[delegates]
+* https://learn.microsoft.com/en-us/dotnet/csharp/events-overview[Introduction to events]
```
antonioaversa, line 89: I would add some content about `field-like events`, linking it in the why section:
https://learn.microsoft.com/en-us/dotnet/csharp/event-pattern#define-and-raise-field-like-events

The "field-like" terminology is new to me, and may be new to the reader.

antonioaversa (changes requested):

rules/S2290/csharp/rule.adoc
```diff
@@ -1,17 +1,14 @@
 == Why is this an issue?
 
-Field-like events are events that do not have explicit ``++add++`` and ``++remove++`` methods. The compiler generates a ``++private++`` ``++delegate++`` field to back the event, as well as generating the implicit ``++add++`` and ``++remove++`` methods.
+Field-like events are events that do not have explicit https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/add[`add`] and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/remove[`remove`] methods. The compiler generates a `private` `delegate` field to back the event, as well as generating the implicit `add` and `remove` methods.
 
+When a `virtual` field-like `event` is overridden by another field-like `event`, the behavior of the C# compiler is to generate a new `private` `delegate` field in the derived class, separate from the parent's field. This results in multiple and separate events being created, which is rarely what's actually intended.
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: I am not sure moving it up will change a lot. At this point the only thing between the noncompliant example and the paragraph you suggest is this short sentence `To prevent this, remove the `virtual` designation from the parent class event.`
Do you think that short sentence disrupts the thinking of the reader that much that it is worth it moving?

rules/S2290/csharp/rule.adoc
```diff
@@ -1,17 +1,14 @@
...
 
-When a ``++virtual++`` field-like ``++event++`` is overridden by another field-like ``++event++``, the behavior of the C# compiler is to generate a new ``++private++`` ``++delegate++`` field in the derived class, separate from the parent's field. This results in multiple and separate events being created, which is rarely what's actually intended.
-
-
-To prevent this, remove the ``++virtual++`` designation from the parent class event.
-
+To prevent this, remove the `virtual` designation from the parent class event.
```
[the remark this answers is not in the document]
csaba-sagi-sonarsource: I moved it though I am afraid people will glimpse over the sentence as it is nested between the examples.

rules/S2290/csharp/rule.adoc
```diff
@@ -1,125 +1,97 @@
...
-  public event EventHandler OnRefueled; // Compliant
+  public event EventHandler OnRefuel; // Compliant
 
   public void Refuel()
   {
-    if (OnRefueled != null)
+    if (OnRefuel != null)
     {
-      OnRefueled(this, null);
+      OnRefuel(this, null);
```
antonioaversa, line 63: You want to pass `EventArgs.Empty` here too

rules/S2290/csharp/rule.adoc
```diff
@@ -1,125 +1,97 @@
...
-      Console.WriteLine("This event will never be called");
-    });
+        Console.WriteLine("This event will be called");
+    };
     r2.Refuel();
   }
 }
 ----
 
+To prevent this, remove the `virtual` designation from the parent class event.
```
antonioaversa, line 49: What about having this sentence after the `=== Compliant solution`, but before the code example? This way it's harder to glimpse over (https://github.com/SonarSource/rspec/pull/2178#discussion_r1229343285).

antonioaversa (approved):
LGTM
Couple of conversation to address before merging

PR #2179, opened 2023-06-13
Modify rule S1534: Adapt to LaYC

victor-diez-sonarsource:


rules/S1534/javascript/rule.adoc
```diff
@@ -1,16 +1,13 @@
 == Why is this an issue?
 
-JavaScript allows duplicate member names in classes and object literals, but only the last instance of a duplicated name determines the actual value that will be used for it. Therefore, changing values of other occurrences of a duplicated name will have no effect and may cause misunderstandings and bugs.
-
-
-Defining a ``++class++`` with a duplicated ``++constructor++`` will generate an error.
+JavaScript allows duplicate property names in classes and object literals since ES2015. The last occurrence will overwrite previous definitions. Therefore, having more than one occurrence will have no effect and may cause misunderstandings and bugs.
```
vilchik-elena, line 3: > After ES2015, duplicate property names are allowed everywhere, including [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#duplicate_property_names).

I think you should either remove `since ES2015` or replace it with `even in strict mode since ES2015`

vilchik-elena, line 3: Add code example after this paragraph

rules/S1534/javascript/rule.adoc
```diff
@@ -1,16 +1,13 @@
...
 
+Defining a `class` with a duplicated `constructor` will generate an error.
```
vilchik-elena, line 5: Add code example after this paragraph

rules/S1534/javascript/rule.adoc
```diff
@@ -1,16 +1,13 @@
...
 
 Before ECMAScript 2015, using duplicate names will generate an error in JavaScript strict mode code.
```
vilchik-elena, line 7: I would drop this if we mention ECMAScript 2015 in the first sentense.

rules/S1534/javascript/rule.adoc
vilchik-elena, line 3: ES2015 -> ECMAScript 2015

rules/S1534/javascript/rule.adoc
```diff
@@ -1,16 +1,13 @@
...
 
 This rule will also report on duplicate properties in JSX.
```
vilchik-elena, line 9: add corresponding code example after this paragraph

rules/S1534/javascript/rule.adoc
```diff
@@ -27,6 +24,11 @@ let data = {
   1: "value" // Noncompliant - duplicate of "1"
 }
 
+class Foo {
```
vilchik-elena, line 27: why did you decide to add this example?

rules/S1534/javascript/rule.adoc
```diff
@@ -1,16 +1,13 @@
 == Why is this an issue?
```
vilchik-elena, line 1: I think compliant part can be dropped for this rule

vilchik-elena (changes requested):

rules/S1534/javascript/rule.adoc
victor-diez-sonarsource: to show that class method and properties are also covered

PR #2195, opened 2023-06-14
Modify rule S2589: LaYC format

sebastien-marichal:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2589/csharp/metadata.json
+++ b/rules/S2589/csharp/metadata.json
@@ -1,3 +1,3 @@
 {
-  
+  "title": "Boolean expressions should not be unnecessary"
 }
--- a/rules/S2589/csharp/metadata.json
+++ b/rules/S2589/csharp/metadata.json
@@ -1,3 +1,3 @@
 {
-  "title": "Boolean expressions should not be unnecessary"
+  
 }
--- a/rules/S2589/csharp/rule.adoc
+++ b/rules/S2589/csharp/rule.adoc
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-An operand of a boolean expression that never changes the result of the expression might not match the programmer's intent and leads to unexpected behavior and potential bugs.
+An operand of a boolean expression that never changes the result of the expression might not match the programmer's intent and can lead to unexpected behavior and potential bugs.
 
 [source,csharp]
 ----
--- a/rules/S2589/csharp/rule.adoc
+++ b/rules/S2589/csharp/rule.adoc
@@ -76,7 +76,7 @@ public void Sample(bool b, bool c)
 
 ==== Compliant solution
 
-* The unnecessary operand is made essential
+* The unnecessary operand is updated
 +
 [source,csharp,diff-type=compliant]
 ----
--- a/rules/S2589/csharp/rule.adoc
+++ b/rules/S2589/csharp/rule.adoc
@@ -1,63 +1,71 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+When an operand of a https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1224-boolean-expressions[boolean expression] does not influence the result in a condition, it increases cognitive complexity.
 
-=== Noncompliant code example
-
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
-public void Sample(bool b, bool c, string s)
+public void Sample(bool b, bool c)
 {
     var a = true;
-    if (a) // Noncompliant
+    if (b && a)     // Noncompliant: "a" is always "true", only "b" will change the expression result
     {
         DoSomething();
     }
 
-    if (b && a) // Noncompliant; "a" is always "true"
+    if (c || !a)    // Noncompliant: "!a" is always "false"
+    {
+        DoSomething();
+    }
+}
+----
+
+Since one of the operands is unnecessary to the condition, the condition should be updated.
+
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+public void Sample(bool b, bool c)
+{
+    if (b)          // Compliant
     {
         DoSomething();
     }
 
-    if (c || !a) // Noncompliant; "!a" is always "false"
+    if (c)          // Compliant
     {
         DoSomething();
     }
+}
+----
 
+This also applies to conditional expression in https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1218-conditional-operator[conditional operator]. If one of the operands always evaluates to `null`, the expression should be updated.
+
+[source,csharp,diff-id=2,diff-type=noncompliant]
+----
+public void Sample()
+{
     string d = null;
-    var v1 = d ?? "value"; // Noncompliant, d is always null
-    var v2 = s ?? d; // Noncompliant, d is always null and the produced value is always equal to s. The condition to check the value of s is gratuitous.
+    var v1 = d ?? "value";  // Noncompliant: "d" is always null
+    var v2 = s ?? d;        // Noncompliant: "d" is always null and the produced value is always equal to s.
 }
 ----
 
-=== Compliant solution
-
-[source,csharp]
+[source,csharp,diff-id=2,diff-type=compliant]
 ----
-public void Sample(bool b, bool c, string s)
+public void Sample()
 {
-    var a = true;
-    if (Foo(a))
-    {
-        DoSomething();
-    }
-
-    if (b)
-    {
-        DoSomething();
-    }
-
-    if (c)
-    {
-        DoSomething();
-    }
-
-    var v1 = "value";
-    var v2 = s;
+    var v1 = "value";       // Compliant
+    var v2 = s;             // Compliant
 }
 ----
 
-include::../see.adoc[]
+== Resources
+
+=== Documentation
+
+* https://cwe.mitre.org/data/definitions/571[MITRE, CWE-571] - Expression is Always True
+* https://cwe.mitre.org/data/definitions/570[MITRE, CWE-570] - Expression is Always False
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1224-boolean-expressions[Boolean expressions in C#]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1218-conditional-operator[Conditional expression with the conditional operator in C#]
 
 ifdef::env-github,rspecator-view[]
 
--- a/rules/S2589/csharp/rule.adoc
+++ b/rules/S2589/csharp/rule.adoc
@@ -11,7 +11,7 @@ if (a)
 }
 ----
 
-This also applies to conditional expression in https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator[conditional operator] when one of the operands always evaluates to `null`.
+This also applies to the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator[null coalescing operator] when one of the operands always evaluates to `null`.
 
 [source,csharp]
 ----
@@ -41,6 +41,7 @@ In these cases, it is obvious the code is as intended.
 == How to fix it
 
 The conditions should be reviewed to decide whether:
+
 * to update the unnecessary operand
 * to remove the unnecessary operand
 
@@ -76,8 +77,8 @@ public void Sample(bool b, bool c)
 
 ==== Compliant solution
 
-* The unnecessary operand is updated
-+
+The unnecessary operand is updated:
+
 [source,csharp,diff-type=compliant]
 ----
 public void Sample(bool b, bool c, string s)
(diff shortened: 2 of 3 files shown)


rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+When an operand of a https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1224-boolean-expressions[boolean expression] does not influence the result in a condition, it increases cognitive complexity.
```
martin-strecker-sonarsource, line 3: I would not link to the C# spec. The spec is hard to read and for language enthusiasts only (You are free to keep the link, if you can explain to me what the `true` operator is, where it originates from, how it is used, how it relates to conditional and/or, and why it shouldn't be used anymore 🤯 ).

It can be shortened and made more precise like so:
```suggestion
An operand of a boolean expression that never changes the result of the expression can be removed.
```
While technically correct, I also think that it is confusing. Maybe you can mix it with the original message or keep the original message.

> If a boolean expression doesn’t change the evaluation of the condition, then it is entirely unnecessary, and can be removed. If it is gratuitous because it does not match the programmer’s intent, then it’s a bug and the expression should be fixed.

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
 
-=== Noncompliant code example
-
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
```
martin-strecker-sonarsource, line 5: Shouldn't this go into the `How to fix` section?

A minimal example for this section for illustration could be
```cs
if (true)
{
}
```
(I hope this raises the issue).

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
 ----
-public void Sample(bool b, bool c, string s)
+public void Sample(bool b, bool c)
 {
     var a = true;
-    if (a) // Noncompliant
```
martin-strecker-sonarsource, line 12: Why was this example removed?

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
-    {
-        DoSomething();
-    }
-
-    if (b && a) // Noncompliant; "a" is always "true"
+    if (b && a)     // Noncompliant: "a" is always "true", only "b" will change the expression result
```
martin-strecker-sonarsource, line 10:
```suggestion
    if (b && a)     // Noncompliant: "a" is always "true" and can be removed
```

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
     {
         DoSomething();
     }
 
-    if (c || !a) // Noncompliant; "!a" is always "false"
+    if (c || !a)    // Noncompliant: "!a" is always "false"
```
martin-strecker-sonarsource, line 15:
```suggestion
    if (c || !a)    // Noncompliant: "!a" is always "false" and can be removed
```

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
     }
-
-    string d = null;
-    var v1 = d ?? "value"; // Noncompliant, d is always null
-    var v2 = s ?? d; // Noncompliant, d is always null and the produced value is always equal to s. The condition to check the value of s is gratuitous.
 }
 ----
 
-=== Compliant solution
+Since one of the operands is unnecessary to the condition, the condition should be updated.
```
martin-strecker-sonarsource, line 22:
```suggestion
Since one of the operands is unnecessary to the conditions, the conditions should be updated.
```

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
+----
+
+== Resources
+
+=== Documentation
+
+* https://cwe.mitre.org/data/definitions/571[MITRE, CWE-571] - Expression is Always True
+* https://cwe.mitre.org/data/definitions/570[MITRE, CWE-570] - Expression is Always False
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1224-boolean-expressions[Boolean expressions in C#]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1218-conditional-operator[Conditional expression with the conditional operator in C#]
```
martin-strecker-sonarsource, line 68: Don't link the spec. Use the Average Joe documentation instead
```suggestion
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-and-operator-[Conditional logical AND operator &&]
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-or-operator-[Conditional logical OR operator ||]
```

martin-strecker-sonarsource: Add the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator link

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
 
-    var v1 = "value";
-    var v2 = s;
+[source,csharp,diff-id=2,diff-type=noncompliant]
+----
+public void Sample()
+{
+    string d = null;
+    var v1 = d ?? "value";  // Noncompliant: "d" is always null
+    var v2 = s ?? d;        // Noncompliant: "d" is always null and the produced value is always equal to s.
```
martin-strecker-sonarsource, line 48: s is undefined

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,71 @@
...
+This also applies to conditional expression in https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1218-conditional-operator[conditional operator]. If one of the operands always evaluates to `null`, the expression should be updated.
 
-    var v1 = "value";
-    var v2 = s;
+[source,csharp,diff-id=2,diff-type=noncompliant]
+----
+public void Sample()
+{
+    string d = null;
+    var v1 = d ?? "value";  // Noncompliant: "d" is always null
```
martin-strecker-sonarsource, line 47:
```suggestion
    var v1 = d ?? "value";  // Noncompliant: "d" is always null and v1 is always "value". 
```

rules/S2589/csharp/rule.adoc
martin-strecker-sonarsource, line 48:
```suggestion
    var v2 = s ?? d;        // Noncompliant: "d" is always null and v2 is always equal to s.
```

martin-strecker-sonarsource (changes requested):

rules/S2589/csharp/rule.adoc
sebastien-marichal: I believe mentioning the operand is important as `boolean expression` could be interpreted as the whole condition expression.

I will try to come up with something less confusing.

rules/S2589/csharp/rule.adoc
sebastien-marichal: The given compliant solution was to use a method within the condition. But another compliant solution could be to remove entirely the `if`.
I can add it back and remove the whole `if` as a better compliant example. Since it was my opinion I preferred to remove it.
WDYT?

rules/S2589/csharp/rule.adoc
sebastien-marichal: Doesn't have to be in the `How to fix` section if not too complete, it is our call.
Since we try to have "blog-post" like description, I tend to have most of the content of the rules in the `why` section.

sebastien-marichal: > ```cs
> if (true)
> {
> }
> ```

It does not raise, it has the same exception as [S2583](https://sonarsource.github.io/rspec/#/rspec/S2583/csharp). I will add the same exception section as S2583.

rules/S2589/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
+  "title": "Boolean expressions should not be unnecessary"
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: > not necessary; with no reasonable cause

[Cambridge dictionary](https://dictionary.cambridge.org/dictionary/english/gratuitous)

Seems very fitting to me

> not be unnecessary

is a double negative. "Boolean expressions should not be redundant" seems preferable to me. 

Gratuitous is used quite some times in different places.
https://github.com/search?q=repo%3ASonarSource%2Frspec%20gratuitous&type=code

But redundant is used much more often.
https://github.com/search?q=repo%3ASonarSource%2Frspec+redundant&type=code

I would keep it for the moment and add an issue to replace "gratuitous" with "redundant" after we have confirmation from the docs team (#guild-education).

rules/S2589/csharp/rule.adoc
martin-strecker-sonarsource: Fair enough, but you have two/three different kinds of examples and two different kinds of fixes (removal vs. correction). This should go into "how to fix", IMO.

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,73 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+An operand of a boolean expression that never changes the result of the expression might not match the programmer's intent and leads to unexpected behavior and potential bugs.
```
martin-strecker-sonarsource, line 3: Looks good to me.
```suggestion
An operand of a boolean expression that never changes the result of the expression might not match the programmer's intent and can lead to unexpected behavior and potential bugs.
```

rules/S2589/csharp/metadata.json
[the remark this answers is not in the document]
sebastien-marichal: What about `Boolean expressions should be essential`?
I like the fact there is no negation anymore but I am not sure about "essential".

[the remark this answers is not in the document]
martin-strecker-sonarsource: > essential

I don't think this works. Let's stick with gracious for the moment and raise the issue in #guild-eduction

rules/S2589/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Boolean expressions should not be gratuitous",
+  "title": "Boolean expressions should not be unnecessary",
```
martin-strecker-sonarsource, line 2: It's probably better to not touch the title as it affects all languages. It's not worth it at the moment. If we want to get rid of gratuitous, we should do so in all rules at once. Just to be clear: I had to look up gratuitous as well. For me, it violates our guidelines:

> Try to use simple English and avoid slang. Assume users have high school-level English, equivalent to CEFR levels B1/B2, at best.

So we need confirmation and then remove it everywhere. That is done best in a dedicated PR.

martin-strecker-sonarsource (changes requested):

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,141 @@
...
+}
+----
+* When the condition is the literal `true` or `false`.
+
+In these cases, it is obvious the code is as intended.
+
+== How to fix it
+
+The conditions should be reviewed to decide whether:
+* to update the unnecessary operand
```
martin-strecker-sonarsource, line 44: NIT: When viewed on GitHub, the bullet points do not show up. A new line is needed:
```suggestion

* to update the unnecessary operand
```

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,141 @@
...
-    var v2 = s;
+    string d = GetStringData();
+    var v1 = d ?? "value";  // Compliant
+    var v2 = s ?? d;        // Compliant
 }
 ----
 
-include::../see.adoc[]
+* The unnecessary operand is removed
++
```
martin-strecker-sonarsource, line 108:
```suggestion
The unnecessary operand is removed:
```

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,141 @@
...
+[source,csharp,diff-type=compliant]
```
martin-strecker-sonarsource, line 109: Let's give it a try.
```suggestion
[source,csharp,diff-id=1,diff-type=compliant]
```

martin-strecker-sonarsource (approved):
Some minor suggestions. All are optional.

rules/S2589/csharp/rule.adoc
```diff
@@ -1,63 +1,141 @@
...
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+* The unnecessary operand is updated
++
+[source,csharp,diff-type=compliant]
```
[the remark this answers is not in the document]
sebastien-marichal: Unfortunately, it does not work 😞 
The main issue is what changed lines to display on the noncompliant example.
I let the `diff-type` if, at some point, this will be possible.

PR #2196, opened 2023-06-14
Modify rule S2583: LaYC format

martin-strecker-sonarsource:
[S2583](https://sonarsource.github.io/rspec/#/rspec/S2583/csharp)

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2583/csharp/rule.adoc
+++ b/rules/S2583/csharp/rule.adoc
@@ -2,35 +2,58 @@
 
 include::../description.adoc[]
 
-=== Noncompliant code example
+=== Exceptions
 
+This rule will not raise an issue in either of these cases:
+
+* When the condition is a single `const bool`
++
 [source,csharp]
+----
+const bool debug = false;
+//...
+if (debug)
+{
+  // Print something
+}
+----
+* When the condition is the literal `true` or `false`.
+
+In these cases, it is obvious the code is as intended.
+
+== How to fix it
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 public void Sample(bool b)
 {
     bool a = false;
-    if (a) // Noncompliant
+    if (a)                  // Noncompliant: The true branch is never visited
     {
-        DoSomething(); // never executed
+        DoSomething();      // Never reached
     }
 
-    if (!a || b) // Noncompliant; "!a" is always "true", "b" is never evaluated
+    if (!a || b)            // Noncompliant: "!a" is always "true" and "b" is never evaluated
     {
         DoSomething();
     }
     else
     {
-        DoSomethingElse(); // never executed
+        DoSomethingElse();  // Never reached
     }
 
     var d = "xxx";
-    var res = d ?? "value"; // Noncompliant, d is always not null, "value" is never used
+    var res = d ?? "value"; // Noncompliant: d is always not null, "value" is never used
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
 public void Sample(bool b)
 {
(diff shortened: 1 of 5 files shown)


rules/S2583/metadata.json
```diff
@@ -34,5 +34,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "partial"
```
martin-strecker-sonarsource, line 37: The `d ?? "value"` case can be fixed.

rules/S2583/csharp/rule.adoc
```diff
@@ -54,27 +77,10 @@ public void Sample(bool b)
...
-----
-
-* When the condition is the literal ``++true++`` or ``++false++``.
-
-In these cases it is obvious the code is as intended.
-
 include::../see.adoc[]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-and-operator-[Microsoft Learn] - Conditional logical AND operator &&
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-or-operator-[Microsoft Learn] - Conditional logical OR operator ||
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator[Microsoft Learn] - ?? and ??= operators - the null-coalescing operators
```
sebastien-marichal, line 83:
```suggestion
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-and-operator-[Microsoft Learn - Conditional logical AND operator &&]
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-or-operator-[Microsoft Learn - Conditional logical OR operator ||]
* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator[Microsoft Learn - ?? and ??= operators - the null-coalescing operators]
```

rules/S2583/description.adoc
```diff
@@ -1 +1 @@
-Conditional expressions which are always ``++true++`` or ``++false++`` can lead to dead code. Such code is always buggy and should never be used in production.
+Conditional expressions which are always `true` or `false` can lead to dead code. Such code is always buggy and should never be used in production.
```
sebastien-marichal, line 1: I would explain what "dead code" is; the user might not be familiar with this term. Also, we could add more details about its impact on an application
A link to the [wiki page](https://en.wikipedia.org/wiki/Dead_code) could be added.
Or maybe, we could change the term to [unreachable code](https://en.wikipedia.org/wiki/Unreachable_code), which is more precise for this rule (based on Wikipedia's definition).

sebastien-marichal, line 1: I don't agree with this statement or the term `buggy` is not appropriate here.

```suggestion
Conditional expressions which are always `true` or `false` can lead to dead code. Such code should never be used in production.
```

rules/S2583/metadata.json
sebastien-marichal, line 37: The quickfix update should be in `csharp/metadata.json`, not sure we can take this decision for all squads.

sebastien-marichal (changes requested):

rules/S2583/description.adoc
martin-strecker-sonarsource: I agree. I changed it now and I'm asking myself what "Such code should never be used in production." is conveying as information to the user? I would say this sentence needs clarification (what is production code) and is also confusing: Why is it okay to do it in non-production code, like test code? What about removing that part as well?

sebastien-marichal: We should remove it, I agree, it does not make sense.
We could replace this part with a small paragraph about the impact with an example such as unreachable code with a `Dispose` call or some other critical statement.

sebastien-marichal (changes requested):
Only left is to add more details in the description, IMO.

PR #2197, opened 2023-06-14
Modify rule S4635: Update title

sebastien-marichal:
## Review

- [ ] text style and tone


costin-zaharia-sonarsource (commented):
It looks good but can you please update the message too as Martin suggested to have `startIndex` instead of `offset`. You can find the message file in the root folder, the problem is that is shared with Java.

costin-zaharia-sonarsource (approved):
LGTM!

PR #2205, opened 2023-06-15
Modify S1479: LAYC format

gregory-paidis-sonarsource:


rules/S1479/csharp/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  
-}
+    "quickfix": "targeted"
```
csaba-sagi-sonarsource, line 2: How would you implement a quickfix for this?

rules/S1479/resources-dotnet.adoc
```diff
@@ -0,0 +1,6 @@
+== Resources
+
+=== Documentation
+
```
csaba-sagi-sonarsource, line 4: I think you could also repeat here the switch-statement link.

rules/S1479/vbnet/rule.adoc
```diff
@@ -1,17 +1,25 @@
 == Why is this an issue?
```
csaba-sagi-sonarsource, line 1: Optional, if you think it is worth it you might try to duplicate the why part like martin did in https://github.com/SonarSource/rspec/pull/2155

rules/S1479/csharp/metadata.json
gregory-paidis-sonarsource: I would add all the cases in a dictionary, except the last one with the `default`. It should be easy to parse all the arms of the `switch` and get their keys as well as theirs values. What do you think?

rules/S1479/resources-dotnet.adoc
gregory-paidis-sonarsource: I do, but only inside c# rule.adoc, since it's language-specific.
In VB i repeat the `select case` one.

rules/S1479/csharp/metadata.json
csaba-sagi-sonarsource: I am not saying it is not doable I am saying it is complicated. As you will also need to think of where you put the dictionary.
Is it going to be static or not, if the rule raises twice with the same cases of two separate statements will you generate two dictionaries? If yes than the user still needs to clean up. 
 I am not a huge fan of doing overly complicated quickfixes that need further user action.

PR #2209, opened 2023-06-15
S107: Update examples for C# and VB.NET

mary-georgiou-sonarsource:


rules/S107/vbnet/rule.adoc
```diff
@@ -16,28 +16,30 @@ The solution can be to:
...
-End Function
+End Sub
 ----
 
 * Find a better data structure for the parameters
 
 [source,vbnet]
 ----
-Private Structure Point ' Point is a concept example that logically groups data'
+Class Point ' Point is a concept example that logically groups data'
```
martin-strecker-sonarsource, line 32:
```suggestion
Class Point ' In geometry, Point is a logical structure to group data
```

martin-strecker-sonarsource: I don't have a strong opinion regarding Class vs Structure in VB. Keep it a mutable class here as turning this into a properly defined Structure requires too much code.

martin-strecker-sonarsource (approved):
I would use a `readonly record struct` but LGTM otherwise.

PR #2220, opened 2023-06-16
Modify rule S4586: Add how to fix it section

martin-strecker-sonarsource:


csaba-sagi-sonarsource (approved):
I am approving it, I just realized that the quickfix field was not updated. Please update it before merging.

PR #2239, opened 2023-06-19
Modify rule S5146: Fix invalid Python sample

John-Clifton-SonarSource:
I couldn't get Sonarcloud to trigger this issue using the provided noncompliant code example. 

I think the code examples as written end up being circular because the local function 'redirect()' will call itself rather than the imported 'redirect()' function of the same name. The fix is to change the local function name to be redirecting(). I changed the API endpoint name as well so that it matched. Once I had made this change, the noncompliant code example did lead to Sonarcloud spotting the issue.

## Review

- [x] text style and tone

### DIFF
--- a/rules/S5146/python/how-to-fix-it/flask.adoc
+++ b/rules/S5146/python/how-to-fix-it/flask.adoc
@@ -12,8 +12,8 @@ from flask import Flask, redirect
 
 app = Flask("example")
 
-@app.route("/redirect")
-def redirect():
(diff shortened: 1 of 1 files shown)


hendrik-buchwald-sonarsource (approved):
Nice finding :+1: Since it is a security rule, feel free to ping AppSec next time for a review.

The change itself looks good, just the title of the PR was not matching yet the format of the readme. I have the title, so feel free to merge.

PR #2249, opened 2023-06-20
Modify rule S6477: Adapt to LaYC

victor-diez-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6477/javascript/rule.adoc
+++ b/rules/S6477/javascript/rule.adoc
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-React expects a unique identifier `key` in list components to optimize rendering. Avoid `Array` indexes since the order of the items may change. When keys are missing, React will default to using indexes. This can negatively impact performance and may cause issues with the component state.
+To optimize the rendering of React list components, a unique identifier (UID) is required for each list item. This UID lets React identify the item throughout its lifetime. To provide it, use the `key` attribute of the list item. When the `key` attribute is missing, React will default to using the item's index inside the list component. If the element ordering changes, it will cause keys to not match up between renders, recreating the DOM. It can negatively impact performance and may cause issues with the component state.
 
 [source,javascript]
 ----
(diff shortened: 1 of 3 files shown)


rules/S6477/javascript/rule.adoc
```diff
@@ -1,8 +1,6 @@
 == Why is this an issue?
 
-React expects a unique identifier `key` in list components to optimize rendering. Missing keys will negatively impact performance and can bring your application to a halt when there are enough elements. Avoid `Array` indexes since they are not stable. Instead, use a unique attribute like an ID or a combination of attributes.
-
-=== Noncompliant code example
+React expects a unique identifier `key` in list components to optimize rendering. Avoid `Array` indexes since the order of the items may change. When keys are missing, React will default to using indexes. This can negatively impact performance and may cause issues with the component state.
```
vilchik-elena, line 3: May be I am missing some React background but I fail to understand this phrase "React expects a unique identifier `key` in list components to optimize rendering." 
What is "identifier" here? should we better use "property" or "attribute"?
What does mean "unique"? How single identifier can be not unique? Should we better speak about it's values may be?

vilchik-elena, line 3: >Avoid `Array` indexes since the order of the items may change.

What is "indexes"? 
What is "`Array` indexes"? 
Avoid for what? 
Which items?

Same for further text, I have questions about almost each word :)

rules/S6477/javascript/rule.adoc
```diff
@@ -1,8 +1,6 @@
...
 
 [source,javascript]
```
vilchik-elena, line 5: Add "\\ Noncompliant" and explain what gonna happen

vilchik-elena (changes requested):

rules/S6477/javascript/rule.adoc
victor-diez-sonarsource: not sure I understand your question. `unique identifier` is different to `unique property` or `unique attribute`. unique identifier refers to the value as it's explained in the later phrase ` database IDs are already unique and are the best option. Otherwise, use a counter or a UUID generator.` And there is also a wiki link about it. I would say using the standard terminology is clearer

victor-diez-sonarsource: indexes and Array indexes refer to same thing, I can repeat Array in both if you prefer. I will make it clearer

vilchik-elena (approved):

PR #2254, opened 2023-06-20
Modify rule S6481: Adapt to LaYC

victor-diez-sonarsource:


rules/S6481/javascript/rule.adoc
```diff
@@ -2,10 +2,6 @@
 
 Providing values like an object to a React `Context` will provoke additional re-renders as
```
vilchik-elena, line 3: `like an object`? very scientific :D Could we be more precise?

rules/S6481/javascript/rule.adoc
```diff
@@ -2,10 +2,6 @@
...
 React doesn't know the object is the same.
```
vilchik-elena, line 4:
```suggestion
React doesn't know the value is the same.
```

rules/S6481/javascript/rule.adoc
```diff
@@ -18,7 +14,8 @@ function Component() {
 }
 ----
 
-=== Compliant solution
+This can significantly impact performance since the whole component tree will be re-rendered each time.
```
vilchik-elena, line 17: Is it intentional to have these 2 sentences in one paragraph? IMHO I would split and move the first one to the previous paragraph as before.

rules/S6481/javascript/rule.adoc
```diff
@@ -18,7 +14,8 @@ function Component() {
...
+Wrapping the value in a `useMemo` hook will avoid additional render passes.
```
vilchik-elena, line 18:
```suggestion
To avoid additional render passes wrap the value in a `useMemo` hook.
```

rules/S6481/javascript/rule.adoc
```diff
@@ -18,7 +14,8 @@ function Component() {
...
 
 [source,javascript]
```
vilchik-elena, line 20: add compliant comment

vilchik-elena (changes requested):
The rule has a double negation, how about simplifying it?

rules/S6481/javascript/rule.adoc
```diff
@@ -1,29 +1,26 @@
 == Why is this an issue?
 
-Providing values like an object to a React `Context` will provoke additional re-renders as
-React doesn't know the object is the same.
-This can significantly impact performance since the whole component tree will be re-rendered each time.
-Wrapping the value in a `useMemo` hook will avoid additional render passes.
+Whenever the `value` property of React Context changes, React will rerender the Context and all its child nodes and Consumers. 
 
-=== Noncompliant code example
+Because each Javascript object carries its identity, object or function expressions get a new identity every run. These should not be used as Context `value` because React will always consider they have changed. This can significantly impact performance since React will scan the component tree to find consumers and rerender the provider.
```
vilchik-elena, line 5: What is `every run` in the first sentence?

rules/S6481/javascript/rule.adoc
```diff
@@ -1,29 +1,26 @@
...
 ----
 
-=== Compliant solution
+To avoid additional rerenders wrap the value in a `useMemo` hook. Use the `useCallback()` hook if the value is a function.
 
 [source,javascript]
 ----
 function Component() {
-  const foo = useMemo(() => ({foo: 'bar'}), []);
+  const foo = useMemo(() => ({foo: 'bar'}), []); { /* Compliant: value is cached by useMemo */ }
```
vilchik-elena, line 23: rename variable so that property and variable name are not the same

rules/S6481/javascript/rule.adoc
```diff
@@ -1,29 +1,26 @@
...
   return (
     <SomeContext.Provider value={foo}>
```
vilchik-elena, line 25: I would better have compliant comment here (while leaving `/* value is cached by useMemo */ ` where it's now)

rules/S6481/javascript/rule.adoc
vilchik-elena, line 23: why comment is in the curly braces?

rules/S6481/javascript/rule.adoc
```diff
@@ -1,29 +1,26 @@
...
 
-=== Noncompliant code example
+Because each Javascript object carries its identity, object or function expressions get a new identity every run. These should not be used as Context `value` because React will always consider they have changed. This can significantly impact performance since React will scan the component tree to find consumers and rerender the provider.
 
 [source,javascript]
 ----
 function Component() {
   return (
-    <SomeContext.Provider value={{foo: 'bar'}}> { /* value is an object expression */ }
+    <SomeContext.Provider value={{foo: 'bar'}}> { /* Noncompliant: value is an object expression */ }
```
vilchik-elena, line 11:
```suggestion
    <SomeContext.Provider value={{foo: 'bar'}}> { /* Noncompliant: value is an object literal */ }
```

vilchik-elena (changes requested):

rules/S6481/javascript/rule.adoc
victor-diez-sonarsource: every execution of the expression? don't know how to put it more clear

rules/S6481/javascript/rule.adoc
victor-diez-sonarsource: typo

vilchik-elena (approved):

vilchik-elena (approved):

PR #2257, opened 2023-06-21
Modify Rule S5719: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5719/python/rule.adoc
+++ b/rules/S5719/python/rule.adoc
@@ -2,17 +2,17 @@ This rule raises an issue when an instance or a class method does not have at le
 
 == Why is this an issue?
 
-Every instance method is expected to have at least one positional parameter. This parameter will reference the object instance on which the method is called. Calling an instance method which doesn't have at least one parameter will raise a ``++TypeError++``. By convention, this first parameter is usually named ``++self++``.
+Every instance method is expected to have at least one positional parameter. This parameter will reference the object instance on which the method is called. Calling an instance method which doesn't have at least one parameter will raise a `TypeError`. By convention, this first parameter is usually named `self`.
 
-Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named ``++cls++``. 
+Class methods, i.e. methods annotated with `@classmethod`, also require at least one parameter. The only differences is that they will receive the class itself instead of a class instance. By convention, this first parameter is usually named `cls`. 
 
 === Exceptions
 
-Static methods (methods annotated with the ``++@staticmethod++`` decorator) do no require any positional parameter and will not raise an issue.
+Static methods (methods annotated with the `@staticmethod` decorator) do not require any positional parameter. This rule will not raise an issue on them.
 
 == How to fix it
 
-Adding ``++self++`` as the first parameter of an instance method or ``++cls++`` as the first parameter of a class method will resolve the issue.
+Adding `self` as the first parameter of an instance method or `cls` as the first parameter of a class method will resolve the issue.
 
 === Code examples
 
@@ -49,7 +49,7 @@ class MyClass:
 
 === Pitfalls
 
-Note that ``++__new__++`` and ``++__init_subclass__++`` take a class as first argument even thought they are not decorated with ``++@classmethod++``.
+Note that `__new__` and `__init_subclass__` take a class as first argument even thought they are not decorated with `@classmethod`.
 
 == Resources
 
--- a/rules/S5719/python/rule.adoc
+++ b/rules/S5719/python/rule.adoc
@@ -49,7 +49,7 @@ class MyClass:
 
 === Pitfalls
 
-Note that `__new__` and `__init_subclass__` take a class as first argument even thought they are not decorated with `@classmethod`.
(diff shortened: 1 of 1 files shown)


guillaume-dequenne-sonarsource (commented):
Left some questions - unsure whether changes need to be made.
In SQ UI, the description is looking good though.

rules/S5719/python/rule.adoc
```diff
@@ -1,31 +1,32 @@
...
+
 == Why is this an issue?
 
 Every instance method is expected to have at least one positional parameter. This parameter will reference the object instance on which the method is called. Calling an instance method which doesn't have at least one parameter will raise a "TypeError". By convention, this first parameter is usually named "self".
 
+Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named "cls". 
 
-Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named "cls". Note that ``++__new__++`` and ``++__init_subclass__++`` take a class as first argument even thought they are not decorated with ``++@classmethod++``.
-
+== How can I fix it?
```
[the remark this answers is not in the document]
joke1196: For me the formatting here is ok. So each section under the correct tab.

rules/S5719/python/rule.adoc
```diff
@@ -34,17 +35,18 @@ class MyClass:
...
     def class_method(cls):
         print("class_method")
-
-
-    @staticmethod
-    def static_method():
-        print("static_method")
 ----
 
+=== Pitfalls
```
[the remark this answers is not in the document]
joke1196: Thank you. I added the exception with the static method.

rules/S5719/python/rule.adoc
```diff
@@ -2,13 +2,13 @@ This rule raises an issue when an instance or a class method does not have at le
 
 == Why is this an issue?
 
-Every instance method is expected to have at least one positional parameter. This parameter will reference the object instance on which the method is called. Calling an instance method which doesn't have at least one parameter will raise a "TypeError". By convention, this first parameter is usually named "self".
+Every instance method is expected to have at least one positional parameter. This parameter will reference the object instance on which the method is called. Calling an instance method which doesn't have at least one parameter will raise a ``++TypeError++``. By convention, this first parameter is usually named ``++self++``.
 
-Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named "cls". 
+Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named ``++cls++``.
```
guillaume-dequenne-sonarsource, line 7: I think `The only difference is that they will` (instead of `it`) is more consistent with the previous sentence.

rules/S5719/python/rule.adoc
```diff
@@ -2,13 +2,13 @@ This rule raises an issue when an instance or a class method does not have at le
...
+Every instance method is expected to have at least one positional parameter. This parameter will reference the object instance on which the method is called. Calling an instance method which doesn't have at least one parameter will raise a ``++TypeError++``. By convention, this first parameter is usually named ``++self++``.
 
-Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named "cls". 
+Class methods, i.e. methods annotated with ``++@classmethod++``, also require at least one parameter. The only differences is that it will receive the class itself instead of a class instance. By convention, this first parameter is usually named ``++cls++``. 
 
 === Exceptions
 
-Static methods (methods annotated with the ``++@staticmethod++`` decorator do no require any positional parameter and will not raise an issue.
+Static methods (methods annotated with the ``++@staticmethod++`` decorator) do no require any positional parameter and will not raise an issue.
```
guillaume-dequenne-sonarsource, line 11: Suggestion: `Static methods (methods annotated with the ``++@staticmethod++`` decorator) do no require any positional parameter. No issue will be raised on them.`
(Mostly because the alternative implies that it's the method which raises the issue)
Feel free to ignore, though. I'm not 100% convinced by what I'm suggesting to be honest.

guillaume-dequenne-sonarsource (approved):
Looks good, left minor suggestions.

rules/S5719/python/rule.adoc
joke1196: Yes, you are right, thanks!

rules/S5719/python/rule.adoc
joke1196: No, I think you are right this could be confusing. Have made a change to highlight the fact that this rule will not do anything on static methods

guillaume-dequenne-sonarsource (approved):
LGTM.
As we are leaving some rules in a single-tab format, I feel maybe this one would also be better in only one tab. Wdyt?

joke1196:
The only issue would be with the Pitfalls, as they should be under the How to fix it section. I think it is nice to have them separated and not only being part of the issue description. I found this makes it clearer on what could go wrong.

jean-jimbo-sonarsource (approved):
lgtm

PR #2259, opened 2023-06-21
Modify rule S6486: Adapt to LaYC

victor-diez-sonarsource:


rules/S6486/javascript/rule.adoc
```diff
@@ -1,21 +1,22 @@
...
-
-=== Noncompliant code example
+To optimize the rendering of React list components, a unique identifier (UID) is required in the `key` attribute for each list item. This UID lets React identify the item throughout its lifetime.  Using generated values like `Math.random()` or `Date.now()` is discouraged as their return value will differ between calls, causing the keys to not match up between renders, recreating the DOM. Also, this may cause bugs if values collide.
 
 [source,javascript]
 ----
 function generateButtons(props) {
   return props.buttons.map((button, index) => {
-    <Button key={Math.random()}>{button.number}</Button>
+    <Button key={Math.random()}>{button.number}</Button> <!-- Noncompliant: 'key' value will change, React will need to recreate the list DOM -->
```
vilchik-elena, line 9: is it appropriate to provide such example? it's a button, not list

vilchik-elena (changes requested):
LGTM, but example is weird

rules/S6486/javascript/rule.adoc
victor-diez-sonarsource: didn't change it as it was already like this. @gabriel-vivas-sonarsource can you confirm the example makes sense when the rule is about react lists?

vilchik-elena: Even if example is relevant, I think it's worth saying something like "The rule reports on React List and other components..."

gabriel-vivas-sonarsource: @vilchik-elena are you talking about the wording of the comment? I would suggest:

```diff
- <!-- Noncompliant: 'key' value will change, React will need to recreate the list DOM -->
+ <!-- Noncompliant: Since the 'key' will be different on each render, React will update the DOM unnecessarily -->
```

Regarding using `<Button>` for the example, I think it is not inaccurate (hiding behind double negation à la French).

It is true that we could come up with something more representative. Because one would rarely render many buttons in a list. Perhaps `<li>` or `<a ...>` would be more common.

victor-diez-sonarsource: on it

vilchik-elena (approved):
Just realized that as we have both code examples on one tab, we should use diff. Could you add it?

PR #2261, opened 2023-06-21
S6640: Set quickfix to infeasible

Tim-Pohlmann:
Modify rule S6640: Set quickfix field to infeasible


martin-strecker-sonarsource (approved):
LGTM

PR #2262, opened 2023-06-21
Modify rule S5714: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5714/python/rule.adoc
+++ b/rules/S5714/python/rule.adoc
@@ -4,6 +4,7 @@ This rule raises an issue when the expression used in an ``++except++`` block is
 
 The only two possible types for an ``++except++``'s expression are a class deriving from ``++BaseException++``, or a tuple composed of such classes.
 
+Trying to catch multiple exception in the same ``++except++`` with a boolean expression of exceptions may not work as intended.
 The result of a boolean expression of exceptions is a single exception class, thus using a boolean expression in an ``++except++`` block will result in catching only one kind of exception.
 
 *Note*: __In Python 2 it is possible to raise exception from and old-style class which do not derive from ``++BaseException++``.__
@@ -23,7 +24,7 @@ try:
 except ValueError or TypeError:  # Noncompliant
     print("Catching only ValueError")
 except ValueError and TypeError:  # Noncompliant
-    print("catching only TypeError")
+    print("Catching only TypeError")
 except (ValueError or TypeError) as exception:  # Noncompliant
     print("Catching only ValueError")
 
--- a/rules/S5714/python/rule.adoc
+++ b/rules/S5714/python/rule.adoc
@@ -7,11 +7,22 @@ The only two possible types for an ``++except++``'s expression are a class deriv
 Trying to catch multiple exception in the same ``++except++`` with a boolean expression of exceptions may not work as intended.
 The result of a boolean expression of exceptions is a single exception class, thus using a boolean expression in an ``++except++`` block will result in catching only one kind of exception.
 
+[source,python]
+----
+error = ValueError or TypeError  
+error is ValueError # True
+error is TypeError # False
+
+error = ValueError and TypeError  
+error is ValueError # False
+error is TypeError # True
+----
+
 *Note*: __In Python 2 it is possible to raise exception from and old-style class which do not derive from ``++BaseException++``.__
 
 == How to fix it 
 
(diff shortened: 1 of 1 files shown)


rules/S5714/python/rule.adoc
```diff
@@ -1,12 +1,10 @@
+This rule raises an issue when the expression used in an ``++except++`` block is a boolean expression of exceptions.
+
 == Why is this an issue?
 
 The only two possible types for an ``++except++``'s expression are a class deriving from ``++BaseException++``, or a tuple composed of such classes (or an old style class if you are using python 2, but this has been removed in python 3).
```
guillaume-dequenne-sonarsource, line 5: `python` should be capitalized here.
I'm not sure whether we should rephrase the Python 2 bit as well, knowing it's less and less common to use Python 2.

rules/S5714/python/rule.adoc
```diff
@@ -23,21 +21,21 @@ foo = ValueError or TypeError  # foo == ValueError
 foo = ValueError and TypeError  # foo == TypeError
 ----
 
-
-=== Compliant solution
+Instead use a tuple of the exceptions that should be caught in the ``++except++`` block.
```
guillaume-dequenne-sonarsource, line 24: We should have an `How to fix it` section here.

rules/S5714/python/rule.adoc
joke1196: I have moved it to a note as it it not as important in order to understand the rule.

rules/S5714/python/rule.adoc
```diff
@@ -1,12 +1,21 @@
+This rule raises an issue when the expression used in an ``++except++`` block is a boolean expression of exceptions.
+
 == Why is this an issue?
 
-The only two possible types for an ``++except++``'s expression are a class deriving from ``++BaseException++``, or a tuple composed of such classes (or an old style class if you are using python 2, but this has been removed in python 3).
+The only two possible types for an ``++except++``'s expression are a class deriving from ``++BaseException++``, or a tuple composed of such classes.
+
+Trying to catch multiple exception in the same ``++except++`` with a boolean expression of exceptions may not work as intended.
+The result of a boolean expression of exceptions is a single exception class, thus using a boolean expression in an ``++except++`` block will result in catching only one kind of exception.
```
guillaume-dequenne-sonarsource, line 8: Here, maybe a code snippet would make things clearer.

rules/S5714/python/rule.adoc
```diff
@@ -1,12 +1,21 @@
...
+
+*Note*: __In Python 2 it is possible to raise exception from and old-style class which do not derive from ``++BaseException++``.__
 
+== How to fix it 
 
-This rule raises an issue when the expression used in an ``++except++`` block is a boolean expression of exceptions. The result of such expression is a single exception class, which is valid but not what the developer intended.
+Instead of using a boolean expression in a ``++except++`` block, try using a tuple of the exceptions that should be caught.
```
guillaume-dequenne-sonarsource, line 14: I think phrasing the recommendation as `Make sure to...` rather than `Try...` sounds a bit stronger and confident. Which is probably what we want, given this is an obvious bug.
I'd also probably focus on what should be done rather than what should not be done (i.e not starting with the "instead" clause).

rules/S5714/python/rule.adoc
joke1196: I have added a small example, tell me if you think this is not enough

guillaume-dequenne-sonarsource (approved):
LGTM.

rules/S5714/python/rule.adoc
```diff
@@ -1,12 +1,32 @@
...
+error is ValueError # True
+error is TypeError # False
 
+error = ValueError and TypeError  
+error is ValueError # False
+error is TypeError # True
+----
 
-This rule raises an issue when the expression used in an ``++except++`` block is a boolean expression of exceptions. The result of such expression is a single exception class, which is valid but not what the developer intended.
+*Note*: __In Python 2 it is possible to raise exception from and old-style class which do not derive from ``++BaseException++``.__
```
jean-jimbo-sonarsource, line 21: Fix grammar:
*Note*: __In Python 2 it is possible to raise an exception from an old-style class that does not derive from ``++BaseException++``.__

jean-jimbo-sonarsource (changes requested):
added a grammar fix

jean-jimbo-sonarsource (approved):
LGTM

PR #2264, opened 2023-06-21
Modify rule S5724: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5724/python/rule.adoc
+++ b/rules/S5724/python/rule.adoc
@@ -31,19 +31,19 @@ class WithPropertyMethod(object):
     foo = property(get_foo, set_foo, del_foo, "'foo' property.")
 ----
 
-This code example allows for such programs:
+The following program illustrates how using the built-in operations will call the custom methods defined above. 
 
 ----
 with_decorator = WithPropertyDecorator()
-with_decorator.foo = 1
-some_var = with_decorator.foo
-del with_decorator.foo
+with_decorator.foo = 1 # the method defined under @foo.setter will be called.
+some_var = with_decorator.foo # the method defined under @foo.getter will be called.
+del with_decorator.foo # the method defined under @foo.deleter will be called.
 
 
 with_method = WithPropertyMethod()
-with_method.foo = 1
-some_var = with_method.foo
-del with_method.foo
+with_method.foo = 1 # the method set_foo will be called.
+some_var = with_method.foo # the method get_foo will be called.
+del with_method.foo # the method del_foo will be called.
 ----
 
 Defining a property this way allows for flexibility when refactoring the implementation of the getters, setters and deleters method, as all the accesses and modifications are done through the Python built-in operators (``++=++``,``++.++``) and keyword (``++del++``).
--- a/rules/S5724/python/rule.adoc
+++ b/rules/S5724/python/rule.adoc
@@ -53,7 +53,7 @@ Property getter, setter and deleter methods are called by the Python interpreter
 * Property getter and deleter methods only require a "self" argument.
 * Property setter methods require a "self" argument as well as a value.
 
-Adding any other parameters, or removing these mandatory parameters will throw a ``++TypeError++`` exception at runtime when trying to access, or modify the property.
+Adding any other parameters, or removing these mandatory parameters will throw a ``++TypeError++`` exception at runtime when trying to access or modify the property.
 
 == How to fix it
 
--- a/rules/S5724/python/rule.adoc
+++ b/rules/S5724/python/rule.adoc
@@ -2,7 +2,7 @@ This rule raises an issue when a property getter, setter or deleter does not hav
 
 == Why is this an issue?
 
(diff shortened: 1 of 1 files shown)


rules/S5724/python/rule.adoc
```diff
@@ -1,53 +1,46 @@
...
-This rule raises an issue when:
+* Property getter and deleter methods only require a "self" argument.
+* Property setter methods require "self" as well as a value.
 
-* too many parameters are defined in a property getter, setter or deleter method.
-* the value parameter is missing in a property setter method.
+Adding any other parameter, or removing these mandatory parameters will throw a TypeError exception at runtime.
 
 
-=== Noncompliant code example
```
guillaume-dequenne-sonarsource, line 17: I think we're missing the `How to fix it` section here.

rules/S5724/python/rule.adoc
```diff
@@ -1,53 +1,46 @@
+This rule raises an issue when a property getter, setter or deleter does not have the correct number of arguments.
+
 == Why is this an issue?
 
 Property getter, setter and deleter methods are called by the python interpreter with a specific number or arguments:
```
guillaume-dequenne-sonarsource, line 5: `python` should be capitalized here.
Probably we could have a small explanation on the `@property` and `x.setter` decorators and what they do.

I'm also unsure how conventional/Pythonic it is to have getters and setters being defined as `get_x`, `set_x`, etc... Maybe we can mention that it is a conventional way to define getters and setters and that the contract of these methods should therefore be predictable (I assume that no `TypeError` would be thrown if I were to create my custom setter that takes multiple values at once - but it might be a weird API).

rules/S5724/python/rule.adoc
```diff
@@ -1,53 +1,46 @@
...
-Adding any other parameter, or removing these mandatory parameters will make method calls fail.
-
-
-This rule raises an issue when:
+* Property getter and deleter methods only require a "self" argument.
+* Property setter methods require "self" as well as a value.
 
-* too many parameters are defined in a property getter, setter or deleter method.
-* the value parameter is missing in a property setter method.
+Adding any other parameter, or removing these mandatory parameters will throw a TypeError exception at runtime.
```
guillaume-dequenne-sonarsource, line 10: Referring to this as the `TypeError` may not be thrown for a custom-defined getter/setter.

rules/S5724/python/rule.adoc
joke1196: It's true if you don't use the `property` method or decorator you can use x.get_x with any number of parameters, just like any normal method. I think the rule makes sense really if we want to have a property accessor and mutator in order to use `x.x = 1` or `del x.x`.

guillaume-dequenne-sonarsource: Oh right, I missed the `foo = property(get_foo, set_foo, del_foo, "'foo' property.") ` of the `class B` example and was afraid the rule was simply reporting on method names...
My bad!

rules/S5724/python/rule.adoc
```diff
@@ -1,53 +1,100 @@
+This rule raises an issue when a property getter, setter or deleter does not have the correct number of arguments.
+
 == Why is this an issue?
 
-Property getter, setter and deleter methods are called by the python interpreter with a specific number or arguments:
+Creating property accessor and mutator is quite common in Object-Oriented Programming. Python provides two main ways of implementing getters, setters and deleters; either with the ``++@property++`` decorator, or with the ``++property++`` method.
```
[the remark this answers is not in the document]
joke1196: Yes good catch thank you

guillaume-dequenne-sonarsource (changes requested):
Looks good overall, I'm wondering if we can make the example a bit more explicit and the wording a bit better though.

rules/S5724/python/rule.adoc
```diff
@@ -1,53 +1,100 @@
...
+        del self._foo
 
-=== Noncompliant code example
+    foo = property(get_foo, set_foo, del_foo, "'foo' property.")
+----
 
-[source,python]
+This code example allows for such programs:
+
+----
```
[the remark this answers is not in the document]
joke1196: Yes good call. It make sense. I have added the comments not sure if it is understantable with the methods under a decorator.

rules/S5724/python/rule.adoc
```diff
@@ -1,53 +1,100 @@
...
 
+    def del_foo(self):
+        del self._foo
 
-=== Noncompliant code example
+    foo = property(get_foo, set_foo, del_foo, "'foo' property.")
+----
 
-[source,python]
+This code example allows for such programs:
```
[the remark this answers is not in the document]
joke1196: Ok so I tried something in the mean time without the `allow`. Hopefully it makes it easier to understand.

guillaume-dequenne-sonarsource (approved):
LGTM.
Left a minor comment on grammar.

jean-jimbo-sonarsource (approved):
Thorough.

PR #2269, opened 2023-06-21
Modify S5527: Learn-As-You-Code migration

loris-s-sonarsource:
## Review

- [x] text style and tone


rules/S5527/python/how-to-fix-it/std.adoc
```diff
@@ -0,0 +1,39 @@
+== How to fix it in Python Standard Library
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+Certificate validation is not enabled by default when
+`_create_unverified_context` or `_create_stdlib_context` is used. It is
```
hendrik-buchwald-sonarsource, line 8: Just a note, I wouldn't actually change this for now: those examples already confused me in another rule. Since the methods start with underscores, those seem to be functions that should not be called directly outside of the module.

rules/S5527/python/how-to-fix-it/std.adoc
```diff
@@ -0,0 +1,39 @@
...
+example.check_hostname = False # Noncompliant
+----
+
+==== Compliant solution
+
+[source,python,diff-id=1,diff-type=compliant]
+----
+import ssl
+
+example = ssl.create_default_context()
```
hendrik-buchwald-sonarsource, line 32: The text above says that `_create_default_https_context` should be called. Though, if `create_default_context` is actually secure that seems like the better approach to me since the method does not start with an underscore.

rules/S5527/rationale.adoc
```diff
@@ -0,0 +1,14 @@
...
+the role of hostname validation, combined with certificate validation, is to
+ensure that a system is indeed the one it claims to be, adding an extra layer
+of trust and security.
+
+When hostname validation is disabled, the client skips this critical check.
+This creates an opportunity for attackers to pose as a trusted entity and
+intercept, manipulate, or steal the data being transmitted.
+
+To do so, an attacker would create, buy, or steal a correct certificate
+authentifying `example.com`, serve it using a totally different hostname, and
```
hendrik-buchwald-sonarsource, line 12:
```suggestion
authentifying `example.com`, serve it using a different hostname, and
```

It could be totally different but it does not have to be totally different. It could also be slightly different. So I just wouldn't mention that part.

hendrik-buchwald-sonarsource, line 12: I don't think `authentifying` is a common word. I am not saying it is wrong (e.g. there is [this](https://en.wiktionary.org/wiki/authentifying)) but if you search for [it](https://www.google.com/search?channel=fs&client=ubuntu&q=%22authentifying%22),  you don't find many "serious" sites. So to keep the text easy, I would recommend a different word.

rules/S5527/rationale.adoc
```diff
@@ -0,0 +1,14 @@
...
+over the internet by encrypting the data sent between them. In this process,
+the role of hostname validation, combined with certificate validation, is to
+ensure that a system is indeed the one it claims to be, adding an extra layer
+of trust and security.
+
+When hostname validation is disabled, the client skips this critical check.
+This creates an opportunity for attackers to pose as a trusted entity and
+intercept, manipulate, or steal the data being transmitted.
+
+To do so, an attacker would create, buy, or steal a correct certificate
```
hendrik-buchwald-sonarsource, line 11:
```suggestion
To do so, an attacker would obtain a valid certificate
```

Buying a certificate would probably still include creating it. How about:

hendrik-buchwald-sonarsource (changes requested):
See the comments for details.

rules/S5527/java/how-to-fix-it/java-ee.adoc
```diff
@@ -0,0 +1,59 @@
...
+:cert_variable_safe_value: true
+
+include::../../common/fix/code-rationale-explicit.adoc[]
+
+==== Noncompliant code example
+
+[source,java,diff-id=1,diff-type=noncompliant]
+----
+import java.util.Properties;
+import javax.mail.PasswordAuthentication;
```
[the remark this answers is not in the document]
loris-s-sonarsource: Yes, I removed the last part to avoid hardcoding a password

rules/S5527/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,57 @@
+== How to fix it in Java SE
```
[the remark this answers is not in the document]
loris-s-sonarsource: I wondered the same, and actually javax is not necessarily equal to java ee (it means java extension). In that case, https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/HttpsURLConnection.html is actually noted as Java SE in the docs

rules/S5527/kotlin/how-to-fix-it/ok-http.adoc
```diff
@@ -0,0 +1,68 @@
...
+import javax.net.ssl.SSLSession
+import javax.net.ssl.HostnameVerifier
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.Response
+
+fun request() {
+    val builder = OkHttpClient.Builder()
+    builder.hostnameVerifier(object : HostnameVerifier {
+      override fun verify(hostname: String?, session: SSLSession?): Boolean { // Noncompliant
```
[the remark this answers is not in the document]
loris-s-sonarsource: Yes, wondered the same and it's actually "verify", the function name, that is highlighted in SQ

loris-s-sonarsource (commented):
acting upon the review

rules/S5527/common/fix/validation.adoc
```diff
@@ -0,0 +1,29 @@
...
+the internal network, with firewall protections that would set up a TLS proxy
+thanks to the DNS access, while itself setting up a TLS certificate to fully
+protect messages.
+
+In case the contacted host is located on a development machine, this quick
+solution is recommended:
+
+* Create a self-signed certificate for that machine.
+* Add this self-signed certificate to the system's trust store.
+* Set the correct host name in the `/etc/hosts` file.
```
hendrik-buchwald-sonarsource, line 29: I don't think the readers would understand what they are supposed to do here and it is probably not even necessary. Since you suggest using a self-signed certificate, they can just use localhost.

hendrik-buchwald-sonarsource (changes requested):
I think there is still a bit of work to do in the fix section.

rules/S5527/common/fix/validation.adoc
loris-s-sonarsource, line 29:
```suggestion
* If the hostname is not `localhost`, set the correct host name in the `/etc/hosts` file.
```

loris-s-sonarsource: I just changed it. Do you think it still make sense or would you still delete it ?

loris-s-sonarsource (commented):
post-review changes

rules/S5527/java/how-to-fix-it/apache-commons-email.adoc
```diff
@@ -0,0 +1,57 @@
...
+    email.setMsg(message);
+    email.setSmtpPort(465);
+    email.setAuthenticator(new DefaultAuthenticator(username, password));
+    email.setSSLOnConnect(true); // Noncompliant
+
+    email.send();
+}
+----
+
+==== Compliant code example
```
loris-s-sonarsource, line 32:
```suggestion
==== Compliant solution
```

rules/S5527/java/how-to-fix-it/java-ee.adoc
```diff
@@ -0,0 +1,55 @@
...
+    props.put("mail.smtp.socketFactory.port", "465");
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory"); // Noncompliant
+    props.put("mail.smtp.auth", "true");
+    props.put("mail.smtp.port", "465");
+    
+    return props;
+}
+----
+
+==== Compliant code example
```
loris-s-sonarsource, line 31:
```suggestion
==== Compliant solution
```

rules/S5527/java/how-to-fix-it/java-se.adoc
```diff
@@ -0,0 +1,57 @@
...
+      public boolean verify(String requestedHost, SSLSession remoteServerSession) {
+        return true;  // Noncompliant
+      }
+    });
+
+    return urlConnection.getInputStream();
+}
+----
+
+==== Compliant code example
```
loris-s-sonarsource, line 36:
```suggestion
==== Compliant solution
```

rules/S5527/kotlin/how-to-fix-it/ok-http.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+    Request request = new Request.Builder()
+            .url("https://example.com")
+            .build()
+
+    Response response = client.newCall(request).execute()
+}
+----
+
+==== Compliant code example
```
loris-s-sonarsource, line 40:
```suggestion
==== Compliant solution
```

rules/S5527/common/fix/validation.adoc
```diff
@@ -0,0 +1,22 @@
...
+match its identity. If the remote host is not under your control, consider replicating its
+service to a server whose certificate you can change yourself.
+
+
+In case the contacted host is located on a development machine, and if there
+is no other choice, try following this solution:
+
+* Create a self-signed certificate for that machine.
+* Add this self-signed certificate to the system's trust store.
+* If the hostname is not `localhost`, set the correct host name in the `/etc/hosts` file.
```
loris-s-sonarsource, line 22:
```suggestion
* If the hostname is not `localhost`, set the correct host name in the `/etc/hosts` file.

```

rules/S5527/common/fix/validation.adoc
```diff
@@ -0,0 +1,23 @@
...
```
loris-s-sonarsource, line 22:
```suggestion
* If the hostname is not `localhost`, add the hostname in the `/etc/hosts` file.
```

rules/S5527/common/fix/validation.adoc
```diff
@@ -0,0 +1,23 @@
+To fix the vulnerability of disabled hostname validation, it is strongly
+recommended to first re-enable the default validation.
+
+To avoid running into problems with invalid hostnames, consider the following
+sections.
```
loris-s-sonarsource, line 5:
```suggestion
recommended to first re-enable the default validation and fix the root cause: the validity of the certificate.
```

rules/S5527/common/fix/validation.adoc
```diff
@@ -0,0 +1,20 @@
...
+If a hostname validation failure prevents connecting to the target server, keep
+in mind that **one system's code should not work around another system's problems**,
+as this creates unnecessary dependencies and can lead to reliability issues. +
+Therefore, the first solution is to change the remote host's certificate to
+match its identity. If the remote host is not under your control, consider replicating its
+service to a server whose certificate you can change yourself.
+
+
+In case the contacted host is located on a development machine, and if there
+is no other choice, try following this solution:
```
loris-s-sonarsource, line 15:
```suggestion
as this creates unnecessary dependencies and can lead to reliability issues.

Therefore, the first solution is to change the remote host's certificate to
match its identity. If the remote host is not under your control, consider replicating its
service to a server whose certificate you can change yourself.

In case the contacted host is located on a development machine, and if there
is no other choice, try following this solution:
```

hendrik-buchwald-sonarsource (approved):
LGTM :+1:  The tests are failing though.

PR #2270, opened 2023-06-21
Make sure that includes are always surrounded by empty lines

frederic-tingaud-sonarsource:
When an include is not surrounded by empty lines, its content is inlined on the same line as the adjacent content. That can lead to broken tags and other display issues.
This PR fixes all such includes and introduces a validation step that forbids introducing the same problem again.


frederic-tingaud-sonarsource:
Force pushed a version with one commit that introduces the script and one commit that fixes the files.

ci/validate_asciidoc.sh
```diff
@@ -57,6 +57,19 @@ do
     fi
   else
     #validate asciidoc
+
+    # Make sure include:: clauses are always more than one line away from the previous content
+    find "$dir" -name "*.adoc" -execdir sh -c 'grep -Pzl "\S\ninclude::" $1  | xargs -r -I@ realpath "$PWD/@"' shell {} \; > stuck
```
necto, line 62: A non-separated line with a trailing whitespace will escape this check
```
space is hereV
=== My header 
include::description.adoc[]
```
```suggestion
    find "$dir" -name "*.adoc" -execdir sh -c 'grep -Pzl "\S[ \t]*\ninclude::" $1  | xargs -r -I@ realpath "$PWD/@"' shell {} \; > stuck
```

ci/validate_asciidoc.sh
```diff
@@ -57,6 +57,19 @@ do
     fi
   else
     #validate asciidoc
+
+    # Make sure include:: clauses are always more than one line away from the previous content
```
necto, line 61: I would comment each of the two following lines:
# Includes that are stuck to the line above

# includes that are stuck to the line below

necto (approved):

PR #2273, opened 2023-06-22
Modify rule S5607: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5607/python/rule.adoc
+++ b/rules/S5607/python/rule.adoc
@@ -1,9 +1,11 @@
+This rule raises an issue when an operator is used on incompatible types. Types are considered incompatible if no built-in operations between those types exist and none of the operands has implemented the corresponding special methods.
+
 == Why is this an issue?
 
-:link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special%20methods#emulating-numeric-types
-:link-with-uscores2: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special%20methods#object.__lt__
+:link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?#emulating-numeric-types
+:link-with-uscores2: https://docs.python.org/3/reference/datamodel.html?#object.__lt__
 
-Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations. It is for example possible to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++`` and such an operation will raise a ``++TypeError++``.
+Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations, for example, to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++``, such an operation will raise a ``++TypeError++``.
 
 
 It is possible to define how an operator will behave with a custom class by defining the corresponding special method. See python documentation for a complete list of operators and their methods: {link-with-uscores1}[arithmetic and bitwise operators], {link-with-uscores2}[comparison operators].
@@ -12,10 +14,13 @@ It is possible to define how an operator will behave with a custom class by defi
 For symmetrical binary operators you need to define two methods so that the order of operands doesn't matter, ex: ``++__add__++`` and ``++__radd__++``.
 
 
-This rule raises an issue when an operator is used on incompatible types. Types are considered incompatible if no built-in operations between those types exist and none of the operands has implemented the corresponding special methods.
+== How to fix it
 
+Implementing the special methods for a specific operator will fix the issue. 
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 [source,python]
 ----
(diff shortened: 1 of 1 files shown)


rules/S5607/python/rule.adoc
```diff
@@ -1,9 +1,11 @@
+This rule raises an issue when an operator is used on incompatible types. Types are considered incompatible if no built-in operations between those types exist and none of the operands has implemented the corresponding special methods.
```
guillaume-dequenne-sonarsource, line 1: About this: `Types are considered incompatible if no built-in operations between those types exist`.
I feel this is a bit too strong. If two types are somehow compatible for `__eq__`, but not for `__add__`, they are considered incompatible in the context of `x + y` despite having built-in operations between them that exist.

I know this wording was already in the rule description and I'm not sure we can improve it while staying concise enough, but I think it's worth considering.

rules/S5607/python/rule.adoc
```diff
@@ -1,9 +1,11 @@
...
 :link-with-uscores2: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special%20methods#object.__lt__
 
-Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations. It is for example possible to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++`` and such an operation will raise a ``++TypeError++``.
+Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations, for example, to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++``, such an operation will raise a ``++TypeError++``.
 
 
 It is possible to define how an operator will behave with a custom class by defining the corresponding special method. See python documentation for a complete list of operators and their methods: {link-with-uscores1}[arithmetic and bitwise operators], {link-with-uscores2}[comparison operators].
```
guillaume-dequenne-sonarsource, line 11: `python` here should be capitalized.

rules/S5607/python/rule.adoc
```diff
@@ -1,9 +1,11 @@
...
+
 == Why is this an issue?
 
 :link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special%20methods#emulating-numeric-types
 :link-with-uscores2: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special%20methods#object.__lt__
 
-Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations. It is for example possible to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++`` and such an operation will raise a ``++TypeError++``.
+Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations, for example, to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++``, such an operation will raise a ``++TypeError++``.
```
guillaume-dequenne-sonarsource, line 8: `python` here should be capitalized.

guillaume-dequenne-sonarsource (commented):
Left some small comments.
More generally, I have the feeling we could enrich this rule description a bit, but not sure how to be honest.

rules/S5607/python/rule.adoc
joke1196: Tried something more context specific, it could be too heavy though

guillaume-dequenne-sonarsource: I think it's good! Thanks!

rules/S5607/python/rule.adoc
```diff
@@ -1,21 +1,32 @@
...
+Python provides a set of built-in operations, for example, to add two integers: ``++1 + 2++``. In this example the special method called is ``++__add__++`` on the type ``++int++``.
 
-Calling an operator in python is equivalent to calling a special method (except for the identity operator ``++is++``). Python provides a set of built-in operations. It is for example possible to add two integers: ``++1 + 2++``. It is however not possible to add a string and an integer: ``++1 + "2"++`` and such an operation will raise a ``++TypeError++``.
+It is however not possible to apply an operation on two types which:
 
+* Do not have a built-in operation defined.
+* Do not implement any of the corresponding operator's special method.
 
-It is possible to define how an operator will behave with a custom class by defining the corresponding special method. See python documentation for a complete list of operators and their methods: {link-with-uscores1}[arithmetic and bitwise operators], {link-with-uscores2}[comparison operators].
+Performing such an operation will raise a ``++TypeError++``.
```
joke1196, line 16: I think this make it more clear on why the issue is raised.

rules/S5607/python/rule.adoc
```diff
@@ -1,21 +1,32 @@
...
 
+Python allows developers to define how an operator will behave with a custom class by implementing the corresponding special method. 
+When defining such methods for symmetrical binary operators you need to define two methods so that the order of operands doesn't matter, ex: ``++__add__++`` and ``++__radd__++``.
 
-For symmetrical binary operators you need to define two methods so that the order of operands doesn't matter, ex: ``++__add__++`` and ``++__radd__++``.
+For a complete list of operators and their methods see python documentation: {link-with-uscores1}[arithmetic and bitwise operators], {link-with-uscores2}[comparison operators].
```
guillaume-dequenne-sonarsource, line 21: Here `python` should be capitalized.

guillaume-dequenne-sonarsource (approved):
There is still one `python` that is not capitalized, LGTM otherwise.

rules/S5607/python/rule.adoc
jean-jimbo-sonarsource: Is this a bit more clear? "...Python provides a set of built-in operators. For example, to add two integers: ```1 + 2```,  the built-in operator ```+``` is equivalent to the special method  ```__add__``` which is called on the type ```int```."

jean-jimbo-sonarsource (approved):
I think this will work for now, I have tried to find a less verbose way of explaining it but it wasn't an improvement without leaving out important descriptions of how this works in Python

PR #2287, opened 2023-06-26
Modify rule S5708: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5708/python/rule.adoc
+++ b/rules/S5708/python/rule.adoc
@@ -1,18 +1,20 @@
+This rule raises an issue when the expression used in an `except` statement is not a class deriving from `BaseException` nor a tuple of such classes.
+
 == Why is this an issue?
 
-In Python 3, attempting to catch in an ``++except++`` statement an object which does not derive from BaseException will raise a ``++TypeError++``. In Python 2 it is possible to raise old-style classes but this shouldn't be done anymore in order to be compatible with Python 3.
+In Python 3, attempting to catch in an `except` statement an object which does not derive from `BaseException` will raise a `TypeError`. 
 
+In order to catch multiple exceptions in an `except` statement, a `tuple` of exception classes should be provided. Trying to catch multiple exceptions with a `list` or a `set` will raise a `TypeError`.
 
-In order to catch multiple exceptions in an ``++except++`` statement, a tuple of exception classes should be provided.
+*Note*: __In Python 2 it is possible to raise old-style classes, which do not inherit from BaseException.__
 
+== How to fix it
 
-If you are about to create a custom Exception class, note that custom exceptions should inherit from ``++Exception++``, not ``++BaseException++``. ``++Exception++`` allows people to catch all exceptions except the ones explicitly asking the interpreter to stop, such as ``++KeyboardInterrupt++`` and https://docs.python.org/3/library/exceptions.html#GeneratorExit[``++GeneratorExit++``] which is not an error. See https://www.python.org/dev/peps/pep-0352/#exception-hierarchy-changes[PEP 352] for more information.
+Make sure the expression used in an `expect` statement is an exception which derives from `BaseException` or a tuple of such exceptions.
 
+=== Code examples
 
-This rule raises an issue when the expression used in an ``++except++`` statement is not a class deriving from ``++BaseException++`` nor a tuple of such classes.
-
-
-=== Noncompliant code example
+==== Noncompliant code example
 
 [source,python]
 ----
@@ -20,22 +22,22 @@ class CustomException:
     """An Invalid exception class."""
 
 try:
-    "a string" * 42
-except CustomException:  # Noncompliant
+    ...
+except CustomException:  # Noncompliant: this custom exception does not derive from Exception or BaseException.
     print("exception")
-except (None, list()):  # Noncompliant * 2
+except (None, list()):  # Noncompliant: None and list() are not valid exceptions deriving from BaseException.
     print("exception")
 
 try:
-    "a string" * 42
-except [TypeError, ValueError]:  # Noncompliant. Lists are not accepted.
+    ...
+except [TypeError, ValueError]:  # Noncompliant: lists are not accepted.
     print("exception")
-except {TypeError, ValueError}:  # Noncompliant. Sets are not accepted.
+except {TypeError, ValueError}:  # Noncompliant: sets are not accepted.
     print("exception")
 ----
 
 
-=== Compliant solution
+==== Compliant solution
 
 [source,python]
 ----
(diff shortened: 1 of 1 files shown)


rules/S5708/python/rule.adoc
```diff
@@ -1,58 +1,69 @@
...
 try:
-    "a string" * 42
+    ...
 except (MyError, TypeError):
     print("exception")
 ----
 
+=== Pitfalls
+
+When creating a custom Exception class, note that it should inherit from `Exception`, not `BaseException`. `Exception` allows people to catch all exceptions except the ones explicitly asking the interpreter to stop, such as https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt[`KeyboardInterrupt`] and https://docs.python.org/3/library/exceptions.html#GeneratorExit[`GeneratorExit`] which are not errors per se but program interruption. See https://www.python.org/dev/peps/pep-0352/#exception-hierarchy-changes[PEP 352] and https://docs.python.org/3/library/exceptions.html#exception-hierarchy[Exception Hierarchy] for more information.
```
guillaume-dequenne-sonarsource, line 55: This sentence was also in the description of S5632 (and I changed it slightly). I think it would be better to stay consistent here as well.

joke1196: Yes, thank you, I have reused the sentence from S5632.

rules/S5708/python/rule.adoc
```diff
@@ -1,58 +1,71 @@
...
 
+== Why is this an issue?
 
-In order to catch multiple exceptions in an ``++except++`` statement, a tuple of exception classes should be provided.
+In Python 3, attempting to catch in an `except` statement an object which does not derive from `BaseException` will raise a `TypeError`. 
 
+In order to catch multiple exceptions in an `except` statement, a `tuple` of exception classes should be provided. Trying to catch multiple exceptions with a `list` or a `set` will raise a `TypeError`.
 
-If you are about to create a custom Exception class, note that custom exceptions should inherit from ``++Exception++``, not ``++BaseException++``. ``++Exception++`` allows people to catch all exceptions except the ones explicitly asking the interpreter to stop, such as ``++KeyboardInterrupt++`` and https://docs.python.org/3/library/exceptions.html#GeneratorExit[``++GeneratorExit++``] which is not an error. See https://www.python.org/dev/peps/pep-0352/#exception-hierarchy-changes[PEP 352] for more information.
+*Note*: __In Python 2 it is possible to raise old-style classes, which do not inherit from BaseException.__
```
guillaume-dequenne-sonarsource, line 9: Should we say to raise, to catch, or both here (to raise and catch)?
Also I feel like the note about `BaseException` vs `Exception` might be more relevant here than the one about Python 2. Wdyt? (I think this is also how it was done for S5632).

rules/S5708/python/rule.adoc
```diff
@@ -1,58 +1,71 @@
-== Why is this an issue?
-
-In Python 3, attempting to catch in an ``++except++`` statement an object which does not derive from BaseException will raise a ``++TypeError++``. In Python 2 it is possible to raise old-style classes but this shouldn't be done anymore in order to be compatible with Python 3.
+This rule raises an issue when the expression used in an `except` statement is not a class deriving from `BaseException` nor a tuple of such classes.
```
guillaume-dequenne-sonarsource, line 1: I'm not sure about the grammar, but isn't neither/nor more correct? (vs not/nor)

rules/S5708/python/rule.adoc
```diff
@@ -1,58 +1,71 @@
...
 
-This rule raises an issue when the expression used in an ``++except++`` statement is not a class deriving from ``++BaseException++`` nor a tuple of such classes.
+Make sure the expression used in an `expect` statement is an exception which derives from `BaseException` or a tuple of such exceptions.
 
+=== Code examples
 
-=== Noncompliant code example
+==== Noncompliant code example
 
 [source,python]
```
guillaume-dequenne-sonarsource, line 19: I'm wondering if we could make the example better and use the diff view?
Maybe by clearly separating the case where we catch an invalid custom exception from the one where we catch a list of exception?
I feel we could drop the case where we have a tuple of random objects, as well as the one with the set. I don't feel they add so much value. Wdyt?

Also, if the resulting example is small, maybe we can consider keeping everything in one tab here as well.

guillaume-dequenne-sonarsource (changes requested):
Looks good overall, left some suggestion for further improvement.

rules/S5708/python/rule.adoc
joke1196: Yes, I think it fits nicely in one tab with the more concise example.

rules/S5708/python/rule.adoc
joke1196: Apparently, both are possible, but not/nor does not seems to be the default way :).

rules/S5708/python/rule.adoc
joke1196: Yes, I have reworked the layout a bit and moved up the `BaseException` vs `Exception`, higher up in the description, and left the Python 2 note lower in the page as it is less relevant.

rules/S5708/python/rule.adoc
```diff
@@ -6,55 +6,53 @@ In Python 3, attempting to catch in an `except` statement an object which does n
...
-except (None, list()):  # Noncompliant: None and list() are not valid exceptions deriving from BaseException.
+except CustomException(object):  # Noncompliant: this custom exception does not derive from Exception or BaseException.
     print("exception")
 
 try:
     ...
-except [TypeError, ValueError]:  # Noncompliant: lists are not accepted.
-    print("exception")
-except {TypeError, ValueError}:  # Noncompliant: sets are not accepted.
+except [TypeError, ValueError]:  # Noncompliant: lists or sets are not accepted.
```
guillaume-dequenne-sonarsource, line 32: Sorry, I realize I could have commented on this before, but I find the word "accepted" a bit odd in this context.
I also don't think we should mention set at all since we got rid of the example anyway (it seems too specific).

Maybe saying "Noncompliant: list of exceptions (only tuples are valid)", with or without the parenthesised bit, would be a bit clearer.

guillaume-dequenne-sonarsource (approved):
Left a final comment.
Looks good otherwise!

rules/S5708/python/rule.adoc
```diff
@@ -1,58 +1,69 @@
...
+If you are about to create a custom exception class, note that custom exceptions should inherit from ``++Exception++``, rather than ``++BaseException++``.
 
+``++BaseException++`` is the base class for all built-in exceptions in Python, including system-exiting exceptions like ``++SystemExit++`` or ``++KeyboardInterrupt++``, which are typically not meant to be caught. On the other hand, ``++Exception++`` is intended for exceptions that are expected to be caught, which is generally the case for user-defined exceptions. See https://www.python.org/dev/peps/pep-0352/#exception-hierarchy-changes[PEP 352] for more information. 
 
-This rule raises an issue when the expression used in an ``++except++`` statement is not a class deriving from ``++BaseException++`` nor a tuple of such classes.
+To fix this issue, make sure the expression used in an `expect` statement is an exception which derives from `BaseException`/`Exception` or a tuple of such exceptions.
```
jean-jimbo-sonarsource, line 13: Change expect to except

jean-jimbo-sonarsource (changes requested):
Found typo (expect vs except)

rules/S5708/python/rule.adoc
```diff
@@ -1,58 +1,69 @@
-== Why is this an issue?
-
-In Python 3, attempting to catch in an ``++except++`` statement an object which does not derive from BaseException will raise a ``++TypeError++``. In Python 2 it is possible to raise old-style classes but this shouldn't be done anymore in order to be compatible with Python 3.
+This rule raises an issue when the expression used in an `except` statement is neither a class deriving from `BaseException` nor a tuple of such classes.
 
+== Why is this an issue?
 
-In order to catch multiple exceptions in an ``++except++`` statement, a tuple of exception classes should be provided.
+In Python 3, attempting to catch in an `except` statement an object which does not derive from `BaseException` will raise a `TypeError`.
```
jean-jimbo-sonarsource, line 5: This sentence can be less complex when the words are reordered. I suggest "Attempting to catch an object that does not derive from `BaseException` in Python 3's `except` statement, will raise a TypeError" or similar.

jean-jimbo-sonarsource (changes requested):
Left a rewording comment

rules/S5708/python/rule.adoc
joke1196: I have tried something in between. Tell me what you think.

jean-jimbo-sonarsource (approved):

PR #2294, opened 2023-06-26
Modify rule S2259: LaYC format

sebastien-marichal:
## Review

- [ ] text style and tone


rules/S2259/csharp/exceptionCode1.adoc
```diff
@@ -0,0 +1,15 @@
+[source,csharp]
+----
+public static class Utils
+{
+    public static string ToUpper(this string value)
```
martin-strecker-sonarsource, line 5: "ToUpper" is a bad example. It is shadowed by the instance method ToUpper and is therefore never called. Use something easy like "RemoveVowels" instead :
```
return Regex.Replace(value, "[aeoui]*","", RegexOptions.IgnoreCase);
```
See [sharplab.io](https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQcAMACOSB0AIgJYCGA5lAPYDOwxAxjfgMJUAmApgIJSkA2ATxrEaAbnRZcBACqcAHsHwAlTuQj9SYAKLyADmE40RVKOPSSAzNIBsuBNhbYA3umzvsbjwG04AdhAAOSpgQI1+AEkAM2DQ8IAKPgBbTioo+IA3AQhOAEpc7ABdL3c4azw7PEwAfmxVJKoMzgA1KgB3Tn4aeOAAC1FpGuws/hzckpcJj2xiKOxM7M5sAF5l7Chw8bRp6ddtnYP/dfCJfYOAXymPI9VyBRVOPU0GTgXRzgAabAAib1JUiDEQoAKm+H2+YLqagUAHk9PRTMwIpQqIYWKQaHlTtNLmdPHjEAgJnsDqUkABOeLfXjATRQegMbAwl6kKDfB4NJqtDpdeL5bGkvCU+LxKrVXIbfj8XIcxotdqdbr8ia485AA)

rules/S2259/csharp/rule.adoc
```diff
@@ -1,80 +1,66 @@
...
+It is important to ensure `o` has a value by any means.
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
 public void Method()
 {
-    var o = new object();
-    Console.WriteLine(o.ToString());
+    var o = GetValue();
```
martin-strecker-sonarsource, line 34: How is o guaranteed to be not null in this case? Use `var o = new object();` instead.

rules/S2259/csharp/rule.adoc
```diff
@@ -1,80 +1,66 @@
...
-
-public static class Guard
+public void Method(object input)
 {
-    public static void NotNull<T>([ValidatedNotNull] this T value, string name) where T : class
-    {
-        if (value == null)
-            throw new ArgumentNullException(name);
-    }
+    Console.WriteLine(input.ToString()); // Noncompliant: 'input' might be null
```
martin-strecker-sonarsource, line 45: We only raise, if there is a code path indicating that `input` might be null:
```cs
if (input is null)
{
  // do something
}
// We learned from the if that input might be null
input.ToString(); // Noncompliant

File: rules/S2259/dotnet-why.adoc [hunk 4]
```diff
@@ -0,0 +1,19 @@
+== Why is this an issue?
+
+Accessing a {null-keyword-link}[{null-keyword}] value will always throw a https://learn.microsoft.com/en-us/dotnet/api/system.nullreferenceexception[NullReferenceException] most likely causing the application to terminate abruptly.
+
+Such termination might expose sensitive information that a malicious third party could exploit to, for instance, bypass security measures.
+
+=== Exceptions
+
+There is some scenario to which this rule won't raise:
```
martin-strecker-sonarsource, line 9:
```suggestion
In the following cases, the rule does not raise:
```

rules/S2259/dotnet-why.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+* calls to extension methods are not reported because they can still operate on `{null-keyword}` values.
++
+include::{exceptionCode1}[]
+* value has been previously validated by creating a `ValidatedNotNull` or `NotNull` attribute and used on method parameter
```
martin-strecker-sonarsource, line 14:
```suggestion
* value has been validated by a helper method annotated with a `ValidatedNotNull` attribute
```

martin-strecker-sonarsource: There are more options on how to suppress the warning:
These attributes are available in .Net core only but can be redefined in custom code and will be recognized by us:
* [NotNullAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute?view=net-7.0) same behavior as ValidatedNotNull
* NotNullWhen [see](https://github.com/SonarSource/sonar-dotnet/blob/8fc3b67ae9e293bf0ed77f68aa4de9fdfe3294ab/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Roslyn/NullPointerDereference.NetCore.cs#L92) same behavior as described in the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=net-7.0) 
* DoesNotReturn [see](https://github.com/SonarSource/sonar-dotnet/blob/8fc3b67ae9e293bf0ed77f68aa4de9fdfe3294ab/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Roslyn/NullPointerDereference.CSharp8.cs#L209) and [TerminatesProgram](https://github.com/SonarSource/sonar-dotnet/blob/8fc3b67ae9e293bf0ed77f68aa4de9fdfe3294ab/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Roslyn/NullPointerDereference.CSharp8.cs#L212C40-L212C40) Same behavior as described in the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnattribute?view=net-7.0) and [here](https://blog.jetbrains.com/dotnet/2012/08/15/contract-annotations-in-resharper-7/)
* [DoesNotReturnIfAttribute](https://github.com/SonarSource/sonar-dotnet/blob/8fc3b67ae9e293bf0ed77f68aa4de9fdfe3294ab/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Roslyn/NullPointerDereference.cs#L1687) behaves like described in the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnifattribute?view=net-7.0)

* Debug.Assert: [see](https://github.com/SonarSource/sonar-dotnet/blob/8fc3b67ae9e293bf0ed77f68aa4de9fdfe3294ab/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Roslyn/NullPointerDereference.cs#L375)

We also support the bang operator:

* Bang operator [see](https://github.com/SonarSource/sonar-dotnet/blob/8fc3b67ae9e293bf0ed77f68aa4de9fdfe3294ab/analyzers/tests/SonarAnalyzer.UnitTest/TestCases/SymbolicExecution/Roslyn/NullPointerDereference.CSharp8.Nullable.cs#L20)

martin-strecker-sonarsource (changes requested):

rules/S2259/dotnet-why.adoc
martin-strecker-sonarsource: The test case comment is wrong. That is a TP as `o1` is known to be null at that point.

rules/S2259/why-dotnet.adoc
```diff
@@ -0,0 +1,33 @@
...
+Such termination might expose sensitive information that a malicious third party could exploit to, for instance, bypass security measures.
+
+=== Exceptions
+
+In the following cases, the rule does not raise:
+
+* calls to extension methods are not reported because they can still operate on `{null-keyword}` values.
++
+include::{exceptionCodeExtensionMethod}[]
+* code is unreachable
```
martin-strecker-sonarsource, line 14: I think this causes errors in the ascii doc validation as it requires a newline around the includes:
> ERROR: These adoc files contain an include that is stuck to other content.
This may result in broken tags and other display issues.
Make sure there is an empty line before and after each include

 Given that we have a lot of exceptions, I would suggest using headings instead of the bullet list to resolve this.

martin-strecker-sonarsource (changes requested):

rules/S2259/csharp/exception-code-extension-method.adoc
```diff
@@ -0,0 +1,24 @@
+[source,csharp]
+----
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
+
+public static class Program
+{
+    public static string? RemoveVowels(this string? value)
```
martin-strecker-sonarsource, line 9:
```suggestion
    public static string RemoveVowels(this string value)
```

rules/S2259/csharp/exception-code-analysis-attr.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+public static class Utils
+{
+    public static string Normalize(string value)
+    {
+        Guard.NotNull(value, nameof(value)); // Will throw if value is null
+        if (value == null)
+        {
+            return value.ToString(); // Compliant: this code is unreachable
+        }
```
martin-strecker-sonarsource, line 26: I find this if quite confusing. You have another example with unreachable code later. I don't see the need to repeat unreachable code here. The `return value.ToUpper();` is enough.
```suggestion
```

rules/S2259/csharp/exception-code-extension-method.adoc
```diff
@@ -0,0 +1,24 @@
+[source,csharp]
+----
+using System;
+using System.Diagnostics.CodeAnalysis;
```
martin-strecker-sonarsource, line 4:
```suggestion
```

rules/S2259/csharp/exception-code-extension-method.adoc
```diff
@@ -0,0 +1,24 @@
...
+        if (value == null)
+        {
+            return null;
+        }
+        return Regex.Replace(value, "[aeoui]*","", RegexOptions.IgnoreCase);
+    }
+
+    public static void Main()
+    {
+        Console.WriteLine("Atlantic Ocean".RemoveVowels());
```
martin-strecker-sonarsource, line 20:
```suggestion
```

rules/S2259/vbnet/exception-code-analysis-attr.adoc
```diff
@@ -0,0 +1,22 @@
...
+        If Value Is Nothing Then Throw New ArgumentNullException(Name)
+    End Sub
+End Module
+
+Public Module Utils
+    Public Function Normalize(Value As String) As String
+        CheckNotNull(Value, nameof(Value)) ' Will throw if 'Value' is Nothing
+        If Value Is Nothing Then
+            Return Value.ToString() ' Compliant: this code is unreachable
+        End If
```
martin-strecker-sonarsource, line 18:
```suggestion
```

rules/S2259/vbnet/exception-code-extension-method.adoc
```diff
@@ -0,0 +1,23 @@
...
+            Return Nothing
+        End If
+        Return Regex.Replace(Value, "[aeoui]*", "", RegexOptions.IgnoreCase)
+    End Function
+
+    Sub Main()
+        Dim StrValue As String = Nothing
+        Console.WriteLine(StrValue.RemoveVowels()) ' Compliant: 'RemoveVowels' is an extension method
+        StrValue = "Atlantic Ocean"
+        Console.WriteLine(StrValue.RemoveVowels())
```
martin-strecker-sonarsource, line 20:
```suggestion
```

rules/S2259/vbnet/rule.adoc
```diff
@@ -1,76 +1,19 @@
-== Why is this an issue?
+:null-keyword: Nothing
+:null-keyword-link: https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/nothing
+:exceptionCodeExtensionMethod: vbnet/exception-code-extension-method.adoc
+:exceptionCodeUnreachable: vbnet/exception-code-unreachable.adoc
+:exceptionCodeAnalysisAttr: vbnet/exception-code-analysis-attr.adoc
+:exceptionCodeRiderAttr: vbnet/exception-code-rider-attr.adoc
+:exceptionCodeDebugAssert: vbnet/exception-code-debug-assert.adoc
+:compliantCode: vbnet/compliant-code.adoc
+:noncompliantCode: vbnet/noncompliant-code.adoc
```
martin-strecker-sonarsource, line 9: You can get rid of all the variables by instead defining `:language: vbnet` (and `:language: csharp` respectively).

rules/S2259/csharp/exception-code-rider-attr.adoc
```diff
@@ -0,0 +1,17 @@
+[source,csharp]
+----
+using JetBrains.Annotations;
+
+public class Utils
+{
+    [TerminatesProgram]
+    public void TerminatesProgram() { }
```
martin-strecker-sonarsource, line 8: What do you think about adding an `Environment.FailFast("A catastrophic failure has occurred.");` call?

https://learn.microsoft.com/en-us/dotnet/api/system.environment.failfast?view=net-7.0

I'm not sure if it is a good idea but it would show what the method is supposed to do. (Same would apply to VB).

martin-strecker-sonarsource (changes requested):
I made some suggestions to make the examples a bit more concise. Feel free to apply the changes.

rules/S2259/csharp/exception-code-rider-attr.adoc
sebastien-marichal: I think this is a good idea as you mention to exactly show what the method does.

rules/S2259/vbnet/rule.adoc
sebastien-marichal: Nice idea, although I don't know if it will pass the validation.
I tried we will see.

sebastien-marichal: It worked!

rules/S2259/csharp/rule.adoc
```diff
@@ -1,80 +1,27 @@
...
-----
-public void Method()
-{
-    object o = null;
-    Console.WriteLine(o.ToString()); // Noncompliant, always null
-}
-----
-
-=== Compliant solution
+Values marked with the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving[null forgiving operator]
```
martin-strecker-sonarsource, line 9:
```suggestion
Expressions marked with the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving[null forgiving operator]
```

rules/S2259/csharp/rule.adoc
```diff
@@ -1,80 +1,27 @@
...
-    {
-        Guard.NotNull(value, nameof(value));
-        if (value == null)
-        {
-            return value.ToString(); // Compliant, this code is not reachable
-        }
-        return value.ToUpper();
-    }
+    object o = null;
+    o!.ToString();    // Compliant: the null forgiving operator suppresses the nullable warning
```
martin-strecker-sonarsource, line 16:
```suggestion
    o!.ToString();    // Compliant: the null forgiving operator suppresses the null reference warning
```

rules/S2259/rule.adoc
```diff
@@ -1,33 +0,0 @@
-== Why is this an issue?
```
martin-strecker-sonarsource, line 1: Is this file really unused or do we need to copy it to the cfamilly folder?

rules/S2259/vbnet/noncompliant-code.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+The parameter `Input` might be `Nothing` as suggested by the `if` condition:
+
+[source,vbnet,diff-id=2,diff-type=noncompliant]
+----
+Public Sub Method(Input As Object)
+    If Input Is Nothing Then
+        ' ...
+    End If
+    Console.WriteLine(Input.ToString()) ' Noncompliant: 'Input' might be Nothing
```
martin-strecker-sonarsource, line 19:
```suggestion
    Console.WriteLine(Input.ToString) ' Noncompliant: 'Input' might be Nothing
```

rules/S2259/why-dotnet.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+==== Extensions Methods
+
+Calls to extension methods can still operate on `{null-keyword}` values.
+
+include::{language}/exception-code-extension-method.adoc[]
+
+==== Unreachable code
+
+Unreachable is by definition not executed, thus `{null-keyword}` values will never be accessed.
```
martin-strecker-sonarsource, line 19:
```suggestion
Unreachable code is not executed, thus `{null-keyword}` values will never be accessed.
```

rules/S2259/why-dotnet.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+include::{language}/exception-code-unreachable.adoc[]
+
+==== Validated value by analysis attributes [[null-analysis-attribute]]
+
+https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis[Nullable analysis attributes] provides more accurate semantic information when performing static analysis when it comes to detecting variable null-state. Thus, potential `{null-keyword}` values validated by one of the following attributes will not raise:
```
martin-strecker-sonarsource, line 25:
```suggestion
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis[Nullable analysis attributes] enable the developer to annotate methods with information about the null-state of its arguments. Thus, potential `{null-keyword}` values validated by one of the following attributes will not raise:
```

rules/S2259/why-dotnet.adoc
```diff
@@ -0,0 +1,56 @@
...
+* https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnattribute[DoesNotReturnAttribute]
+* https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnifattribute[DoesNotReturnIfAttribute]
+
+It is important to note those attributes are only available starting .NET Core 3. As a workaround, it is possible to define those attributes manually in a custom class:
+
+include::{language}/exception-code-analysis-attr.adoc[]
+
+==== Validated value by Debug.Assert
+
+When values are validated to not be `{null-keyword}` with https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert[Debug.Assert] it is safe to access it.
```
martin-strecker-sonarsource, line 38:
```suggestion
A value validated with https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert[Debug.Assert] to not be `{null-keyword}`  is safe to access.
```

rules/S2259/why-dotnet.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+include::{language}/exception-code-debug-assert.adoc[]
+
+==== Validated value by IDE-specific attributes
+
+Like with <<null-analysis-attribute,nullable analysis attributes>>, potential `{null-keyword}` values validated by one of the following IDE-specific attributes will not raise
+
+===== Visual Studio
+
+* https://learn.microsoft.com/en-us/dotnet/api/microsoft.validatednotnullattribute[ValidatedNotNullAttribute]
```
martin-strecker-sonarsource, line 48:
```suggestion
* https://learn.microsoft.com/en-us/dotnet/api/microsoft.validatednotnullattribute[ValidatedNotNullAttribute] (The attribute is interpreted the same as the https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute[NotNullAttribute])
```

rules/S2259/csharp/exception-code-analysis-attr.adoc
```diff
@@ -0,0 +1,26 @@
+[source,csharp]
+----
+using System;
+
+public sealed class NotNullAttribute : Attribute { }
```
martin-strecker-sonarsource, line 5:
```suggestion
public sealed class NotNullAttribute : Attribute { } // The alternative name 'ValidatedNotNullAttribute' is also supported
```

rules/S2259/why-dotnet.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+===== JetBrains Rider
+
+* https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#NotNullAttribute[NotNullAttribute]
+* https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#TerminatesProgramAttribute[TerminatesProgramAttribute]
++
```
martin-strecker-sonarsource, line 54:
```suggestion
```

rules/S2259/vbnet/exception-code-rider-attr.adoc
```diff
@@ -0,0 +1,18 @@
+[source,vbnet]
+----
+Imports System
+Imports JetBrains.Annotations
+
+Public Class Utils
+    <TerminatesProgram>
+    Public Sub TerminatesProgram()
```
martin-strecker-sonarsource, line 8:
```suggestion
    Public Sub TerminateProgram()
```

rules/S2259/vbnet/exception-code-rider-attr.adoc
```diff
@@ -0,0 +1,18 @@
...
+        Environment.FailFast("A catastrophic failure has occurred.")
+    End Sub
+
+    Public Sub Method()
+        Dim MyObject As Object = Nothing
+        TerminatesProgram()
```
martin-strecker-sonarsource, line 14:
```suggestion
        TerminateProgram()
```

martin-strecker-sonarsource (changes requested):

rules/S2259/rule.adoc
sebastien-marichal: It was not included at all, cfamily `rule.adoc` is almost the same but better (it includes the right cpp attribute for code snippets for instance).

rules/S2259/why-dotnet.adoc
sebastien-marichal: It is intended to have the Rider example at the same indentation level as the bullet point.
The current example is about the `TerminatesProgramAttribute`, that's why.
I am also OK to have it at the root level as a single example for the whole Rider section if you think that's OK.

martin-strecker-sonarsource (approved):
There are two unresolved comments above. Otherwise LGTM.

PR #2297, opened 2023-06-27
Modify S2734: LaYC format

guillaume-dequenne-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2734/python/rule.adoc
+++ b/rules/S2734/python/rule.adoc
@@ -1,3 +1,5 @@
+This rule raises an issue when the ``++__init__++`` method of a class contains a `return` or a `yield` statement.
+
 == Why is this an issue?
 
 By contract, every Python function returns something, even if it is the ``++None++`` value, which can be returned implicitly by omitting the ``++return++`` statement, or explicitly. 
--- a/rules/S2734/python/rule.adoc
+++ b/rules/S2734/python/rule.adoc
@@ -1,31 +1,39 @@
 == Why is this an issue?
 
-By contract, every Python function returns something, even if it's the ``++None++`` value, which can be returned implicitly by omitting the ``++return++`` statement, or explicitly. 
+By contract, every Python function returns something, even if it is the ``++None++`` value, which can be returned implicitly by omitting the ``++return++`` statement, or explicitly. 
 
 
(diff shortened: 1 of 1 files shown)


rules/S2734/python/rule.adoc
```diff
@@ -1,31 +1,39 @@
 == Why is this an issue?
 
-By contract, every Python function returns something, even if it's the ``++None++`` value, which can be returned implicitly by omitting the ``++return++`` statement, or explicitly. 
+By contract, every Python function returns something, even if it is the ``++None++`` value, which can be returned implicitly by omitting the ``++return++`` statement, or explicitly. 
 
 
-The ``++__init__++`` method is required to return ``++None++``. A ``++TypeError++`` will be raised if the ``++__init__++`` method either ``++yield++``s or ``++return++``s any expression other than ``++None++``. Returning some expression that evaluates to ``++None++`` will not raise an error, but is considered bad practice.
+The ``++__init__++`` method is required to return ``++None++``. A ``++TypeError++`` will be raised if the ``++__init__++`` method either yields or returns any expression other than ``++None++``. Returning an expression that evaluates to ``++None++`` will not raise an error, but is considered bad practice.
 
+To fix this issue, make sure that the ``++__init__++`` method does not return anything.
```
joke1196, line 8: I think this sentence is a bit confusing because of the sentence: `The __init__ method is required to return None.`  Maybe we could rephrase the previous sentence something like:  A ``++TypeError++`` will be raised if the ``++__init__++`` method either yields or returns any expression other than ``++None++``. While returning None will not raise an error, it is considered a bad practice. Removing the return statement altogether is preferred. 

Than  it makes more sense with the __init__ method should not have a return statement

joke1196 (changes requested):
Looks good! Not sure if changes are needed, but I left a small comment for some potential clarification.

rules/S2734/python/rule.adoc
```diff
@@ -1,31 +1,39 @@
 == Why is this an issue?
```
joke1196, line 1: I noticed the description is missing.

joke1196 (changes requested):
Looks good like this! I think it works well with this wording. I have just added a comment as I only notice now the description is missing, sorry about that.

rules/S2734/python/rule.adoc
guillaume-dequenne-sonarsource: Good catch, thanks!

joke1196 (approved):
LGTM!

joke1196 (approved):
LGTM!

jean-jimbo-sonarsource (approved):
LGTM

PR #2299, opened 2023-06-27
Modify rule S2688: LaYC format

gregory-paidis-sonarsource:


rules/S2688/csharp/metadata.json
```diff
@@ -1,5 +1,3 @@
 {
-  "defaultQualityProfiles": [
-    "Sonar way"
-  ]
+  "quickfix": "partial"
```
gregory-paidis-sonarsource, line 2: `a != NaN` and `a == NaN` can be replaced in-place with `double.IsNaN`, but if someone uses `>=, <=, <, >` with `double.NaN`, it is not clear what they are trying to do. Therefore, I chose partial.

Tim-Pohlmann: That makes sense. But why did you remove the `defaultQualityProfiles` property?

rules/S2688/csharp/rule.adoc
```diff
@@ -1,43 +1,57 @@
...
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 var a = double.NaN;
 
-if (a == double.NaN) // Noncompliant; always false
+if (a == double.NaN) // Noncompliant: always false
```
Tim-Pohlmann, line 19: Do we have a standardized format for noncompliant reasons? I now have seen `;`, `:` and `,`.

rules/S2688/csharp/rule.adoc
```diff
@@ -1,43 +1,57 @@
...
+When anything is compared with `NaN` using one of the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/comparison-operators[comparison operators] `>`, `>=`, `<`, `<=` or the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#equality-operator-[equality operator] `==`, the result will always be `false`. In contrast, when anything is compared with `NaN` using the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#inequality-operator-[inequality operator] `!=`, the result will always be `true`.
 
-[source,csharp]
+Instead, the best way to see whether a variable is equal to `NaN` is to use the https://learn.microsoft.com/en-us/dotnet/api/system.single.isnan[float.IsNaN] and https://learn.microsoft.com/en-us/dotnet/api/system.single.isnan[double.IsNaN] methods, which work as expected.
+
+== How to fix it
+
+=== Code examples
+
+==== Noncompliant code example
```
Tim-Pohlmann, line 13: Since it is the only section, can we omit `=== Code examples`? Or will validation fail if we do that?

Tim-Pohlmann (commented):
Some questions for clarification.

rules/S2688/csharp/metadata.json
gregory-paidis-sonarsource: Actually, partial does not mean what I thought it means. The documentation says:

```
some of the issues produced by the rule propose a quick fix, but not all.
```

So it suposes we already have a partially-working quickfix. I will change it to `targeted`.

The `defaultQualityProfiles` is SonarWay in the generic metadata.json, so I think it is redundant to also have it here.

rules/S2688/csharp/rule.adoc
gregory-paidis-sonarsource: I am not sure if it is standarized at LT-level, but during the previous LAYC sprint we decided to use:

```csharp
// Noncompliant: blabla
```

rules/S2688/csharp/rule.adoc
gregory-paidis-sonarsource: Validation fails. 

From what I know, `Code examples` is stripped from the SQ/SC anyways 🤷

Tim-Pohlmann (approved):
LGTM!

rules/S2688/csharp/rule.adoc
```diff
@@ -1,43 +1,57 @@
...
 
-include::../description.adoc[]
+https://learn.microsoft.com/en-us/dotnet/api/system.double.nan[double.NaN] and https://learn.microsoft.com/en-us/dotnet/api/system.single.nan[float.NaN] are not equal to anything, not even themselves. 
 
-=== Noncompliant code example
+When anything is compared with `NaN` using one of the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/comparison-operators[comparison operators] `>`, `>=`, `<`, `<=` or the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#equality-operator-[equality operator] `==`, the result will always be `false`. In contrast, when anything is compared with `NaN` using the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#inequality-operator-[inequality operator] `!=`, the result will always be `true`.
 
-[source,csharp]
+Instead, the best way to see whether a variable is equal to `NaN` is to use the https://learn.microsoft.com/en-us/dotnet/api/system.single.isnan[float.IsNaN] and https://learn.microsoft.com/en-us/dotnet/api/system.single.isnan[double.IsNaN] methods, which work as expected.
```
Tim-Pohlmann, line 7: I just noticed that both links are the same because they both inherit from `Single.IsNan`. @gregory-paidis-sonarsource

PR #2304, opened 2023-06-27
Modify rule S3862: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3862/python/rule.adoc
+++ b/rules/S3862/python/rule.adoc
@@ -107,7 +107,7 @@ def generator():
     yield from my_iterable
 ----
 
-Make sure to use the async keyword when iterating over async generators.
+Make sure to use the `async` keyword when iterating over async generators.
 
 [source,python,diff-id=2,diff-type=compliant]
 ----
--- a/rules/S3862/python/rule.adoc
+++ b/rules/S3862/python/rule.adoc
@@ -1,47 +1,53 @@
+This rule raises an issue when a non iterable object is used in a `for-in` loop, in a `yield from` or when it is unpacked.
+
 == Why is this an issue?
 
-For-in loops, https://docs.python.org/3/whatsnew/3.3.html#pep-380-syntax-for-delegating-to-a-subgenerator[``++yield from++``] and iterable unpacking only work with https://docs.python.org/3/glossary.html#term-iterable[iterable objects]. In order to be iterable, an object should have either an ``++__iter__++`` method or a ``++__getitem__++`` method implementing the https://docs.python.org/3/glossary.html#term-sequence[Sequence] semantic.
+`for-in` loops, https://docs.python.org/3/whatsnew/3.3.html#pep-380-syntax-for-delegating-to-a-subgenerator[`yield from`] and iterable unpacking only work with https://docs.python.org/3/glossary.html#term-iterable[iterable objects]. In order to be iterable, an object should have either an ``++__iter__++`` method or a ``++__getitem__++`` method implementing the https://docs.python.org/3/glossary.html#term-sequence[Sequence] semantic.
 
 
 Note also that iterating over an https://docs.python.org/3/glossary.html#term-asynchronous-iterable[asynchronous iterable], i.e. an object having the ``++__aiter__++`` method, requires the use of https://docs.python.org/3/reference/compound_stmts.html#the-async-for-statement[``++async for ... in++``] instead of ``++for ... in++``.
 
+== How to fix it
 
-This rule raises an issue when a non iterable object is used in a ``++for-in++`` loop, in a ``++yield from++`` or when it is unpacked.
+Make sure your object is an iterable when using it in `for-in` loops,`yield from` and unpacking statements, by implementing ``++__iter__++`` or ``++__getitem__++``. To iterate over an asynchronous iterable, make sure to use the `async` keyword, i.e `async for ... in`.
 
-=== Noncompliant code example
+=== Code examples
 
-[source,python]
+==== Noncompliant code example
+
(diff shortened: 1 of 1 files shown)


rules/S3862/python/rule.adoc
```diff
@@ -1,47 +1,89 @@
+This rule raises an issue when a non iterable object is used in a `for-in` loop, in a `yield from` or when it is unpacked.
+
 == Why is this an issue?
 
-For-in loops, https://docs.python.org/3/whatsnew/3.3.html#pep-380-syntax-for-delegating-to-a-subgenerator[``++yield from++``] and iterable unpacking only work with https://docs.python.org/3/glossary.html#term-iterable[iterable objects]. In order to be iterable, an object should have either an ``++__iter__++`` method or a ``++__getitem__++`` method implementing the https://docs.python.org/3/glossary.html#term-sequence[Sequence] semantic.
+`for-in` loops, https://docs.python.org/3/whatsnew/3.3.html#pep-380-syntax-for-delegating-to-a-subgenerator[`yield from`] and iterable unpacking only work with https://docs.python.org/3/glossary.html#term-iterable[iterable objects]. In order to be iterable, an object should have either an ``++__iter__++`` method or a ``++__getitem__++`` method implementing the https://docs.python.org/3/glossary.html#term-sequence[Sequence] semantic.
```
guillaume-dequenne-sonarsource, line 5: Should we replace "semantic" with "protocol" to be a bit more precise?

rules/S3862/python/rule.adoc
```diff
@@ -1,47 +1,89 @@
...
+        self._values = [1,2,3,4]
+
+    def __iter__(self):
+        return iter(self._values)
+----
+
+Here is a similar example with ``++__getitem__++``:
 
+----
+[source,python]
```
guillaume-dequenne-sonarsource, line 22: This should be above the separator line (it currently appears in the code snippet).

rules/S3862/python/rule.adoc
guillaume-dequenne-sonarsource, line 5: I have the feeling this section does a great job at explaining how an iterator actually works, but doesn't really answer the question "Why is this an issue?".

I believe a simple sentence explaining which kind of exception would be thrown if trying to iterate from a non iterable object would do the trick.

rules/S3862/python/rule.adoc
```diff
@@ -1,47 +1,89 @@
...
+----
+
+Note also that iterating over an https://docs.python.org/3/glossary.html#term-asynchronous-iterable[asynchronous iterable], i.e. an object having the ``++__aiter__++`` method, requires the use of https://docs.python.org/3/reference/compound_stmts.html#the-async-for-statement[``++async for ... in++``] instead of ``++for ... in++``.
+
+== How to fix it
+
+Make sure your object is an iterable when using it in `for-in` loops,`yield from` and unpacking statements, by implementing ``++__iter__++`` or ``++__getitem__++``. To iterate over an asynchronous iterable, make sure to use the `async` keyword, i.e `async for ... in`.
 
-empty = Empty()
+=== Code examples
```
guillaume-dequenne-sonarsource, line 50: I have the feeling these examples do too much and are a bit noisy overall.
Maybe we could consider only keeping the `for ... in` example for the non-async case. And maybe we could have 2 different snippets clearly separated to tackle the `async` case.
Not sure what would be best.

guillaume-dequenne-sonarsource (changes requested):
I think the explanation how the iterator works is quite nice.
I left some comments for small improvements I think we could do.

rules/S3862/python/rule.adoc
joke1196: Yes, thank you!

rules/S3862/python/rule.adoc
joke1196: I think it is nice having an example of unpacking and yield as well. I will reduce the number of examples for unpacking, it would be cleaner. Tell me if it is still too noisy. I also tried a few different layouts. Not sure I can have multiple Noncompliant code examples sections so I just split the examples with a bit of text.

rules/S3862/python/rule.adoc
joke1196: Yes, you are right!  I also added a small sentence for the async iterator part.

jean-jimbo-sonarsource (approved):
LGTM

rules/S3862/python/rule.adoc
```diff
@@ -1,47 +1,92 @@
...
+my_iterator = IterExample()
+for i in my_iterator:
+    print(i) # Prints 1,2,3,4
+
+my_iterator = GetItemExample()
+for i in my_iterator:
+    print(i) # Prints 1,2,3,4
+----
+
+Note also that iterating over an https://docs.python.org/3/glossary.html#term-asynchronous-iterable[asynchronous iterable], i.e. an object having the ``++__aiter__++`` method, requires the use of https://docs.python.org/3/reference/compound_stmts.html#the-async-for-statement[``++async for ... in++``] instead of ``++for ... in++``. Failing to provide the `async` keyword will result in a `TypeError`, stating the object is not iterable.
```
guillaume-dequenne-sonarsource, line 46: I feel like the comma is unneeded in the sentence "Failing to provide the `async` keyword will result in a `TypeError` stating the object is not iterable."

guillaume-dequenne-sonarsource (approved):
I left one last suggestion regarding grammar.

I still think the Noncompliant/Compliant code examples are a bit noisy, but I don't see an obvious way to improve it without losing information.
So I think it's good as it is. LGTM

PR #2306, opened 2023-06-27
Modify rule S6478: Adapt to LaYC

victor-diez-sonarsource:


rules/S6478/javascript/rule.adoc
```diff
@@ -1,38 +1,32 @@
 == Why is this an issue?
 
-Nested components are defined in an enclosing one and can be simple functions or arrow expressions. React recreates them systematically during the render pass because it doesn't know they haven't changed. This hurts performance as the components are recreated too many times. However, this can also hide surprising bugs where the state of the nested components is lost between renders. Trying to use `useCallback` hooks for child components is also discouraged. You should actually refactor this code to define a component on its own, passing props if needed.
+React components should not be nested, as their state will be lost on each re-render of their parent component, possibly introducing bugs. This will also impact performance as child components will be recreated unnecessarily.
```
yassin-kammoun-sonarsource, line 3:
```suggestion
React components automatically re-render whenever there is a change in their state or props. This is not true for nested React components, as their state will be lost on each re-render of their parent component. This can cause unexpected behaviors. This will also impact performance as child components will be recreated unnecessarily.
```

yassin-kammoun-sonarsource: As discussed, you can ignore this suggestion.

rules/S6478/javascript/rule.adoc
```diff
@@ -1,38 +1,32 @@
...
+
+If the goal is to have the state reset, use a `key` instead.
```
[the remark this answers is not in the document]
victor-diez-sonarsource: the point i want to make is that nesting components will reset state, so you should not do that. if that is in fact the goal, use a 'key' field (instead of nesting the component)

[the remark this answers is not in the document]
yassin-kammoun-sonarsource: 👍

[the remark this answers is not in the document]
victor-diez-sonarsource: I expanded the sentence to make it clearer

rules/S6478/javascript/rule.adoc
```diff
@@ -85,13 +77,31 @@ class Component extends React.Component {
...
+    <div>
+      <OtherComponent renderFooter={() => <div />} />
+    </div>
+  );
+}
+----
+
 == Resources
 
 * https://reactjs.org/docs/reconciliation.html#elements-of-different-types[Elements Of Different Types] - React documentation
```
yassin-kammoun-sonarsource, line 106: The hyphen here is outside of the square brackets.

yassin-kammoun-sonarsource (approved):
LGTM. I left a last comment about a tiny hiphen thing.

rules/S6478/javascript/rule.adoc
victor-diez-sonarsource: ouch, good catch!

PR #2307, opened 2023-06-27
[APPSEC-801] Migrate S4433 to LaYC format.

gaetan-ferry-sonarsource:
## Review

- [x] text style and tone


gaetan-ferry-sonarsource:
That was on purpose. Considering this section is optional and that the explanation would probably look like "Here the code does not use anonymous authentication" I chose not to add it. I open to be challenged here.

rules/S4433/common/impact.adoc
```diff
@@ -0,0 +1,30 @@
...
+In such a case, all users configured in the directory might see their identity
+and privileges taken over.
+
+==== Sensitive information leak
+
+If the attacker gets read-only access to the directory, he will be able to read
+the data it stores. That data might include security-sensitive pieces of
+information.
+
+Typically, an attacker might get access to user accounts lists that he can use
```
hendrik-buchwald-sonarsource, line 22:
```suggestion
Typically, an attacker might get access to user account lists that he can use
```

rules/S4433/common/impact.adoc
```diff
@@ -0,0 +1,30 @@
...
+in further intrusion steps. For example, the attacker could use such lists to
+perform password spraying, or related attacks, on all systems that rely on the
+affected directory as an identity provider.
+
+If the directory contains some Personally Identifiable Information, an attacker
+accessing it might represent a violation of regulatory requirements in some
+countries. For example, this kind of security event would go against the
+European RGPD law.
```
hendrik-buchwald-sonarsource, line 30:
```suggestion
European GDPR law.
```

RGPD is the French name for it.

hendrik-buchwald-sonarsource (changes requested):
Okay, good point. The fix is so simple in this case ("use a password") that it does not need a text about that.

All in all it looks very nice already. I left some comments. Also, can you remove the "Compliant" comments from the samples? It was done sometimes in old rules but we are not supposed to have compliant comments anymore.

rules/S4433/common/impact.adoc
gaetan-ferry-sonarsource: oopsy, french guy spotted

hendrik-buchwald-sonarsource (approved):
LGTM!

PR #2310, opened 2023-06-28
Modify rule S1764: LaYC format

mary-georgiou-sonarsource:
## Review

- [ ] text style and tone


rules/S1764/csharp/metadata.json
```diff
@@ -1,3 +1,5 @@
 {
-  "scope": "Main"
+  "title": "Identical expressions should not be used on both sides of operators",
+  "scope": "Main",
+  "quickfix": "targeted"
```
gregory-paidis-sonarsource, line 4: Why do you think this is targeted? What would you do for redundant if conditions?
We could go from this:

```csharp
if (a == a)
{
 DoWork();
}
```

to this:

```csharp
DoWork();
```

...but I think that this would not be correct. Usually, these weird always-true/always-false conditions should not be deleted, but fixed. Giving someone the ability to auto-fix it might lead to more mistakes.

About cases like:

```csharp
var x = 5/5; // always 0 :)
```

...I think we could indeed auto-fix it. Maybe `partial` would be better?

gregory-paidis-sonarsource (changes requested):
Left a few suggestions

rules/S1764/csharp/rule.adoc
```diff
@@ -1,8 +1,36 @@
...
+
+This rule ignores the following operators:
+
+* Multiplication (*)
+* Addition (+)
+* Assignment (=)
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#left-shift-operator-[Left-shift (<<)]
+* https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#right-shift-operator-[Right-shift (>>)]
+
+=== Code examples
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: There's no fix in these examples - it's mostly an illustration of what to avoid and what's the result.
Since also the fix is not really straight forward - it depends on the intention of the developer.

IMO they should stay in `Why this is an issue`.

rules/S1764/csharp/rule.adoc
```diff
@@ -1,8 +1,36 @@
...
+
+==== Noncompliant code example
 
 [source,csharp]
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: Same comment here, there's no compliant example. IMO these tags do not really have a meaning here.

rules/S1764/csharp/metadata.json
mary-georgiou-sonarsource: We discussed offline.
`Partial` indicates that there's an implementation - `targeted` means that there can be one (`partial` or `covered`.)

mary-georgiou-sonarsource: To expand - my idea matches @gregory-paidis-sonarsource  comment here is that there can be a partial quickfix for certain cases.

rules/S1764/csharp/rule.adoc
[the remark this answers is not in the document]
gregory-paidis-sonarsource: Sounds reasonable!

rules/S1764/csharp/metadata.json
gregory-paidis-sonarsource: Agreed, I misunderstood the semantics of `partial`. Thanks!

gregory-paidis-sonarsource (approved):
LGTM! I like the new two-level lists :)

PR #2311, opened 2023-06-28
Modify S1862: LaYC format

gregory-paidis-sonarsource:


rules/S1862/description.adoc
```diff
@@ -1,4 +1,3 @@
-A chain of ``++if++``/``++else if++`` statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to ``++true++``. 
-
+A chain of `if`/`else if` statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to `true`.
```
gregory-paidis-sonarsource, line 1: This is used in a billion other languages, so I just cleaned up the asciidoc format

rules/S1862/csharp/rule.adoc
```diff
@@ -1,10 +1,17 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+A chain of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-if-statement[if/else if] statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to `true`. 
+Therefore, duplicating a condition automatically leads to dead code inside the duplicated condition block. Usually, this is due to a copy/paste error.
```
mary-georgiou-sonarsource, line 4:
```suggestion
A chain of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-if-statement[if/else if] statements are evaluated from top to bottom. At most, only one branch will be executed: the first whose statement with the condition that evaluates to `true`. 
Therefore, duplicating a condition leads to unreachable code inside the duplicated condition block. Usually, this is due to a copy/paste error. 
```

rules/S1862/csharp/rule.adoc
```diff
@@ -1,10 +1,17 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+A chain of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-if-statement[if/else if] statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to `true`. 
+Therefore, duplicating a condition automatically leads to dead code inside the duplicated condition block. Usually, this is due to a copy/paste error. 
 
-=== Noncompliant code example
+At best, it's simply dead code and at worst, it's a bug that could lead to unexpected behavior.
```
mary-georgiou-sonarsource, line 6:
```suggestion
The result of such duplication can lead to unreachable code or even to unexpected behavior. 
```

rules/S1862/csharp/rule.adoc
```diff
@@ -14,15 +21,15 @@ else if (param == 2)
 {
   CloseWindow();
 }
-else if (param == 1) // Noncompliant
+else if (param == 1) // Noncompliant: condition has already been checked
 {
-  MoveWindowToTheBackground();
+  MoveWindowToTheBackground(); // dead code
```
mary-georgiou-sonarsource, line 26:
```suggestion
  MoveWindowToTheBackground(); // unreachable code
```

rules/S1862/vbnet/rule.adoc
```diff
@@ -1,26 +1,30 @@
...
+[source,vbnet,diff-id=1,diff-type=noncompliant]
 ----
 If param = 1 Then
   OpenWindow()
 ElseIf param = 2 Then
   CloseWindow()
-ElseIf param = 1 Then ' Noncompliant
-  MoveWindowToTheBackground()
+ElseIf param = 1 Then ' Noncompliant: condition has already been checked
+  MoveWindowToTheBackground() ' dead code
```
mary-georgiou-sonarsource, line 21: same changes here.

mary-georgiou-sonarsource (changes requested):

rules/S1862/csharp/rule.adoc
gregory-paidis-sonarsource: I don't agree with grammarly about the `is` -> `are`. I think the `(chain of statements)` as a unit is one thing, so it should be `is evaluated` and not `are evaluated`.

rules/S1862/csharp/rule.adoc
gregory-paidis-sonarsource: I think at some point we were told to use semi-realistic examples (e.g. not use `foo` and `bar` with no meaning that much).

This example seems straightforward to me, and I cannot think of a simpler one that makes sense. If you have any idea, please let me know :)

rules/S1862/vbnet/rule.adoc
```diff
@@ -1,26 +1,30 @@
 == Why is this an issue?
 
-A chain of ``++If++``/``++ElseIf++`` statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to ``++True++``.
+A chain of https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/if-then-else-statement[If/ElseIf] statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to `True`.
```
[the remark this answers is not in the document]
gregory-paidis-sonarsource: I did the changes in-line, I prefer simplicity+duplication from splitting the logic with parameters for <=2 files.

rules/S1862/csharp/rule.adoc
mary-georgiou-sonarsource: True - I aggree.

rules/S1862/csharp/rule.adoc
mary-georgiou-sonarsource: I find `MoveWindowToTheBackground` for example a bit out of place. But I'm not too strong on that - it still passes the message and is simple enought.

rules/S1862/description.adoc
mary-georgiou-sonarsource: Please revert this.
Replacing ``++method name++`` with `method name` will not work for all languages. Amelie told me for example, in the CFamily, they have attributes like [[nodiscard]] that need the format ``++ to be displayed correctly.

mary-georgiou-sonarsource (approved):

rules/S1862/vbnet/rule.adoc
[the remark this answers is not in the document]
mary-georgiou-sonarsource: It's an easy change IMO - and it will save us a lot of time in the future when we have to change one description and in practice, we'll change two.

In summary it will keep things easier in sync.

I leave it on you.

PR #2315, opened 2023-06-28
Modify rule S5553: CPP-4409 new cases + LayC

necto:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5553/cfamily/rule.adoc
+++ b/rules/S5553/cfamily/rule.adoc
@@ -342,6 +342,7 @@ std::cout <<c <<std::endl; // c refers to a the memory formerly occupied by temp
 === Documentation
 
 - https://en.cppreference.com/w/cpp/language/lifetime[Lifetime]
+- https://en.cppreference.com/w/cpp/language/range-for[Range-based for loop - Temporary range expression]
 - https://en.cppreference.com/w/cpp/header/string_view[std::string_view]
 
 === Articles & blog posts
--- a/rules/S5553/cfamily/rule.adoc
+++ b/rules/S5553/cfamily/rule.adoc
@@ -48,7 +48,7 @@ Accessing a dangling reference or pointer causes undefined behavior.
 This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.
 
 Practially this has wide range of effects.
-In many cases the access works and succeeds at writing or reading a value.
+In many cases the access works by accident and succeeds at writing or reading a value.
 However, it can start misbehaving at anytime.
 If compilation flags, compiler, platform, or runtime environment change,
 the same code can crash the application, corrupt memory, or leak a secret.
@@ -263,8 +263,22 @@ The interface of `Wrapper::getOr` from the example above is dangerous,
 since it is bound to create subtle bugs.
 Instead of fixing every call, it is better to make the interface less error-prone.
 
-For example, use `std::string_view` (or the equivalent if you can't use {cpp}17)
-to make this interface both safe and efficient:
+For example, always return the `std::string` by value.
+
+[source,cpp]
+----
+  std::string Wrapper::getOr(std::string backup) const {
+    return contents.empty() ? backup : contents;
+  }
+----
+
+There is no reference involved, so no risk of having it dangling.
+
+This safer solution, does involve some performance overhead of potentially unncessarily
+copying the `w.contents` string but in many cases it is acceptable.
+
+Alternatively, you can use `std::string_view` (or the equivalent if you can't use {cpp}17)
+to make this interface more efficient:
 
 [source,cpp]
 ----
@@ -278,7 +292,8 @@ and the temporary `std::string_view` stores a pointer to the string literal `"st
 When `getOr` returns `contents` it copies the `backup` value, which is fast,
 and the result is copied to a variable in the caller's stack frame.
 
-Note, that the following code would still create a dangling reference
+Note, that the interface is only slightly safer.
+The following code would still create a dangling pointer
 because it creates a temporary `std::string`
 and the `std::string_view` refers to this temporary object
 that gets destroyed at the end of the full expression:
--- a/rules/S5553/cfamily/rule.adoc
+++ b/rules/S5553/cfamily/rule.adoc
@@ -353,8 +353,6 @@ std::cout <<c <<std::endl; // c refers to the memory formerly occupied by a temp
 ifdef::env-github,rspecator-view[]
 '''
 
-TODO: check if this rule should still belong to "MISRA C++ 2008 recommended"
-
 == Comments And Links
 (visible only on this page)
 
--- a/rules/S5553/cfamily/rule.adoc
+++ b/rules/S5553/cfamily/rule.adoc
@@ -34,7 +34,7 @@ Dereferencing and even copying such dangling pointer or reference causes *undefi
 
 === Exceptions [[exceptions]]
 
-A reference-to-`const` bound to a temporary object directly extends lifetime of this object.
+A reference-to-`const` directly bound to a temporary object extends lifetime of this object.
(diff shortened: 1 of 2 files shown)


rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
...
+
+A reference-to-`const` bound to a temporary object directly extends lifetime of this object.
+The object enjoys the same lifetime as the reference it is bound to.
+
+However, lifetime extension is not transitive.
+If the definition of the temporary relies on another temporary,
+this second temporary will still be destroyed at the end of the full expression,
+creating an immediately dangling reference.
+
+=== What is the potential impact?
```
tomasz-kaminski-sonarsource, line 45: I was wondering about describing potential impact of the undefined behavior. This one sounds ok, but I wonder what value it adds.

rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
...
+
+Accessing a dangling reference or pointer causes undefined behavior.
+This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.
+
+Practially this has wide range of effects.
+In many cases the access works and succeeds at writing or reading a value.
```
tomasz-kaminski-sonarsource, line 51: I would rewrite that in a way that would suggest that in many cases in works by accident, but not simply working,

rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
-== Why is this an issue?
 
-Binding a temporary object to a reference to const usually lengthens the lifetime of the temporary: instead of being destroyed at the end of the full expression, the temporary will have the same lifetime as the reference itself.
+References and pointers to temporary objects become dangling after these objects are destroyed.
```
tomasz-kaminski-sonarsource, line 2: Can we actually put a text before Why this is an issue, I tought that there are simply tabs, and this text should be part of Why this is an issue.

rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
...
+    }
+};
+
+Stats getStats();
+
+int getTotalScore() {
+  int sum = 0;
+  auto stats = getStats();
+  // stats is not a temporary and it is alive throughout the loop
+  for (auto score : stats.getPlayerScores()) { // Compliant
```
tomasz-kaminski-sonarsource, line 153: I would mention if with an initializer here as most popural option.

rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
...
+----
+#include <string>
+#include <memory>
+struct MyStrRef {
+  std::string const* target;
+  MyStrRef(std::string const& target):target (&target) {}
+};
+
+void f(Wrapper const& w) {
+  auto s = std::make_unique<std::string>("string");
```
tomasz-kaminski-sonarsource, line 215: This example seem very questionable in isolation, as you can just declare string that you will later use. I think we should not place code with smells in complain solutions.
Maybe to present the point use shared_ptr and store shared_ptr<std::String const> in MyStrRef?

tomasz-kaminski-sonarsource (changes requested):
Looks good, a lot of informative content here. I have added few comments and general questions on shape.

rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
...
+----
+std::string_view c = w.getOr(std::string{"str"}); // Noncompliant
+std::cout <<c <<std::endl; // c refers to a the memory formerly occupied by temporary string
+----
+
+== Resources
+
+=== Documentation
+
+- https://en.cppreference.com/w/cpp/language/lifetime[Lifetime]
```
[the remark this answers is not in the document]
necto: Apparently we have used links to cppreference in S5502, S1912, S5798, S5800, S5417, S5524, S3229, S5403, S2815, S5000.
We only mentioned an abseil reference in S5995
I think the cpprefenrece is a reliable resource, well regarded by the C++ community.
abseil is maybe less so, but still on the trustworthy side.
What would be the problem in using these reources?

rules/S5553/cfamily/rule.adoc
necto: Perhaps, we should have a common shared section explaining why UB is bad, cited from many rules that detect UB.
However, I think it is worth a quick paragraph emphasizing the practical impact rather than theoretical "there be dragons" scare tale.
Otherwise, the users might thing, well it might a UB, but my grandfather did it, my father did it, and I did it for 10 years, so why would I care?

tomasz-kaminski-sonarsource: Ok, good point. I will also include that section in my rule.

rules/S5553/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Great, thanks for pointing this out.

rules/S5553/cfamily/rule.adoc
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: We are endorsing these sources as the company if we link them, and that has marketing consequences. This is not related to the quality of the blogs, but Abseil is clearly connected to Google. And I am not sure if the company, in general, wants to make such references. We may ask Amelie about it.

rules/S5553/cfamily/rule.adoc
necto: added "by accident". Did you have in mind something more than that?

rules/S5553/cfamily/rule.adoc
necto: I am not entirely sure what you suggested, is ac019745e close?

rules/S5553/cfamily/rule.adoc
necto: I gues you meant `for` with an initializer. Done in 6ce3edaef

rules/S5553/cfamily/rule.adoc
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Thanks for clarying with the docs.

rules/S5553/cfamily/rule.adoc
```diff
@@ -1,33 +1,310 @@
...
+
+void f(Wrapper const& w) {
+  std::string c = w.getOr("<empty>"); // Compliant, c is an independent copy
+  std::cout <<c <<std::endl;
+}
+----
+
+== Going the extra mile
+
+The interface of `Wrapper::getOr` from the example above is dangerous,
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: I am still concerned that going the extra mile presents solutions that still have the same issues. What would think about starting the paragraph, with `Note that the interface using `std::string_view` still mostly are prone to the same problems, however, they avoid the performance impact on std::string creation.

rules/S5553/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Yes, thanks. This looks like things that are actually used in code.

rules/S5553/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Yes, thanks.

rules/S5553/cfamily/rule.adoc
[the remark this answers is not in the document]
necto: added

tomasz-kaminski-sonarsource (approved):

tomasz-kaminski-sonarsource (approved):

PR #2316, opened 2023-06-28
Modify rule S2114: LaYC format

gregory-paidis-sonarsource:


rules/S2114/csharp/rule.adoc
```diff
@@ -1,6 +1,9 @@
 == Why is this an issue?
 
-Passing a collection as an argument to the collection's own method is either an error - some other argument was intended - or simply nonsensical code.
+Passing a collection as an argument to the collection's own method is a code defect. There is no reason to add or concatenate a list to itself, and it can lead to reduced readability, as well as unexpected side effects.
```
sebastien-marichal, line 3: I don't like the sentence `There is no reason...`, I find it condescending.

I feel like the description could be a bit more generic, it seems odd to me to talk about only 3-4 examples when in the example there are more cases.

```suggestion
Passing a collection as an argument to the collection's own method is a code defect. Doing so might either have unexpected side effects or always have the same result.
```

sebastien-marichal (changes requested):
I feel the why section could be improved.

sebastien-marichal (approved):
LGTM!

PR #2317, opened 2023-06-28
Modify rule S2319: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2319/python/rule.adoc
+++ b/rules/S2319/python/rule.adoc
@@ -2,20 +2,25 @@
 
 Python has no pre/post increment/decrement operator. For instance, ``x{plus}{plus}`` and ``++x--++`` will fail to parse. More importantly, ``{plus}{plus}x`` and ``++--x++`` will do nothing. To increment a number, simply write ``++x += 1++``.
 
+=== Code examples 
 
-=== Noncompliant code example
+==== Noncompliant code example
 
-[source,python]
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
-++x # Noncompliant
+++x # Noncompliant: pre and post increment operators do not exist.
+
+x-- # Noncompliant: pre and post decrement operators do not exist.
 ----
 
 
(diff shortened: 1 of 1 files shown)


rules/S2319/python/rule.adoc
```diff
@@ -1,21 +1,28 @@
...
+=== Code examples 
 
-=== Noncompliant code example
+==== Noncompliant code example
 
-[source,python]
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
-++x # Noncompliant
+++x # Noncompliant: pre and post increment operators do not exist.
```
guillaume-dequenne-sonarsource, line 13: I'd probably add "in Python." (I know it's obvious), just because the sentence seems a bit odd without it.

rules/S2319/python/rule.adoc
```diff
@@ -1,21 +1,28 @@
+This rule raises an issue when an increment or decrement operator is used.
```
guillaume-dequenne-sonarsource, line 1: I guess we could say "a non-existing increment or decrement operator" or mention pre/post explicitly, so that there cannot be a confusion with `+=`.

guillaume-dequenne-sonarsource (approved):
Left minor comments.
Also, I see that the legacy key `PreIncrementDecrement ` is replaced with `S2319` when running rule-api.
Not sure what we should do about that, but I think it is generally disruptive to change rule keys, so we should make sure what should be the correct behavior here before merging.

rules/S2319/python/rule.adoc
jean-jimbo-sonarsource: I like the idea of adding pre/post, "non-existing" isn't a typical enough phrase to use here.

jean-jimbo-sonarsource (changes requested):
Added some comments

jean-jimbo-sonarsource (approved):
LGTM

PR #2321, opened 2023-06-28
Modify rule S2711: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2711/python/rule.adoc
+++ b/rules/S2711/python/rule.adoc
@@ -12,12 +12,11 @@ If the goal is to break out of a loop, use `break` instead of `return`.
 
 [source,python,diff-id=1,diff-type=noncompliant]
 ----
-class MyClass: 
-    a = 1
-    while a < 3:
-        if a % 2 == 0:
-            return # Noncompliant: return is outside of a function
-        a += 1
+a = 1
+while a < 3:
+    if a % 2 == 0:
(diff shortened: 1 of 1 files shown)


rules/S2711/python/rule.adoc
```diff
@@ -1,18 +1,50 @@
...
+
+If the goal is to break out of a loop, use `break` instead of `return`.
+
+=== Code examples
+
+==== Noncompliant code example
 
+[source,python,diff-id=1,diff-type=noncompliant]
+----
+class MyClass:
```
guillaume-dequenne-sonarsource, line 15: I understand we have a class here to highlight the fact that we're not in a function, but having a loop here feels pretty weird as well. Wdyt?
Would it make sense to do away with the class and pretend this is module-level code?

guillaume-dequenne-sonarsource (approved):
I have the feeling the code example is weird, with a function inside a class.
Not requesting changes, but if you think this can be made a bit better, please don't hesitate.
Looks good otherwise.

rules/S2711/python/rule.adoc
joke1196: Yes, I think you are right it makes enough sense without the Class.

jean-jimbo-sonarsource (approved):
LGTM

PR #2323, opened 2023-06-29
Modify rule S3249: LaYC format 

mary-georgiou-sonarsource:


rules/S3249/rule.adoc
```diff
@@ -1,62 +1,93 @@
 == Why is this an issue?
 
-Making a ``++base++`` call in an overriding method is generally a good idea, but not in ``++GetHashCode++`` and ``++Equals++`` for classes that directly extend ``++object++`` because those methods are based on the object reference. Meaning that no two ``++objects++`` that use those ``++base++`` methods will ever be equal or have the same hash.
+Making a `base` call when overriding a method is generally a good idea, but not in the case of https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode[`GetHashCode`] and https://learn.microsoft.com/en-us/dotnet/api/system.object.equals[`Equals`] for classes that directly extend `Object`. These methods are based on the object reference, which means that there are no two objects that use those `base` methods and can ever be equal or have the same hash.
```
Tim-Pohlmann, line 3:
```suggestion
Making a `base` call when overriding a method is generally a good idea, but not in the case of https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode[`GetHashCode`] and https://learn.microsoft.com/en-us/dotnet/api/system.object.equals[`Equals`] for classes that directly extend `Object`. These methods are based on the object's reference, meaning that no two objects using those `base` methods can be equal or have the same hash.
```

rules/S3249/rule.adoc
```diff
@@ -1,62 +1,93 @@
...
-----
+== How to fix it
 
+=== Code examples
 
-=== Compliant solution
+==== Noncompliant code example
 
-[source,text]
+[source,csharp]
```
Tim-Pohlmann, line 15:
```suggestion
[source,csharp,diff-id=1,diff-type=noncompliant]
```

rules/S3249/rule.adoc
```diff
@@ -1,62 +1,93 @@
...
+
+    public override int GetHashCode() =>
+        x.GetHashCode() ^ base.GetHashCode(); // Noncompliant, base.GetHashCode returns a code based on the objects reference
 }
 ----
 
+==== Compliant solution
 
-=== Exceptions
+[source,csharp]
```
Tim-Pohlmann, line 40:
```suggestion
[source,csharp,diff-id=1,diff-type=compliant]
```

rules/S3249/rule.adoc
```diff
@@ -1,62 +1,93 @@
...
     {
       return true;
     }
     ...
   }
+
+  public override int GetHashCode() =>
+    x.GetHashCode()
 }
 ----
```
Tim-Pohlmann, line 84: I would move this to the `Exception` section. It is more relevant there. Also, chaining code blocks like this [is causing issues and should be avoided](https://sonarsource.slack.com/archives/C059FQ512EA/p1688030499161979).
Alternatively, you could include the guard statement in the `Noncompliant` code.

rules/S3249/rule.adoc
```diff
@@ -1,62 +1,93 @@
...
 
+
+== Resources
+
+
+=== Documentation
+
+* https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-7.0[Object.GetHashCode Method]
+* https://learn.microsoft.com/en-us/dotnet/api/system.object.equals[Object.Equals Method]
```
Tim-Pohlmann, line 93:
```suggestion
* Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-7.0[Object.GetHashCode Method]
* Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/api/system.object.equals[Object.Equals Method]
```

Tim-Pohlmann (changes requested):

rules/S3249/rule.adoc
mary-georgiou-sonarsource: I did not apply the `using`.
I think it's better to use simple perfect tense as it's something that is generally true and we don't try to say that is happening right now.

rules/S3249/rule.adoc
mary-georgiou-sonarsource: I did not add this because we have 2 compliant examples and I remember that it does not work.
@gregory-paidis-sonarsource  is this the case?

rules/S3249/rule.adoc
mary-georgiou-sonarsource: Same

rules/S3249/rule.adoc
mary-georgiou-sonarsource: Sounds good - I'll move it to the exception and then I'll apply the comments above too.

rules/S3249/rule.adoc
gregory-paidis-sonarsource: I am pretty sure it does not work.
[click here (internal)](https://discuss.sonarsource.com/t/support-for-multiple-compliant-code-snippets-in-the-rspec-code-diffs/14644)

rules/S3249/rule.adoc
Tim-Pohlmann: How about:
```
These methods are based on the object's reference, meaning that no two objects that use those `base` methods can be equal or have the same hash.
```

Tim-Pohlmann (approved):
LGTM! Before merging, look at the last suggestion I made in the unresolved comment thread and see if you like it.

PR #2325, opened 2023-06-29
Modify rule S5712: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5712/python/rule.adoc
+++ b/rules/S5712/python/rule.adoc
@@ -105,7 +105,7 @@ MyClass() + MyOtherClass()  # This returns 42
 
 === Pitfalls
 
-The ``++__length_hint__++`` special method also requires to return a `NotImplemented`. Its behaviour differs from the other methods, because when this special method returns `NotImplemented` the caller will return a default value. See https://peps.python.org/pep-0424/[PEP 424] for more information.
+The ``++__length_hint__++`` special method also requires to return a `NotImplemented`. Its behavior differs from the other methods, because when it returns `NotImplemented`, a default value will be returned instead. See https://peps.python.org/pep-0424/[PEP 424] for more information.
 
 == Resources
 
--- a/rules/S5712/python/rule.adoc
+++ b/rules/S5712/python/rule.adoc
@@ -1,61 +1,61 @@
-This rule raises an issue when some special methods raises `NotImplementedError` instead of returning `NotImplemented`.
+This rule raises an issue when a special methods raises `NotImplementedError` instead of returning `NotImplemented`.
 
 == Why is this an issue?
 
-In Python, special methods corresponding to numeric operators, rich comparison operators and the ``++__length_hint__++`` method should return ``++NotImplemented++`` when the operation is not supported. 
+In Python, special methods corresponding to numeric operators and rich comparison operators should return ``++NotImplemented++`` when the operation is not supported. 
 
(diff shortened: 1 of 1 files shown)


rules/S5712/python/rule.adoc
```diff
@@ -1,12 +1,12 @@
-== Why is this an issue?
-
-In Python, special methods corresponding to numeric operators, rich comparison operators and the ``++__length_hint__++`` method should return ``++NotImplemented++`` when the operation is not supported. These methods should not raise ``++NotImplementedError++`` as callers don't expect it and won't catch this exception.
+This rule raises an issue when some special methods raises `NotImplementedError` instead of returning `NotImplemented`.
```
guillaume-dequenne-sonarsource, line 1: This should probably read "when a special method raises" (or "when some special methods raise")

rules/S5712/python/rule.adoc
```diff
@@ -1,12 +1,12 @@
...
 
+== Why is this an issue?
 
-For example ``++A + B++`` is equivalent to calling ``++A.__add__(B)++``. If this binary operation is not supported by class A, ``++A.__add__(B)++`` should return ``++NotImplemented++``. The interpreter will then try the reverse operation, i.e. ``++B.__radd__(A)++``. This enables adding new operations by changing only one class instead of two.
+In Python, special methods corresponding to numeric operators, rich comparison operators and the ``++__length_hint__++`` method should return ``++NotImplemented++`` when the operation is not supported.
```
guillaume-dequenne-sonarsource, line 5: I'm not sure whether `__length_hint__ ` is the most obvious example to give. In fact, I didn't even know this existed.

rules/S5712/python/rule.adoc
```diff
@@ -58,9 +58,15 @@ This rule raises an issue when one of the following methods raises ``++NotImplem
...
+Make sure special methods return `NotImplemented` instead of raising a `NotImplementedError`.
 
-[source,python]
+=== Code examples
+
+==== Noncompliant code example
+
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
 class MyClass:
```
guillaume-dequenne-sonarsource, line 71: Next to the "# Noncompliant" (can't point to that specific line as it wasn't changed), we could have something like:
`# Noncompliant: the exception will be propagated`

guillaume-dequenne-sonarsource (changes requested):
Looks good overall.
Left a small comment on grammar and some minor suggestions.

rules/S5712/python/rule.adoc
joke1196: I think the singular is more appropriate as we will raise an issue for each special methods.

rules/S5712/python/rule.adoc
joke1196: I think this is more like a pitfall as __length_hint__ is the only non-numeric and non-comparison special method, also the behavior is different than the other special methods. I will move it to the appropriate section.

rules/S5712/python/rule.adoc
```diff
@@ -105,7 +105,7 @@ MyClass() + MyOtherClass()  # This returns 42
 
 === Pitfalls
 
-The ``++__length_hint__++`` special method also requires to return a `NotImplemented`. Its behaviour differs a bit from the other methods, because if this special method is not implemented a default value is returned. 
+The ``++__length_hint__++`` special method also requires to return a `NotImplemented`. Its behaviour differs from the other methods, because when this special method returns `NotImplemented` the caller will return a default value. See https://peps.python.org/pep-0424/[PEP 424] for more information.
```
guillaume-dequenne-sonarsource, line 108: I believe we should consistently use American English, so we should use "behavior".
I'd suggest rewriting:
"..., because when it returns `NotImplemented`, a default value will be returned. See https://peps.python.org/pep-0424/[PEP 424] for more information."

guillaume-dequenne-sonarsource (approved):
LGTM

jean-jimbo-sonarsource (approved):
LGTM

PR #2328, opened 2023-06-29
Modify rule S3005: LaYC format

mary-georgiou-sonarsource:


mary-georgiou-sonarsource:
there's another PR open for this.

PR #2330, opened 2023-06-29
Modify rule S3005: LaYC format 

mary-georgiou-sonarsource:


rules/S3005/csharp/rule.adoc
```diff
@@ -47,7 +50,11 @@ public class MyClass
 }
 ----
 
+== Resources
 
+=== Documentation
+* Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/api/system.threadstaticattribute[ThreadStaticAttribute Class]
+* Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadlocal-1[ThreadLocal<T> Class]
```
sebastien-marichal, line 57:
```suggestion
=== Documentation

* Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/api/system.threadstaticattribute[ThreadStaticAttribute Class]
* Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadlocal-1[ThreadLocal<T> Class]
```

rules/S3005/csharp/rule.adoc
```diff
@@ -1,12 +1,15 @@
 == Why is this an issue?
 
-When a non-``++static++`` class field is annotated with ``++ThreadStatic++``, the code seems to show that the field can have different values for different calling threads, but that's not the case, since the ``++ThreadStatic++`` attribute is simply ignored on non-``++static++`` fields. 
+When you annotate a field with the https://learn.microsoft.com/en-us/dotnet/api/system.threadstaticattribute[`ThreadStatic` attribute], it is an indication that the value of this field is unique for each thread. But if you don’t mark the field as `static`,  then the `ThreadStatic` attribute is ignored.
 
 
-So ``++ThreadStatic++`` should either be removed or replaced with a use of the ``++ThreadLocal<T>++`` class, which gives a similar behavior for non-``++static++`` fields.
+== How to fix it
```
sebastien-marichal, line 6: Educational: Could this rule be a candidate for a single `Why is this an issue`?

This rule seems to be quite short to me at least for its description part.

sebastien-marichal (changes requested):
Left minor comments.

rules/S3005/csharp/rule.adoc
mary-georgiou-sonarsource: Yes - I think so. 
Indeed - I will remove the `How to fix it`.

sebastien-marichal: I am now, not really convinced anymore this rule should be on one page.
I worked on [S2252](https://github.com/SonarSource/rspec/pull/2340) which I rewrite to have only a `Why is this an issue?` section.

The code examples in this rule are a bit complex since it provides 2 solutions.

If you prefer to keep the structure within a single section (which is fine for me), I would like to suggest integrating the code example in the description as it would be in a blog post. And get ride of the structure `Code examples` and `Noncompliant code example` / `Compliant solution`

sebastien-marichal (changes requested):
I kind of changed my opinion about having everything in the same section 🙊 
We can discuss it.

sebastien-marichal (approved):
LGTM!

PR #2336, opened 2023-06-29
 S2688: Fix a broken link

gregory-paidis-sonarsource:


Tim-Pohlmann (approved):
Good catch with the .log :D

PR #2344, opened 2023-06-30
Modify rule S3244: LaYC format

gregory-paidis-sonarsource:


rules/S3244/csharp/rule.adoc
```diff
@@ -1,17 +1,64 @@
...
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+event EventHandler myEvent;
+void LogEvent(object s, EventArgs e) => Console.WriteLine($"Event raised with sender {s} and arguments {e}!");
+
+void DoWork()
+{
+        myEvent += LogEvent;
+        // ...
+        myEvent -= LogEvent;
```
sebastien-marichal, line 50: Pedantic: add a `// Compliant: ...` comment

rules/S3244/csharp/rule.adoc
```diff
@@ -1,17 +1,64 @@
...
+Func<int, int> lambda2 = x => x + 1;
 
-'''
-== Comments And Links
-(visible only on this page)
+var result = lambda1 == lambda2; // result is false here
+----
 
-include::../comments-and-links.adoc[]
+This is even more apparent when working with https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/[events] since they are effectively https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/delegates/how-to-combine-delegates-multicast-delegates[multicast delegates] that offer ways of https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-subscribe-to-and-unsubscribe-from-events[subscribing and unsubscribing] to them. If an anonymous function is used to subscribe to an event, it is virtually impossible to unsubscribe from it. This happens because to remove the entry from the subscription list, a reference to the original method is needed, but if the anonymous function has not been stored before subscribing, a reference to it is nowhere to be found.
```
sebastien-marichal, line 15: I think some words might be too high level for non-native speakers. Feel free to apply the suggestion partially.

```suggestion
This is even more true when working with https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/[events] since they are https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/delegates/how-to-combine-delegates-multicast-delegates[multicast delegates] that offer ways of https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-subscribe-to-and-unsubscribe-from-events[subscribing and unsubscribing] to them. If an anonymous function is used to subscribe to an event, it is impossible to unsubscribe from it. This happens because to remove the entry from the subscription list, a reference to the original method is needed, but if the anonymous function has not been stored before subscribing, there is no way to find a reference to it.
```

sebastien-marichal (approved):
LGTM!
Left pedantic comments

PR #2345, opened 2023-06-30
Modify rule S4428: LaYC format

mary-georgiou-sonarsource:


mary-georgiou-sonarsource:
this is solved here: https://github.com/SonarSource/rspec/pull/2346

PR #2347, opened 2023-06-30
Modify rule S2995: LaYC format

sebastien-marichal:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2995/csharp/metadata.json
+++ b/rules/S2995/csharp/metadata.json
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "targeted"
+  "quickfix": "infeasible"
 }
--- a/rules/S2995/csharp/metadata.json
+++ b/rules/S2995/csharp/metadata.json
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "targeted"
 }
--- a/rules/S2995/csharp/rule.adoc
+++ b/rules/S2995/csharp/rule.adoc
@@ -6,7 +6,7 @@ Due to value type variables containing directly an instance of the type, they ca
 
 == How to fix it
 
-When comparing value types, prefer using the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators[equality operators] or the https://learn.microsoft.com/en-us/dotnet/api/system.object.equals(system-object)[`Object.Equals`].
+When comparing value types, prefer using the https://learn.microsoft.com/en-us/dotnet/api/system.object.equals[`Object.Equals`].
 
 Note that in the case of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct[structure types], it is recommended to https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type#struct-example[implement value equality]. If not, https://rules.sonarsource.com/csharp/RSPEC-3898[S3898] might raise.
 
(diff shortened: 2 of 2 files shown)


rules/S2995/csharp/metadata.json
```diff
@@ -24,5 +24,5 @@
   "defaultQualityProfiles": [
     "Sonar way"
   ],
-  "quickfix": "unknown"
+  "quickfix": "targeted"
```
sebastien-marichal, line 27: Not sure if we can make it, I might update it later.

rules/S2995/csharp/rule.adoc
```diff
@@ -1,34 +1,48 @@
 == Why is this an issue?
 
-Using ``++Object.ReferenceEquals++`` to compare the references of two value types simply won't return the expected results most of the time because such types are passed by value, not by reference.
+In C#, the https://learn.microsoft.com/en-us/dotnet/api/system.object.referenceequals[`Object.ReferenceEquals`] method is used to compare two https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/reference-types[reference type] variables. However, this method should not be used to compare https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-types[value types], such as `int`, `float`, or `bool`. This is because value types variable contains an instance of the type.
```
mary-georgiou-sonarsource, line 4:
```suggestion
In C#, the https://learn.microsoft.com/en-us/dotnet/api/system.object.referenceequals[`Object.ReferenceEquals`] method is used to compare two https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/reference-types[reference type] variables. If you use this method to compare two https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-types[value types], such as `int`, `float`, or `bool` you will not get the expected results because value type variables contain an instance of the type and not a reference to it.
```

rules/S2995/csharp/rule.adoc
```diff
@@ -1,34 +1,48 @@
...
 
+Thus, value types variables can't have the same reference and using `Object.ReferenceEquals` to compare them will always return `false` even if the compared variables have the same value.
```
mary-georgiou-sonarsource, line 5:
```suggestion
Due to value type variables containing directly an instance of the type, they can't have the same reference, and using `Object.ReferenceEquals` to compare them will always return `false` even if the compared variables have the same value.
```

rules/S2995/csharp/rule.adoc
```diff
@@ -1,34 +1,48 @@
...
-    int a = 1;
-    int b = 1;
+==== Noncompliant code example
 
-    if (Object.ReferenceEquals(myStruct, s1))  // Noncompliant; this can never be true
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
+public void Method(MyStruct struct1, MyStruct struct2)
+{
+    if (Object.ReferenceEquals(struct1, struct2)) // Noncompliant: this will be always false
```
mary-georgiou-sonarsource, line 17:
```suggestion
    if (Object.ReferenceEquals(struct1, struct2)) // Noncompliant: this is always false
```

rules/S2995/csharp/rule.adoc
```diff
@@ -1,34 +1,48 @@
...
+}
+----
+
+==== Compliant solution
+
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+public void Method(MyStruct struct1, MyStruct struct2)
+{
+    if (struct1.Equals(struct2)) // Compliant: value are compared
```
mary-georgiou-sonarsource, line 30: If I remember correctly in case it's not a default struct one should define the equality by overloading the comparison operators.

Might be better to give a more complete example in this case + add a sentence in "how to fix it?" WDTY?

See here: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type#struct-example

Also please add this link to the resources - I think it's pretty useful.

mary-georgiou-sonarsource (changes requested):

rules/S2995/csharp/rule.adoc
sebastien-marichal: I updated the code examples to be more clear regarding value equality for structs.

mary-georgiou-sonarsource (approved):
LGTM

PR #2352, opened 2023-07-03
Modify rule S6079: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6079/javascript/rule.adoc
+++ b/rules/S6079/javascript/rule.adoc
@@ -23,7 +23,7 @@ describe('My test suite', function() {
 });
 ----
 
-Since the ``++console.log++`` statement is executed after calling ``++done()++`` , there is no guarantee that it will run after the test has fully completed. In fact, it is likely that the ``++console.log++`` statement will execute before the assertions inside the ``++setTimeout++`` function, leading to misleading and inaccurate output.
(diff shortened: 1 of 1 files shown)


rules/S6079/javascript/rule.adoc
```diff
@@ -1,104 +1,54 @@
...
+        expect('hello').to.have.lengthOf(5);
+        done();
+
+        console.log('Test has completed.'); // Noncompliant: Code after calling done, which may produce unexpected behavior
+    }, 1000);
+  });
 });
 ----
 
+Since the ``++console.log++`` statement is executed after calling ``++done()++`` , there is no guarantee that it will run after the test has fully completed. In fact, it is likely that the ``++console.log++`` statement will execute before the assertions inside the ``++setTimeout++`` function, leading to misleading and inaccurate output.
```
victor-diez-sonarsource, line 26: please double check this is the case. seems very strange to me

victor-diez-sonarsource (changes requested):

rules/S6079/javascript/rule.adoc
yassin-kammoun-sonarsource: I will.

yassin-kammoun-sonarsource:
Do you think we should just keep the initial description? After re-reading, it seems quite good in fact. WDYT?

rules/S6079/javascript/rule.adoc
yassin-kammoun-sonarsource: Good catch 👍

victor-diez-sonarsource (approved):

PR #2353, opened 2023-07-03
 Modify rule S2997: LaYC format

mary-georgiou-sonarsource:


rules/S2997/rule.adoc
```diff
@@ -1,27 +1,32 @@
...
-[source,text]
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 public FileStream WriteToFile(string path, string text)
 {
   using (var fs = File.Create(path)) // Noncompliant
```
sebastien-marichal, line 17: Pedantic: I would remove the comment on line 21 to have it on the same line as the `Noncompliant` one.

```suggestion
  using (var fs = File.Create(path)) // Noncompliant: 'fs' is disposed at the end of the using scope
```

rules/S2997/rule.adoc
```diff
@@ -32,3 +37,9 @@ public FileStream WriteToFile(string path, string text)
 }
```
sebastien-marichal, line 37: I would add a compliant comment:
```
 return fs; // Compliant: 'fs' is not disposed once the end of the scope is reached
```

sebastien-marichal (approved):
Left pedantic comments feel free to apply them!

PR #2360, opened 2023-07-03
Update LaYC description with "Related rules" subtitle

amelie-renard-sonarsource:


ilia-kebets-sonarsource (approved):
LGTM, can you verify that the validation doesn't need to be updated?

amelie-renard-sonarsource:
I have already checked, and everything is OK (there are already rules that use this subtitle)

PR #2371, opened 2023-07-03
Modify rule S3610: LaYC format

gregory-paidis-sonarsource:


rules/S3610/csharp/rule.adoc
```diff
@@ -1,22 +1,51 @@
 == Why is this an issue?
 
-Calling ``++GetType()++`` on a nullable object returns the underlying value type. Thus, comparing the returned ``++Type++`` object to ``++typeof(Nullable<SomeType>)++`` doesn't make sense. The comparison either throws an exception or the result can be known at compile time.
+Calling https://learn.microsoft.com/en-us/dotnet/api/system.object.gettype[GetType()] on a https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types[nullable value type] object returns the underlying value type. Thus, comparing the returned https://learn.microsoft.com/en-us/dotnet/api/system.type[`Type`] object to `typeof(Nullable<SomeType>)` doesn't make sense, as the comparison will either throw an https://learn.microsoft.com/en-us/dotnet/api/system.nullreferenceexception[NullReferenceException] or the result can be known at compile time.
```
mary-georgiou-sonarsource, line 3:
```suggestion
Calling https://learn.microsoft.com/en-us/dotnet/api/system.object.gettype[GetType()] on a https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types[nullable value type] object returns the underlying value type. Therefore, comparing the returned https://learn.microsoft.com/en-us/dotnet/api/system.type[`Type`] object to `typeof(Nullable<SomeType>)` will result in either an https://learn.microsoft.com/en-us/dotnet/api/system.nullreferenceexception[NullReferenceException] or the result will always be `true` or `false` and can be seen already at compile time.
```

rules/S3610/csharp/rule.adoc
```diff
@@ -1,22 +1,51 @@
...
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
+void DoChecks<T>(Nullable<T> nullable)
+{
+    bool areEqual = nullable.GetType() == typeof(Nullable<int>); // Noncompliant: always false
+    bool areNotEqual = nullable.GetType() != typeof(Nullable<int>); // Noncompliant: always true
+
+    nullable = null;
```
mary-georgiou-sonarsource, line 18: I'd remove the argument and declare this variable directly in the body, since you are also setting it there - I find it more consistent and clearer this way.
```suggestion
    Nullable<T> nullable = null;
```

mary-georgiou-sonarsource (changes requested):

rules/S3610/csharp/rule.adoc
gregory-paidis-sonarsource: I think it's valuable for it to be an argument, since that way we do not know the type of `T`, and the type-check can be either `true` or `false` in the compliant case (but not in the noncompliant case, hence the rule's purpose).

I changed the second example that mutates the value to use a new one.
Let me know if you disagree.

mary-georgiou-sonarsource (approved):
LGTM

PR #2402, opened 2023-07-05
Modify rule S3531: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3531/javascript/rule.adoc
+++ b/rules/S3531/javascript/rule.adoc
@@ -27,7 +27,7 @@ function* range(start, end) {
 }
 ----
 
-Yielding without a value makes it harder for the generator consumer to understand the purpose or context of the yielded value. Instead, provide a specific value with ``++yield++`` to make the generated sequence more meaningful and informative.
+Yielding without a value makes it harder for the generator consumer to understand the purpose or context of the yielded value. Instead, one should always provide an explicit value with ``++yield++`` (using `undefined` when that is the intention) to make the generated sequence more meaningful and informative.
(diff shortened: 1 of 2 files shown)


rules/S3531/javascript/rule.adoc
```diff
@@ -1,32 +1,49 @@
...
-    yield answer;
+function* range(start, end) {
+  while (start < end) {
+    yield; // Noncompliant: The generator yields undefined
+    start++;
   }
 }
 ----
 
+Yielding without a value makes it harder for the generator consumer to understand the purpose or context of the yielded value. Instead, provide a specific value with ``++yield++`` to make the generated sequence more meaningful and informative.
```
victor-diez-sonarsource, line 30:
```suggestion
Yielding without a value makes it harder for the generator consumer to understand the purpose or context of the yielded value. Instead, one should always provide an explicit value with ``++yield++`` (using `undefined` when that is the intention) to make the generated sequence more meaningful and informative.
```

I would use explicit rather than specific.

victor-diez-sonarsource: you also use it in the rule title now, I think is  better wording

victor-diez-sonarsource (approved):
just a minor comment, feel free to ignore

rules/S3531/javascript/rule.adoc
yassin-kammoun-sonarsource: Makes sense 👍

PR #2404, opened 2023-07-05
Modify rule S2123: LaYC format

gregory-paidis-sonarsource:


rules/S2123/csharp/rule.adoc
```diff
@@ -1,36 +1,63 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+When using the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/arithmetic-operators#postfix-increment-operator[postfix increment] operator, it is important to know that the result of the expression `x++` is the value **before** the operation `x`.
 
-=== Noncompliant code example
+This means that in some cases, the result might not be what you expect:
 
-[source,csharp]
+* When assigning `x++` to `x`, it's practically the same as assigning `x` to itself, since the value is assigned before the increment takes place
```
sebastien-marichal, line 7:
```suggestion
* When assigning `x++` to `x`, it's the same as assigning `x` to itself, since the value is assigned before the increment takes place
```

rules/S2123/csharp/rule.adoc
```diff
@@ -1,36 +1,63 @@
...
+* When returning `x++`, the returning value is `x`, not `x+1`
+
+To solve the first scenario, eliminate the assignment, since `x\++` mutates `x` anyways. On the other hand, to solve the second scenario consider using the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/arithmetic-operators#prefix-increment-operator[prefix increment] operator, since it works in reverse: the result of the expression `++x` is the value **after** the operation, which is `x+1`, as one might expect.
```
sebastien-marichal, line 10: Why the `x\++` here?

sebastien-marichal, line 10: This seems to be more about the solution than the issue. Maybe it could be under the `How to fix it` section.
Might I suggest adding the corresponding scenarios in that case?

sebastien-marichal (changes requested):
Content is good.
Left some small comments and suggestions.

rules/S2123/csharp/rule.adoc
gregory-paidis-sonarsource:
```
Value1 is `x++` and value2 is `++x`
```

For some reason, it captures the first and last backtick and it looks like this:

```
Value1 is x` and value2 is `
```

So i need to escape the `++`

gregory-paidis-sonarsource: You are right, I moved some of the description to `How to fix it`

sebastien-marichal (approved):
LGTM!

PR #2408, opened 2023-07-05
Fix the rule-id link format

necto:
Rule IDs are [automatically transformed](https://github.com/SonarSource/sonar-rule-api/blob/master/src/main/java/com/sonarsource/ruleapi/asciidoctor/AsciiDoctorConverter.java#L111) into link placeholders by rule-api. These placeholders are translated into autolinks by the products.
If these IDs appear within other links, it might [break some products](https://discuss.sonarsource.com/t/cross-rule-references-inside-a-tags-in-html-rule-descriptions/14971/4).

Moreover, the generated autolinks are product-specific, and cannot be encoded statically in the rule description.


necto:
[RULEAPI-790](https://sonarsource.atlassian.net/browse/RULEAPI-790) tracks an automatic check to block this pattern

John-Clifton-SonarSource (approved):
Looks good to me. Thank you for the clarification.

PR #2421, opened 2023-07-06
Modify rule S3504: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3504/javascript/rule.adoc
+++ b/rules/S3504/javascript/rule.adoc
@@ -1,5 +1,8 @@
 == Why is this an issue?
 
+Variables declared with ``++var++`` are function-scoped, meaning they are accessible within the entire function in which they are defined. If a variable is declared using ``++var++`` outside of any function, it becomes a global variable and is accessible throughout the entire JavaScript program.
+
+
 ECMAScript 2015 introduced the ``++let++`` and ``++const++`` keywords for block-scope variable declaration. Using ``++const++`` creates a read-only (constant) variable.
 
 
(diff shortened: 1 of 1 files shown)


victor-diez-sonarsource (changes requested):
I think there is space in this rule to explain more concepts. 

When talking about var: https://developer.mozilla.org/en-US/docs/Glossary/Hoisting

Let and const: non-hoisted but you face the [temporal dead zone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#temporal_dead_zone_tdz)

Probably explain a little bit about js [scopes](https://developer.mozilla.org/en-US/docs/Glossary/Scope)

yassin-kammoun-sonarsource:
@victor-diez-sonarsource Is it better now ?

rules/S3504/javascript/rule.adoc
```diff
@@ -1,6 +1,20 @@
...
+
+
+Variables declared with ``++const++`` are also block-scoped, similar to  ``++let++``. However, ``++const++`` variables are immutable, meaning their value cannot be changed after assignment. This applies to the binding between the variable name and its value, but it does not mean the value itself is immutable if it is an object or an array.
+
+
+A variable declared with ``++let++`` or ``++const++`` is said to be in a "temporal dead zone", meaning the period between entering a scope and declaring a ``++let++`` or ``++const++`` variable. During this phase, accessing the variable results in a ``++ReferenceError++``. This helps catch potential errors and encourages proper variable declaration.
+
+
+Unlike ``++let++`` and ``++const++``, variables declared with ``++var++`` are subject to "hoisting," which means that they are moved to the top of their scope during the compilation phase, even if the actual declaration is placed lower in the code.
```
victor-diez-sonarsource, line 14:
```suggestion
Unlike ``++let++`` and ``++const++``, variables declared with ``++var++`` are subject to "hoisting", which means that they are moved to the top of their scope during the compilation phase, even if the actual declaration is placed lower in the code.
```

victor-diez-sonarsource (approved):
excellent, thanks! :)

just a small typo with a comma

rules/S3504/javascript/rule.adoc
yassin-kammoun-sonarsource: 👍

PR #2426, opened 2023-07-07
Modify rule S6441: Adapt to LaYC

victor-diez-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6441/javascript/rule.adoc
+++ b/rules/S6441/javascript/rule.adoc
@@ -1,8 +1,8 @@
 == Why is this an issue?
 
-This rule reports the methods of React component classes that are neither used nor React lifecycle methods. 
+Methods that are never executed are dead code and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
 
-This rule checks only usages of the method inside the scope of a component class. It will also raise an issue if the method is only used outside the class. React encourages data-driven components. To modify a child component, re-render it with new `props`, instead of calling internal methods from outside the component.
+When using React class components, all non-React lifecycle methods should be called within the scope of the component. If a method is only called from outside the class, consider using `props` to interact with the component and re-render if needed, as React encourages data-driven components.
 
 [source,javascript,diff-id=1,diff-type=noncompliant]
 ----
@@ -11,13 +11,13 @@ class Profile extends React.Component {
     return <h1>{ props.name }</h1>;
   }
 
-  getDefaultName() { // Noncompliant, this method is never used and is a dead code
+  getDefaultName() { // Noncompliant: this method is never used and is a dead code
     return 'John Smith';
   }
 }
 ----
 
-Methods that are never executed are dead code and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
+To fix the issue, remove the dead code or call the method from within the component scope.
 
 [source,javascript,diff-id=1,diff-type=compliant]
 ----
--- a/rules/S6441/javascript/rule.adoc
+++ b/rules/S6441/javascript/rule.adoc
@@ -1,14 +1,10 @@
 == Why is this an issue?
 
-This rule reports the methods of React component classes which are neither used nor React lifecycle methods. Note that this rule checks only the scope of a component class, which means that issues will be raised on the methods that are only used outside of the class.
+This rule reports the methods of React component classes that are neither used nor React lifecycle methods. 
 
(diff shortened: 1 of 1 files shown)


rules/S6441/javascript/rule.adoc
```diff
@@ -38,4 +34,5 @@ class Profile extends React.Component {
 
 == Resources
 
-* https://reactjs.org/docs/components-and-props.html[Components and Props] - React Documentation
\ No newline at end of file
+* https://reactjs.org/docs/components-and-props.html[React Documentation - Components and Props]
+* https://legacy.reactjs.org/docs/state-and-lifecycle.html#adding-lifecycle-methods-to-a-class[React Documentation - ]
```
yassin-kammoun-sonarsource, line 38: Should we use links to the new React website instead?

yassin-kammoun-sonarsource, line 38:
```suggestion
* https://legacy.reactjs.org/docs/state-and-lifecycle.html#adding-lifecycle-methods-to-a-class[React Documentation - Adding Lifecycle Methods to a Class]
```

rules/S6441/javascript/rule.adoc
```diff
@@ -38,4 +34,5 @@ class Profile extends React.Component {
 
 == Resources
```
yassin-kammoun-sonarsource, line 36:
```suggestion
=== Documentation
```

rules/S6441/javascript/rule.adoc
```diff
@@ -1,14 +1,10 @@
...
 
-Methods that are never executed are dead code: unnecessary, inoperative code that should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
+This rule checks only usages of the method inside the scope of a component class. It will also raise an issue if the method is only used outside the class. React encourages data-driven components. To modify a child component, re-render it with new `props`, instead of calling internal methods from outside the component.
 
-React encourages data-driven components. To modify a child component, re-render it with new `props`.
-
-=== Noncompliant code example
-
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
```
yassin-kammoun-sonarsource, line 7: Can you fix the noncompliant comment below, meaning `// Noncompliant: ...`?

rules/S6441/javascript/rule.adoc
```diff
@@ -1,14 +1,10 @@
 == Why is this an issue?
 
-This rule reports the methods of React component classes which are neither used nor React lifecycle methods. Note that this rule checks only the scope of a component class, which means that issues will be raised on the methods that are only used outside of the class.
+This rule reports the methods of React component classes that are neither used nor React lifecycle methods. 
 
-Methods that are never executed are dead code: unnecessary, inoperative code that should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
+This rule checks only usages of the method inside the scope of a component class. It will also raise an issue if the method is only used outside the class. React encourages data-driven components. To modify a child component, re-render it with new `props`, instead of calling internal methods from outside the component.
```
yassin-kammoun-sonarsource, line 5: I think we should rework this part because it's essentially describing what the rule does, not what the problem actually is.

Here is a suggestion on how to tackle this:
1. First, introduce the problem about methods never executed or dead code, basically the text between the two snippets below
2. Second, give some context about what kind of methods it is good practice to define in a React component: either lifecycle methods or methods used for rendering
3. Third, summarize what the rule is going to check and report

yassin-kammoun-sonarsource, line 5: > This rule checks only usages of the method inside the scope of a component class. It will also raise an issue if the method is only used outside the class. React encourages data-driven components. To modify a child component, re-render it with new `props`, instead of calling internal methods from outside the component.

This paragraph is confusing because it starts with what the rule checks. Then it jumps to an unrelated topic and talks about data-driven components. And it somehow concludes with child components and rendering.

Could you clarify how the parts _"This rule checks only usages of the method inside the scope of a component class. It will also raise an issue if the method is only used outside the class."_ and _"React encourages data-driven components. To modify a child component, re-render it with new `props`, instead of calling internal methods from outside the component."_ are related to one another?

I feel like something is missing in-between to connect the two parts.

yassin-kammoun-sonarsource (changes requested):

rules/S6441/javascript/rule.adoc
victor-diez-sonarsource: didn't find anything convincing in the new website about lifecycle

victor-diez-sonarsource: oops

yassin-kammoun-sonarsource: Let's keep them then :)

PR #2430, opened 2023-07-10
Modify rule S4619: Adapt to LaYC

victor-diez-sonarsource:


rules/S4619/javascript/rule.adoc
```diff
@@ -1,42 +1,26 @@
...
-    if (expectedValue in arr) { // Noncompliant, will be always true because the array is made of 3 elements and the #1 is always there whatever its value
+    if (expectedValue in arr) { // Noncompliant: will be always false
         return expectedValue + " found in the array";
     } else {
         return expectedValue + " not found";
     }
 }
 ----
 
+Use the Array method `includes()` to determine whether an array contains a certain value. If the actual intention is to check for an array slot, use ``++hasOwnProperty++``.
```
yassin-kammoun-sonarsource, line 21:
```suggestion
Use the method `Array.prototype.includes()` to determine whether an array contains a certain value. If the actual intention is to check for an array slot, use ``++ Object.prototype.hasOwnProperty() ++``.
```

yassin-kammoun-sonarsource: Not convinced about this suggestion. Feel free to ignore it.

rules/S4619/javascript/rule.adoc
```diff
@@ -1,42 +1,26 @@
...
 
-=== Compliant solution
-
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
 function func() {
     let arr = ["a", "b", "c"];
```
yassin-kammoun-sonarsource, line 26: Let's maybe use `const` instead.

rules/S4619/javascript/rule.adoc
```diff
@@ -1,42 +1,26 @@
...
-If checking for an array slot is indeed desired, using ``++hasOwnProperty++`` makes the code intention clearer.
-
-
-=== Noncompliant code example
-
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
 function func1() {
     let arr = ["a", "b", "c"];
```
yassin-kammoun-sonarsource, line 10: Let's maybe use `const` instead.

yassin-kammoun-sonarsource (approved):
LGTM.

rules/S4619/javascript/rule.adoc
victor-diez-sonarsource: what would you suggest?

yassin-kammoun-sonarsource: I meant I am not convinced by what I suggested. What you wrote is fine :)

victor-diez-sonarsource: ah, ok. I like it, I will change it

PR #2435, opened 2023-07-10
Modify rule S2392: Adapt to LaYC

victor-diez-sonarsource:


rules/S2392/javascript/rule.adoc
```diff
@@ -1,22 +1,21 @@
 == Why is this an issue?
 
-Variables that are declared inside a block but used outside of it (which is possible with a ``++var++``-style declaration) should be declared outside the block.
+The `var` statement declares variables that are function-scoped or globally-scoped. `var` declarations are hoisted, meaning declaring a variable anywhere in the code is equivalent to declaring it at the top of the function or the script.
```
yassin-kammoun-sonarsource, line 3:
```suggestion
The `var` statement declares variables that are function-scoped or globally-scoped. `var` declarations are hoisted, meaning declaring a variable anywhere in the code is equivalent to declaring it at the top of the function or the file. 
```

rules/S2392/javascript/rule.adoc
```diff
@@ -1,22 +1,21 @@
 == Why is this an issue?
 
-Variables that are declared inside a block but used outside of it (which is possible with a ``++var++``-style declaration) should be declared outside the block.
+The `var` statement declares variables that are function-scoped or globally-scoped. `var` declarations are hoisted, meaning declaring a variable anywhere in the code is equivalent to declaring it at the top of the function or the script. 
 
-
-=== Noncompliant code example
+Even if hosted, it is still recommended to declare it inside the block it is used. This improves readability and maintainability because it makes it clear where the variable is being used, making the code easier to understand and follow, especially for other developers who may be working on the same codebase.
```
yassin-kammoun-sonarsource, line 5:
```suggestion
Even if hoisted, it is still recommended to declare the variable inside the block it is used. This improves readability and maintainability because it makes it clear where the variable is being used. The code then becomes easier to understand and follow, especially for other developers who may be working on the same codebase.
```

rules/S2392/javascript/rule.adoc
```diff
@@ -1,22 +1,21 @@
...
 
 [source,javascript]
```
yassin-kammoun-sonarsource, line 7: Missing diff-tag

yassin-kammoun-sonarsource (approved):
Tiny things to address. Otherwise looks good to me.

rules/S2392/javascript/rule.adoc
victor-diez-sonarsource: here I prefer script, which works for both files and for HTML tags. I think it's more idiomatic.

PR #2440, opened 2023-07-11
Modify rule S4140: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4140/javascript/rule.adoc
+++ b/rules/S4140/javascript/rule.adoc
@@ -1,23 +1,34 @@
 == Why is this an issue?
 
-An array declared with missing ("sparse") elements is likely to be an error: an extra comma was inserted or perhaps the developer meant to insert the missing value and forgot.
+A sparse array is an array in which the elements do not occupy a contiguous range of indices. In other words, there are gaps (or "holes") between the elements in the array, where some indices have no corresponding value assigned to them.
 
+Including an extra comma in an array literal signifies an `undefined` or empty element in the array, thus creating a sparse array. The JavaScript engine automatically assigns `undefined` to the omitted element.
 
-=== Noncompliant code example
+While this is a well-defined behavior, it can be misleading and raises suspicions about the original intent: an extra comma was intentionally inserted, or perhaps the developer meant to insert the missing value but forgot.
 
 [source,javascript]
 ----
-let a = [1, , 3, 6, 9];  // Noncompliant
+let a = [1, , 3, 6, 9]; // Noncompliant: Extra comma maybe denoting an oversight
 ----
 
-
-=== Compliant solution
+You should either remove the extra comma if this was a mistake or add the value you meant to insert initially.
 
 [source,javascript]
 ----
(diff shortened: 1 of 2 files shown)


rules/S4140/javascript/rule.adoc
```diff
@@ -1,23 +1,34 @@
...
-
-=== Compliant solution
+You should either remove the extra comma if this was a mistake or add the value you meant to insert initially.
 
 [source,javascript]
 ----
 let a = [1, 3, 6, 9];
 ----
 
+Alternatively, if the purpose was deliberately creating a sparse array, this should be made explicit with an `undefined` element.
```
victor-diez-sonarsource, line 21:
```suggestion
Alternatively, if the purpose was deliberately creating a sparse array, this should be made explicit with an `undefined` element.
```

Careful here, setting explicitly undefined does not create an sparse array, but an array with the value set to defined at that position.

victor-diez-sonarsource (changes requested):

victor-diez-sonarsource:
I would mention something that the docs are explaining very well:

- Sparse arrays
Arrays can contain "empty slots", which are not the same as slots filled with the value undefined. In some operations, empty slots behave as if they are filled with `undefined`. But in others (most notably array iteration methods), empty slots are skipped.

yassin-kammoun-sonarsource:
> * Arrays can contain "empty slots", which are not the same as slots filled with the value undefined. In some operations, empty slots behave as if they are filled with `undefined`. But in others (most notably array iteration methods), empty slots are skipped.

Is that a suggestion that I copy-paste directly or is it taken from some documentation?

victor-diez-sonarsource:
it's copy/pasted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays

rules/S4140/javascript/rule.adoc
yassin-kammoun-sonarsource: Indeed. I will just remove that part and clarify further what sparse arrays are with what you mentioned from the documentation.

rules/S4140/javascript/rule.adoc
```diff
@@ -2,29 +2,22 @@
...
 
 While this is a well-defined behavior, it can be misleading and raises suspicions about the original intent: an extra comma was intentionally inserted, or perhaps the developer meant to insert the missing value but forgot.
 
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
 let a = [1, , 3, 6, 9]; // Noncompliant: Extra comma maybe denoting an oversight
 ----
 
 You should either remove the extra comma if this was a mistake or add the value you meant to insert initially.
```
victor-diez-sonarsource, line 14: I would mention that you can use undefined if the additional comma was intended (not saying that this will create a sparse array)

victor-diez-sonarsource (approved):

PR #2441, opened 2023-07-11
Create a shared_content folder at root

frederic-tingaud-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/ci/validate_asciidoc.sh
+++ b/ci/validate_asciidoc.sh
@@ -145,6 +145,8 @@ else
   echo "No new asciidoc file changed"
 fi
 
+find rules -name "tmp*.adoc" -delete
+
 if (( exit_code == 0 )); then
     echo "Success"
 else
--- a/ci/validate_asciidoc.sh
+++ b/ci/validate_asciidoc.sh
@@ -92,6 +92,7 @@ do
           TMP_ADOC="$language/tmp_$(basename "${dir}")_${language##*/}.adoc"
           echo "== Description" > "$TMP_ADOC"
           cat "$RULE" >> "$TMP_ADOC"
+          rm "${TMP_ADOC}"
         else
           echo "ERROR: no asciidoc file $RULE"
           exit_code=1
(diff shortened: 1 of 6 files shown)


rules/S5553/cfamily/rule.adoc
```diff
@@ -42,16 +42,7 @@ If the definition of the temporary relies on another temporary,
...
-
-Accessing a dangling reference or pointer causes undefined behavior.
-This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.
-
-Practically this has a wide range of effects.
-In many cases, the access works by accident and succeeds at writing or reading a value.
-However, it can start misbehaving at any time.
-If compilation flags, compiler, platform, or runtime environment change,
-the same code can crash the application, corrupt memory, or leak a secret.
+include::../../../shared_content/cfamily/dangling_reference_impact.adoc[]
```
tomasz-kaminski-sonarsource, line 45: I would prefer if that would not be relative. Does asciidoctor has any notion of `include` dirs?
I think we are already doing some transformation of the includes, so we could replace `shared_content/...` with dedicated path. This has additional benefit of people being unable to use `shared_content` inside the rule.

ci/validate_asciidoc.sh
```diff
@@ -106,9 +107,15 @@ do
     # Which is why we need to create this list and cannot do anything with the paths it contains until we find the corresponding include
     find "$dir" -name "*.adoc" -execdir sed -r -n -e 's/^:(\w+):\s+([A-Za-z0-9\/._-]+)$/\1\t\2/p' {} \; > vars
     # Directly included
-    find "$dir" -name "*.adoc" -execdir sh -c 'grep -Eh "include::" "$1" | grep -Ev "{\w+}" | grep -v "rule.adoc" | sed -r "s/include::(.*)\[\]/\1/" | xargs -r -I@ realpath "$PWD/@"' shell {} \; > included
+    find "$dir" -name "*.adoc" -execdir sh -c 'grep -h "include::" "$1" | grep -Ev "{\w+}" | grep -v "rule.adoc" | sed -r "s/include::(.*)\[\]/\1/" | xargs -r -I@ realpath "$PWD/@"' shell {} \; > included
     # Included through variable
     VARS_FULL_PATH=$(realpath vars) PATH_WITH_VARIABLE=${PATH_WITH_VARIABLE} find "$dir" -name "*.adoc" -execdir sh -c 'grep -Eh "include::.*\{" "$1" | xargs -r -I@ $PATH_WITH_VARIABLE $VARS_FULL_PATH "@" | xargs -r -I@ realpath "$PWD/@"' shell {} \; >> included
+    # There should be no cross inclusion between rules
```
tomasz-kaminski-sonarsource, line 113: Isn't this subset of safe mode?

tomasz-kaminski-sonarsource (changes requested):

rules/S5553/cfamily/rule.adoc
frederic-tingaud-sonarsource: Sadly asciidoctor doesn't have include directories. That would have helped a lot. I thought about having a magic attribute to simplify the includes. The problem is that at the moment we have no custom behavior for includes and breaking that would break some workflows (GitHub previews, VSCode preview and so on). We discussed it with @necto and decided that the benefit was not good enough compared to what we would lose. But I'm open to rediscussing it because these paths are indeed ugly.
I don't understand your last point about forbidding people to use `shared_content` inside the rule. Are there places where you don't want people to access `shared_content`?

ci/validate_asciidoc.sh
frederic-tingaud-sonarsource: We currently build with unsafe mode. I don't know exactly why, but that is good for us because otherwise the `shared_content` folder couldn't be included.

rules/S5553/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Ok. The script could have special handling for shared folders, but all other normal tools will break because of that. Makes sense.

ci/validate_asciidoc.sh
```diff
@@ -106,9 +107,15 @@ do
...
+    cross_references=$(grep -Eh "S[0-9]{3,}" included | grep -v ${dir})
```
tomasz-kaminski-sonarsource, line 114: I think we should check all `includes` that does not contain `shared_content` directory here. Not only rule includes.

tomasz-kaminski-sonarsource: If of course this is not already breaking something.

necto: Agreed

necto (approved):
I would also mention this capability in https://github.com/SonarSource/rspec/blob/master/docs/description.adoc
And the "shared_content" folder in
https://github.com/SonarSource/rspec#rules-directory-structure

tomasz-kaminski-sonarsource (approved):

PR #2445, opened 2023-07-11
Modify rule S2176(Java): Update rule according to the LayC

alban-auzeill:
Before:
https://next.sonarqube.com/sonarqube/coding_rules?q=s2176&open=java%3AS2176
After:
image
image
image

### DIFF
--- a/rules/S2176/java/rule.adoc
+++ b/rules/S2176/java/rule.adoc
@@ -20,8 +20,8 @@ public class User {
 }
 ----
 
-But it is an issue when a class has the same name as the superclass it extends or the interfaces it implements.
-This issue is called class name shadowing. It's an issue because it can be unclear which class is being referred to
+However, this is an issue when a class has the same name as the superclass it extends or the interfaces it implements.
+This issue is called class name shadowing. It is an issue because it can be unclear which class is being referred to
 in the code, leading to ambiguity and potential bugs.
 
 Therefore, it is recommended to use unique and descriptive class names that do not conflict with the names of the
--- a/rules/S2176/java/rule.adoc
+++ b/rules/S2176/java/rule.adoc
@@ -20,9 +20,9 @@ public class User {
 }
 ----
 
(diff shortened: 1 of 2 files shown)


rules/S2176/java/rule.adoc
```diff
@@ -1,27 +1,67 @@
...
-package my.mypackage;
+package org.foo.presentation;
 
-public class Foo implements a.b.Foo { // Noncompliant
+public class User {
+  // ..
+}
 ----
 
+But it is an issue when a class has the same name as the superclass it extends or the interfaces it implements.
```
johann-beleites-sonarsource, line 23:
```suggestion
This becomes an issue when a class has the same name as the superclass it extends or the interfaces it implements.
```

rules/S2176/java/rule.adoc
```diff
@@ -1,27 +1,67 @@
...
+This issue is called class name shadowing. It's an issue because it can be unclear which class is being referred to
```
johann-beleites-sonarsource, line 24: AFAIK, one should avoid contractions in formal writing
```suggestion
This issue is called class name shadowing. It is an issue because it can be unclear which class is being referred to
```

rules/S2176/java/rule.adoc
```diff
@@ -1,27 +1,67 @@
...
+package org.foo.presentation;
 
-public class Foo implements a.b.Foo { // Noncompliant
+public class User {
+  // ..
+}
 ----
 
+However, this is an issue when a class has the same name as the superclass it extends or the interfaces it implements.
+This issue is called class name shadowing. It is an issue because it can be unclear which class is being referred to
```
johann-beleites-sonarsource, line 24: We use the word "issue" a lot here. My proposal:
```suggestion
However, this becomes an issue when a class has the same name as the superclass it extends or the interfaces it implements, also known as class name shadowing.
It is problematic because it can be unclear which class is being referred to
```

johann-beleites-sonarsource (approved):
One small suggestion, otherwise LGTM!

PR #2449, opened 2023-07-11
Modify rule S6427: LaYC format

amelie-renard-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S6427/cfamily/rule.adoc
+++ b/rules/S6427/cfamily/rule.adoc
@@ -1,10 +1,19 @@
 == Why is this an issue?
 
-The class `std::optional` stores an optional value: a `std::optional<T>` can either contain a value of type `T` or be empty.
+The class `std::optional<T>` either stores a value of type `T` or is empty.
 
-One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. Assigning a value to the dereferenced optional or the result of the function `value()` is undefined behavior if the optional has no value. That is why the preferred way to assign a value to an optional is to assign it directly: it will work in any case.
+One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. If the optional is uninitialized, assigning a value to the dereferenced optional is undefined behavior.
+
+Another way to access the value of a non-empty optional is the function `value()`. But assigning a value to the optional object through this function will throw an exception (`std::bad_optional_access`) if the optional has no value, and the assignment will not happen.
+
+For the assignment of an optional to happen correctly, whatever its state, it is better to:
+
+* Assign the value directly with the `operator=`: e.g. `std::optional<int> opt = 3;`
+* Use the `emplace` function (for example, when the move or copy operation is expensive or forbidden).
 
 == How to fix it
+
+The rule raises an issue when the `operator=` or the `value()` function initializes an optional.
 
 === Code examples
 
--- a/rules/S6427/cfamily/rule.adoc
+++ b/rules/S6427/cfamily/rule.adoc
@@ -1,25 +1,29 @@
 == Why is this an issue?
 
-The class `std::optional` stores an optional value: a `std::optional<T>` can either contain a value of type `T` or be empty. One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer.
+The class `std::optional` stores an optional value: a `std::optional<T>` can either contain a value of type `T` or be empty.
 
-However, the similarity ends there. In particular, the preferred way to assign a value to an optional is to assign it directly (as opposed to assigning it to the dereferenced value or to the result of the function `value()`). In that case, the assignment works even if the optional does not have a prior value. 
+One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. Assigning a value to the dereferenced optional or the result of the function `value()` is undefined behavior if the optional has no value. That is why the preferred way to assign a value to an optional is to assign it directly: it will work in any case.
 
-=== Noncompliant code example
+== How to fix it
 
-[source,cpp]
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
 void g(std::optional<int> &val, bool b) {
     if (b) {
-      *val = 314; // Noncompliant, will throw if the optional was previously empty
+      *val = 314; // Noncompliant, it is undefined behavior if the optional is empty
     } else {
-      val.value() = 42; // Noncompliant, will throw if the optional was previously empty
+      val.value() = 42; // Noncompliant, it will throw if the optional is empty
     }
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=compliant]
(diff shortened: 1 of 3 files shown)


rules/S6427/cfamily/rule.adoc
```diff
@@ -1,25 +1,29 @@
...
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
 void g(std::optional<int> &val, bool b) {
     if (b) {
-      *val = 314; // Noncompliant, will throw if the optional was previously empty
+      *val = 314; // Noncompliant, it is undefined behavior if the optional is empty
```
marco-antognini-sonarsource, line 17:
```suggestion
      *val = 314; // Noncompliant; the behavior is undefined if the optional is empty.
```

rules/S6427/cfamily/rule.adoc
```diff
@@ -1,25 +1,29 @@
...
     } else {
-      val.value() = 42; // Noncompliant, will throw if the optional was previously empty
+      val.value() = 42; // Noncompliant, it will throw if the optional is empty
```
marco-antognini-sonarsource, line 19:
```suggestion
      val.value() = 42; // Noncompliant; it will throw if the optional is empty.
```

rules/S6427/cfamily/rule.adoc
```diff
@@ -1,25 +1,29 @@
 == Why is this an issue?
 
-The class `std::optional` stores an optional value: a `std::optional<T>` can either contain a value of type `T` or be empty. One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer.
+The class `std::optional` stores an optional value: a `std::optional<T>` can either contain a value of type `T` or be empty.
```
marco-antognini-sonarsource, line 3: I feel this sentence is very repetitive. Maybe something like this:
```suggestion
The class `std::optional<T>` is either in an empty state or it stores an value of type `T`.
```

rules/S6427/cfamily/rule.adoc
```diff
@@ -1,25 +1,29 @@
...
 
-However, the similarity ends there. In particular, the preferred way to assign a value to an optional is to assign it directly (as opposed to assigning it to the dereferenced value or to the result of the function `value()`). In that case, the assignment works even if the optional does not have a prior value. 
+One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. Assigning a value to the dereferenced optional or the result of the function `value()` is undefined behavior if the optional has no value. That is why the preferred way to assign a value to an optional is to assign it directly: it will work in any case.
```
marco-antognini-sonarsource, line 5: > Assigning a value to [...] the result of the function value() is undefined behavior if the optional has no value.

Technically, this is well defined: `.value()` will throw and the assignement will never happen.

The connection with the pointer semantics works relatively well, I think, when using `*` or `->` to access the value: it is undefined on `nullptr` and it is undefined on `nullopt`.

The connection with pointers stops when `.value()` is involved: there is simply no counterpart for pointers.

> the preferred way [...] is to assign it directly

This could be more explicit and refer to the `operator=` of the optional, or its `emplace` function -- the later is also interesting when the move or copy operation is expensive or forbidden.

marco-antognini-sonarsource (changes requested):
I think it would be worth mentioning `emplace`. There is also a wording issue: `opt.value() = x` is well-defined even if `opt` is empty (`std::bad_optional_access`).

rules/S6427/cfamily/rule.adoc
amelie-renard-sonarsource: Modified, let me know what you think

rules/S6427/cfamily/rule.adoc
amelie-renard-sonarsource: Updated in https://github.com/SonarSource/rspec/pull/2449/commits/3533d1f1e18dbf179e7b464c7cc93208cdd7321d

rules/S6427/cfamily/rule.adoc
marco-antognini-sonarsource: Much better! 👍

rules/S6427/cfamily/rule.adoc
```diff
@@ -1,11 +1,20 @@
 == Why is this an issue?
 
-The class `std::optional` stores an optional value: a `std::optional<T>` can either contain a value of type `T` or be empty.
+The class `std::optional<T>` either stores a value of type `T` or is empty.
 
-One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. Assigning a value to the dereferenced optional or the result of the function `value()` is undefined behavior if the optional has no value. That is why the preferred way to assign a value to an optional is to assign it directly: it will work in any case.
+One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. If the optional is uninitialized, assigning a value to the dereferenced optional is undefined behavior.
```
marco-antognini-sonarsource, line 5: Technically, just calling `operator*` on an empty optional is UB. This suggestion removes the notion of "assignment" since it's already past the point of UB. I tried to keep it short but maybe there's a better alternative.

```suggestion
One way to access the value of a non-empty optional is the `operator*`, making an optional look like a pointer. However, the similarity ends there: an optional object models an object, not a pointer. Furthermore, dereferencing an empty optional is undefined behavior.
```

rules/S6427/cfamily/rule.adoc
```diff
@@ -1,11 +1,20 @@
...
+
+Another way to access the value of a non-empty optional is the function `value()`. But assigning a value to the optional object through this function will throw an exception (`std::bad_optional_access`) if the optional has no value, and the assignment will not happen.
+
+For the assignment of an optional to happen correctly, whatever its state, it is better to:
+
+* assign the value directly with the `operator=`: e.g. `std::optional<int> opt = 3;`
```
marco-antognini-sonarsource, line 11: Nit: the example may give the impression we need a new variable.

```suggestion
* assign the value directly with the `operator=`: e.g. `myOptionalInteger = 3;`
```

rules/S6427/cfamily/rule.adoc
```diff
@@ -1,25 +1,38 @@
...
 
-[source,cpp]
+For the assignment of an optional to happen correctly, whatever its state, it is better to:
+
+* assign the value directly with the `operator=`: e.g. `std::optional<int> opt = 3;`
+* use the `emplace` function (for example, when the move or copy operation is expensive or forbidden).
+
+== How to fix it
+
+The rule raises an issue when the `operator=` or the `value()` function initializes an optional.
```
marco-antognini-sonarsource, line 16:
```suggestion
The rule raises an issue when the `operator*` or the `value()` function are used to assign a new value to an optional.
```

marco-antognini-sonarsource (commented):
Looks much better already, a couple of smaller nits and I think it's ready.

rules/S6427/cfamily/rule.adoc
amelie-renard-sonarsource: I have rewritten this paragraph a bit in https://github.com/SonarSource/rspec/pull/2449/commits/ebf7f7963022f73f824203a3f1359f9f29c452f7
WDYT?

marco-antognini-sonarsource (approved):
LGTM

PR #2451, opened 2023-07-11
Modify rule S3863: Adapt to LaYC

victor-diez-sonarsource:


rules/S3863/javascript/rule.adoc
```diff
@@ -1,24 +1,26 @@
...
-
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
 import { B1 } from 'b';
-import { B2 } from 'b'; // Noncompliant
+import { B2 } from 'b'; // Noncompliant: there is already an import from module 'b'.
 ----
 
+Instead, one should consolidate the imports into a single statement. By consolidating all imports from the same module in a single `import` statement, the code becomes more concise and easier to read, as there is only one import statement to keep track of. Additionally, it can make it easier to identify which modules are used in the code.
```
yassin-kammoun-sonarsource, line 11:
```suggestion
Instead, one should consolidate the same imports into a single statement. By consolidating all imports from the same module in a single `import` statement, the code becomes more concise and easier to read, as there is only one import statement to keep track of. Additionally, it can make it easier to identify which modules are used in the code.
```

yassin-kammoun-sonarsource (approved):
Looks good to me.

I wonder whether we should update the rule title in the JSON file from

`Imports from the same modules should be merged`

to

`Imports from the same module should be merged`

WDYT?

PR #2452, opened 2023-07-11
Modify rule S4634: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4634/javascript/rule.adoc
+++ b/rules/S4634/javascript/rule.adoc
@@ -72,15 +72,12 @@ When you have a promise chain and want to introduce an intermediate step with an
 
 [source,javascript]
 ----
-fetchData()
-  .then(data => processData(data))
-  .then(processedData => Promise.resolve(processedData)) // Intermediate step with resolved value
-  .then(finalResult => {
-    console.log(finalResult);
-  })
-  .catch(error => {
-    console.error(error);
-  });
+const data = cache ? cache : fetchData();
+
+return Promise.resolve(data) // data may be a Promise or not, we need to wrap it
+    .then(data => {
+        return sanitizeData(data);
+    })
 ----
 
 Using `Promise.resolve` and `Promise.reject` is particularly useful when you want to simplify the creation of promises with immediately resolved or rejected states. They provide a cleaner and more direct approach compared to creating a new promise with the `Promise` constructor and manually calling `resolve` or `reject`.
@@ -90,6 +87,7 @@ Using `Promise.resolve` and `Promise.reject` is particularly useful when you wan
 
 * https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous[MDN - Asynchronous JavaScript]
 * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise[MDN - Promise]
+* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then[MDN - Promise.prototype.then()]
 * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve[MDN - Promise.resolve()]
 * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject[MDN - Promise.reject()]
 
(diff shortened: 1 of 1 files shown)


rules/S4634/javascript/rule.adoc
```diff
@@ -1,27 +1,29 @@
 == Why is this an issue?
 
-When a ``++Promise++`` needs to only "resolve" or "reject", it's more efficient and readable to use the methods specially created for such use cases: ``++Promise.resolve(value)++`` and ``++Promise.reject(error)++``.
+When a promise only needs to "resolve" or "reject", it's more efficient and readable to use the methods specially created for such use cases: ``++Promise.resolve()++`` and ``++Promise.reject()++``.
```
victor-diez-sonarsource, line 3: Not sure about the reasoning. I would rather start with a short explanation are Promises. And then continue with: 

Avoid converting a non-promise value to a promise with the Promise constructor if the resolved value is already known. If you have a value that is not a promise and does not depend on an asynchronous operation, but you want to use it in a promise chain, you can use Promise.resolve() to convert it to a promise. Similarly, use Promise.reject...

victor-diez-sonarsource (changes requested):

rules/S4634/javascript/rule.adoc
yassin-kammoun-sonarsource: Probably too much explanation of what a promise is. Let me know if I should cut branches here and there.

victor-diez-sonarsource: no need, I would say that this rule is a perfect candidate to add a 'how to fix' tab. I would just add a noncompliant snippet in the general explanation and put all alternatives you mention under the new 'how to fix it?' title

rules/S4634/javascript/rule.adoc
```diff
@@ -1,27 +1,90 @@
...
+const promise = Promise.reject(error);
 ----
 
+If you have a condition and want to create a promise that is either resolved or rejected based on that condition, using `Promise.resolve` or `Promise.reject` helps make the code more readable and concise.
+
+[source,javascript]
+----
+function fetchData() {
+  if (shouldFetchData()) {
+    return Promise.resolve(fetchDataFromServer());
```
victor-diez-sonarsource, line 56: this example is redundant as `fetchDataFromServer()` looks like a promise already.  probably can be rewritten as:

```js
function fetchData() {
  if (cache) {
    return Promise.resolve(cache);
  } else {
  if (shouldFetchData()) {
    return fetchDataFromServer()
        .then(data => {
            cache = data;
            return data;
        })
  } else {
    return Promise.reject(new Error('Data fetch is not required'));
  }
}
```

yassin-kammoun-sonarsource: 👍

rules/S4634/javascript/rule.adoc
yassin-kammoun-sonarsource: 👍

rules/S4634/javascript/rule.adoc
```diff
@@ -1,27 +1,98 @@
...
+}
 ----
 
+When you have a promise chain and want to introduce an intermediate step with an immediately resolved value, using `Promise.resolve` allows you to continue the chain without introducing unnecessary complexity.
+
+[source,javascript]
+----
+fetchData()
+  .then(data => processData(data))
+  .then(processedData => Promise.resolve(processedData)) // Intermediate step with resolved value
```
victor-diez-sonarsource, line 77: no need for `Promise.resolve()` here neiter, thenables can resolve with values. Probably another good reference:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

The usecase in promise chains for Promise.resolve() is when you want to initiate the promise chain with something which is not a promise.

```js
const data = cache ? cache : fetchData();

return Promise.resolve(data) // data may be a Promise or not, we need to wrap it
    .then(data => {
        return sanitizeData(data);
    })

```

yassin-kammoun-sonarsource: 👍

rules/S4634/javascript/rule.adoc
yassin-kammoun-sonarsource: 👍

rules/S4634/javascript/rule.adoc
yassin-kammoun-sonarsource: 👍

victor-diez-sonarsource:
Last thing, I think we are actually the 'why is this an issue?' reason. We are defining promises and Promise.resolve(), but we are missing why we should not use the constructor for non-async values

victor-diez-sonarsource (changes requested):
Last thing, I think we are actually the 'why is this an issue?' reason. We are defining promises and Promise.resolve(), but we are missing why we should not use the constructor for non-async values

victor-diez-sonarsource (approved):

PR #2453, opened 2023-07-11
Modify rule S6491: LaYC format

marco-antognini-sonarsource:
Keep the whole description in "Why is this an issue" because reading it once should be enough and fix is trivial. Also, aimed for conciseness.

## Review

- [x] text style and tone

### DIFF
--- a/rules/S6491/cfamily/rule.adoc
+++ b/rules/S6491/cfamily/rule.adoc
@@ -25,16 +25,16 @@ This rule raises an issue when a format specification uses `0` as a custom paddi
 
 === Code examples
 
-Numeric padding also works as expected with other specifiers to change the type representation, set the precision, apply locale-specific formatting, etc.
+This example shows how custom character padding and numeric padding differ when formatting a floating-point variable `num`. It uses a sign specifier (`{plus}`) and the alternative form (`#`).
 
-The following example illustrates the difference in padding behaviors when a sign specifier (such as `{plus}`) and the alternative form (`#`) are used to format the floating-point variable `num`.
+As a side note, numeric padding also works with other specifiers to change the type representation, set the precision, apply locale-specific formatting, etc.
 
 ==== Noncompliant code example
 
 [source,cpp,diff-id=1,diff-type=noncompliant]
 ----
 auto stringify(float num) {
-  return std::format("{:0>+#5", num); // Noncompliant: custom character padding with "0"
+  return std::format("{:0>+#5}", num); // Noncompliant: custom character padding with "0"
 }
 ----
 
--- a/rules/S6491/cfamily/rule.adoc
+++ b/rules/S6491/cfamily/rule.adoc
@@ -19,7 +19,102 @@ std::format("{:0>5} {:0>#5}", -10, -1.0);
 std::format("{:05} {:#05}", -10, -1.0);
 ----
 
+== How to fix it
+
 This rule raises an issue when a format specification uses `0` as a custom padding character to right-align a value instead of using the numeric padding.
+
+=== Code examples
+
+Use the numeric padding when right-aligning a numeric variable `num` with zeros.
+
+It works with default sign and representation:
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+std::format("{:0>5}", num)
+----
+
+==== Compliant solution
+
+A compliant solution is to remove the `>`:
+
+[source,cpp,diff-id=1,diff-type=compliant]
+----
+std::format("{:05}", num)
+----
+
+Before, the `0` was used as a custom character padding. +
+Now, it means numeric padding is used.
+
+Here are some examples of outputs for these two code snippets:
+
+// Note: Asciidoc collapses multiple regular whitespaces in literals.
+// To avoid this issue in the tables below, we use EN SPACE Unicode: U+2002.
+
+[cols="1,1,1"]
+|===
+| Value of `num` | Noncompliant output | Compliant output
+
+| `10`
+| `"00010"`
+| `"00010"`
+
+| `-10`
+| `"00-10"`
+| `"-0010"`
+
+| `NAN`
+| `"00nan"`
+| `"  nan"`
+
+| `-INFINITY`
+| `"0-inf"`
+| `" -inf"`
+
+|===
+
+Numeric padding also works with a sign specifier (such as `{plus}`) and the alternative form (`#`):
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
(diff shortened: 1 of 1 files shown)


rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,29 @@
```
amelie-renard-sonarsource, line 2: I am not sure about not filling in the "How can I fix it" tab. Don't you think the user would need this information every time (s)he sees the issue message?
I think I would, but that may be because I'm not familiar with `std::format`...

amelie-renard-sonarsource (changes requested):
I thought you wanted to start with an easy one ;)

rules/S6491/cfamily/rule.adoc
marco-antognini-sonarsource: Initially, I thought "since the fix is trivial, we don't need it". But, while the fix is only about changing a character or two, it is not always trivial. So now I agree with you.

I added the "how to fix it" section with some more examples. Let me know what you think of it.

marco-antognini-sonarsource (commented):
It _looked_ easy at first 🙃 
Thanks for the great feedback.

rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,80 @@
 
 `std::format` and the related formatting functions provide two different options to pad numerical values up to a specific width:
 
-* The usual character padding that applies to many types and looks like: `std::format(“{:*>5}”, num);`. This adds as many '*' characters as necessary to reach a width of 5.
+* Custom character padding: it applies to many types and can align the text to the left, the center, or the right. +
+  For example, `std::format("{:*>5}", num)` aligns `num` to the right (`>`) by inserting as many `*` characters as necessary to reach a width of 5.
 
-* Adding the number 0 just before the width to declare numeric padding: `std::format(“{:05}”, num);`. This numeric padding adds 0 before the number and therefore takes no alignment option.
+* Numeric padding with `0`: it applies to most arithmetic types by adding `0` before the width specifier. +
+  For example, `std::format("{:05}", num)` adds enough `0` before `num` to align it to the right to reach a width of 5.
```
amelie-renard-sonarsource, line 9: The first bullet point describes the outcome of the custom character option while the second bullet point describes how to enable the numeric padding. I did not find it explicit enough so I have asked Island AI to help with that :)
```suggestion
* Numeric padding with `0`: this option is available for most arithmetic types and is enabled by adding `0` before the width specifier. +
  For example, `std::format("{:05}", num)` adds enough `0` before `num` to align it to the right and reach a width of 5.
```

rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,80 @@
...
 
-When working with numeric values, one almost always wants to use the numeric padding as it works correctly with signed numbers:
+`0` can also be used as a custom character padding.
+Not only is the syntax confusing but the results are also surprising when aligning native values to the right:
 
 [source,cpp]
 ----
-std::format("{:0>5}", -10); // Usual padding, prints 00-10
-std::format("{:05}", -10);  // Numeric padding, -0010
+// Noncompliant: the use of character padding results in "00-10"
```
amelie-renard-sonarsource, line 16: To make the two comments more consistent
```suggestion
// Noncompliant: character padding results in "00-10"
```

rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,80 @@
...
+std::format("{:0>5}", -10) // "0" is a custom character padding here.
+
+// Compliant: use numeric padding to produce "-0010"
```
amelie-renard-sonarsource, line 19: For consistency
```suggestion
// Compliant: numeric padding results in "-0010"
```

rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,80 @@
...
 
-=== Noncompliant code example
+This rule raises an issue when a format specification uses `0` as a custom padding character to right-align a value instead of using the numeric padding.
 
-[source,cpp]
+=== Code examples
+
+Numeric padding also works as expected with other specifiers to change the type representation, set the precision, apply locale-specific formatting, etc.
+
+The following example illustrates the difference in padding behaviors when a sign specifier (such as `{plus}`) and the alternative form (`#`) are used to format the floating-point variable `num`.
```
amelie-renard-sonarsource, line 31: I like the ideas of the sentences but I find them a bit hard to follow within the logic of the rule. WDYT about (with the help of Island AI):
```suggestion
This example shows how custom character padding and numeric padding differ when formatting a floating-point variable `num`. It uses a sign specifier (`{plus}`) and the alternative form (`#`).

As a side note, numeric padding also works with other specifiers to change the type representation, set the precision, apply locale-specific formatting, etc.
```

rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,80 @@
...
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-void formatted(int i, double d) {
- std::cout << std::format("{:0>5}\n", i); // Noncompliant
- std::cout << std::format("{:0>#5}\n", d); // Noncompliant
+auto stringify(float num) {
+  return std::format("{:0>+#5", num); // Custom character padding with "0"
```
amelie-renard-sonarsource, line 38:
```suggestion
  return std::format("{:0>+#5", num); // Noncompliant: custom character padding with "0"
```

rules/S6491/cfamily/rule.adoc
```diff
@@ -2,36 +2,80 @@
...
 ----
-void formatted(int i, double d) {
- std::cout << std::format("{:05}\n", i); // Compliant
- std::cout << std::format("{:#05}\n", d); // Compliant
+auto stringify(float num) {
+  return std::format("{:+#05}", num); // Numeric padding.
 }
 ----
+
+Here are some examples of outputs for these two code snippets:
```
amelie-renard-sonarsource, line 53:
```suggestion
Here are examples of outputs for the two code snippets:
```

amelie-renard-sonarsource (changes requested):
I like this new version, good work!
Some suggestions to rewrite sentences a bit (with the help of Island AI :) )

marco-antognini-sonarsource:
@amelie-renard-sonarsource Thanks for the great suggestions! I should try this AI :-)
Before asking for another review, I'll consider [this other point](https://github.com/SonarSource/rspec/pull/2503/files#r1266557731) (link external resources) from #2503.

amelie-renard-sonarsource (approved):

PR #2455, opened 2023-07-11
Modify rule S3799: Adapt to LaYC

victor-diez-sonarsource:


rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
...
-
-
-=== Noncompliant code example
-
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
-var {a: {}, b} = myObj; // Noncompliant
-function foo({first: [], second}) { // Noncompliant
+var {a: {}} = myObj; // Noncompliant: this does not create any variable
```
yassin-kammoun-sonarsource, line 8:
```suggestion
let {a: {}} = myObj; // Noncompliant: this does not define any variable 'a'
```

rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
...
 
-=== Compliant solution
+If that is not the intention, fix the left-side pattern to contain the variables to create.
 
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
-var {a = {}, b} = myObj;
-function foo({first = [], second}) {
+var {a = {}} = myObj;
```
yassin-kammoun-sonarsource, line 20:
```suggestion
let {a = {}} = myObj;
```

rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
 == Why is this an issue?
 
-Destructuring is a convenient way of extracting multiple values from data stored in (possibly nested) objects and arrays. However, it is possible to create an empty pattern that has no effect. When empty curly braces or brackets are used to the right of a property name most of the time the intent was to use a default value instead.
+Destructuring is a convenient way of extracting multiple values from data stored in (possibly nested) objects and arrays. However, it is possible to create an empty pattern that has no effect, where no variables are assigned to the resulting values.
```
yassin-kammoun-sonarsource, line 3:
```suggestion
Destructuring is a convenient way of extracting multiple values from data stored in (possibly nested) objects and arrays. It can make code more concise and expressive by directly extracting values or properties needed from arrays or objects. However, it is possible to define an empty pattern that has no effect, where no variables are bound to the destructured values.
```

rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
...
-
-=== Noncompliant code example
-
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
-var {a: {}, b} = myObj; // Noncompliant
-function foo({first: [], second}) { // Noncompliant
+var {a: {}} = myObj; // Noncompliant: this does not create any variable
+function foo({first: []}) { // Noncompliant: this does not define any parameter
```
yassin-kammoun-sonarsource, line 9:
```suggestion
function foo({p: []}) { // Noncompliant: this does not define any parameter 'p'
```

rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
...
   // ...
 }
 ----
 
+When empty curly braces or brackets are used to the right of a property name, the intent is to use a default value. To fix it, use the assignment operator (`=`).
```
yassin-kammoun-sonarsource, line 14:
```suggestion
When empty curly or square brackets are bound to a pattern with a colon (`:`), like `{ pattern: [] }` or `{ pattern: {} }`, the intent is likely to define a default value. To properly define such a default value, use the assignment operator (`=`) instead.
```

rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
...
-=== Compliant solution
+If that is not the intention, fix the left-side pattern to contain the variables to create.
 
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
-var {a = {}, b} = myObj;
-function foo({first = [], second}) {
+var {a = {}} = myObj;
+function foo({first = []}) {
```
yassin-kammoun-sonarsource, line 21:
```suggestion
function foo({p = []}) {
```

rules/S3799/javascript/rule.adoc
```diff
@@ -1,32 +1,32 @@
...
+var {a: {}} = myObj; // Noncompliant: this does not create any variable
+function foo({first: []}) { // Noncompliant: this does not define any parameter
   // ...
 }
 ----
 
+When empty curly braces or brackets are used to the right of a property name, the intent is to use a default value. To fix it, use the assignment operator (`=`).
 
-=== Compliant solution
+If that is not the intention, fix the left-side pattern to contain the variables to create.
```
yassin-kammoun-sonarsource, line 16: Let's move this sentence after the snippet and provide a specific one to illustrate this possible fix, because, to be honest, I don't understand how the fix would look like with what you wrote :)

yassin-kammoun-sonarsource (changes requested):

rules/S3799/javascript/rule.adoc
victor-diez-sonarsource: I did not want to mention 'a' explicitly, because that would only the case where the user just wanted to set default value. It can also be the case where the user forgot to fill the pattern inside the curly braces

rules/S3799/javascript/rule.adoc
victor-diez-sonarsource: same as before, I didn't want to only cover the default assignment case

rules/S3799/javascript/rule.adoc
victor-diez-sonarsource: I meant to fix the destructuring to actually create variables:

```js

let {a: {b, c}} = myObj;
```

How would you phrase it?

victor-diez-sonarsource: what about now?

rules/S3799/javascript/rule.adoc
yassin-kammoun-sonarsource: 👍

rules/S3799/javascript/rule.adoc
yassin-kammoun-sonarsource: 👍

rules/S3799/javascript/rule.adoc
yassin-kammoun-sonarsource: Perfect, thanks!

PR #2460, opened 2023-07-12
Modify rule S2990: Adapt to LaYC

victor-diez-sonarsource:


rules/S2990/javascript/rule.adoc
```diff
@@ -1,14 +1,13 @@
 == Why is this an issue?
 
-When the keyword ``++this++`` is used outside of an object, it refers to the global ``++this++`` object, which is the same thing as the ``++window++`` object in a standard web page. Such uses could be confusing to maintainers. Instead, simply drop the ``++this++``, or replace it with ``++window++``; it will have the same effect and be more readable.
+When a function is called without an explicit object context, the `this` keyword refers to the global object. This also applies when it is used outside of an object. The global `this` object refers to the global context in which the JavaScript code is executed. This can cause problems when the code is executed in different contexts, such as in a browser or in a Node.js environment, where the global object is different. Such uses could be confusing to maintainers as the actual value depends on the execution context and it can be unclear what object the "this" keyword is referring to.
 
+In JavaScript "strict mode" the global "this" will always be `undefined`.
```
yassin-kammoun-sonarsource, line 5:
```suggestion
In JavaScript's "strict mode", the global "this" will always be `undefined`.
```

rules/S2990/javascript/rule.adoc
```diff
@@ -1,14 +1,13 @@
 == Why is this an issue?
 
-When the keyword ``++this++`` is used outside of an object, it refers to the global ``++this++`` object, which is the same thing as the ``++window++`` object in a standard web page. Such uses could be confusing to maintainers. Instead, simply drop the ``++this++``, or replace it with ``++window++``; it will have the same effect and be more readable.
+When a function is called without an explicit object context, the `this` keyword refers to the global object. This also applies when it is used outside of an object. The global `this` object refers to the global context in which the JavaScript code is executed. This can cause problems when the code is executed in different contexts, such as in a browser or in a Node.js environment, where the global object is different. Such uses could be confusing to maintainers as the actual value depends on the execution context and it can be unclear what object the "this" keyword is referring to.
```
yassin-kammoun-sonarsource, line 3: I think we could add an introductory sentence or two about the fact that `this` in JavaScript is different from other languages and that `this` behaves differently depending on the context. We can then enumerate which contexts (function, arrow, class, global) without developing them since the rule focuses only on the global context, and finally proceed with what you wrote.

Would that make sense?

yassin-kammoun-sonarsource, line 3:
```suggestion
When a function is called without an explicit object context, the `this` keyword refers to the global object. This also applies when it is used outside of an object. The global `this` object refers to the global context in which the JavaScript code is executed. This can cause problems when the code is executed in different contexts, such as in a browser or in a Node.js environment, where the global object is different. Such uses could confuse maintainers as the actual value depends on the execution context, and it can be unclear what object the ``++this++`` keyword is referring to.
```

rules/S2990/javascript/rule.adoc
```diff
@@ -1,14 +1,13 @@
...
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
-this.foo = 1;   // Noncompliant
-console.log(this.foo); // Noncompliant
+this.foo = 1;   // Noncompliant: 'this' refers to global 'this'
+console.log(this.foo); // Noncompliant: 'this' refers to global 'this'
 
 function MyObj() { 
   this.foo = 1; // Compliant
```
yassin-kammoun-sonarsource, line 13: Let's replace the operator `==` with its strict variant `===` here and in the other snippet.

rules/S2990/javascript/rule.adoc
yassin-kammoun-sonarsource, line 5: Since we are not explaining what strict mode is all about, should we at least add a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode ?

yassin-kammoun-sonarsource (changes requested):

rules/S2990/javascript/rule.adoc
victor-diez-sonarsource: probably i extended too much, it's hard to just enumerate without developing, there are many caveats

yassin-kammoun-sonarsource (approved):
Looks great!

PR #2461, opened 2023-07-12
Focus the search input on search-page load

necto:


frederic-tingaud-sonarsource (approved):
🙏
That's a tiny thing that has been bothering me many times!

PR #2462, opened 2023-07-12
Create rule 5686: Function call arguments should be passed in proper order

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone


rules/S5686/python/rule.adoc
```diff
@@ -0,0 +1,70 @@
+When the names of parameters in a function/method call match the names of the method arguments, it contributes to clearer, more readable code. However, when the names match, but are passed in a different order than the function/method arguments, it indicates a mistake in the parameter order which will likely lead to unexpected results.
```
guillaume-dequenne-sonarsource, line 1: I think the format for the high level description is more: "This rule raises an issue when..."
Maybe something like:
`This rule raises an issue when the arguments passed to a function call match the names of its parameters but are passed in a different order.`

I still think the original sentence makes sense and could be moved to the `Why is this an issue` section.

rules/S5686/python/rule.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+== Why is this an issue?
+
+Python functions typically expect arguments to be provided in a specific order to ensure the correct execution of the function. When arguments are passed in a different order, the function may receive incorrect values for its parameters, resulting in semantic discrepancies. This can lead to incorrect computations, unexpected behavior, or even errors.
```
guillaume-dequenne-sonarsource, line 5: This bit is okay, but it assumes the parameters are in the wrong order.
Maybe the code is just confusing because the naming itself is not consistent.
Not sure whether we should, and how to mention this. Happy to discuss.

rules/S5686/python/rule.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+== How to fix it
+
+Make sure that function call arguments in correct order
```
guillaume-dequenne-sonarsource, line 9: Rephrase to:
`Make sure that function call arguments are in the correct order.`

rules/S5686/python/rule.adoc
```diff
@@ -0,0 +1,70 @@
...
+speed = (1, 2)
+
+move_point(coord, speed)
+----
+
+== Resources
+
+=== Documentation
+
+* https://docs.python.org/3/tutorial/controlflow.html#defining-functions[Defining functions] - Python function understanding
```
guillaume-dequenne-sonarsource, line 47: I believe the standard way of adding links to documentation should be something like:

`* Python Documentation - https://docs.python.org/3/tutorial/controlflow.html#defining-functions[Defining functions]`

guillaume-dequenne-sonarsource (changes requested):
Left some minor comments.

rules/S5686/python/rule.adoc
```diff
@@ -1,12 +1,15 @@
...
+When the names of parameters in a function/method call match the names of the method arguments, it contributes to clearer, more readable code. However, when the names match, but are passed in a different order than the function/method arguments, it indicates a mistake in the parameter order which will likely lead to unexpected results.
 
 Python functions typically expect arguments to be provided in a specific order to ensure the correct execution of the function. When arguments are passed in a different order, the function may receive incorrect values for its parameters, resulting in semantic discrepancies. This can lead to incorrect computations, unexpected behavior, or even errors.
 
+In case of when parameters with matching names set in different order are correct from a code execution perspective, it could be confusing from code understanding and should be avoided.
+
 == How to fix it
 
-Make sure that function call arguments in correct order
+Make sure that function call arguments are in the correct order.
```
guillaume-dequenne-sonarsource, line 12: Sorry, after a re-read, this still sounds a bit off to me.
What about:
`Make sure that the arguments of this call are in the correct order.`?

I left out "function" because nothing prevents this rule from raising on methods as well.

guillaume-dequenne-sonarsource (approved):
LGTM.

maksim-grebeniuk-sonarsource:
Not actual

PR #2473, opened 2023-07-13
Modify Rule S3329: Add VB.NET rule description

zsolt-kolbay-sonarsource:
## Review

- [ ] text style and tone


Tim-Pohlmann:
While you touch this: Set the hotifx field.

mary-georgiou-sonarsource (approved):
LGTM - please set the quickfix field, as Tim mentioned, since we are touching this rule. 

I'd say that it might be partial - we can always propose when using the CreateEncryptor overload with two parameters, to transform the second to `object.IV`.

PR #2498, opened 2023-07-17
Modify rule S3686: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3686/javascript/rule.adoc
+++ b/rules/S3686/javascript/rule.adoc
@@ -1,9 +1,26 @@
 == Why is this an issue?
 
+JavaScript has a ``++new++`` keyword that is used in conjunction with constructor functions to create new instances of objects. When you use the ``++new++`` keyword with a function, it signifies that the function is intended to be used as a constructor function to create objects.
+
+Any function can be used as a constructor function by convention. Constructor functions are used to create new objects with the same structure or properties. They are typically named with an initial capital letter to distinguish them from regular functions.
+
+[source,javascript]
+----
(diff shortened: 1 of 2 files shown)


rules/S3686/javascript/rule.adoc
```diff
@@ -1,9 +1,26 @@
 == Why is this an issue?
 
-Constructor functions, which create new object instances, must only be called with ``++new++``. Non-constructor functions must not. Mixing these two usages could lead to unexpected results at runtime.
+JavaScript has a ``++new++`` keyword that is used in conjunction with constructor functions to create new instances of objects. When you use the ``++new++`` keyword with a function, it signifies that the function is intended to be used as a constructor function to create objects.
```
victor-diez-sonarsource, line 3:
```suggestion
JavaScript has the ``++new++`` keyword that is used in conjunction with constructor functions to create new instances of objects. When you use the ``++new++`` keyword with a function, it signifies that the function is intended to be used as a constructor function to create objects.
```

victor-diez-sonarsource: just to avoid ambiguity when reading, "Javascript has a new keyword"

victor-diez-sonarsource (approved):

rules/S3686/javascript/rule.adoc
yassin-kammoun-sonarsource: Ah, indeed :)

PR #2500, opened 2023-07-17
Modify rule S4043: Adapt to LaYC

victor-diez-sonarsource:


rules/S4043/javascript/rule.adoc
```diff
@@ -1,32 +1,55 @@
...
+const sorted = b.toSorted();
 ----
-const reversed = [...a].reverse();  // spread the contents of 'a' into a new array, so reverse doesn't impact 'a'
-const reversed2 = a.toReversed(); // ES2023 method that copies and reverses the array
-a.reverse(); // reverse the array in place
 
-const sorted = [...b].sort(); // spread the contents of 'b' into a new array, so sort doesn't impact 'b'
-const sorted2 = b.toSorted(); // ES2023 method that copies and sorts the array
-b.sort(); // sort the array in place
+Alternatively, change a mutating method into a non-mutating alternative using the `spread syntax` or `slice()` to create a copy first.
```
yassin-kammoun-sonarsource, line 29:
```suggestion
Alternatively, change a mutating method into a non-mutating alternative using the spread syntax (`...`) or `slice()` to create a copy first.
```

rules/S4043/javascript/rule.adoc
```diff
@@ -1,32 +1,55 @@
...
+
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
+const reversed = [...a].reverse();  
+const sorted = [...b].sort();
+----
+[source,javascript,diff-id=1,diff-type=compliant]
```
yassin-kammoun-sonarsource, line 36: I am surprised that using the same diff-id multiple times is allowed. Does the tool allow it? I am asking because I vaguely remember facing an issue with that.

rules/S4043/javascript/rule.adoc
```diff
@@ -1,32 +1,55 @@
...
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
-const reversed = a.reverse(); // Noncompliant
-const sorted = b.sort(); // Noncompliant
+a.reverse();
+b.sort();
 ----
 
+Or use non-mutating alternatives ``toReversed`` and ``toSorted``
```
yassin-kammoun-sonarsource, line 21:
```suggestion
Or use non-mutating alternatives ``toReversed`` and ``toSorted``.
```

yassin-kammoun-sonarsource (approved):

rules/S4043/javascript/rule.adoc
victor-diez-sonarsource: tbh, I took for granted that it's possible, let put those check in our todo list when checking UI at the end of the sprint

yassin-kammoun-sonarsource: 👍

PR #2502, opened 2023-07-17
Modify rule S955: LaYC format

amelie-renard-sonarsource:
## Review

- [x] text style and tone


michael-jabbour-sonarsource (approved):
LGTM! I agree with the new wording that this is "implementation-defined" rather than "undefined" behavior. This has been the case [since C++11 ](https://timsong-cpp.github.io/cppwp/n3337/lex.header#2). This MISRA wording is based on older versions of the standard and this seems to have been also updated in MISRA C++ 2023. :smile:

PR #2503, opened 2023-07-17
Modify rule S6490: LaYC format

marco-antognini-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S6490/cfamily/rule.adoc
+++ b/rules/S6490/cfamily/rule.adoc
@@ -1,23 +1,27 @@
 == Why is this an issue?
 
-With `std::format` and the related formatting functions,
-you can control many aspects of the textual representation of a value.
-Its width and precision can be specified in two ways:
+The `std::format` function and related formatting functions allow you to control how to display a value as text, including its width and precision.
 
-* Statically: the desired values are directly embedded in the format specification. +
-+
-In the following example, the padding character is `_`, the alignment is to the right (`>`), the width specifier is `10`, and the precision specifier is `.2` with a fixed decimal notation (`f`).
-The formatted floating-point value (3.14159) is displayed with two digits after the decimal point and padded with underscores to a minimum width of 10 characters.
-The resulting string is `+"______3.14"+`.
-+
+For example, you can convert the number `3.14159` to a string with a minimum width of `10` and a precision of `2` using:
 [source,cpp]
 ----
 std::format("{:_>10.2f}", 3.14159)
 ----
 
-* Dynamically: `{}` or `{arg-id}` is used to refer to a complementary function argument. +
-+
-The next example results in the same value as in the previous example.
+The format string uses:
+
+* `_` as the padding character
+* `>` to align the value to the right
+* `10` to specify the width
+* `.2` to specify the precision
+* `f` to use the fixed decimal notation
+
+The resulting string is `+"______3.14"+`.
+
+Furthermore, there are two ways to specify the width and the precision:
+
+* Statically, by embedding the desired values in the format specification, as in the previous example.
+* Dynamically, by referring to complementary function arguments using `{}` or `{arg-id}`. The previous example becomes:
 +
 [source,cpp]
 ----
@@ -29,7 +33,11 @@ On the other hand, using dynamic values can result in runtime exceptions and une
 
 Specifically, a `std::format_error` exception is thrown when a dynamic value for the width or the precision specifiers is either negative or not a built-in integer.
 
-This rule focuses on the latter problem: it raises an issue when the specifier used for the width or the precision is not an integer.
+== How to fix it
+
+The rule raises an issue when the width or precision specifier is not an integer.
+
+=== Code examples
 
 [source,cpp]
 ----
--- a/rules/S6490/cfamily/rule.adoc
+++ b/rules/S6490/cfamily/rule.adoc
@@ -6,19 +6,18 @@ The value of each of these options can be specified:
 * statically, by placing the desired value in a format specification,
 * or dynamically, by using `{}` or `{arg-id}` to refer to a complementary function argument.
 
-A `std::format_error` exception is thrown when a dynamic value for the width or the precision is either not a built-in integer or negative.
+A `std::format_error` exception is thrown when a dynamic value for the width or the precision specifiers is either not a built-in integer or negative.
 
-This rule focuses on the first problem and raises an issue when the argument used for the width or precision is not an integer.
+This rule focuses on the first problem: it raises an issue when the specifier used for the width or the precision is not an integer.
 
 === Noncompliant code example
 
 [source,cpp]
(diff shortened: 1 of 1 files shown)


rules/S6490/cfamily/rule.adoc
```diff
@@ -1,34 +1,36 @@
 == Why is this an issue?
```
amelie-renard-sonarsource, line 1: The rule raises an issue when the user specifies a width or a precision that is not an integer. It seems like a rookie mistake from somebody who did not understand what options he set or what these options do.
So, IMO, we should add an explanation here about these options (maybe based on what is described here: https://en.cppreference.com/w/cpp/utility/format/formatter#Standard_format_specification)
(BTW, I would add this link to the resources section)

amelie-renard-sonarsource (changes requested):

rules/S6490/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks, that's a good point. I'll try to improve this.

Re. the resource section: Initially, I didn't want to add it because I felt the page miss one big introductory explanation (or a link to it): where should I write these specifiers? The page for [`std::format`](https://en.cppreference.com/w/cpp/utility/format/format) introduces the syntax `{ arg-id (optional) : format-spec }`. Maybe a good solution is to link both pages... I'll see what I can come up with, and we may apply the same in #2453.

marco-antognini-sonarsource: I reworked the rule description, any feedback is appreciated.

rules/S6490/cfamily/rule.adoc
```diff
@@ -1,34 +1,65 @@
...
+}
 ----
 
+== Resources
+
+=== Documentation
+
+* {cpp} reference -
+  https://en.cppreference.com/w/cpp/utility/format/format[`std::format`] and
+  https://en.cppreference.com/w/cpp/utility/format/formatter#Standard_format_specification[standard format specification]
```
marco-antognini-sonarsource, line 65: I wasn't sure if I should put it on two lines... Happy to add this one to #2453 (with any adjustments).

amelie-renard-sonarsource: I would put it on two lines as it is what we usually do:
```suggestion
* {cpp} reference -
  https://en.cppreference.com/w/cpp/utility/format/format[`std::format`]
* https://en.cppreference.com/w/cpp/utility/format/formatter#Standard_format_specification[standard format specification]
```

rules/S6490/cfamily/rule.adoc
amelie-renard-sonarsource: I think the content is good, it gives the user all the necessary information.
However, I would rewrite it a little to make it easier to follow. I have written some suggestions, to give you an idea of what I have in mind.

rules/S6490/cfamily/rule.adoc
```diff
@@ -1,34 +1,65 @@
...
+This rule focuses on the latter problem: it raises an issue when the specifier used for the width or the precision is not an integer.
 
 [source,cpp]
 ----
-std::cout << std::format("{0:*^{1}}", s, 5) << std::endl; // Compliant 
-std::cout << std::format("{0:*^10.3}", d) << std::endl; // Compliant
-std::cout << std::format("{0:*^{1}.{2}}", d, 12, 4) << std::endl; // Compliant
-std::cout << std::format("{0:*^17.5}", s, "17", 5.0) << std::endl; // Compliant
-std::cout << std::format("{0:*^{2}.{3}} {1}", s, d, maxLen, maxLen + 2) << std::endl; // Compliant
+void example(std::string_view s, double d, int maxLen) {
```
amelie-renard-sonarsource, line 36: I would use the second tab here:
```suggestion
== How to fix it

The rule raises an issue when the width or precision specifier is not an integer.

=== Code examples

==== Noncompliant code example
[source,cpp]
----
void example(std::string_view s, double d, int maxLen) {
// ...
```

rules/S6490/cfamily/rule.adoc
```diff
@@ -1,34 +1,65 @@
...
-* statically, by placing the desired value in a format specification
-* or dynamically, by using `{}` or `{arg-id}` to refer to a complementary function argument.
-
-When dynamic values are used either for the width or precision, an `std::format_error` exception is thrown if the corresponding argument is not a built-in integer.
-This rule raises an issue when the argument used for the width or precision is not an integer.
-
-=== Noncompliant code example
+* Statically: the desired values are directly embedded in the format specification. +
++
+In the following example, the padding character is `_`, the alignment is to the right (`>`), the width specifier is `10`, and the precision specifier is `.2` with a fixed decimal notation (`f`).
```
amelie-renard-sonarsource, line 9: The paragraph contains a lot of useful information, but I don't think it's easy to follow are we are in the middle of understanding how the static and dynamic notations work

rules/S6490/cfamily/rule.adoc
```diff
@@ -1,34 +1,65 @@
 == Why is this an issue?
 
-`std::format` and the related formatting functions provide the possibility to specify the desired length
-and padding of the textual representation of the argument through the width and precision options.
-The value of each of these options can be specified:
+With `std::format` and the related formatting functions,
```
amelie-renard-sonarsource, line 3: I would start by explaining the format string:
```suggestion
The `std::format` function and related formatting functions allow you to control how a value is displayed as text, including its width and precision.

For example, you can modify the number `3.14159` to have a minimum width of `10` and a precision of `2` using the format string "{:_>10.2f}".
[source,cpp]
----
std::format("{:_>10.2f}", 3.14159)
----

The format string uses:

* `_` as the padding character
* `>` to align the value to the right
* `10` to specify the width
* `.2` to specify the precision
* `f` to indicate fixed decimal notation

The resulting string is `+"______3.14"+`.
```

rules/S6490/cfamily/rule.adoc
```diff
@@ -1,34 +1,65 @@
...
+you can control many aspects of the textual representation of a value.
+Its width and precision can be specified in two ways:
```
amelie-renard-sonarsource, line 5: And then I would mention the two possible notations:
```suggestion
The width and the precision can be specified in two different ways:

* Statically, by embedding the desired values in the format specification, as in the previous example.
* Dynamically, by referring to complementary function arguments using `{}` or `{arg-id}`. +
With dynamic notation, the previous example becomes:
+
[source,cpp]
----
std::format("{0:_>{1}.{2}f}", 3.14159, 10, 2)
----
```

amelie-renard-sonarsource (changes requested):

amelie-renard-sonarsource (approved):
LGTM :D
Good job!

PR #2505, opened 2023-07-17
Modify rule S6459: LaYC format

michael-jabbour-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S6459/cfamily/rule.adoc
+++ b/rules/S6459/cfamily/rule.adoc
@@ -87,5 +87,5 @@ auto func2(T&& arg)
 
  * CPP reference - https://en.cppreference.com/w/cpp/language/reference#Forwarding_references[Forwarding references]
  * CPP reference - https://en.cppreference.com/w/cpp/language/constraints[Constraints and concepts]
- * CPP reference - https://en.cppreference.com/w/cpp/concepts/copy_constructible[`std::copy_constructible` concept]
- * CPP reference - https://en.cppreference.com/w/cpp/concepts/copyable[`std::copyable` concept]
+ * CPP reference - https://en.cppreference.com/w/cpp/concepts/copy_constructible[std::copy_constructible concept]
+ * CPP reference - https://en.cppreference.com/w/cpp/concepts/copyable[std::copyable concept]
--- a/rules/S6459/cfamily/rule.adoc
+++ b/rules/S6459/cfamily/rule.adoc
@@ -87,5 +87,5 @@ auto func2(T&& arg)
 
  * CPP reference - https://en.cppreference.com/w/cpp/language/reference#Forwarding_references[Forwarding references]
  * CPP reference - https://en.cppreference.com/w/cpp/language/constraints[Constraints and concepts]
- * CPP reference - https://en.cppreference.com/w/cpp/concepts/copy_constructible[std::copy_constructible concept]
- * CPP reference - https://en.cppreference.com/w/cpp/concepts/copyable[std::copyable concept]
+ * CPP reference - https://en.cppreference.com/w/cpp/concepts/copy_constructible[`std::copy_constructible` concept]
+ * CPP reference - https://en.cppreference.com/w/cpp/concepts/copyable[`std::copyable` concept]
--- a/rules/S6459/cfamily/rule.adoc
+++ b/rules/S6459/cfamily/rule.adoc
@@ -42,7 +42,7 @@ void f() {
 template <std::copy_constructible T> void func(T&& t);  // Overload #1.
 template <typename T> void func(T&& t);  // Overload #2 (unconstrained).
 
-int main() {
+void f() {
     // Call with an rvalue argument:
     // Calls #2 (no surprise): T is 'std::unique_ptr<int>', which doesn't satisfy 'std::copy_constructible'.
     func(std::make_unique<int>(42));
--- a/rules/S6459/cfamily/rule.adoc
+++ b/rules/S6459/cfamily/rule.adoc
@@ -2,41 +2,110 @@
 
 _Type-constraints_ provide a terse way to express constraints on the type deduced for a given template parameter or auto placeholder.
 In a situation when a type-constraint is applied to a forwarding reference parameter (`T&&`), the corresponding concept will be checked
-against the _lvalue reference_ (if the argument is an _lvalue_) or the plain type (if the argument is an _rvalue_).
-
-Even if it is possible to write a check that works for both plain types and references, it requires a dedicated effort,
-and a naive attempt may silently fail for one or the other. For instance, a `std::copyable` constraint is never satisfied for references,
-regardless of the referenced type, while a `std::copy_constructible` constraint always is.
-
-This rule detects forwarding reference parameters that are constrained by the standard-provided concepts using _type-constraint_ syntax.
-
-=== Noncompliant code example
+against the _lvalue reference_ (if the argument is an _lvalue_) or the plain type (if the argument is an _rvalue_):
 
 [source,cpp]
 ----
-auto func(std::copy_constructible auto&& arg) // noncompliant
-{ /* … */ }
+template <SomeConcept T> void func(T&& x);
+void f() {
+  func(SomeType{});  // Argument is an rvalue -> T is deduced as 'SomeType'
+  SomeType obj;
+  func(obj);  //  Argument is lvalue -> T is deduced as 'SomeType&'
+}
+----
 
+Even if it is possible to write `SomeConcept` in a way that works for both plain types and references, it requires a dedicated effort and a
+naive attempt may silently fail for one or the other.
+
+Many standard-provided constraints change their behavior when the type is a reference. For instance, a `std::copyable` constraint is never
+satisfied for references, regardless of the referenced type, while a `std::copy_constructible` constraint always is. Consider the following
+examples:
+
+[source,cpp]
+----
+template <std::copyable T> void func(T&& t);  // Overload #1.
+template <typename T> void func(T&& t);  // Overload #2 (unconstrained).
+
+void f() {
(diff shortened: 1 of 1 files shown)


rules/S6459/cfamily/rule.adoc
```diff
@@ -2,41 +2,110 @@
...
+    // Call with an lvalue argument:
+    std::string s{""};
+    // Calls #2! T is 'std::string&' which doesn't satisfy 'std::copyable'.
+    func(s);
+}
+----
+
+[source,cpp]
+----
+template <std::copy_constructible T> void func(T&& t);  // Overload #1.
```
amelie-renard-sonarsource, line 42: I like the idea of illustrating that standard-provided constraints do not work well with forwarding references. Still, I think one example is enough for the user to understand the point.

rules/S6459/cfamily/rule.adoc
```diff
@@ -2,41 +2,110 @@
...
+template <typename T> void func(T&& t);  // Overload #2 (unconstrained).
+
+void f() {
+    // Call with an rvalue argument:
+    // Calls #1 (no surprise): T is 'std::string', which satisfies 'std::copyable'.
+    func(std::string{""});
+    // Call with an lvalue argument:
+    std::string s{""};
+    // Calls #2! T is 'std::string&' which doesn't satisfy 'std::copyable'.
+    func(s);
```
amelie-renard-sonarsource, line 36: I would avoid writing "no surprise" as it may make the user feel stupid if it is not obvious.
```suggestion
    // Call with an rvalue argument:
    func(std::string{""}); // Calls #1: T is 'std::string', which satisfies 'std::copyable'.
    
    // Call with an lvalue argument:
    std::string s{""};
    func(s);  // Calls #2: T is a reference type ('std::string&') which never satisfies 'std::copyable'.
```

rules/S6459/cfamily/rule.adoc
```diff
@@ -2,41 +2,110 @@
...
+void f() {
+  func(SomeType{});  // Argument is an rvalue -> T is deduced as 'SomeType'
+  SomeType obj;
+  func(obj);  //  Argument is lvalue -> T is deduced as 'SomeType&'
+}
+----
 
-This rule detects forwarding reference parameters that are constrained by the standard-provided concepts using _type-constraint_ syntax.
+Even if it is possible to write `SomeConcept` in a way that works for both plain types and references, it requires a dedicated effort and a
+naive attempt may silently fail for one or the other.
```
amelie-renard-sonarsource, line 18: I am not fond of the word "naive". What about:
```suggestion
Even if it is possible to write `SomeConcept` in a way that works for both plain types and references, it is not straightforward and requires a dedicated effort.
```

amelie-renard-sonarsource (changes requested):
Nice work! I think the content you added will be helpful for the user.

rules/S6459/cfamily/rule.adoc
```diff
@@ -2,41 +2,110 @@
...
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+auto func(std::copy_constructible auto&& arg)
+{ /* … */ }
+----
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Looks much better. Fixed in 67dd3536861f6a0ddf1bcb2b31f4cb89f75665a6. :smile:

rules/S6459/cfamily/rule.adoc
```diff
@@ -2,41 +2,110 @@
...
+=== Exceptions
+
+The `std::ranges::range` concept and its refinements (like `std::ranges::forward_range`, `std::ranges::bidirectional_range`)
+are designed to handle forwarding references parameters, and will not raise issues for this rule.
+
+== How to fix it
 
+Either wrap the deduced type in `std::remove_cvref_t` and apply the standard-provided constraint on the result in a `requires` clause, or
+design a custom constraint that works equally for reference and non-reference types (in case you have to deal with this problem often for
+a certain constraint).
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Good point, I slightly changed the suggested wording, and addressed that in 273878771dda02ba5bc937bf3dd56322315b4a06.

rules/S6459/cfamily/rule.adoc
```diff
@@ -2,41 +2,110 @@
 
 _Type-constraints_ provide a terse way to express constraints on the type deduced for a given template parameter or auto placeholder.
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Agreed, changed in 54f20a8c259c1aca8fed58b1e56cbd7fcb243d3a.

rules/S6459/cfamily/rule.adoc
michael-jabbour-sonarsource: You are right, I removed the second example in 7268c96e7f34b5aeecad638646b025ffbaba2bea.

rules/S6459/cfamily/rule.adoc
```diff
@@ -36,7 +80,11 @@ auto func2(T&& arg)
...
 
-=== Exceptions
 
-The `std::ranges::range` concept and its refinements (like `std::ranges::forward_range`, `std::ranges::bidirectional_range`)
-are designed to handle forwarding references parameters, and will not raise issues for this rule.
+== Resources
+
+=== Documentation
+
+ * CPP reference - https://en.cppreference.com/w/cpp/language/reference#Forwarding_references[Forwarding references]
```
amelie-renard-sonarsource, line 88: What about adding a link to the c++ constraints and concepts page?
```suggestion
 * CPP reference - https://en.cppreference.com/w/cpp/language/reference#Forwarding_references[Forwarding references]
 * CPP reference - https://en.cppreference.com/w/cpp/language/constraints[Constraints and concepts]
```

rules/S6459/cfamily/rule.adoc
```diff
@@ -1,32 +1,76 @@
...
-auto func(std::copy_constructible auto&& arg) // noncompliant
+template <std::copyable T> void func(T&& t);  // Overload #1.
+template <typename T> void func(T&& t);  // Overload #2 (unconstrained).
+
+void f() {
+    // Call with an rvalue argument:
+    func(std::string{""});  // Calls #1: T is 'std::string', which satisfies 'std::copyable'.
+    // Call with an lvalue argument:
+    std::string s{""};
+    func(s);  // Calls #2: T is a reference type ('std::string&') and therefore doesn't satisfy 'std::copyable'.
```
amelie-renard-sonarsource, line 34:
```suggestion
    func(s);  // Calls #2: T is a reference type ('std::string&') and therefore does not satisfy 'std::copyable'.
```

rules/S6459/cfamily/rule.adoc
```diff
@@ -1,32 +1,76 @@
...
+  issue with a specific constraint.
+
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+auto func(std::copy_constructible auto&& arg)
```
amelie-renard-sonarsource, line 61: I would add the "noncompliant" comment to highlight that the same block contains two different examples:
```suggestion
auto func(std::copy_constructible auto&& arg) // Noncompliant
```

rules/S6459/cfamily/rule.adoc
```diff
@@ -1,32 +1,76 @@
...
 { /* … */ }
 
 template<std::copyable T>
-auto func2(T&& arg)  // noncompliant
+auto func2(T&& arg)
```
amelie-renard-sonarsource, line 65:
```suggestion
auto func2(T&& arg) //  Noncompliant
```

rules/S6459/cfamily/rule.adoc
```diff
@@ -1,32 +1,76 @@
...
+    func(s);  // Calls #2: T is a reference type ('std::string&') and therefore doesn't satisfy 'std::copyable'.
+}
+----
+
+The rule raises an issue when a forwarding reference parameter is constrained by a standard-provided concept using _type-constraint_ syntax.
+
+=== Exceptions
+
+The `std::ranges::range` concept and its refinements (like `std::ranges::forward_range`, `std::ranges::bidirectional_range`)
+are designed to handle forwarding reference parameters and will not raise issues for this rule.
```
amelie-renard-sonarsource, line 43: We usually phrase it the following way:
```suggestion
The rule does not raise an issue for the `std::ranges::range` concept and its refinements (like `std::ranges::forward_range`, `std::ranges::bidirectional_range`) which are designed to handle forwarding reference parameters.
```
i.e.: 
the rule raises issues or not 
vs 
the behavior raises issues or not

But your sentence is correct, so I let you choose what you prefer :)

amelie-renard-sonarsource (changes requested):
Some nitpicks, looks good already!

rules/S6459/cfamily/rule.adoc
michael-jabbour-sonarsource: Good point

amelie-renard-sonarsource:
I would keep the backticks: see the example in the [docs](https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc)
Maybe indeed some of our rules don't follow this guideline, but they should :)

michael-jabbour-sonarsource:
> I would keep the backticks: see the example in the [docs](https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc) Maybe indeed some of our rules don't follow this guideline, but they should :)

True, thanks for pointing that out.

amelie-renard-sonarsource (approved):

PR #2512, opened 2023-07-18
Modify rule S956: LaYC format

amelie-renard-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S956/cfamily/rule.adoc
+++ b/rules/S956/cfamily/rule.adoc
@@ -1,34 +1,26 @@
 == Why is this an issue?
 
-These are the only forms for the ``++#include++`` directive permitted by the standard. The behavior is undefined when other forms are used.
+The `#include` directive is a preprocessor directive that tells the compiler to insert the contents of a file in the source code.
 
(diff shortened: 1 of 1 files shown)


rules/S956/cfamily/rule.adoc
```diff
@@ -1,36 +1,28 @@
 == Why is this an issue?
 
-These are the only forms for the ``++#include++`` directive permitted by the standard. The behavior is undefined when other forms are used.
+The `#include` directive is a preprocessor directive that tells the compiler to insert the contents of a file in the source code.
 
+However, the standard only allows the `#include` directive to be followed by angle brackets (`<filename.h>`) or double quotes (`"filename.h"`).
 
-=== Noncompliant code example
+If the `#include` contains macro names, the replacement list of preprocessing tokens must also follow this rule:
```
michael-jabbour-sonarsource, line 7:
```suggestion
If the `#include` directive contains macro names, then the result of expanding them must also follow this rule:
```
I didn't know about "replacement-lists" (which seems to be the term used by the standard to describe the `#define` directive grammar). Maybe macro expansion is more common here? WDYT?

amelie-renard-sonarsource: I totally agree, thanks!

PR #2524, opened 2023-07-18
Modify rule S2871: Fix snippet format

yassin-kammoun-sonarsource:


victor-diez-sonarsource:
wrong title? :P

victor-diez-sonarsource (approved):
oops, i missed that, thx!

PR #2528, opened 2023-07-18
Modify rule S3416: add C# and VB.NET

antonioaversa:
## Review

- [ ] text style and tone

# Details

## Testing frameworks included in the rule:
- Microsoft Extensions Logging
- Apache log4net
- NLog

### System.Diagnostics.Tracing.EventSource (EWT Logging)
  - No mechanism defining logging name

### System.Diagnostics.Trace
  - No mechanism defined for source
  - `System.Diagnostics.TraceSource` should be used instead

### System.Diagnostics.TraceSource
  - Doesn't seem relevant: people seem not to respect the equation "TraceSource name == class name"
  - Sometimes dedicated classes for logging are created:
    - see https://sourcegraph.com/github.com/mono/mono/-/blob/mcs/class/referencesource/System.Workflow.ComponentModel/AuthoringOM/Trace.cs?L61
    - Unlike for `ILogger.CreateLogger`, the name used is often non-fully qualified
    - see  https://sourcegraph.com/github.com/dotnet/roslyn/-/blob/src/Features/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/BrokeredServices/BrokeredServiceTraceListener.cs?L24
    - sometimes `nameof` is used
    - I am afraid including this scenario would make the rule quite noisy

```cs
class EnclosingType
{
	private static readonly TraceSource _source = new TraceSource("EnclosingType");
}
```

## Further analysis on Log4net

Some people do the following:

```cs
private static readonly ILog log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
```

Ref: https://stackoverflow.com/a/43074497

All GetLogger overloads are defined [here](https://logging.apache.org/log4net/release/sdk/html/T_log4net_LogManager.htm):
- Public methodStatic member GetLogger(String)	
- Public methodStatic member GetLogger(Type)	
- Public methodStatic member GetLogger(Assembly, String)	
- Public methodStatic member GetLogger(Assembly, Type)	
- Public methodStatic member GetLogger(String, String)	
- Public methodStatic member GetLogger(String, Type)	

The following methods have been excluded:
`GetLogger(String)` and `GetLogger(String, String)`: They are often used to have loggers with custom names:
  - Ref: https://sourcegraph.com/search?q=context:global+LogManager.GetLogger+lang:c%23+&patternType=standard&sm=1&groupBy=repo
 
`GetLogger(String, Type)`, `GetLogger(Assembly, Type)`
  - If the user has decided to go for a more sophisticated way of retrieving a logger
  - they probably have a complex use case, which justifies the non-standard use.

Not sure about `GetLogger(Assembly, String)`
  - Not clear to me what's the role of the `repositoryAssembly` in this case
  - Not sure it's the same as for `GetLogger(Assembly, Type)`

## Further analysis on NLog

Default example:

```cs
var log = LogManager.GetCurrentClassLogger();
```

Some use `NLog.LogManager.GetLogger(GetType().ToString());`

Some use nameof, or typeof().FullName
- Ref: https://sourcegraph.com/search?q=context:global+NLog.LogManager.GetLogger+lang:c%23+&patternType=standard&sm=1&groupBy=repo

However, often names don't seem to necessarily correspond:
- Ref: https://sourcegraph.com/github.com/izhaorui/Zr.Admin.NET/-/blob/ZR.Admin.WebApi/Controllers/System/SysLoginController.cs

Some have switches between full name and simple name: 
- Ref: https://sourcegraph.com/github.com/ServiceStack/ServiceStack/-/blob/ServiceStack.Logging/src/ServiceStack.Logging.NLog/NLogLogger.cs

Some use reflection and call stack inspection:
- Ref: https://sourcegraph.com/github.com/Fr33dan/GPSaveConverter/-/blob/GPSaveConverter/LogHelper.cs

## Further analysis on Serilog
Log.Logger Serilog loggers don't have a standard way of defining logger names based on class
The doc says: 

> This is typically done once at application start-up, and the logger saved for later use by application classes. Multiple loggers can be created and used independently if required.

Ref: https://github.com/serilog/serilog/wiki/Getting-Started

The `ILogger` interface would be used, instead.


rules/S3416/rule-dotnet.adoc
```diff
@@ -0,0 +1,52 @@
...
+----
+
+However, if the convention is not in place, the override would not affect logs from `MyNamespace.EnclosingType`, since they are made via a logger with a different name.
+
+Moreover, using the same logger name for multiple types prevents the granular configuration of each type's logger, since there is no way to distinguish them in configuration.
+
+=== Exceptions
+
+The rule doesn't raise issues when custom handling of logging names is in place. For example, when:
+- the non-generic version of `ILoggerFactory.CreateLogger`, taking a string as its input parameter, is used
```
csaba-sagi-sonarsource, line 30: Here you say it does not raise for taking a string as its input parameter, however you have examples that state that
` _logger = loggerFactory.CreateLogger(typeof(AnotherType).FullName);` is noncompliant.
I either misunderstand something or it is contradictory.
+ what about nameof(AnotherType)? I guess that should also raise.

rules/S3416/vbnet/how.adoc
```diff
@@ -0,0 +1,79 @@
+== How to fix it in Microsoft Extensions Logging
```
csaba-sagi-sonarsource, line 1: Btw I know we have this hot to fix in framework stuff, but I am not entirely sure we should use it here.
Yes, we will cover different logging frameworks, but the rule is quite straight forward about what you are messing up and the fix is the same in all of them. Just that the method is called differently.
IMHO I would simplify the rule as much as possible.

csaba-sagi-sonarsource (changes requested):
I left a clarification question about the string parameter there, and also another question to think about it. The second one is completely optional, it is just my opinion.

rules/S3416/rule-dotnet.adoc
csaba-sagi-sonarsource: Yes, I agree with you for both that it should raise and also that the bullet points should be a bit more clear. 
For me, the first one is the more confusing, since it says the issue will ignore the overloads taking a string as an argument but in some cases, it does not. However, both the `typeof` and `nameof` variants depend on a Type, but at the end of the day, they are strings.

rules/S3416/vbnet/how.adoc
antonioaversa: As discussed, I have removed all logging framework-specific examples, made more generic one-fit-all examples, and simplified the How to fix section.

rules/S3416/rule-dotnet.adoc
```diff
@@ -0,0 +1,50 @@
...
+}
+----
+
+However, if the convention is not in place, the override would not affect logs from `MyNamespace.EnclosingType`, since they are made via a logger with a different name.
+
+Moreover, using the same logger name for multiple types prevents the granular configuration of each type's logger, since there is no way to distinguish them in configuration.
+
+=== Exceptions
+
+The rule doesn't raise issues when custom handling of logging names is in place, and the logger name is not derived from a `Type`.
```
csaba-sagi-sonarsource, line 29: What about nested private classes?
Is it OK in that case to use the type of the enclosing class?

rules/S3416/vbnet/why-code-example.adoc
```diff
@@ -0,0 +1,11 @@
+[source,vbnet]
+----
+Class EnclosingType
+    Private ReadOnly _logger As ILogger
+
+    Public Sub New(loggerFactory As ILoggerFactory)
+        _logger = loggerFactory.CreateLogger(Of AnotherType)()   ' Noncompliant
```
csaba-sagi-sonarsource, line 7: Not a very strong opinion but I don't like when the names start with a special char. And I am not sure how many of our rules have examples where the names start with an underscore.

csaba-sagi-sonarsource (changes requested):
I left a couple of comments + I think it would be great to add a list of frameworks that we target in the rule.

rules/S3416/vbnet/why-code-example.adoc
antonioaversa: That's a long debate :-)

All examples from Microsoft Learn, at least the ones for logging that are mentioned in the resources, follow this convention for private fields. 
Coming from Java, I always appreciated how this convention helps with constructor assignments, which have always been a bit cumbersome to me in Java, due to Java not following the convention: `this.logger = logger;`. 
But there are also people that say that the demonstrative pronoun `this` is actually more readable.

In school, I have was taught to limit the use of `this` and `that`, as they are "sloppy" and assume reader location.
And that's where my aversion to `this` probably comes from :-)

I would also mention that the underscore has the advantage of not creating casing problems in case-insensitive languages such as VB.NET, when you have both a property `Logger` and a variable or an input parameter of a method, `logger`. But that's not so relevant I think.

Either way, I have changed, because my little grammatical idiosyncrasies should stay with me only :-)

antonioaversa:
> I left a couple of comments + I think it would be great to add a list of frameworks that we target in the rule.

List of supported frameworks added.

rules/S3416/rule-dotnet.adoc
csaba-sagi-sonarsource: I was just wondering if one would actually want to do that, but yeah, it is configurable on the outer class too. So all is good!

PR #2547, opened 2023-07-19
Modify rule S2230: LaYC review [SONARJAVA-4490]

dorian-burihabwa-sonarsource:


rules/S2230/java/rule.adoc
```diff
@@ -1,17 +1,40 @@
 == Why is this an issue?
 
-Marking a non-public method ``++@Transactional++`` is both useless and misleading because Spring doesn't "see" non-``++public++`` methods, and so makes no provision for their proper invocation. Nor does Spring make provision for the methods invoked by the method it called.
+Marking a non-`public` method `@Transactional` is both useless and misleading because Spring doesn't "see" non-`public` methods, and so makes no provision for their proper invocation.
```
dorian-burihabwa-sonarsource, line 3:
```suggestion
Marking a non-`public` method `@Transactional` is both useless and misleading because Spring does not recognize non-`public` methods, and so makes no provision for their proper invocation.
```

kaufco (approved):
LGTM!

PR #2553, opened 2023-07-19
Modify rule S6223: LaYC format

marco-antognini-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S6223/cfamily/rule.adoc
+++ b/rules/S6223/cfamily/rule.adoc
@@ -12,17 +12,21 @@ This may lead to invalid program states, data corruption, or crashes.
 
 == How to fix it
 
-The rule raises an issue when you invoke `remove`, `remove_if`, or `unique` from the `std` or `std::ranges` namespace and ignore their returned value.
+The rule raises an issue when the result of `remove`, `remove_if`, or `unique` from the `std` or `std::ranges` namespace is ignored.
 
 There are essentially three ways to fix this problem:
 
-. Use the returned iterator or subrange to consider only the relevant elements.
+. Use the returned iterator or subrange to consider only the relevant elements. +
+  This solution can be sufficient.
+  It is also efficient but non-trivial and error-prone, as explained in the first example below.
 . Use the _erase-remove_ idiom to drop elements from a container. +
-  This technique consists in calling the `erase` member function of the container using the produced iterator or subrange.
+  It consists of calling the container's `erase` member function using the produced iterator or subrange.
 . Use other algorithms:
 * Since {cpp}20, `remove` and `remove_if` can be replaced with the free functions `std::erase` and `std::erase_if`, respectively.
-* The member functions of `std::list` and `std::forward_list` of the same name do erase the desired elements from the container.
-* `std::partition` and `std::stable_partition` (and their `std::ranges` siblings) reorder elements without resulting in unspecified values.
+* The member functions of `std::list` and `std::forward_list` of the same name _do_ erase the desired elements from the container.
+* `std::remove_copy` and `std::remove_copy_if` (and their `std::ranges` siblings) avoid the problem by copying the relevant values to another container. +
+  This solution requires more memory and usually requires more computation since elements are not moved but copied.
+* `std::partition` and `std::stable_partition` (and their `std::ranges` siblings) reorder elements inside the same container without resulting in unspecified values.
 
 Which technique to use depends on the context, as we illustrate with the following examples.
 
@@ -114,7 +118,7 @@ void filterOutEvenNumbers(std::vector<int>& v) {
 }
 ----
 +
-This solution is sufficient to fix the bug and is more readable than the original code.
+This solution is sufficient, reliable, and more readable than the original code.
 
 * Before {cpp}20, we can apply the _erase-remove_ idiom and use the `erase` member function of the container:
 +
@@ -126,6 +130,8 @@ void filterOutEvenNumbers(std::vector<int>& v) {
   v.erase(newEnd, v.end());
 }
 ----
++
+This solution is semantically identical to the {cpp}20 version but more verbose.
 
 ==== Noncompliant code example
 
--- a/rules/S6223/cfamily/rule.adoc
+++ b/rules/S6223/cfamily/rule.adoc
@@ -1,55 +1,171 @@
 == Why is this an issue?
 
-Despite their names, the standard remove algorithms (`std::remove`, `std::remove_if`, `std::unique`) do not remove the elements. Instead, they rearrange elements so that the preserved (not removed) elements are stored first and return an iterator to the end of that sub-sequence (the content of the remainder of the container is unspecified). The responsibility for removing the remaining elements lies in the hands of the programmer, leading to what is called the _erase-remove_ idiom.
+Despite their names, the standard remove algorithms (`std::remove`, `std::remove_if`, `std::unique`) do not erase elements from a given range.
+Instead, they shift the preserved (not removed) elements to the beginning of the range and return an iterator after the last preserved element.
+The "removed" elements have unspecified values.
 
+{cpp}20 introduced functions in the `std::ranges` namespace with the same names.
+Aside from returning a subrange instead of an iterator, they exhibit the same behavior.
 
-Ignoring the return of one of these functions is an indication of a bug, as further operations on the container may access elements with unspecified values.
+Ignoring the result of any of these functions indicates a bug: operations on the container may access elements with unspecified values.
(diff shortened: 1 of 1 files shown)


rules/S6223/cfamily/rule.adoc
```diff
@@ -1,55 +1,178 @@
...
 
-* S6165 - replacing erase-remove idiom with `std::erase`/`std::erase_if`
+=== Documentation
+
+* Wikipedia - https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom[Erase-remove idiom]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/remove[std::remove, std::remove_if]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/unique[std::unique]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/ranges/remove[std::ranges::remove, std::ranges::remove_if]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/ranges/unique[std::ranges::unique]
+* {cpp} reference - https://en.cppreference.com/w/cpp/container/vector/erase2[std::erase, std::erase_if (std::vector)]
```
michael-jabbour-sonarsource, line 174:
```suggestion
* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/remove[`std::remove`, `std::remove_if`]
* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/unique[`std::unique`]
* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/ranges/remove[`std::ranges::remove`, `std::ranges::remove_if`]
* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/ranges/unique[`std::ranges::unique`]
* {cpp} reference - https://en.cppreference.com/w/cpp/container/vector/erase2[`std::erase`, `std::erase_if` (for `std::vector`)]
```
Similar to the examples listed [here](https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc)?

rules/S6223/cfamily/rule.adoc
```diff
@@ -1,55 +1,178 @@
...
+There are essentially three ways to fix this problem:
 
-The rule raises an issue when the result of the invocation of `remove`, `remove_if` or `unique` from namespace `std` or `std::ranges` is ignored.
+. Use the returned iterator or subrange to consider only the relevant elements. +
+  This solution can be sufficient.
+  It is also efficient but non-trivial and error-prone, as explained in the first example below.
+. Use the _erase-remove_ idiom to drop elements from a container. +
+  It consists of calling the container's `erase` member function using the produced iterator or subrange.
+. Use other algorithms:
+* Since {cpp}20, `remove` and `remove_if` can be replaced with the free functions `std::erase` and `std::erase_if`, respectively.
```
michael-jabbour-sonarsource, line 25: Since there is a semantic difference between `std::remove` and `std::erase` (i.e., what if the user really wanted to do a `std::remove`?), I would instead suggest merging C++20 `std::erase` into the previous point about the "erase-remove" idiom, WDYT?

marco-antognini-sonarsource (commented):
Thanks for the nice feedback.

rules/S6223/cfamily/rule.adoc
```diff
@@ -1,55 +1,178 @@
...
+Despite their names, the standard remove algorithms (`std::remove`, `std::remove_if`, `std::unique`) do not erase elements from a given range.
+Instead, they shift the preserved (not removed) elements to the beginning of the range and return an iterator after the last preserved element.
+The "removed" elements have unspecified values.
 
+{cpp}20 introduced functions in the `std::ranges` namespace with the same names.
+Aside from returning a subrange instead of an iterator, they exhibit the same behavior.
 
-Ignoring the return of one of these functions is an indication of a bug, as further operations on the container may access elements with unspecified values.
+Ignoring the result of any of these functions indicates a bug: operations on the container may access elements with unspecified values.
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: I agree with you. The sentence you added looks much better! :smile:

rules/S6223/cfamily/rule.adoc
```diff
@@ -1,55 +1,211 @@
...
+  if (std::ranges::empty(validNames)) {
+    // No valid names.
+    return nullptr;
+  }
+
+  auto longestNameIterator = std::ranges::max_element(
+    validNames, // Correct: the following lambda is only called with valid names.
+    [](auto f1, auto f2) { return std::strlen(f1) < std::strlen(f2); }
+  );
+  return *longestNameIterator;
```
michael-jabbour-sonarsource, line 122:
```suggestion
  // Compliant: std::views::filter is used instead.
  auto validNames = names | std::views::filter([](char const* e) { return e != nullptr; });

  auto longestNameIterator = std::ranges::max_element(
    validNames, // Correct: std::strlen is only called with valid names.
    std::ranges::less{},
    [](char const* name) { return std::strlen(name); }
  );

  if (longestNameIterator == validNames.end())
    return nullptr;  // No names passed the filter.

  return *longestNameIterator;
```

Please note that I am still new to ranges, so my suggestions might be wrong. I just had the following points in mind while reading the code:

- Is the call to `std::views::all` necessary?
- As far as I understand, checking if the range is empty before calling `std::ranges::max_element` requires an extra pass, and each element has to be checked against the predicate twice as a result.
- Maybe we can use a projection function instead of custom comparator?

michael-jabbour-sonarsource (commented):
Thanks for your responses Marco. I added a two more minor comments, feel free to discard if you think they are irrelevant...

rules/S6223/cfamily/rule.adoc
```diff
@@ -1,55 +1,211 @@
...
 
+There are essentially three ways to fix this problem:
 
-The rule raises an issue when the result of the invocation of `remove`, `remove_if` or `unique` from namespace `std` or `std::ranges` is ignored.
+. Use the returned iterator or subrange to consider only the relevant elements. +
+  This solution can be sufficient.
+  It is also efficient but non-trivial and error-prone, as explained in the first example below.
+. Use the _erase-remove_ idiom to drop elements from a container.
+* It consists of calling the container's `erase` member function using the produced iterator or subrange. +
+* Since {cpp}20, `remove` and `remove_if` can be replaced with the free functions `std::erase` and `std::erase_if`, respectively.
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: love it, thanks

rules/S6223/cfamily/rule.adoc
marco-antognini-sonarsource: > Is the call to std::views::all necessary?

Indeed, it's not. We could also write it like this:

```cpp
auto validNames = std::views::filter(names, [](char const* e) { return e != nullptr; });
```

But I guess your version is more idiomatic.

> As far as I understand, checking if the range is empty before calling `std::ranges::max_element` requires an extra pass, and each element has to be checked against the predicate twice as a result.

It's fairly complex, I had to say, but as far as I understood, it's O(1). https://eel.is/c++draft/range.prim.empty#2 says:
* 2.1 is not applicable (not an array)
* 2.2 seems applicable: it results in https://en.cppreference.com/w/cpp/ranges/view_interface/empty and thus roughly in a simple `begin == sentinel_end` check.
* 2.3 is not applicable (not sized)
* 2.4 works too: `begin` is always precomputed and `end` is a sentinel.

I initially wanted to use `empty` to keep somewhat the same code structure as the noncompliant example.

> Maybe we can use a projection function instead of custom comparator?

This makes sense. Yet, it changes the code structure. So I'm not sure.

TBH, this example is a bit artificial: a for loop may be simpler to read, and it could also be rewritten as:
```cpp
char const* longestName(std::vector<char const*> const& names) {
  if (names.empty()) {
    return nullptr;
  }
  return *std::ranges::max_element(
    names,
    [](auto f1, auto f2) {
      return f1 != nullptr && f2 != nullptr && std::strlen(f1) < std::strlen(f2);
    }
  );
}
```

So in the end I'd like to use your suggestion for `names | filter` and keep the rest as close as the noncompliant code.

michael-jabbour-sonarsource: You are right, `std::ranges::empty` ends up calling `std::ranges::begin` only here, which is cached and reused inside `max_element`. Thanks for the explanation. :smile: 
I now agree with you, it is best to keep the compliant code as close as possible to the non-compliant one...

michael-jabbour-sonarsource (approved):

PR #2557, opened 2023-07-19
Modify rule S6458: LaYC format

michael-jabbour-sonarsource:
## Review

- [ ] text style and tone


michael-jabbour-sonarsource:
I didn't add diff view to this one because we provide two compliant solutions (one for C++20, and one for pre-C++20) for the first non-compliant example. I can raise this limitation in a discuss post if the reviewer agrees that it is necessary, but I want to think about a concrete proposal first. Diff views were disabled for the rest of the examples, to keep the look consistent, and I relied on manual "Compliant" / "Noncompliant" comments to highlight the relevant parts of the code for now.

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,7 +1,7 @@
 == Why is this an issue?
```
marco-antognini-sonarsource, line 1: Something that I find missing in this section is: what is the consequence of this issue?

In a nutshell, I'd say something along these lines: even if the code compiles now, introducing another template instantiation later could result in a compile-time error that is hard to understand and debug.

marco-antognini-sonarsource (changes requested):
I have a couple of open discussion points. My suggestions should probably be better phrased, so consider them just as ideas.

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,7 +1,7 @@
...
 
 _Forwarding references_ (also known as _universal references_) provide the ability to write a template that can deduce and accept any kind of reference to the object (_rvalue_/_lvalue_ _mutable_/_const_).
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Do you mean providing the syntax for a regular wrapper function that uses `std::forward`? If this is the case, I wouldn't agree because adding something like:

```cpp
template <typename T>
void func_wrapper(T&& arg) {
   // Note that arg is an lvalue here.
   // conditionally move arg, depending on whether the caller of func_wrapper provided an lvalue or an rvalue.
   func(std::forward<T>(arg));
}
```
I think such a snippet would distract a user who is already familiar with forwarding references from quickly understanding the issue the rule is trying to address. Also, we have many rules about forwarding references, such information might be a good candidate for some shared content, but I wouldn't place such information in the opening paragraph, because I would like the first few lines to be sufficient to give the necessary context for a user who is already familiar with the concepts. WDYT?

Not sure if I misinterpreted your suggestion, sorry if this were the case :smile:

rules/S6458/cfamily/rule.adoc
```diff
@@ -50,11 +53,13 @@ If `Wrapper` were a base class, those checks would become:
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Good catch, fixed in 00e47e4e06e74e30c8e904ce36dab3f17837a78c.

rules/S6458/cfamily/rule.adoc
michael-jabbour-sonarsource: Totally true, I made an attempt in 2f591ae8502f924a122047db02c3e40803d12fab.

rules/S6458/cfamily/rule.adoc
```diff
@@ -63,107 +68,132 @@ public:
...
+[source,cpp]
+----
+template<typename T>
+class OtherWrapper {
+public:
+  OtherWrapper(OtherWrapper const& other) = default;
+
+  template<typename U>
+    requires std::constructible_from<T, U>
+  OtherWrapper(U&& u) // Noncompliant: constructible_from check is not sufficient in general
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: I added more explanations around examples in e3cdb5b6118edca98f3b1c035e643db855f6cf0b.

rules/S6458/cfamily/rule.adoc
```diff
@@ -63,107 +68,132 @@ public:
...
-    : value(std::forward<Args>(args)...)
-  {} 
-
 private:
   T value;
 };
 ----
 
-=== Compliant solution
+==== Compliant solution
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Addressed in e3cdb5b6118edca98f3b1c035e643db855f6cf0b.

rules/S6458/cfamily/rule.adoc
```diff
@@ -20,23 +20,26 @@ private:
...
 Wrapper<int> const cw;
-Wrapper<int> w2(cw); // ok, calls Wrapper(Wrapper const& other)
-Wrapper<int> w3(w); // ill-formed, calls Wrapper(U&& u) with U = Wrapper<int>&
+Wrapper<int> w2(cw); // Ok: calls Wrapper(Wrapper const& other)
+Wrapper<int> w3(w); // Ill-formed: calls Wrapper(U&& u) with U = Wrapper<int>&
 ----
 
-The greediness of forwarding references is a general problem, but it is even stronger for constructors that can be called with a single argument of _forwarding reference_ type
-(including variadic template constructors), because in that case they compete with copy or move constructors. 
+The greediness of forwarding references is a general problem, but it is even more serious for constructors that can be called with a single argument of _forwarding reference_ type
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: I agree with you. I changed the phrasing a bit in c0bddbc0d16f73026fc76b409de89b609a2fce22. WDYT?

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,7 +1,7 @@
 == Why is this an issue?
 
 _Forwarding references_ (also known as _universal references_) provide the ability to write a template that can deduce and accept any kind of reference to the object (_rvalue_/_lvalue_ _mutable_/_const_).
-This enables the creation of a perfect forwarding constructor for wrapper types, which allows a wrapper to be constructed with any type that the underlying type can be constructed with:
+This enables the creation of a perfect forwarding constructor for wrapper types, which allows a wrapper to forward its constructor arguments into the initializer of the underlying type:
 [source,cpp]
 ----
 template<typename T>
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: This is a very good point. I somehow overlooked this possibility. Fixed in 0bd382de66ce3076e5ff8b08a3691a3d0f1a9d95.

rules/S6458/cfamily/rule.adoc
```diff
@@ -20,23 +20,26 @@ private:
...
-(including variadic template constructors), because in that case they compete with copy or move constructors. 
+The greediness of forwarding references is a general problem, but it is even more serious for constructors that can be called with a single argument of _forwarding reference_ type
+(including variadic template constructors) because in that case, they compete with copy or move constructors.
 
-To eliminate this pitfall, such constructors should be constrained in a way that it is not considered a candidate when `U&&` is a reference to the class type
-or to a class derived from the class type. In the previous example, this can be achieved by adding to the forwarding reference constructor:
+== How to fix it
+
+The rule reports forwarding constructors without proper constraints if they can be called with a single argument.
+To eliminate this pitfall, add constraints to such constructors so that they are not considered an overload candidate when `U&&` is a reference to the class type
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: I found the section to be already long, so I only rephrased in a2776ad905a180e8b1d23f95ab65da0717567af8.

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,169 +1,218 @@
...
+  {} 
 
 private:
   T value;
 };
+----
+
+Using `std::derived_from` instead of `std::same_as` is only meant for demonstration purposes here. `std::derived_from` is necessary only if
+`TemplateWrapper` has derived classes, to ensure that the copy constructors of these derived classes don't end up calling the forwarding
+constructor.
```
marco-antognini-sonarsource, line 174: > only meant for demonstration purposes here

I think it is fine not to go into more detail -- we would need a full blog post to explain this properly.
At the same time, I feel this whole point about inheritance deserves the attention of some of our users. What do you think about moving this example to the "Going the extra mile" section? I would structure it a bit differently, like so (very rough sentences, please adjust):

---

### Going the extra mile

One has to be careful when inheritance is involved to preserve the usual slicing semantics: the forwarding constructor should be further constrained with `!std::derived_from`:

    code example

Eric Niebler explains this in further detail in [Universal References and the Copy Constructor](https://ericniebler.com/2013/08/07/universal-references-and-the-copy-constructo/).


---

What do you think?

marco-antognini-sonarsource (commented):
Thanks for addressing my comments. I have a couple of small suggestions about style and one about the extra mile section.

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,169 +1,218 @@
...
 ----
-Wrapper<int> w;
-Wrapper<int> const cw;
-Wrapper<int> w2(cw); // ok, calls Wrapper(Wrapper const& other)
-Wrapper<int> w3(w); // ill-formed, calls Wrapper(U&& u) with U = Wrapper<int>&
+Wrapper w("str1");
+Wrapper const cw("str2");
+
+Wrapper w2(cw);  // Ok: calls Wrapper(Wrapper const& other)
+Wrapper w3(s);   // Ill-formed: calls Wrapper(T&& str) with T = Wrapper&
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: Good catch :smile:

I addressed that in 5dba9cd81f27094e373d4ecef20596cf9fd2af35. Note that I slightly changed the suggested wording to make it fit into one line, let me know what you think about it... :smile:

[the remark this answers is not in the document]
marco-antognini-sonarsource: Looks great!

rules/S6458/cfamily/rule.adoc
michael-jabbour-sonarsource: This is a great idea. I had some hard time thinking of where to properly explain this, and adding another example just for that felt unnecessary. The "Going the extra mile" section sounds relevant. I made an attempt in 8d8588399e52e153ead85f1f1e53fd99eff6e4b3.

Thanks for sharing! :bow:

michael-jabbour-sonarsource (commented):
Thanks for the nice feedback for the great ideas. I added a few commits. Let me know what you think. :smile: 

Please note that the intermediate commits sometimes don't reflect the final state, so it might be easier to do the review without looking at individual commits this time. Sorry about that...

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,169 +1,218 @@
...
 ----
 
-The greediness of forwarding references is a general problem, but it is even stronger for constructors that can be called with a single argument of _forwarding reference_ type
-(including variadic template constructors), because in that case they compete with copy or move constructors. 
+It can be argued that the greediness of forwarding references is a general problem. However, this rule specifically targets constructors
+that can be called with a single argument of _forwarding reference_ type. This is because in such cases, they can compete with copy or move
+constructors including those that are implicitly generated by the compiler.
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: Makes sense. 👍

rules/S6458/cfamily/rule.adoc
```diff
@@ -24,50 +24,52 @@ and will therefore be selected. This is usually not the expected behavior.
...
-or to a class derived from the class type. This can be achieved by adding any of the following checks to the forwarding reference constructor:
+To eliminate this pitfall, add constraints to such constructors so that they are not considered an overload candidate when the argument is
+a reference to the class itself. This can be achieved by adding any of the following checks to the forwarding reference constructor:
 
-* A check of the concept `!same_as<remove_cvref_t<U>, Wrapper>`
-* A check of type predicate  `!is_same_v<remove_cvref_t<U>, Wrapper>`, or
-* An `enable_if` with the equivalent condition.
+* a check of the concept `!std::same_as<std::remove_cvref_t<U>, Wrapper>`, or
+* a check of type predicate  `!std::is_same_v<std::remove_cvref_t<U>, Wrapper>`, or
+* an `std::enable_if` with the equivalent condition
```
marco-antognini-sonarsource, line 51:
```suggestion
* an `std::enable_if` with the equivalent condition.
```

rules/S6458/cfamily/rule.adoc
```diff
@@ -24,50 +24,52 @@ and will therefore be selected. This is usually not the expected behavior.
...
 
-If `Wrapper` were a base class, those checks would become:
+Note that special care has to be taken when `Wrapper` is a base class. This is explained in more detail in the "Going the extra mile"
```
marco-antognini-sonarsource, line 53:
```suggestion
Note that special care has to be taken when `Wrapper` is a base class. This is explained in more detail in "Going the extra mile"
```

rules/S6458/cfamily/rule.adoc
```diff
@@ -24,50 +24,52 @@ and will therefore be selected. This is usually not the expected behavior.
...
+below. In this case, those checks would become:
 
-* The concept `!derived_from<std::remove_cvref_t<U>, Wrapper>`
-* A type-predicate `!is_base_of_v<Wrapper, std::remove_cvref_t<U>>`, or
-* An `enable_if` with the equivalent condition.
+* the concept `!std::derived_from<std::remove_cvref_t<U>, Wrapper>`, or
+* a type-predicate `!std::is_base_of_v<Wrapper, std::remove_cvref_t<U>>`, or
+* an `std::enable_if` with the equivalent condition
```
marco-antognini-sonarsource, line 58:
```suggestion
* an `std::enable_if` with the equivalent condition.
```

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,169 +1,218 @@
...
+----
+class Wrapper {
+public:
+  template <typename T,
+      typename std::enable_if<
+      !std::is_same<
+        Wrapper, 
+        typename std::remove_cv<typename std::remove_reference<T>::type>::type
+      >::value, 
+      int>::type /* Unnamed */ = 0>
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: Looks good. And thanks for sharing this info.

rules/S6458/cfamily/rule.adoc
```diff
@@ -221,6 +224,53 @@ private:
 };
 ----
 
+=== Going the extra mile
+
+When the forwarding constructor belongs to a base class, using the `same_as` constraint check solely is not sufficient. The problem is that
+the forwarding constructor may still get selected when we are copying from a derived object.
```
marco-antognini-sonarsource, line 230:
```suggestion
When the forwarding constructor belongs to a base class, using the `same_as` constraint check is not sufficient:
The forwarding constructor can still get selected when we are copying from a derived object.
```

rules/S6458/cfamily/rule.adoc
michael-jabbour-sonarsource: I removed the full stop yesterday because I learned that the [guideline](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc#lists) says we should avoid them in lists. :smile:
WDYT?

rules/S6458/cfamily/rule.adoc
```diff
@@ -221,6 +224,53 @@ private:
...
+
+[source,cpp]
+----
+class Base {
+public:
+  template <typename T>
+  requires (!std::same_as<std::remove_cvref_t<T>, Base>)  // Bad: same_as is not sufficient for base classes.
```
marco-antognini-sonarsource, line 237: I'm not sure why, but "incorrect" sounds less harsh than "bad" to me. Feel free to dismiss the suggestion if you disagree.

```suggestion
  requires (!std::same_as<std::remove_cvref_t<T>, Base>)  // Incorrect: same_as is not sufficient for base classes.
```

rules/S6458/cfamily/rule.adoc
michael-jabbour-sonarsource: Same as [this comment](https://github.com/SonarSource/rspec/pull/2557/files#r1274487244). I can restore them if you think I misinterpreted the guideline here. :smile:

rules/S6458/cfamily/rule.adoc
```diff
@@ -221,6 +224,53 @@ private:
...
+  Base(T&& str) : str(std::forward<T>(str)) {}
+private:
+  std::string str;
+};
+
+class Derived : public Base {};
+----
+
+Then the following would result in a compilation error:
```
marco-antognini-sonarsource, line 246:
```suggestion
Then the following results in a compilation error:
```

michael-jabbour-sonarsource (commented):
Thanks again for the review @marco-antognini-sonarsource. I responded to your new comments. Let me know what you think. :smile:

rules/S6458/cfamily/rule.adoc
```diff
@@ -221,6 +224,53 @@ private:
...
+
+[source,cpp]
+----
+class Derived2 : public Base {
+  // [...]
+public:
+  Derived2(Derived2 const& d)
+    // d is of Derived2 type and it therefore satisfies the same_as constraint for the forwarding constructor
+    : Base(d) {  // Error: Calls the forwarding constructor instead of the base copy constructor
+    // [...]
```
marco-antognini-sonarsource, line 265: We usually use this notation:
```suggestion
    // ...
```

rules/S6458/cfamily/rule.adoc
```diff
@@ -221,6 +224,53 @@ private:
...
+// Note that the constraint is satisfied when T is Derived&
+Base b(d);  // Calls the forwarding constructor instead of the usual "slicing" behavior
+----
+
+Additionally, subclasses can run into trouble when they try to define their copy constructors:
+
+[source,cpp]
+----
+class Derived2 : public Base {
+  // [...]
```
marco-antognini-sonarsource, line 260:
```suggestion
  // ...
```

rules/S6458/cfamily/rule.adoc
marco-antognini-sonarsource: Nicely done!

rules/S6458/cfamily/rule.adoc
```diff
@@ -172,6 +224,62 @@ private:
...
+
+
 == Resources
 
-* Effective Modern {cpp} item 29: Avoid overloading on universal references
+=== Documentation
+
+* CPP reference - https://en.cppreference.com/w/cpp/utility/forward[`std::forward`]
+* CPP reference - https://en.cppreference.com/w/cpp/language/overload_resolution#Ranking_of_implicit_conversion_sequences[Ranking of implicit conversion sequences during overload resolution]
+* CPP reference - https://en.cppreference.com/w/cpp/utility/optional/optional[`std::optional` constructor]
```
marco-antognini-sonarsource, line 280: When looking at the resources section in isolation, it feels a bit weird to have a link to `std::optional`'s ctor. I'd suggest inlining this link when talking about `std::optional` and removing it from here.

```suggestion
```

rules/S6458/cfamily/rule.adoc
```diff
@@ -1,169 +1,221 @@
...
 private:
   T value;
 };
+----
+
+==== Compliant solution
 
+In this case, we can use a type tag to allow the user to explicitly choose the emplace constructor. This approach is simpler to implement
+and offers greater flexibility. It is the same approach used by many wrapper types in the standard library, such as `std::optional` and
+`std::expected`.
```
marco-antognini-sonarsource, line 207:
```suggestion
In this case, we can use a type tag to allow the user to explicitly choose the emplace constructor.
This approach is simpler to implement and offers greater flexibility.
It is the same approach used by many wrapper types in the standard library,
such as https://en.cppreference.com/w/cpp/utility/optional/optional[`std::optional`]
and https://en.cppreference.com/w/cpp/utility/expected/expected[`std::expected`].
```

rules/S6458/cfamily/rule.adoc
marco-antognini-sonarsource: I feel this guideline is misleading: from a grammar point of view, we need punctuation here. It would be different for a list of simple items (like a grocery shopping list).

marco-antognini-sonarsource (approved):
Sorry for the multiple reviews -- I think I confused GitHub somehow and it published my review too early.

Anyhow, I'm done reviewing it. It looks good, with a couple of minor grammar suggestions & other small nits.

Feel free to pick what you agree with, dismiss the rest, and merge. :)

rules/S6458/cfamily/rule.adoc
michael-jabbour-sonarsource: Good point, I was under the impression we are not allowed to pul external links outside of the "Resources" section, but you are right, a few rules seem to do that... :smile:

michael-jabbour-sonarsource:
Thanks for the nice review @marco-antognini-sonarsource. :smile:

PR #2571, opened 2023-07-20
Modify S6333(Cfn): Improve samples

loris-s-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S6333/cloudformation/rule.adoc
+++ b/rules/S6333/cloudformation/rule.adoc
@@ -12,7 +12,7 @@ A public API that doesn't have access control implemented:
 ----
 AWSTemplateFormatVersion: 2010-09-09
 Resources:
-  NoncompliantApiGatewayMethod:
+  ExampleMethod:
     Type: AWS::ApiGateway::Method
     Properties:
       AuthorizationType: NONE # Sensitive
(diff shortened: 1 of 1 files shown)


rules/S6333/cloudformation/rule.adoc
```diff
@@ -10,7 +10,9 @@ A public API that doesn't have access control implemented:
 
 [source,yaml]
 ----
-NoncompliantApiGatewayMethod:
+AWSTemplateFormatVersion: 2010-09-09
+Resources:
+  NoncompliantApiGatewayMethod:
```
hendrik-buchwald-sonarsource, line 15: I think it would be nice to use more generic names for the resources that do not change from sensitive to compliant samples. To avoid the unnecessary diff.

rules/S6333/cloudformation/rule.adoc
```diff
@@ -21,7 +23,9 @@ A Serverless Application Model (SAM) API resource that is public by default:
 
 [source,yaml]
 ----
-OpenApiDefault: # Sensitive
+AWSTemplateFormatVersion: 2010-09-09
+Resources:
+  OpenApiDefault: # Sensitive
```
hendrik-buchwald-sonarsource, line 28: I think it would be nice to use more generic names for the resources that do not change from sensitive to compliant samples. To avoid the unnecessary diff.

hendrik-buchwald-sonarsource (approved):
I would suggest to also update the resource names, otherwise it looks fine.

PR #2590, opened 2023-07-20
Modify rule S2167: reworked rule description for LayC format [SONARJAVA-4490]

kaufco:


rules/S2167/java/rule.adoc
```diff
@@ -1,29 +1,72 @@
 == Why is this an issue?
 
-It is the sign, rather than the magnitude of the value returned from ``++compareTo++`` that matters. Returning ``++Integer.MIN_VALUE++`` does _not_ convey a higher degree of inequality, and doing so can cause errors because the return value of ``++compareTo++`` is sometimes inversed, with the expectation that negative values become positive. However, inversing ``++Integer.MIN_VALUE++`` yields ``++Integer.MIN_VALUE++`` rather than ``++Integer.MAX_VALUE++``.
+The `Comparable.compareTo` method returns a negative integer, zero, or a positive integer to indicate whether the object is less than, equal to, or greater than the specified object.
```
dorian-burihabwa-sonarsource, line 3:
```suggestion
The `Comparable.compareTo` method returns a negative integer, zero, or a positive integer to indicate whether the object is less than, equal to, or greater than the parameter.
```

rules/S2167/java/rule.adoc
```diff
@@ -1,29 +1,72 @@
...
+The sign of the return value or whether it is zero is what matters, not its magnitude.
 
+Returning a positive or negative constant value other than the basic ones (-1, 0, or 1) lacks justification and indicates a misunderstanding of the principle.
+It also misleads readers to think that the concrete value of the returned constant is somehow essential here.
```
dorian-burihabwa-sonarsource, line 7: We should change the tone of this text to nudge the user into changing the value, not into highlighting their misunderstanding
```suggestion
Since the contract of `compareTo` does not assign meaning to a change in magnitude returning a positive or negative constant value other than the basic ones (-1, 0, or 1) does not bring additional information to the caller.
```

rules/S2167/java/rule.adoc
```diff
@@ -1,29 +1,72 @@
...
 
-=== Noncompliant code example
+When returning `Integer.MIN_VALUE`, an additional issue is 32-bit two's complement range overflow.
+If a caller were to negate the result of `compareTo()`, which may happen to invert the tested condition, this attempt would result in unexpected behavior where the condition is not inverted as intended.
```
dorian-burihabwa-sonarsource, line 10: Does it make sense to assume that the caller is going to use the return value when we justify the rule by saying that the caller should not use anything but the sign?

rules/S2167/java/rule.adoc
```diff
@@ -1,29 +1,72 @@
...
-=== Compliant solution
+==== Compliant solution
 
 [source,java]
 ----
-public int compareTo(MyClass) {
+public int compareTo(Name name) {
   if (condition) {
-    return -1;
+    return hashCode() - name.hashCode(); // Compliant, not a constant
```
dorian-burihabwa-sonarsource, line 58: Specifically, because we have values that can change as we make the value change over time, I would remove the text at lines 9-10

rules/S2167/java/rule.adoc
```diff
@@ -1,29 +1,72 @@
...
   if (condition) {
-    return Integer.MIN_VALUE;  // Noncompliant
+    return -1; // Compliant
   }
+}
 ----
 
+==== Compliant solution
+
+[source,java]
```
dorian-burihabwa-sonarsource, line 43: This code example and the following are missing diff-id and type attributes

dorian-burihabwa-sonarsource (changes requested):

rules/S2167/java/rule.adoc
kaufco: No diff at all, I'd like to explicitly show here which case is allowed by the rule, and which has no noncompliant counterpart. An option would be to duplicate the noncompliant code, but then we have 3 copies of it ...

rules/S2167/java/rule.adoc
kaufco: Nope. Removed as discussed.

rules/S2167/java/rule.adoc
```diff
@@ -1,29 +1,82 @@
...
+  }
+}
+----
+
+==== Compliant solution
+
+It is compliant to return other values than -1, 0 or 1 if they are not constants:
 
 [source,java]
 ----
```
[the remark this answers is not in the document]
kaufco: Fixed.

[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: Is it?

dorian-burihabwa-sonarsource (changes requested):

PR #2596, opened 2023-07-21
Modify rule S836: Adapt to LaYC

tomasz-kaminski-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S836/cfamily/rule.adoc
+++ b/rules/S836/cfamily/rule.adoc
@@ -269,7 +269,7 @@ auto [px, py, pz] = [&] {
 
 === External coding guidelines
 
-* MITRE, CWE-457 - https://cwe.mitre.org/data/definitions/457[Use of Uninitialized Variable]
+* MITRE - https://cwe.mitre.org/data/definitions/457[CWE-457 Use of Uninitialized Variable]
 * MISRA C:2004, 9.1 - All automatic variables shall have been assigned a value before being used.
 * MISRA {cpp}:2008, 8-5-1 - All variables shall have a defined value before they are used.
 
--- a/rules/S836/cfamily/rule.adoc
+++ b/rules/S836/cfamily/rule.adoc
@@ -63,7 +63,7 @@ void usingNew() {
 === What is the potential impact?
 
 Using garbage values will cause the program to behave
-non-deterministically at runtime.
+nondeterministically at runtime.
 The program may produce a different output or crash depending
 on the run.
 
@@ -88,7 +88,7 @@ object of the given type is created. Such a constructor is called even
 if a variable is declared without any initializer.
 However, if the constructor code omits the initialization
 of a member that itself does not have the default constructor,
-the member will remain uninitialized. And reading
+the member will remain uninitialized  (See also S2107). And reading
 from it will produce a garbage value:
 
 [source,cpp]
@@ -101,20 +101,21 @@ struct Partial {
   float y;
 };
 
-void initialized() {
+int initialized() {
   Partial p; // constructor is called
   // or even Partial p{};
-  return p.x; // Non-complaint: reading an uninitialized variable
+  return p.x; // Non-compliant: reading an uninitialized variable
 }
 ----
 
 
-=== Exceptions: global, static, and thread-local variables.
+=== Exceptions
 
-All the variables with static storage duration, meaning: global, static,
-or thread-local variables, are zero-initialized before the initializer
-is evaluated. As a consequence, any such variable or member of such
-object has a defined value even if no initializer is specified.
+This rule does not flag the variables with static storage duration, meaning:
+global, static, and thread-local variables.
+
+All the variables with static storage duration are zero-initialized before the initializer is evaluated.
+As a consequence, any variable or member of such an object has a defined value even if no initializer is specified.
 
 [source,cpp]
 ----
@@ -123,7 +124,7 @@ int globTab[10];
 Aggregate globAggr;
 Partial globPart; // x member is zero-initialized
 
-void uses() {
+int uses() {
   static int staticInt;
   return globInt     // Compliant: all zero-initialized
        + globTab[2]
@@ -137,7 +138,7 @@ void uses() {
 
 Commonly, the use of an uninitialized object is an indication of a defect in the
 code, where either variable initialization was skipped on some code paths, or
-the object is used by mistake. Generally, such problems can be addressed by:
+the object is used by mistake. Generally, you can address such problems by:
 
 * Initializing variables on the declaration with a valid value
 * Assigning to the variable on the code path(s) that was missing initialization
@@ -163,7 +164,7 @@ int function(int flag, int b) {
 
 ==== Compliant solution
 
-Initializing variable in all code paths:
+Initializing variable on all code paths:
 
 [source,c,diff-id=1,diff-type=compliant]
 ----
@@ -189,7 +190,7 @@ int function(int flag, int b) {
   } else {
     return 10;
   }
-  return a; // Complaint
+  return a; // Compliant
 }
 ----
 
@@ -202,7 +203,7 @@ int function(int flag, int b) {
   if (flag) {
     a = b;
   }
-  return a; // Complaint
+  return a; // Compliant
 }
 ----
 
@@ -212,7 +213,7 @@ Initializing value in the definition:
 ----
 int function(int flag, int b) {
   int const a = flag ? b : 10;
-  return a; // Complaint
+  return a; // Compliant
 }
 ----
 
@@ -220,17 +221,17 @@ int function(int flag, int b) {
 
 Initializing the variable to zero at the declaration is not always the right
 solution to fix the issue, as it may lead to logic errors if such a value
-is not handled correctly. For example, setting an `age` field of the `Employee`
-the structure may break assumptions of retirement handling code. Or more commonly,
-setting a pointer to `NULL`, will turn dereference of an uninitialized value
-into a null pointer dereference.
+is not handled correctly. For example, setting an `age` field of an `Employee`
+structure may break assumptions of retirement handling code. Or more commonly,
+setting a pointer to `NULL` will turn dereference of an uninitialized value
+into a null-pointer dereference.
 
 
 === Going the extra mile
 
 With the addition of lambdas in {cpp}11, it is possible to initialize
 variables on declaration without creating a separate function
-to the compute value:
+to compute the value:
 
 [source,cpp]
 ----
@@ -265,7 +266,7 @@ auto [px, py, pz] = [&] {
 
 === Documentation
 
-- C++ reference - https://en.cppreference.com/w/cpp/language/storage_duration[Storage class specifiers]
+* C++ reference - https://en.cppreference.com/w/cpp/language/storage_duration[Storage class specifiers]
 
 === External coding guidelines
 
--- a/rules/S836/cfamily/rule.adoc
+++ b/rules/S836/cfamily/rule.adoc
@@ -3,10 +3,10 @@ behavior due to garbage values.
 
 == Why is this an issue?
 
-A local variable of any built-in type (like `int`, `float`, and also pointers),
-declared without an initial value, is not initialized to any particular value.
-Consequently, code that uses such a variable without being previously assigned to
-does not have defined behavior. For example:
+A local variable of any built-in type (such as `int`, `float`, and pointers),
+declared without an initial value is not initialized to any particular value.
+Consequently, if no value is assigned to such a variable first, the code that
+uses it has no defined behavior.
 
 [source,cpp]
 ----
@@ -15,14 +15,14 @@ int addition() {
   return x + 10; // Noncompliant: x has grabage value
 }
 
-int derefence() {
+int dereference() {
   int* p; // p is not initialized
   return *p; // Noncompliant: p has garbage value
 }
 ----
 
 Similarly, structures that simply aggregate variables of built-in types, such as arrays or `struct`/`class`
-types without a constructor, will not initialize their members, when declared without
+types without a constructor, will not initialize their members when declared without
 an initializer:
 
 [source,cpp]
@@ -35,11 +35,11 @@ struct Aggregate {
 void aggregates() {
   int* intArray[5]; // each element of array is not initializer
   Aggregate aggr; // members aggr.i, agrr.f are not initialized
-  Aggregate aggrArray[2]; // members of each element are not initialied
+  Aggregate aggrArray[2]; // members of each element are not initialized
 }
 ----
 
-Finally, allocating objects of such builtin or aggregates types on the heap,
+Finally, allocating objects of builtin or such aggregates types on the heap,
 also does not initialize their values:
 
 [source,c]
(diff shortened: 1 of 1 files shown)


rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+    return 10;
+  }
+  return a; // Complaint
+}
+----
 
-=== Compliant solution
+Providing valid initial value:
 
 [source,cpp]
```
necto, line 197: Why did you change the language to `cpp`?

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+----
 int function(int flag, int b) {
   int a;
   if (flag) {
     a = b;
+  } else {
+    a = 10;
   }
-  return a; // Noncompliant - "a" has not been initialized in all paths
+  return a; // Complaint
```
necto, line 176:
```suggestion
  return a; // Compliant
```
Please find&replace it in other places too

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+to compute value:
+
+[source,cpp]
+----
+int x = [&] { // capture all context by reference
+  if (someCondition())
+    return computeVar1();
+  /* peform more computations */
+  return other;
+}(); // invoke lambda just (immediately)  after creation
```
necto, line 231:
```suggestion
}(); // invoke the lambda immediately (right after creation)
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+----
+
+Such pattern is referred as _Immediately invoked function expression_
+(IIFE) or _Imediately invoked lambda_.
+Furthermore, with the addition of structured binding in {cpp}17, it is
+possible to declare multiple variables whose values are coupled:
+
+[source,cpp]
+----
+auto [px, py. pz] = [&] {
```
necto, line 241:
```suggestion
auto [px, py, pz] = [&] {
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+  if (x_dir) {
+    return std::make_tuple(1.0, 0, 0);
+  } else if (y_dir) {
+    return std::make_tuple(0, 1.0, 0);
+  } else {
+    assert(z_dir);
+    return std::make_tuple(0, 0, 1.0);
```
necto, line 248:
```suggestion
    return std::make_tuple(1, 0, 0);
  } else if (y_dir) {
    return std::make_tuple(0, 1, 0);
  } else {
    assert(z_dir);
    return std::make_tuple(0, 0, 1);
```
To make it compile, and avoid the hairy `0., 1.0, 0.`

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+
+=== External coding guidelines
+
 * https://cwe.mitre.org/data/definitions/457[MITRE, CWE-457] - Use of Uninitialized Variable
 * MISRA C:2004, 9.1 - All automatic variables shall have been assigned a value before being used.
 * MISRA {cpp}:2008, 8-5-1 - All variables shall have a defined value before they are used.
 
+=== Related rules
+
+* S2107 detect fields being left uninitialized after invocation of constructor
```
necto, line 267:
```suggestion
* S2107 detects fields being left uninitialized after the invocation of a constructor
```

necto (changes requested):

tomasz-kaminski-sonarsource:
Do you think that adding Pitfall paragraph, that just initializing the value to zero is not always the right solution? Like changing an uninitialized pointer to a null pointer.

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
+Variables should be initialized before their use to avoid unexpected
+behaviors due to garbage values.
+
 == Why is this an issue?
 
-Variables should be initialized before their use to avoid unexpected behaviors due to garbage values.
+A local variable of the built-in type (like `int`, `float`, and also pointers),
+declared without initial value, is not initialized to any particular value.
+Consequently, code that uses such a variable without previously assigned to
+them does not have defined behavior. For example:
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Good suggestion.

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+  Aggregate* aggrPtr = new Aggregate; // members of allocated Aggregate are not initialized
+  Aggregate* aggrArr = new Aggregate[5]; // members of each of 5 Aggregate objects are not initialized
+}
+----
+
+
+=== What is the potential impact?
+
+In the best-case scenario, using garbage value will cause the
+program to behave non-deterministically at runtime.
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: If you ship it user, crashing it worse. But I understand this depends, and should not be in rule.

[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: I just removed it and have:
> Using garbage values will cause the program to behave
non-deterministically at runtime.

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
+The program may produce a different output or crash depending
+on the run.
+
+In some situations, loading a variable may expose a sensitive
+data, like a password, that was previously stored in the same
+location. Leading to vulnerability, where such defect
+is used as a gadget for extracting information from the instance
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: I believe that in vulnerability reasearch world, the exploit word is avoided. And they refers to gadget. I will change it to "that uses this defect" if that is ok with you/

rules/S836/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Accident, I was having all examples marked as `cpp` in first place.

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,272 @@
...
 
 == Resources
 
+=== Documentation
+
+- C++ reference - https://en.cppreference.com/w/cpp/language/storage_duration[Storage class specifiers]
+
+=== External coding guidelines
+
 * https://cwe.mitre.org/data/definitions/457[MITRE, CWE-457] - Use of Uninitialized Variable
```
necto, line 272: The link should be applied to the document, leaving the standard (source) as plain text

tomasz-kaminski-sonarsource: Could you elaborate. I used that exact spelling in other rule.

tomasz-kaminski-sonarsource: OK, I checked the link formatting doc, and I think got it correct now.

necto: I think it should be:
source - https://link[document]
```
* MITRE - https://cwe.mitre.org/data/definitions/457[CWE-457 Use of Uninitialized Variable]
```
But I can be wrong

tomasz-kaminski-sonarsource: I used the first form. Grep in RSPEC shows only the old forms.

tomasz-kaminski-sonarsource: I have checked with Best Practices, and we should just CWE as source:
https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+  Aggregate* aggrPtr = new Aggregate; // members of allocated Aggregate are not initialized
+  Aggregate* aggrArr = new Aggregate[5]; // members of each of 5 Aggregate objects are not initialized
+}
+----
+
+
+=== What is the potential impact?
+
+Using garbage values will cause the program to behave
+non-deterministically at runtime.
```
necto, line 66:
```suggestion
nondeterministically at runtime.
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
 ----
+struct Partial {
+  // x is not initialized
+  Partial() : y(10.0) {}
+
+  int x;
+  float y;
+};
+
+void initialized() {
```
necto, line 104:
```suggestion
int initialized() {
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+object has a defined value even if no initializer is specified.
+
+[source,cpp]
+----
+int globInt;
+int globTab[10];
+Aggregate globAggr;
+Partial globPart; // x member is zero-initialized
+
+void uses() {
```
necto, line 126:
```suggestion
int uses() {
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+  Partial() : y(10.0) {}
+
+  int x;
+  float y;
+};
+
+void initialized() {
+  Partial p; // constructor is called
+  // or even Partial p{};
+  return p.x; // Non-complaint: reading an uninitialized variable
```
necto, line 107:
```suggestion
  return p.x; // Non-compliant: reading an uninitialized variable
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+[source,c]
+----
+int function(int flag, int b) {
+  int a;
+  if (flag) {
+    a = b;
+  } else {
+    return 10;
+  }
+  return a; // Complaint
```
necto, line 192:
```suggestion
  return a; // Compliant
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
 [source,cpp]
 ----
 int function(int flag, int b) {
-  int a = 0;
+  int a = 10;
   if (flag) {
     a = b;
   }
-  return a;
+  return a; // Complaint
```
necto, line 205:
```suggestion
  return a; // Compliant
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
 }
 ----
 
+Initializing value in the definition:
+
+[source,c]
+----
+int function(int flag, int b) {
+  int const a = flag ? b : 10;
+  return a; // Complaint
```
necto, line 215:
```suggestion
  return a; // Compliant
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
 
+=== Why is there an issue for a class with a default constructor?
 
-=== Noncompliant code example
+In {cpp}, a class can define a default constructor invoked when an
+object of the given type is created. Such a constructor is called even
+if a variable is declared without any initializer.
+However, if the constructor code omits the initialization
+of a member that itself does not have the default constructor,
+the member will remain uninitialized. And reading
```
necto, line 91:
```suggestion
the member will remain uninitialized  (See also S2107). And reading
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+}
+----
+
+
+=== Exceptions: global, static, and thread-local variables.
+
+All the variables with static storage duration, meaning: global, static,
+or thread-local variables, are zero-initialized before the initializer
+is evaluated. As a consequence, any such variable or member of such
+object has a defined value even if no initializer is specified.
```
necto, line 117: I would lean towards the standard section title "Exceptions", and  clarify it inside the section:

```suggestion
=== Exceptions

This rule does not flag the variables with static storage duration, meaning:
global, static, and thread-local variables.

All the variables with static storage duration are zero-initialized before the initializer is evaluated.
As a consequence, any variable or member of such an object has a defined value even if no initializer is specified.
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+       + globPart.x
+       + staticInt;
+}
+----
+
+== How to fix it
+
+Commonly, the use of an uninitialized object is an indication of a defect in the
+code, where either variable initialization was skipped on some code paths, or
+the object is used by mistake. Generally, such problems can be addressed by:
```
necto, line 140: This is quite subjective, so feel free to discard it.

I notice that you use quite a lot of passive voice.
In general, it hinders comprehension and leads to more complex sentences.
It is also against [our guidelines](https://github.com/SonarSource/rspec/blob/master/docs/tone_guide.adoc#guidelines).

In most places in this rule description, however, it makes sense. Indeed, it is clearer to say "constructor is called" or "variable is initialized", rather than "C++ abstract machine skips the constructor call", or "initializes a variable".

But here, I think you could improve the readability by using active voice:
```suggestion
the object is used by mistake. Generally, you can address such problems by:
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
     a = b;
   }
-  return a; // Noncompliant - "a" has not been initialized in all paths
+  return a; // Noncompliant: "a" has not been initialized in all paths
 }
 ----
 
+==== Compliant solution
+
+Initializing variable in all code paths:
```
necto, line 166:
```suggestion
Initializing variable on all code paths:
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+
+Initializing the variable to zero at the declaration is not always the right
+solution to fix the issue, as it may lead to logic errors if such a value
+is not handled correctly. For example, setting an `age` field of the `Employee`
+the structure may break assumptions of retirement handling code. Or more commonly,
+setting a pointer to `NULL`, will turn dereference of an uninitialized value
+into a null pointer dereference.
+
+
+=== Going the extra mile
```
necto, line 229: I think it is worth mentioning that in this case you can often make your variable `const`. And also that using `const` variables makes you more disciplined about your initialization.

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+  return a; // Complaint
+}
+----
+
+=== Pitfalls
+
+Initializing the variable to zero at the declaration is not always the right
+solution to fix the issue, as it may lead to logic errors if such a value
+is not handled correctly. For example, setting an `age` field of the `Employee`
+the structure may break assumptions of retirement handling code. Or more commonly,
```
necto, line 224:
```suggestion
is not handled correctly. For example, setting an `age` field of an `Employee`
structure may break assumptions of retirement handling code. Or more commonly,
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+setting a pointer to `NULL`, will turn dereference of an uninitialized value
```
necto, line 225:
```suggestion
setting a pointer to `NULL` will turn dereference of an uninitialized value
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+into a null pointer dereference.
```
necto, line 226: With no `-` both adjectives apply to the "dereference", i.e., it is equivalent to "pointer null dereference"
```suggestion
into a null-pointer dereference.
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+
+
+=== Going the extra mile
+
+With the addition of lambdas in {cpp}11, it is possible to initialize
+variables on declaration without creating a separate function
+to the compute value:
```
necto, line 233:
```suggestion
to compute the value:
```

rules/S836/cfamily/rule.adoc
```diff
@@ -1,42 +1,283 @@
...
+  }
+}();
+----
 
 == Resources
 
-* https://cwe.mitre.org/data/definitions/457[MITRE, CWE-457] - Use of Uninitialized Variable
+=== Documentation
+
+- C++ reference - https://en.cppreference.com/w/cpp/language/storage_duration[Storage class specifiers]
```
necto, line 268: Not sure if this makes any difference. for consistency
```suggestion
* C++ reference - https://en.cppreference.com/w/cpp/language/storage_duration[Storage class specifiers]
```

necto (approved):

rules/S836/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Good suggestion. Applied.

rules/S836/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Making variable `const` prevents moving them, or (if I recall correctly) prevents named return initialization. I agree that using const when possible is better, but I think it is not necessary if you want to avoid uninitialized variables.

PR #2598, opened 2023-07-21
Modify rule S2154: reworked rule description for LayC format [SONARJAVA-4490]

kaufco:


rules/S2154/java/rule.adoc
```diff
@@ -1,27 +1,67 @@
 == Why is this an issue?
 
-If wrapped primitive values (e.g. ``++Integers++`` and ``++Floats++``) are used in a ternary operator (e.g. ``++a?b:c++``), both values will be unboxed and coerced to a common type, potentially leading to unexpected results. To avoid this, add an explicit cast to a compatible type.
+Using wrapped primitive values in a ternary operator does not simply return one operand or the other based on the condition.
```
dorian-burihabwa-sonarsource, line 3:
```suggestion
Using boxed values in a ternary operator does not simply return one operand or the other based on the condition.
```

rules/S2154/java/rule.adoc
```diff
@@ -1,27 +1,67 @@
...
+Instead, the primitive values are unboxed and coerced to a common type, which can result in a loss of precision when converting one operand from `int` to `float` or from `long` to `double`.
```
dorian-burihabwa-sonarsource, line 4:
```suggestion
Instead, the values are unboxed and coerced to a common type, which can result in a loss of precision when converting one operand from `int` to `float` or from `long` to `double`.
```

dorian-burihabwa-sonarsource (approved):
Feel free to use the last suggestions if you feel like it makes sense

PR #2599, opened 2023-07-21
Modify rule S4158: Adapt to LaYC

victor-diez-sonarsource:


rules/S4158/javascript/rule.adoc
```diff
@@ -1,19 +1,38 @@
...
-if (strings.includes("foo")) {}  // Noncompliant
+if (strings.includes("foo")) {}  // Noncompliant: strings is always empty
 
-for (str of strings) {}  // Noncompliant
+for (const str of strings) {}  // Noncompliant
 
 strings.forEach(str => doSomething(str)); // Noncompliant
 ----
+
+Make sure your code provides some way to populate the collection if their elements are to be accessed.
```
yassin-kammoun-sonarsource, line 16:
```suggestion
Make sure that the collection is populated first if their elements are to be accessed.
```

rules/S4158/javascript/rule.adoc
```diff
@@ -1,19 +1,38 @@
...
+
+[source,javascriptdiff-id=1,diff-type=compliant]
```
yassin-kammoun-sonarsource, line 18:
```suggestion
[source,javascript,diff-id=1,diff-type=compliant]
```

yassin-kammoun-sonarsource (approved):

rules/S4158/javascript/rule.adoc
victor-diez-sonarsource: not sure about this one, implementation just checks for population methods used in the variable, but they may be called on event, request, or whatever. so we allow accessing empty arrays if there is a way to populate it

yassin-kammoun-sonarsource: Yeah, I dunno how to properly express the fix to suggest. Feel free to ignore what I wrote.

victor-diez-sonarsource: this case is valid for example:
```js
let globalArray = [];

router.get('/', (req,res) => {
   res.send(globalArray.forEach(element => `${element} is here`))
})

router.post('/', (req, res) => {
   globalArray.push(req.body.element);
   res.send('ok!');
})
```

yassin-kammoun-sonarsource: I see. Okay, let's go with your suggestion then. Looking at this example, I am convinced.

victor-diez-sonarsource:
```suggestion
"Make sure there exists a path in your code that populates your collection if their elements are to be accessed."
```

PR #2600, opened 2023-07-21
Modify S2078(java+common): Improved fix

loris-s-sonarsource:


rules/S2078/java/how-to-fix-it/java-se.adoc
```diff
@@ -43,6 +43,7 @@ For Java, OWASP's functions
 https://www.javadoc.io/doc/org.owasp.esapi/esapi/latest/org/owasp/esapi/Encoder.html#encodeForDN-java.lang.String-[`encodeForDN`]
 and
 https://www.javadoc.io/doc/org.owasp.esapi/esapi/latest/org/owasp/esapi/Encoder.html#encodeForLDAP-java.lang.String-[`encodeForLDAP`]
-allow sanitizing these characters automatically.
+allow sanitizing these characters automatically. +
```
hendrik-buchwald-sonarsource, line 46: Does the "automatically" make sense? If you actually have to pass the user input into it and then get the escaped version back, I would not call that automatic.

rules/S2078/java/how-to-fix-it/java-se.adoc
```diff
@@ -43,6 +43,7 @@ For Java, OWASP's functions
...
+Otherwise, here are the details showing how a popular framework implements it: https://github.com/bcgit/bc-java/blob/r1rv74/prov/src/main/java/org/bouncycastle/jce/provider/X509LDAPCertStoreSpi.java#L52-L70[Bouncy Castle Java].
```
hendrik-buchwald-sonarsource, line 47: The "Otherwise" seems unfitting in this context. I would rather write something along the lines of "If you can not use these libraries, you might want to consider implementing your own encoder. ((Do we really want to recommend this? I don't feel too comfortable doing it)) You can find a sample implementation in the framework Bouncy Castle Java".

hendrik-buchwald-sonarsource (approved):
LGTM

PR #2613, opened 2023-07-24
Modify rule S5890: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5890/python/rule.adoc
+++ b/rules/S5890/python/rule.adoc
@@ -1,28 +1,48 @@
+This rule raises an issue when assigned value type is not compatible with typehint of assigning variable.
+
 == Why is this an issue?
 
+Type hints in Python allow you to specify the expected types of variables and function return values. While type hints are not enforced at runtime, they serve as documentation and can be checked using static type checkers to catch type-related errors during development.
+
+When an assigned value type is incompatible with the type hint of the assigning variable, it can lead to several issues:
+
+* *Type-related bugs:* Assigning a value of an incompatible type to a variable with a specific type hint may lead to unexpected behavior or errors at runtime. For example, if a variable is expected to be an integer but is assigned a string, operations that assume the variable is an integer will raise exceptions or produce incorrect results.
+
+* *Readability and maintainability:* Type hints improve code readability by explicitly stating the intended types of variables and functions. When the assigned value type doesn't match the hint, it can confuse other developers and make the code harder to maintain.
+
+* *False sense of correctness:* Type hints can provide a sense of confidence that the code is handling specific types, but if the assigned values don't adhere to those hints, it can create a false sense of correctness and may lead to runtime errors that could have been caught during static analysis.
+
+* *Integration with tools:* When using static type checkers, incorrect type annotations can cause the checker to produce false positives or negatives. If the assigned value types don't match the type hints, it can lead to misleading error messages or make it difficult to detect actual type-related issues.
 Type hints can be used to communicate the intended type of a given variable. These are not enforced at runtime and not respecting them might not necessarily lead to runtime errors.
 
+== How to fix it
 
-It is however confusing and could lead to maintainability issues.
+Assign to variable value of type same as type hint type or update the type hint.
 
+=== Code examples
 
-=== Noncompliant code example
+==== Noncompliant code example
 
-[source,python]
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
 def my_function():
     my_int: int = "string"  # Noncompliant
 ----
 
 
-=== Compliant solution
+==== Compliant solution
 
-[source,python]
+[source,python,diff-id=1,diff-type=compliant]
 ----
 def my_function():
     my_str: str = "string" 
 ----
 
+== Resources
+
+=== Documentation
+
+* Python documentation - https://docs.python.org/3/library/typing.html[typing — Support for type hints]
 
 ifdef::env-github,rspecator-view[]
 
--- a/rules/S5890/python/rule.adoc
+++ b/rules/S5890/python/rule.adoc
@@ -12,7 +12,7 @@ When an assigned value type is incompatible with the type hint of the assigning 
 
 == How to fix it
 
-Assign the variable to the value compatible with or change the type hint.
+Assign the variable to the value compatible with the type hint or change the type hint to be compatible with the variable's type.
 
 === Code examples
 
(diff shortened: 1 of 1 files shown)


rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,48 @@
...
+
+When an assigned value type is incompatible with the type hint of the assigning variable, it can lead to several issues:
+
+* *Type-related bugs:* Assigning a value of an incompatible type to a variable with a specific type hint may lead to unexpected behavior or errors at runtime. For example, if a variable is expected to be an integer but is assigned a string, operations that assume the variable is an integer will raise exceptions or produce incorrect results.
+
+* *Readability and maintainability:* Type hints improve code readability by explicitly stating the intended types of variables and functions. When the assigned value type doesn't match the hint, it can confuse other developers and make the code harder to maintain.
+
+* *False sense of correctness:* Type hints can provide a sense of confidence that the code is handling specific types, but if the assigned values don't adhere to those hints, it can create a false sense of correctness and may lead to runtime errors that could have been caught during static analysis.
```
joke1196, line 13: I am not sure about this point. It feels like adding type hints makes the code less good as it gives a false sense of correctness, which I don't agree with. 
Also, I am not sure I get the last part of this sentence, because having type hints will allow static type checker to catch these issues. If I understand correctly; this last part is saying that having a correct type assigned will help the type check to catch an error, but there are no errors.
I feel this point goes against adding type hints, more than applying the correct type. Maybe I didn't fully understand what you meant.

rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,48 @@
...
+
+* *Integration with tools:* When using static type checkers, incorrect type annotations can cause the checker to produce false positives or negatives. If the assigned value types don't match the type hints, it can lead to misleading error messages or make it difficult to detect actual type-related issues.
 Type hints can be used to communicate the intended type of a given variable. These are not enforced at runtime and not respecting them might not necessarily lead to runtime errors.
```
joke1196, line 16: Maybe we could merge this part with the first sentence of the description as they say something similar.

rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,48 @@
+This rule raises an issue when assigned value type is not compatible with typehint of assigning variable.
```
joke1196, line 1: Small typo, I think it should be: when an assigned value type is not compatible with the type hint of the assigning variable.

rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,48 @@
...
+
+* *False sense of correctness:* Type hints can provide a sense of confidence that the code is handling specific types, but if the assigned values don't adhere to those hints, it can create a false sense of correctness and may lead to runtime errors that could have been caught during static analysis.
+
+* *Integration with tools:* When using static type checkers, incorrect type annotations can cause the checker to produce false positives or negatives. If the assigned value types don't match the type hints, it can lead to misleading error messages or make it difficult to detect actual type-related issues.
 Type hints can be used to communicate the intended type of a given variable. These are not enforced at runtime and not respecting them might not necessarily lead to runtime errors.
 
+== How to fix it
 
-It is however confusing and could lead to maintainability issues.
+Assign to variable value of type same as type hint type or update the type hint.
```
joke1196, line 20: Maybe we could have a bit less repetition of the word `type` something like: Assign to the variable a value of the type proposed by the type hint or update the type hint.

joke1196 (changes requested):
I added a few comments. I may not have understood properly each point so don't hesitate to correct me, especially for the last two points of the list.

rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,43 @@
+This rule raises an issue when assigned value type is not compatible with the type hint of assigning variable.
```
joke1196, line 1: Still think we are missing some articles here: when `an` assigned value type is not compatible with the type hint of `the` assigning variable.

rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,43 @@
...
+When an assigned value type is incompatible with the type hint of the assigning variable, it can lead to several issues:
+
+* *Type-related bugs:* Assigning a value of an incompatible type to a variable with a specific type hint may lead to unexpected behavior or errors at runtime.
 
+* *Readability and maintainability:* Type hints improve code readability by explicitly stating the intended types of variables and functions. When the assigned value type doesn't match the hint, it can confuse other developers and make the code harder to maintain.
 
-It is however confusing and could lead to maintainability issues.
+== How to fix it
 
+Assign to variable value compatible with the type hint or change it.
```
joke1196, line 15: Here as well small grammar issues: missing articles. Assign to `a` variable `a` value compatible with `its` type hint. After re-reading it I would maybe drop the `change it` only because I have the feeling it would push people to maybe drop the type hint to fix the issue or maybe go for `Any`, not 100% sure on this one I let you decide.

joke1196 (changes requested):
Looks good! Small comments on some grammar issues.

rules/S5890/python/rule.adoc
```diff
@@ -1,28 +1,43 @@
...
+When an assigned value type is incompatible with the type hint of the assigning variable, it can lead to several issues:
+
+* *Type-related bugs:* Assigning a value of an incompatible type to a variable with a specific type hint may lead to unexpected behavior or errors at runtime.
 
+* *Readability and maintainability:* Type hints improve code readability by explicitly stating the intended types of variables and functions. When the assigned value type doesn't match the hint, it can confuse other developers and make the code harder to maintain.
 
-It is however confusing and could lead to maintainability issues.
+== How to fix it
 
+Assign the variable to the value compatible with or change the type hint.
```
joke1196, line 15: I am not sure we can say it that way I think we need a repetition of `type hint` in the end, but I will let @jean-jimbo-sonarsource  have a go at it.

joke1196 (approved):
LGTM. One thing I was not sure about so I will ask for Jean's review.

rules/S5890/python/rule.adoc
jean-jimbo-sonarsource: "Assign the variable to the value compatible with or change the type hint." sounds incomplete. I don't see any harm in clarifying by being more verbose. 
For example: "Assign the variable to the value compatible with the type hint or change the type hint to be compatible with the variable's type."

jean-jimbo-sonarsource (changes requested):
I have made a suggestion, but would like an extra pair of eyes on it. It may not disambiguate the sentence enough.

jean-jimbo-sonarsource (approved):

PR #2615, opened 2023-07-24
Modify rule S5781: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5781/rule.adoc
+++ b/rules/S5781/rule.adoc
@@ -5,7 +5,7 @@ This rule raises an issue when the same value is used multiple times when instan
 By definition a set cannot hold two identical values. 
 When instantiating a set litteral with the same value repeated multiple times, 
 only the last occurence of the duplicated value will remain. 
-Duplicated values should be either modified or removed.
+The duplicated values should be either modified or removed.
 
 === Noncompliant code example
 
--- a/rules/S5781/rule.adoc
+++ b/rules/S5781/rule.adoc
@@ -1,25 +1,25 @@
+This rule raises an issue when the same value is used multiple times when instantiating a set literal.
+
 == Why is this an issue?
 
-A set cannot have two identical values. When a value is repeated in a set literal, only the last occurence will remain. Thus duplicate values should be either modified or removed.
-
-
-This rule raises an issue when the same value is used multiple times as a value in a set literal.
-
+By definition a set cannot hold two identical values. 
+When instantiationg a set litteral with the same value repeated multiple times, 
+only the last occurence of the duplicated value will remain. 
+Duplicated values should be either modified or removed.
 
 === Noncompliant code example
 
-[source,text]
+[source,text,diff-id=1,diff-type=noncompliant]
 ----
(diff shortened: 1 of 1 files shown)


maksim-grebeniuk-sonarsource (commented):
LGTM

rules/S5781/rule.adoc
```diff
@@ -1,25 +1,25 @@
...
-A set cannot have two identical values. When a value is repeated in a set literal, only the last occurence will remain. Thus duplicate values should be either modified or removed.
-
+This rule raises an issue when the same value is used multiple times when instantiating a set literal.
 
-This rule raises an issue when the same value is used multiple times as a value in a set literal.
+== Why is this an issue?
 
+By definition a set cannot hold two identical values. 
+When instantiating a set litteral with the same value repeated multiple times, 
+only the last occurence of the duplicated value will remain.
```
guillaume-dequenne-sonarsource, line 7: We probably want to add something to explain why this might happen and why this is bad:
* maybe the duplication comes from a copy paste that didn't get replaced (an actual value is missing from the set)
* maybe it's just a simple duplication, in which case it's just confusing/difficult to maintain code

As far as I know we don't raise on function calls, so we wouldn't detect any actual redundant computation (as we're not sure 2 identical calls would lead to identical results), so I wouldn't mention that, but it's also a legitimate potential impact.

guillaume-dequenne-sonarsource (changes requested):
Looks good, minor comments

rules/S5781/rule.adoc
joke1196: Yes, that's right, I have added more details on why this is an issue.

rules/S5781/rule.adoc
```diff
@@ -1,25 +1,25 @@
...
-
+This rule raises an issue when the same value is used multiple times when instantiating a set literal.
 
-This rule raises an issue when the same value is used multiple times as a value in a set literal.
+== Why is this an issue?
 
+By definition a set cannot hold two identical values. 
+When instantiating a set litteral with the same value repeated multiple times, 
+only the last occurence of the duplicated value will remain. 
+The duplicated values should be either modified or removed.
```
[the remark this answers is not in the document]
joke1196: I see what you mean. I have reworked this part a bit.

guillaume-dequenne-sonarsource (approved):
LGTM

rules/S5781/rule.adoc
```diff
@@ -1,25 +1,31 @@
+This rule raises an issue when the same value is used multiple times when instantiating a set literal.
+
 == Why is this an issue?
 
-A set cannot have two identical values. When a value is repeated in a set literal, only the last occurence will remain. Thus duplicate values should be either modified or removed.
+By definition a set cannot hold multiple times the same value.
```
jean-jimbo-sonarsource, line 5: "By definition, a set cannot hold the same value multiple times." is clearer

jean-jimbo-sonarsource (approved):
Left one comment

PR #2619, opened 2023-07-24
Modify rule S1046: LaYC format

amelie-renard-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1046/cfamily/rule.adoc
+++ b/rules/S1046/cfamily/rule.adoc
@@ -41,7 +41,7 @@ void f() {
 
 === Documentation
 
-* C{plus}{plus} reference - https://en.cppreference.com/w/cpp/language/try_catch[try-block]
+* {cpp} reference - https://en.cppreference.com/w/cpp/language/try_catch[try-block]
 
 === External coding guidelines
 
--- a/rules/S1046/cfamily/rule.adoc
+++ b/rules/S1046/cfamily/rule.adoc
@@ -1,55 +1,49 @@
 == Why is this an issue?
 
-The catch-all handler should come last in a chain of ``++catch++`` or ``++@catch++`` statements because it catches everything, and any more-specific ``++catch++``/``++@catch++`` that comes after it will never be used, even when the relevant condition occurs. 
+The catch-all handler, written `catch(...)` in C{plus}{plus}, or `@catch(...)` in Objective-C, catches every type of exception. If there is another catch statement for a specific exception after the catch-all handler, it will not be executed because the catch-all handler will already have handled the exception.
 
+== How to fix it
 
-This {cpp} code sample is very similar to the Objective-C equivalent with ``++@try++`` and ``++@catch++``.
+=== Code examples
 
+The following C{plus}{plus} examples also work in Objective-C by replacing `try` and `catch` with their equivalents `@try` and `@catch`.
 
-=== Noncompliant code example
+==== Noncompliant code example
 
 [source,cpp]
 ----
-void f1()
-{
-  try
-  {
+void f() {
+  try {
     // ...
-  }
-  catch (...)
-  {
+  } catch (...) {
     // Handle all exception types
-  }
-  catch (std::exception const &e)  // Noncompliant - handler will never be called
-  {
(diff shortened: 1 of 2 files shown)


rules/S1046/cfamily/rule.adoc
```diff
@@ -1,56 +1,50 @@
 == Why is this an issue?
 
-The catch-all handler should come last in a chain of ``++catch++`` or ``++@catch++`` statements because it catches everything, and any more-specific ``++catch++``/``++@catch++`` that comes after it will never be used, even when the relevant condition occurs. 
+The catch-all handler, written `catch(...)` in C{plus}{plus}, or `@catch(...)` in Objective-C, catches every type of exception. If there is another catch statement for a specific exception after the catch-all handler, it will not be executed because the catch-all handler will already have handled the exception.
```
marco-antognini-sonarsource, line 3: The previous Asciidoc version was also incorrect: the `...` needs to be escaped with `` `+ bla +` `` or it is transformed to the Unicode equivalent … .

```suggestion
The catch-all handler, written `+catch(...)+` in {cpp}, or `+@catch(...)+` in Objective-C, catches every type of exception. If there is another catch statement for a specific exception after the catch-all handler, it will not be executed because the catch-all handler will already have handled the exception.
```

BTW, Grammarly suggests splitting the second sentence for clarity:
```suggestion
The catch-all handler, written `+catch(...)+` in {cpp}, or `+@catch(...)+` in Objective-C, catches every type of exception. Suppose there is another catch statement for a specific exception after the catch-all handler. In that case, it will not be executed because the catch-all handler will already have handled the exception.
```

rules/S1046/cfamily/rule.adoc
```diff
@@ -1,56 +1,50 @@
...
 
+== How to fix it
 
-This {cpp} code sample is very similar to the Objective-C equivalent with ``++@try++`` and ``++@catch++``.
+=== Code examples
 
+The following C{plus}{plus} examples also work in Objective-C by replacing `try` and `catch` with their equivalents `@try` and `@catch`.
```
marco-antognini-sonarsource, line 9: To be extremely pedantic, the exception type should also be changed (`std::exception const &e` --> `NSException *e`). It's probably not worth going into so much detail here though. I'd just change the sentence to be accurate, maybe something like this:

```suggestion
The following {cpp} example is very similar in Objective-C: the `try` and `catch` equivalents are `@try` and `@catch`, respectively.
```

rules/S1046/cfamily/rule.adoc
```diff
@@ -1,56 +1,50 @@
...
   }
 }
 ----
 
 
 == Resources
 
+=== Documentation
+
+* C{plus}{plus} reference - https://en.cppreference.com/w/cpp/language/try_catch[try-block]
```
marco-antognini-sonarsource, line 44: I'm not 100% convinced it's necessary but we could link to https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Tasks/HandlingExceptions.html for completeness.

WDYT?

marco-antognini-sonarsource, line 44:
```suggestion
* {cpp} reference - https://en.cppreference.com/w/cpp/language/try_catch[try-block]
```

For consistency with the idiomatic way of spelling C++ in Asciidoc.

marco-antognini-sonarsource (approved):
LGTM with some small nits. I don't feel we need another round unless you want to -- I'll let you decide if my suggestions make sense and apply them if needed.

rules/S1046/cfamily/rule.adoc
amelie-renard-sonarsource: Oh indeed! I don't know why I forgot about it... Thanks!

amelie-renard-sonarsource: "library/archive" :D
I will add it for the unlucky users who deal with Obj-C

rules/S1046/cfamily/rule.adoc
amelie-renard-sonarsource: Modified

PR #2623, opened 2023-07-25
Modify rule S3520: Convert to LaYC format

LoghinVladDev:
## Review

- [ ] text style and tone


rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
+Do not deallocate the same object twice to avoid "double-free" undefined behavior.
+
 == Why is this an issue?
 
-Using ``++free(...)++`` or ``++delete++`` releases the reservation on a memory location, making it immediately available for another purpose. So releasing the same memory location twice can lead to corrupting the program's memory. 
+Using `delete` or `free(...)` releases the reservation on a memory location, making it immediately available for another purpose.
+Releasing the same pointer twice is undefined behavior and will usually crash the program.
```
necto, line 6:
```suggestion
Releasing the same memory location twice leads to undefined behavior and can often crash the program.
```
"releasing the same pointer" is not correct, as you are not releasing the pointer per see.
If you take it as "releasing the memory pointed to by the same pointer" it is also not correct, because if I change the pointer in the meantime, it is fine to call `free` on it again.

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
+
+=== What is the potential impact
+
+The effects of a "double-free" heavily depend on how the program's memory management data structures are implemented.
```
necto, line 10: Is the "data structures" detail important to describe the impact?
I feel omitting it unburdens the object of the sentence
```suggestion
The effects of a "double-free" heavily depend on how the program's memory management is implemented.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
+In the case of such a scenario, one of the following can be observed:
 
+- The program's memory management data structures can become corrupted.
+This will usually cause a crash.
+- In specific circumstances, corruption of mentioned data structures can cause two later calls to malloc to return the same pointer.
+This can lead to a buffer overflow vulnerability.
```
necto, line 16: How?

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
+- The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Any access to that memory past its lifetime end, including another `delete`, is undefined behavior.
+As a result, a second call can be optimized out, leading to hard-to-identify bugs.
```
necto, line 19: In this circumstance, how skipping this second call can lead to a hard-to-identify bug?

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
+
+The effects of a "double-free" heavily depend on how the program's memory management data structures are implemented.
+In the case of such a scenario, one of the following can be observed:
 
+- The program's memory management data structures can become corrupted.
+This will usually cause a crash.
+- In specific circumstances, corruption of mentioned data structures can cause two later calls to malloc to return the same pointer.
+This can lead to a buffer overflow vulnerability.
+- The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Any access to that memory past its lifetime end, including another `delete`, is undefined behavior.
```
necto, line 18: Is it UB in C?

In any case, I would start by saying that UBs are bad regardless. And then only went into the details of particular misbehavior observed in the field

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
 
-A best practice to avoid this bug calls for setting just-freed pointers to ``++NULL++``, and always null-testing before a ``++free++`` or ``++delete++``.
+== How to fix it
 
+Besides just ensuring that every release happens once, the following best practices can also be used:
 
-=== Noncompliant code example
+- After a `delete` or `free`, set the pointer's value to `nullptr` or `NULL`.
+`delete` of `nullptr` and `free` of `NULL` are defined as having no effect.
+- Attempt to release any allocated memory in the function it was allocated in.
```
necto, line 27: > Do. Or do not. There is no try. (C)

Seriously, this bullet point was prepended by "best practice". It is already in the context of "try your best to follow this, but you can't be required to do so"

```suggestion
- To release any allocated memory in the function it was allocated in.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
+- In specific circumstances, corruption of mentioned data structures can cause two later calls to malloc to return the same pointer.
+This can lead to a buffer overflow vulnerability.
+- The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Any access to that memory past its lifetime end, including another `delete`, is undefined behavior.
+As a result, a second call can be optimized out, leading to hard-to-identify bugs.
 
-A best practice to avoid this bug calls for setting just-freed pointers to ``++NULL++``, and always null-testing before a ``++free++`` or ``++delete++``.
+== How to fix it
 
+Besides just ensuring that every release happens once, the following best practices can also be used:
```
necto, line 23: Except for the "set to null", I would move the rest of this list into the "Going the extra mile" subsection.

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
-=== Noncompliant code example
+- After a `delete` or `free`, set the pointer's value to `nullptr` or `NULL`.
+`delete` of `nullptr` and `free` of `NULL` are defined as having no effect.
+- Attempt to release any allocated memory in the function it was allocated in.
+- If lifetime of memory is tied to an object, use smart pointers.
+These will manage the lifetime of the required memory without any "double-free" risk.
 
+=== Code examples
+
+==== Noncompliant code example
```
necto, line 33: Looks like the obvious fix of "remove one of the `free` calls" is missing

rules/S3520/cfamily/rule.adoc
```diff
@@ -22,13 +47,30 @@ void doSomething(int size) {
 }
 ----
 
+In {cpp}:
```
necto, line 50: Let's have one non-compliant example per "Noncompliant code example" subsection. This will make it easier to find which noncompliant code snippet a "Compliant solution" corresponds to.

You can repeat "Noncompliant code example" and "Compliant solution" multiple times.

necto (changes requested):

rules/S3520/cfamily/rule.adoc
```diff
@@ -41,12 +83,50 @@ void doSomething(int size) {
...
+    delete p;
+    p = nullptr; // This will prevent freeing the same memory again
+  }
+
+  delete p; // This is OK: if the memory was freed in the if-block above, operator delete of nullptr is a no-op
+  p = nullptr; // This will prevent freeing the same memory again
+}
+----
+
+{cpp}, using smart pointers:
```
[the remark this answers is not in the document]
necto: Yes. Although, I am not sure about the diff view. and I think you don't need to use "Noncompliant code example" / "Compliant solution" sections

rules/S3520/cfamily/rule.adoc
LoghinVladDev: Calling `free` twice with the same parameter is UB in both C and CPP.
CPP does not define double delete as UB directly, but it does define the lifetime of heap-allocated memory.

rules/S3520/cfamily/rule.adoc
LoghinVladDev: Added as comment in examples, to keep repetitions low and diffs available

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,17 +1,42 @@
...
+In the case of such a scenario, one of the following can be observed:
 
+- The program's memory-management data structures can become corrupted.
+This will usually cause a crash.
+- In specific circumstances, corruption of mentioned data structures can cause two later calls to malloc to return the same pointer.
+This can lead to a buffer overflow vulnerability.
+- The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Any access to that memory past its lifetime end, including another `delete`, is undefined behavior.
+Similar to free, this will usually cause a crash.
+Although this does not mean it is guaranteed, it can lead to worse issues.
```
necto, line 20: I do not understand this sentence: what is `it` in the first and the second parts?

rules/S3520/cfamily/rule.adoc
necto: Right now the paragraph speaks only about C++ standard.
Can you rephrase it to include C as well? (potentially in another bullet point)

rules/S3520/cfamily/rule.adoc
```diff
@@ -22,30 +47,92 @@ void doSomething(int size) {
...
 void doSomething(int size) {
-  char *cp = (char *) malloc(sizeof(char) * size);
+  char* cp = (char*) malloc(sizeof(char) * size);
 
   // ...
   if (condition) {
     free(cp);
-    cp = NULL; // This will prewent freeing the same memory again
+    cp = NULL; // This will prevent freeing the same memory again
+    // Additionally, this call to free here can be removed since another will happen later
```
necto, line 61: "Additionally" suggests that you can do both: set it to null and remove the `free`. But doing so will introduce a memory leak.
```suggestion
    // Alternatively, the call to free here can be removed since another will happen later
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -22,30 +47,92 @@ void doSomething(int size) {
...
+[source,cpp,diff-id=2,diff-type=compliant]
+----
+void doSomething(int size) {
+  std::string* p = new std::string;
+
+  // ...
+  if (condition) {
+    delete p;
+    p = nullptr; // This will prevent freeing the same memory again
+    // Additionally, this call to delete here can be removed since another will happen later
```
necto, line 98: "Additionally" suggests that you can do both: set it to null and remove the `delete`. But doing so will introduce a memory leak.
```suggestion
    // Alternatively, the invocation of delete here can be removed since another will happen later
```

rules/S3520/cfamily/rule.adoc
necto, line 61: Personally, I feel that "remove call to free" is the primary fix, and setting to null is kind of a hack. It is also mentioned first in the list of fixes above.
However, presently this fix is hidden in a comment. I would add at least one example that just removes the call to `free` or invocation of `delete`.
BTW, it would work better with the diff highlighting, because at the moment our products highlight on the line-level granularity and not token-level

necto (changes requested):

rules/S3520/cfamily/rule.adoc
LoghinVladDev: Rephrased, together with the next comment about memory lifetime in C.
Moved UB descriptions to "why is this an issue"
Described impact of UB more clearly in "what is the potential impact"

rules/S3520/cfamily/rule.adoc
```diff
@@ -5,25 +5,32 @@ Deallocating a memory location more than once leads to undefined behavior.
...
+Both the C and {cpp} standards describe it as such.
+Due to this, there is no guarantee that a crash will happen on a "double-free".
+
+The effects of a "double-free" depend entirely on the program's memory management implementation.
+In the case of such an event, one of the following can be observed:
 
 - The program's memory-management data structures can become corrupted.
 This will usually cause a crash.
-- In specific circumstances, corruption of mentioned data structures can cause two later calls to malloc to return the same pointer.
+- Demonstrative Example 2 on CWE - https://cwe.mitre.org/data/definitions/415[415 Double Free] presents a set of circumstances where a crash does not occur.
```
necto, line 24: The citation style from "Resources" section does not work when used inside of a sentence
```suggestion
- Demonstrative Example 2 on https://cwe.mitre.org/data/definitions/415[CWE-415] presents a set of circumstances where a crash does not occur.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -5,25 +5,32 @@ Deallocating a memory location more than once leads to undefined behavior.
...
+
+The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Access to memory past its lifetime end, including another `delete`, is undefined behavior.
+
 === What is the potential impact
 
-The effects of a "double-free" heavily depend on how the program's memory management is implemented.
-In the case of such a scenario, one of the following can be observed:
+The danger of a "double-free" comes directly from the fact that it is undefined behavior.
+Both the C and {cpp} standards describe it as such.
```
necto, line 16: The article use here is inconsistent
```suggestion
Both C and {cpp} standards describe it as such.
```
Or
```suggestion
Both the C and the {cpp} standards describe it as such.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -5,25 +5,32 @@ Deallocating a memory location more than once leads to undefined behavior.
...
+Due to this, there is no guarantee that a crash will happen on a "double-free".
```
necto, line 17: What do you mean by 
> crash will happen on a "double-free"
?
Is it that there is no guarantee the crash will happen at all, or that it will happen immediately when you call the second `free`?

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
+Deallocating a memory location more than once leads to undefined behavior.
+
 == Why is this an issue?
 
-Using ``++free(...)++`` or ``++delete++`` releases the reservation on a memory location, making it immediately available for another purpose. So releasing the same memory location twice can lead to corrupting the program's memory. 
+Using `delete` or `free(...)` releases the reservation on a memory location, making it immediately available for another purpose.
```
necto, line 5: Here is a[ "convention" I made up yesterday](https://docs.google.com/document/d/1hEN48Qqh4UcOjdslhNAtzRQyhU247Mk7MYr1YIuWHtM/edit#heading=h.6ujwyp74oeqh): 
> cite function names without parentheses

I am not claiming this is objectively better per se. It is better to be consistent across rule descriptions, though.
"free" without parentheses is used in S3946, S3529, S1232, S1262, S984. With parentheses only in this rule.

Also you have it in the next paragraph

Feel free to challenge this convention. So far no one did.

```suggestion
Using `delete` or `free` releases the reservation on a memory location, making it immediately available for another purpose.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
...
 
-[source,cpp]
+The effects of a "double-free" depend entirely on the program's memory management implementation.
+In the case of such an event, one of the following can be observed:
+
+- The program's memory-management data structures can become corrupted.
+This will usually cause a crash.
+- Demonstrative Example 2 on CWE - https://cwe.mitre.org/data/definitions/415[415 Double Free] presents a set of circumstances where a crash does not occur.
+In these circumstances, the corruption of the mentioned data structures causes two later calls to malloc to return the same pointer.
+This can lead to a buffer overflow vulnerability.
```
necto, line 26: if two calls to malloc return the same pointer, it can also lead to attacker accessing directly the secret buffer without any overflow

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
...
+    p = nullptr; // This will prevent freeing the same memory again
+  }
+
+  delete p; // Compliant: if the memory was freed in the if-block above, operator delete of nullptr is a no-op
+}
+----
+
+=== Going the extra mile
+
+{cpp}-11 introduced smart pointers.
```
necto, line 147:
```suggestion
{cpp}11 introduced smart pointers.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
...
+Using these instead of manual memory management can mitigate most dynamic memory management issues.
+This includes the issue of "double-free".
```
necto, line 149: Stylistic, to avoid overusing of "this/these"
```suggestion
Using these instead of manual memory management can mitigate most dynamic memory management issues
including "double-free".
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
...
+----
+void doSomething(int size) {
+  auto p = std::make_unique<std::string>();
+
+  // ...
+  if (condition) {
+    p.reset(); // Delete happens here.
+  }
+
+  // delete not called by std::unique_ptr destructor.
```
necto, line 161:
```suggestion
  // delete not called by std::unique_ptr destructor if the branch above was taken.
```

necto (changes requested):

rules/S3520/cfamily/rule.adoc
LoghinVladDev: Both apply, a crash immediately and a crash at all, since there must be unpredictable scenarios

rules/S3520/cfamily/rule.adoc
LoghinVladDev: Will add it, need to look up the name, but I believe it was "read-what-where"

LoghinVladDev:
In the [review comment](https://github.com/SonarSource/rspec/pull/2629#pullrequestreview-1560663191) for S3529, you stated a good point about the "Going the extra mile" section. 

While not applicable in the exact same format, the info is useful. I will do a minor modification to at least specify RAII firsthand, before going into smart pointers, since it uses the idiom.

It would be really cool to have something of the sort of "general best practices", since:
- As an example, this rule, S3529, and other memory allocation rules will mention smart pointers/RAII in the "going the extra mile"
- Best practices could also refer back to the rules that they cover. "Using the RAII idiom" would refer to both memory allocation, and, for example, resource acquisition (files, locks, etc.)

Just randomly throwing it here, but I see it as a potential great initiative:
- Cross squad collab. effort - best practices
- Either done on blog posts/generic rules?

Would this be possible/feasible? I would open a discuss thread for this, please advise.

LoghinVladDev:
Have a look if the pitfalls section is too long.
It is yet incomplete.

shared content will be removed from commit once merge with master is done

rules/S3520/cfamily/rule.adoc
necto: What is the point of this sentence?
Why would the reader care if there is a guarantee of a crash?

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,274 @@
...
+
+- The program's memory-management data structures can become corrupted.
+This will usually cause a crash.
+- Demonstrative Example 2 on https://cwe.mitre.org/data/definitions/415[CWE-415] presents a set of circumstances where a crash does not occur.
+In these circumstances, the corruption of the mentioned data structures causes two later calls to `malloc` to return the same pointer.
+This can lead to a sensitive data exposure vulnerability or a buffer overflow vulnerability.
+
+== How to fix it
+
+Besides just ensuring that every release happens once, the following best practices can also be used:
```
necto, line 30: 1. unnecessary passive voice in "practices can be used"
2. the best practices are specifically meant to ensure every release happens once, so you can reformulate this sentence:
```suggestion
To ensure that every release happens once, you can follow these best practices:
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
 
-  // delete not called by std::unique_ptr destructor.
+  // delete not called by std::unique_ptr destructor if the branch above was taken.
+}
+----
+
+=== Pitfalls
+
+However, this is not a perfect solution.
+Any raw pointers of references to memory held by _RAII_ objects may still lead to a "double-free".
```
necto, line 169: I am not sure where this section will be placed, and whether it will follow directly the "going the extra mile" section. Because of that, I think it is better to avoid coupling:
```suggestion
However, any raw pointers of references to memory held by _RAII_ objects may still lead to a "double-free".
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
+Since `std::make_*` accepts construction arguments valid only for the underlying type, it will not allow a pointer to the same type to be passed as a parameter.
+This helps to avoid the confusion created by the constructor overloads.
+
+Compliant solution
+[source,cpp,diff-id=3,diff-type=compliant]
+----
+void doSomething(char const* data) {
+  auto p1 = std::make_shared<std::string>(data);
+  auto p2 = p1; /// Compliant
 }
```
necto: I think this explanation is too detailed. Remember this is already a tangent (pitfalls) of a tangent (going the extra mile). At this point it should be enough to mention the pitfall (i.e., the non-compliant and compliant code snippets + 2 lines explanation of why) and point to an article that can provide more details.

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
 ----
 
+==== Initialization of `std::unique_ptr`
+
+Similar to `std::unique_ptr`, a "double-free" can occur from an erroneous initialization.
```
necto, line 211:
```suggestion
Similar to `std::shared_ptr`, a "double-free" can occur from an erroneous initialization.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
+When the owner counter reaches 0, the dynamically allocated memory area pointed to is released.
+
+The following example will demonstrate two `std::shared_ptr` objects being created with the intention of pointing to the same object.
+
+Noncompliant code example
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
+void doSomething(char const* data) {
+  std::shared_ptr<std::string> p1(new std::string(data));
+  std::shared_ptr<std::string> p2(p1.get()); /// Noncompliant: Intended: std::shared_ptr<string> p2(p1);
```
necto, line 188: What is the purpose of the third `/` in the `/// ...` comments?

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
+
+Noncompliant code example
+[source,cpp,diff-id=4,diff-type=noncompliant]
+----
+void doSomething(char const* data) {
+  // Copy intended
+  std::unique_ptr<std::string> p1(new std::string(data));
+  // ...
+  std::unique_ptr<std::string> p2(p1.get()); /// Noncompliant: Since the copy constructor of std::unique_ptr is deleted, an user might attempt this.
+                                             /// If intention was to create another pointer - std::unqiue_ptr<std::string> p2(*p1);
```
necto, line 222: I don't think you need to explain the possible intention and the reasoning in the head of a user. This is not didactic material, but a self-help rule description. The intended reader of this example is not going for the reason why this example is here in the first place.

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
+
+  // Ownership transfer intended
+  std::unqiue_ptr<std::string> p3(new std::string(data));
+  // ...
+  std::unique_ptr<std::string> p4(p3.get()); /// Noncompliant: The intention was to transfer ownership, “double-free” will happen.
+                                             /// Possibly intended code: std::unique_ptr<std::string> p4(p3.release());
```
necto, line 228: Again, I don't think this adds any value to the reader trying to get the pitfalls of using smart pointers, given the compliant example below

rules/S3520/cfamily/rule.adoc
```diff
@@ -158,14 +159,110 @@ void doSomething(int size) {
...
+}
+----
+
+==== Out-of-scope access
+
+Extra care must be taken when passing the memory address contained in a RAII object.
+
+Both `std::shared_ptr` and `std::unique_ptr` provide the `get` method to obtain a raw pointer to the contained memory.
+This should only be used to read the data.
+Passing this pointer to a call that could release the memory could cause a "double-free", amongst other problems.
```
necto, line 256: I suggest the following reformulating, addressing the readability issues:
1. "could release the memory could cause" - the repetition of "could" while grammatically correct, didn't parse for me the first time
2. "amongst other problems" is too vague and makes me think "what other problems?" "why are they not mentioned explicitly?", while I don't think you want the reader thinking about this. I can think only of the problems with "use-after-free". Let me know if you had in mind others.

```suggestion
Passing this pointer to a function that might release it can lead to a "double-free" and "use-after-free".
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,274 @@
...
 
-A best practice to avoid this bug calls for setting just-freed pointers to ``++NULL++``, and always null-testing before a ``++free++`` or ``++delete++``.
+The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Access to memory past its lifetime end, including another `delete`, is undefined behavior.
 
+=== What is the potential impact
 
-=== Noncompliant code example
+The danger of a "double-free" comes directly from the fact that it is undefined behavior.
+Both C and {cpp} standards describe it as such.
```
necto, line 16: As an option; to make it shorter without losing information, and arguably with no loss of readability. Feel free to disregard it.
```suggestion
The danger of a "double-free" comes directly from the fact that it is undefined behavior (in both C and {cpp}).
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,274 @@
...
+- Demonstrative Example 2 on https://cwe.mitre.org/data/definitions/415[CWE-415] presents a set of circumstances where a crash does not occur.
+In these circumstances, the corruption of the mentioned data structures causes two later calls to `malloc` to return the same pointer.
+This can lead to a sensitive data exposure vulnerability or a buffer overflow vulnerability.
+
+== How to fix it
+
+Besides just ensuring that every release happens once, the following best practices can also be used:
+
+- Release any allocated memory in the function it was acquired in.
+Ideally, the release should be independent of any additional conditions.
```
necto, line 33: Given this is already a recommendation, no need for the qualifier
```suggestion
The release should be independent of any additional conditions.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,274 @@
...
-  cp = NULL; // This will prevent freeing the same memory again
+  free(cp); // Compliant: if the memory was freed in the if-block above, free(NULL) is a no-op
 }
 ----
 
+Example for {cpp}:
+
+==== Noncompliant code example
+
+Remove the unnecessary invocation of `delete`:
```
necto, line 96: Since this applies only to the first snippet in the "Compliant solution", it is better placed in that section

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,274 @@
...
+  // ...
+  if (condition) {
+    // ...
+  }
+
+  delete p; // Compliant: no previous call to free in the above branch
+}
+----
+
+Set the pointer to `nullptr` after any invocation of `delete`:
```
necto, line 128: If it is really "after **any** `delete`", the "compliant solution" does not follow this advice, because it does not nullify `p` at the end of the function
```suggestion
Set the pointer to `nullptr` after invoking `delete`:
```

necto (changes requested):
Please, also update the quick fix status in the metadata

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,274 @@
...
+  }
+
+  delete p; // Compliant: if the memory was freed in the if-block above, operator delete of nullptr is a no-op
+}
+----
+
+=== Going the extra mile
+
+
+include::../../../shared_content/cfamily/memory_raii_extra_mile.adoc
```
necto, line 148:
```suggestion
include::../../../shared_content/cfamily/memory_raii_extra_mile.adoc[]
```

rules/S3520/cfamily/rule.adoc
LoghinVladDev: None, habbit. Will remove

rules/S3520/cfamily/rule.adoc
LoghinVladDev: Intention was to exemplify how can one reach that pitfall. Regardless, with more attentiveness, they can also be understood from the compliant/noncompliant, and from the now reduced expl.

So yeah, reduced the explanation, confirm if it is better as succint.

rules/S3520/cfamily/rule.adoc
LoghinVladDev: For a while I assumed a double free is a guaranteed crash, and I'm pretty sure a number of cpp users assume the same.
Having that crash guarantee allows a certain group of people go with "I'll investigate if it's a crash". While in this case, I'm trying to emphasise that - just because in most cases it does, you might have a special scenario where it does not crash, and you should investigate regardless.

necto: I see, thanks for the explanation

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
 
-A best practice to avoid this bug calls for setting just-freed pointers to ``++NULL++``, and always null-testing before a ``++free++`` or ``++delete++``.
+The {cpp} standard defines the first `delete` call as the end of the lifetime for dynamically allocated memory.
+Access to memory past its lifetime end, including another `delete`, is undefined behavior.
 
+=== What is the potential impact
 
-=== Noncompliant code example
+The danger of a "double-free" comes directly from the fact that it is undefined behavior (in both C and {cpp}).
+Due to this, there is no guarantee that a crash will happen on a "double-free" when the resource is released or at all until the end of the program's execution.
```
necto, line 16: It is not clear what "this" refers to here (I guess an artifact of rewriting)
```suggestion
Note that there is no guarantee that a crash will happen on a "double-free" when the resource is released or at all until the end of the program's execution.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
+This also comes from confusion raised by the constructor overloads of this class.
+
+Noncompliant code example
+[source,cpp,diff-id=4,diff-type=noncompliant]
+----
+void doSomething(char const* data) {
+  // Creating another unique_ptr
+  std::unique_ptr<std::string> p1(new std::string(data));
+  // ...
+  std::unique_ptr<std::string> p2(p1.get()); // Noncompliant: Since the copy constructor of std::unique_ptr is deleted, an user might attempt this.
```
necto, line 211: Same as with the shared_ptr
```suggestion
  std::unique_ptr<std::string> p2(p1.get()); // Noncompliant: both p1 and p2 own the pointer now.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
+
+  // Transfering ownership
+  std::unqiue_ptr<std::string> p3(new std::string(data));
+  // ...
+  std::unique_ptr<std::string> p4(p3.get()); // Noncompliant: The intention was to transfer ownership, “double-free” will happen.
```
necto, line 216:
```suggestion
  std::unique_ptr<std::string> p4(p3.get()); // Noncompliant: both p3 and p4 own the pointer now.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
+----
+void doSomething(char const* data) {
+  std::shared_ptr<std::string> p1(new std::string(data));
+  std::shared_ptr<std::string> p2(p1.get()); // Noncompliant: Intended: std::shared_ptr<string> p2(p1);
+}
+----
+
+Both objects are created from the raw pointer, leading them to have independent control blocks. Both will attempt to release the original memory area.
+
+To avoid this, only use copy/move construction and the `std::make_*` family of functions on initialization.
```
necto, line 188: Is there another function from std::make_* that is applicable here?
```suggestion
To avoid this, only use copy/move construction for sharing ownership or the `std::make_shared` for initialization.
```

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
 
-[source,cpp]
+The effects of a "double-free" depend entirely on the program's memory management implementation.
+In the case of such an event, one of the following can be observed:
+
+- The program's memory-management data structures can become corrupted.
+This will usually cause a crash.
+- Demonstrative Example 2 on https://cwe.mitre.org/data/definitions/415[CWE-415] presents a set of circumstances where a crash does not occur.
+In these circumstances, the corruption of the mentioned data structures causes two later calls to `malloc` to return the same pointer.
+This can lead to a sensitive data exposure vulnerability or a buffer overflow vulnerability.
```
necto, line 25:
```suggestion
This can lead to a sensitive-data-exposure vulnerability or a buffer-overflow vulnerability.
```

necto (changes requested):

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
+
+  // Transfering ownership
+  auto p3 = std::make_unique<std::string>(data);
+  auto p4 = std::move(p3); // Compliant: Will invoke move constructor, which is present in std::unique_ptr.
+}
+----
+
+==== Out-of-scope access
+
+Extra care must be taken when passing the memory address contained in a RAII object.
```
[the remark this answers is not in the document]
LoghinVladDev: I was really oscillating between the two

rules/S3520/cfamily/rule.adoc
```diff
@@ -1,51 +1,267 @@
...
+Passing this pointer to a function that might release it can lead to a "double-free" and "use-after-free".
 
 == Resources
 
-* https://cwe.mitre.org/data/definitions/415[MITRE, CWE-415] - Double Free
-* https://owasp.org/www-community/vulnerabilities/Doubly_freeing_memory[OWASP, Doubly freeing memory]
+=== Documentation
+
+* {cpp} reference - https://en.cppreference.com/w/cpp/memory/unique_ptr[std::unique_ptr]
+* {cpp} reference - https://en.cppreference.com/w/cpp/memory/shared_ptr[std::shared_ptr]
```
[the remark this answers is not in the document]
LoghinVladDev: Okay, assumed as they're also referenced in std::unique/shared_ptr pages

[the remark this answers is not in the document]
necto: True, so maybe the links will be redundant. your call

rules/S3520/cfamily/rule.adoc
```diff
@@ -164,7 +164,7 @@ void doSomething(int size) {
 
 === Pitfalls
 
-However, any raw pointers of references to memory held by _RAII_ objects may still lead to a "double-free".
+However, keeping a raw pointer or reference to memory held by _RAII_ objects may still lead to a "double-free".
```
necto, line 167: Sorry, my previous suggestion was incomplete (and Grammarly didn't help :( )
```suggestion
However, keeping a raw pointer or a reference to memory held by _RAII_ objects may still lead to a "double-free".
```

LoghinVladDev:
> That was a long one, but we've done it!

Not expecting memory issues to be quick, glad to be able to dive into it

PR #2629, opened 2023-07-25
Modify rule S3529: Adapt to LaYC

tomasz-kaminski-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3529/cfamily/rule.adoc
+++ b/rules/S3529/cfamily/rule.adoc
@@ -1,27 +1,151 @@
+Accessing block of the memory that was already has undefined behavior.
+This rule flag access via pointer or references to released heap memory.
+
 == Why is this an issue?
 
-Once a block of memory has been ``++free++``d, it becomes available for other memory requests. Whether it's re-used immediately, some time later, or not at all is random, and may vary based on load. Because of that randomness, tests may pass when running locally, but the odds are that such code will fail spectacularly in production by returning strange values, executing unexpected code, or causing a program crash.
+A program may allocated additional block of the memory, using the the `malloc`.
+When no longer needed such block of memory are released using `free` funciton.
+Reading or writting to a heap allocator memory after it was released leads to undefined behavior.
 
+[source,c]
+----
+char *cp = malloc(sizeof(char)*10); // memory is allocated
+// all bytes in cp can be used here
+free(cp); // memory is released
+cp[9] = 0; // Noncompliant: memory is used after it was released
+----
 
-=== Noncompliant code example
+In addition to `malloc` and `free` pair, in `C++` a heap memory may be acquired by use of operator `new`,
+and later released using operator `delete`.
 
 [source,cpp]
 ----
-char *cp = malloc(sizeof(char)*10);
-
-// ...
-free(cp);
-
-cp[9] = 0;  // Noncompliant
+int *intArray = new int[20]; // memory is allocated
+// elemements of intArray can be written or read here
+delete[] intArray; // memory is released
+intArray[3] = 10; // Noncompliant: memory is used after it was released
 ----
 
+Releasing a memory block by invoking `free` or operator `delete`,
+indicates the memory managment system that given block is no longer used by the program.
+Depending on the state and load of the program, such block can be then:
+ * reused, i.e., same pointer is returned by allocation funciton,
+ * released to operating system, making it uncessible to the program.
+
+=== What is the potential impact?
+
+Accessing an released memory causes undefined behavior.
+This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.
+
+Pratically this has a wide range of the effects.
+
+Firstly, the program may crash due the memory no longer being accessible,
+or due unexpected value being read or written via the pointer.
+
+Moreover, reading from the released memory, may produce a garbage value,
+or in case when such memory was already reused to store sensitve data, such a pasword,
+may lead to vulnerability that uses such a defect to extract information from intance of the program.
+
+Finally, writting to released memory, may change value of the unrelated object in remote part of the code,
+if the memory was reused by it.
+As different object may reuse same block of memory between runs, this leads to unintuitive and hard diagnose bugs in programs.
+
+
+== How to fix it
+
+In most situations, the use of uninintialized object is an strong indication of the defect in the code,
+and fixing it requires review of the object allocation and deallocaiton strategies.
+Generally, fixing the issues, requires adjusting the allocation code, so either:
+
+* Moving accesses to memory before the deallocation
+* Moving the deallocaiton so it happens after all the uses
+
+If possible, it is desired to remove manual memory allocations,
+and replace them with stack allocated object, or in case of {cpp},
+an stack object that manages memory (that uses RAII idiom).
+
+=== Noncompliant code example
+
+[source,c,diff-id=1,diff-type=noncompliant]
+----
+int *intArray = malloc(sizeof(int)*10);
+// ...
+free(intArray);
+intArray[9] = 0;  // Noncompliant
+----
+
+=== Compliant solution
+
+Releasing memory after all of its uses.
+
+[source,c,diff-id=1,diff-type=compliant]
+----
+int *intArray = malloc(sizeof(int)*10);
+// ...
+intArray[9] = 0;  // Compliant
+free(intArray);
+----
+
+Alternatively, if the size of array is know at compile-time:
+
+[source,c]
+----
+int intArray[10];
+// ...
+intArray[9] = 0;  // Compliant
+----
+
+In {cpp}, `std::vector` can be used with arbitrally number of elements:
+
+[source,cpp]
+----
+std::vector<int> intArray;
+// ...
+intArray[9] = 0;  // Compliant
+----
+
+=== Going the extra mile
+
+In {cpp}, using manually allocation and deallocating memory is considered a code smell.
+It is recommended to follow a _RAII_ idiom, and create that manage the memory by allocating it in constructor and freeing in destructor.
+Such object can be then passed between the functions (either as argument or by being returned) instead of raw pointers.
+Passing such object by value gaurantees that the heap memory remains accessible, when accessed via the object.
+
+[source,cpp]
+----
+std::vector<int> intArray; // manages array on heap
+std::unique_ptr<Class> objPtr = std::make_unique<Class>(); // manages object on heap
+
+intArray[5]; // OK
+objPtr->foo(); // OK
+----
+
+However, any raw pointers or references to memory hold by such object may still lead to use after freee:
+[source,cpp]
+----
+int* p1 = &intArray[0];    // becomes dangling when intArray is destroyed
+int* p2 = intArray.data(); // same as above
+Class* p3 = objPtr.get();  // becomes dangling, when objPtr releases the pointer
+----
 
 == Resources
 
-* https://cwe.mitre.org/data/definitions/416[MITRE, CWE-416] - Use After Free
-* https://wiki.sei.cmu.edu/confluence/x/GdYxBQ[CERT, MEM30-C.] - Do not access freed memory
-* https://wiki.sei.cmu.edu/confluence/x/onw-BQ[CERT, MEM50-CPP.] - Do not access freed memory
-* https://wiki.sei.cmu.edu/confluence/x/OXw-BQ[CERT, EXP54-CPP.] - Do not access an object outside of its lifetime
+=== Documentation
+
+- C++ reference - https://en.cppreference.com/w/cpp/language/raii[RAII]
+- C++ reference - https://en.cppreference.com/w/cpp/memory/unique_ptr[std::unique_ptr]
+- C++ reference - https://en.cppreference.com/w/cpp/memory/shared_ptr[std::shared_ptr]
+
+=== Standards
+
+* CWE - https://cwe.mitre.org/data/definitions/416[416 - Use After Free]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/GdYxBQ[MEM30-C - Do not access freed memory]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/onw-BQ[MEM50-CPP - Do not access freed memory]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/OXw-BQ[EXP54-CPP - Do not access an object outside of its lifetime]
+
+=== Related rules
+
+* S5025 recommends avoiding manual memory management
 
 ifdef::env-github,rspecator-view[]
 


rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
+Accessing a block of the memory that was already freed has undefined behavior.
+This rule flag access via pointer or references to released heap memory.
```
LoghinVladDev, line 2:
```suggestion
Accessing a memory block that was already freed is undefined behavior.
This rule flags access via pointer or references to released heap memory.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+
 == Why is this an issue?
 
-Once a block of memory has been ``++free++``d, it becomes available for other memory requests. Whether it's re-used immediately, some time later, or not at all is random, and may vary based on load. Because of that randomness, tests may pass when running locally, but the odds are that such code will fail spectacularly in production by returning strange values, executing unexpected code, or causing a program crash.
+A program may allocate an additional block of memory, using the `malloc` function.
+When no longer needed such blocks of memory are released using the `free` function.
+Reading or writing to a heap allocator memory after it was released leads to undefined behavior.
```
LoghinVladDev, line 8:
```suggestion 
A program may allocate an additional memory block using the `malloc` function.
When no longer needed, such memory blocks are released using the `free` function.
After it is released, reading or writing to a heap-allocated memory block leads to undefined behavior.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
 ----
-char *cp = malloc(sizeof(char)*10);
+int *intArray = new int[20]; // memory is allocated
+// elements of intArray can be written or read here
+delete[] intArray; // memory is released
+intArray[3] = 10; // Noncompliant: memory is used after it was released
+----
+
+Releasing a memory block by invoking `free` or operator `delete`,
+indicates the memory management system that the given block is no longer used by the program.
```
LoghinVladDev, line 30:
```suggestion
Releasing a memory block by invoking `free` or operator `delete`,
informs the memory management system that the given block is no longer in use by the program.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+Depending on the state and load of the program, such block can be then:
+ * reused, i.e., the same pointer is returned by allocation function,
```
LoghinVladDev, line 32:
```suggestion
 * reused, i.e., the same pointer is returned by the allocation function,
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
-char *cp = malloc(sizeof(char)*10);
+int *intArray = new int[20]; // memory is allocated
+// elements of intArray can be written or read here
+delete[] intArray; // memory is released
+intArray[3] = 10; // Noncompliant: memory is used after it was released
+----
+
+Releasing a memory block by invoking `free` or operator `delete`,
+indicates the memory management system that the given block is no longer used by the program.
+Depending on the state and load of the program, such block can be then:
```
LoghinVladDev, line 31: Missing new line between "...then:" and * causes list not to render on my end.

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+Accessing a released memory causes undefined behavior.
+This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.
+
+Practically this has a wide range of effects:
+
+* The program may crash due to the memory no longer being accessible,
+  or due to unexpected value being read or written via the pointer.
+* Reading from the released memory may produce a garbage value,
+  or in the case when such memory was already reused to store sensitive data, such as passwords,
+  may lead to a vulnerability that uses such a defect to extract information from an instance of the program.
```
LoghinVladDev, line 46: I would separate the two
```suggestion
* Reading from the released memory may produce a garbage value.
* When such memory was already reused to store sensitive data, such as passwords, it may lead to a vulnerability that uses such a defect to extract information from an instance of the program.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+* Writing to released memory, may change the value of the unrelated object in a remote part of the code if the memory was reused by it.
+  As different objects may reuse same block of memory between runs, this leads to unintuitive and hard diagnose bugs in programs.
```
LoghinVladDev, line 48:
```suggestion
  As different objects may reuse same the block of memory between runs, this leads to unintuitive and hard diagnose bugs in programs.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+[source,cpp]
+----
+std::vector<int> intArray;
+// ...
+intArray[9] = 0;  // Compliant
+----
+
+=== Going the extra mile
+
+In {cpp}, using manually allocating and deallocating memory is considered a code smell.
```
LoghinVladDev, line 106: Either
```suggestion
In {cpp}, manually allocating and deallocating memory is considered a code smell.
```
Or
```suggestion
In {cpp}, using manual memory management is considered a code smell.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+objPtr->foo(); // OK
+----
+
+However, any raw pointers or references to memory held by such object may still lead to a use after free:
+[source,cpp]
+----
+int* p1 = &intArray[0];    // becomes dangling when intArray is destroyed
+int* p2 = intArray.data(); // same as above
+Class* p3 = objPtr.get();  // becomes dangling, when objPtr releases the pointer
+----
```
LoghinVladDev, line 127: Maybe specify smart pointers release & move for such cases when transfer to out of scope of heap allocated memory is required?

LoghinVladDev (changes requested):
Also be sure to do another Grammarly pass for minor findings

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+
+
+== How to fix it
+
+In most situations, the use of an uninitialized object is a strong indication of a defect in the code,
+and fixing it requires a review of the object allocation and deallocation strategies.
+Generally, fixing the issues requires adjusting the allocation code, so either:
 
+* Moving accesses to memory before the deallocation
+* Moving the deallocation so it happens after all the uses
```
necto, line 58: Here in the preambule you are speaking about "allocation code", but in both bullet points you are only mentioning the "deallocation"
```suggestion
Generally, fixing the issues requires adjusting the allocation code, so either:

* Moving accesses to memory before the deallocation
* Moving the deallocation so it happens after all the uses
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,148 @@
...
+intArray[9] = 0;  // Compliant
+----
+
+In {cpp}, `std::vector` can be used with an arbitrally number of elements:
+
+[source,cpp]
+----
+std::vector<int> intArray;
+// ...
+intArray[9] = 0;  // Compliant
```
necto, line 101: This does not mention that you've preallocated vector to hold at least 10 elements, so if the `// ...` does not resize the vector, this "compliant solution" will crash

necto (changes requested):
Do you think there is a case where a quick fix is feasible?
If not, I'd suggest updating it in the metadata

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Could you elaborate a bit more. I was thinking mostly about cases, where you escape raw poitner to object owner by unique_ptr, and then use it after unique_ptr was destroyed.

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I just remove the workd allocation from the sentence before the list. Now it says "adjusting the code, so either".

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Good catch. I added line with resize.

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,149 @@
...
+[source,c]
+----
+char *cp = malloc(sizeof(char)*10); // memory is allocated
+// all bytes in cp can be used here
+free(cp); // memory is released
+cp[9] = 0; // Noncompliant: memory is used after it was released
+----
 
-=== Noncompliant code example
+In addition to the `malloc` and `free` pair, in `C++` a heap memory may be acquired by use of the operator `new`,
```
necto, line 18:
```suggestion
In addition to the `malloc` and `free` pair, in {cpp} a heap memory may be acquired by use of the operator `new`,
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,149 @@
...
 ----
-char *cp = malloc(sizeof(char)*10);
+int *intArray = new int[20]; // memory is allocated
+// elements of intArray can be written or read here
+delete[] intArray; // memory is released
+intArray[3] = 10; // Noncompliant: memory is used after it was released
+----
+
+Releasing a memory block by invoking `free` or operator `delete`,
+informs the memory management system that the given block is no longer in use by the program.
```
necto, line 30: Comma between the subject and the verb breaks the sentence
```suggestion
Releasing a memory block by invoking `free` or operator `delete`
informs the memory management system that the given block is no longer in use by the program.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,149 @@
...
+  As different objects may reuse same the block of memory between runs, this leads to unintuitive and hard diagnose bugs in programs.
+
+
+== How to fix it
+
+In most situations, the use of an uninitialized object is a strong indication of a defect in the code,
+and fixing it requires a review of the object allocation and deallocation strategies.
+Generally, fixing the issues requires adjusting the code, so either:
+
+* Moving accesses to memory before the deallocation
```
necto, line 57:
```suggestion
* Moving accesses to the memory before the deallocation
```

necto (changes requested):

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,149 @@
...
+* Moving the deallocation so it happens after all the uses
+
+If possible, it is desired to remove manual memory allocations,
+and replace them with stack-allocated objects, or in case of {cpp},
+a stack object that manages memory (that uses RAII idiom).
+
+=== Noncompliant code example
```
necto, line 64: This and "Compliant solution" should be nested under "Code examples" subsection

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,149 @@
...
+intArray[9] = 0;  // Compliant
+----
+
+In {cpp}, `std::vector` can be used with an arbitrally number of elements:
 
-cp[9] = 0;  // Noncompliant
+[source,cpp]
+----
+std::vector<int> intArray;
+intArray.resize(10);
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: I went for resizing, because there is huge difference in meaning between:
```
std::vector<int> intArray(10); // 10 elements,
std::vector<int> intArray{10}; // 1 element
```
And I didn't want this rule to go into that trap, and potentially confuse someone. Which would defeat the point of this being better.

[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: So, I went for something that is unambigous.

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Thanks. Good catch.

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+Releasing a memory block by invoking `free` or operator `delete`
+informs the memory management system that the given block is no longer in use by the program.
+Depending on the state and load of the program, such block can be then:
+
+ * reused, i.e., the same pointer is returned by the allocation function,
+ * released to the operating system, making it inaccessible to the program.
+
+=== What is the potential impact?
+
+Accessing a released memory causes undefined behavior.
```
necto, line 38: "computer memory" is uncountable.
"memory" that is countable is "remembering",
e.g.,  I have good memorie**s** of my vacation trip. (countable) vs. My memory is good. (uncountable)
```suggestion
Accessing released memory causes undefined behavior.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.
+
+Practically this has a wide range of effects:
+
+* The program may crash due to the memory no longer being accessible,
+  or due to unexpected value being read or written via the pointer.
+* Reading from the released memory may produce a garbage value.
+* When such memory was already reused to store sensitive data, such as passwords, it may lead to a vulnerability that uses such a defect to extract information from an instance of the program.
```
necto, line 46: stylistic - avoids repeating word "such" in the sentence
```suggestion
* When the memory was already reused to store sensitive data, such as passwords, it may lead to a vulnerability that uses this defect to extract information from an instance of the program.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+* Writing to released memory, may change the value of the unrelated object in a remote part of the code if the memory was reused by it.
```
necto, line 47: Comma between subject and verb breaks the sentence
```suggestion
* Writing to released memory may change the value of the unrelated object in a remote part of the code if the memory was reused by it.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+  As different objects may reuse same the block of memory between runs, this leads to unintuitive and hard diagnose bugs in programs.
```
necto, line 48: IMO, it is obvious that we are only talking about bugs in programs and no other kinds of bugs
```suggestion
  As different objects may reuse same the block of memory between runs, this leads to unintuitive and hard diagnose bugs.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+
+
+== How to fix it
 
+In most situations, the use of an uninitialized object is a strong indication of a defect in the code,
+and fixing it requires a review of the object allocation and deallocation strategies.
+Generally, fixing the issues requires adjusting the code, so either:
```
necto, line 55: Stylistic, to avoid 3 subsequent gerunds ("fixing -> adjusting -> moving")
```suggestion
Generally, the fix requires adjusting the code, so either:
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+
+* Moving accesses to the memory before the deallocation
+* Moving the deallocation so it happens after all the uses
+
+If possible, it is desired to remove manual memory allocations,
+and replace them with stack-allocated objects, or in case of {cpp},
+a stack object that manages memory (that uses RAII idiom).
```
necto, line 62: Here the plurality of "stack-allocated objects"/"stack object" does not match (last two subjects are "object" vs "objects").

How about this?
```suggestion
If possible, it is desired to remove manual memory allocation,
and replace them with stack-allocated objects, or in case of {cpp},
stack objects that manage memory (that uses RAII idiom).
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+int *intArray = (int*)malloc(sizeof(int)*10);
 // ...
-free(cp);
+free(intArray);
+intArray[9] = 0;  // Noncompliant
+----
+
+=== Compliant solution
+
+Releasing memory after all of its uses.
```
necto, line 78: Subjective. IMO, an imperative is more straightforward than a description
```suggestion
Release the memory after all of its uses.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
-cp[9] = 0;  // Noncompliant
+[source,c,diff-id=1,diff-type=compliant]
+----
+int *intArray = (int*)malloc(sizeof(int)*10);
+// ...
+intArray[9] = 0;  // Compliant
+free(intArray);
 ----
 
+Alternatively, if the size of the array is known at compile-time:
```
necto, line 88: Since you give the description of the previous solution "release the memory", why not summarize this one?
```suggestion
Alternatively, allocate the array on the stack,
if the size of the array is known at compile-time:
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+
+[source,c]
+----
+int intArray[10];
+// ...
+intArray[9] = 0;  // Compliant
+----
+
+In {cpp}, `std::vector` can be used with an arbitrally number of elements:
```
necto, line 97: If we go with imperative
```suggestion
In {cpp}, use `std::vector` with an arbitrary number of elements:
```
Or, at least fix the typo
```suggestion
In {cpp}, `std::vector` can be used with an arbitrary number of elements:
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+std::vector<int> intArray;
+intArray.resize(10);
+// ...
+intArray[9] = 0;  // Compliant
+----
+
+=== Going the extra mile
+
+In {cpp}, manually allocating and deallocating memory is considered a code smell.
+It is recommended to follow an _RAII_ idiom and create a class, that manages the memory by allocating it in the constructor and freeing it in the destructor.
```
necto, line 110: Depending on how you pronounce RAII, it may require "a" or "an" for indefinite article. At the same time, isn't it a single relatively well-known idiom? Let's use the definite article
Also, no need for comma before "that"
```suggestion
It is recommended to follow the _RAII_ idiom and create a class that manages the memory by allocating it in its constructor and freeing it in its destructor.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+Such an object can be then passed between the functions (either as an argument or by being returned) instead of raw pointers.
```
necto, line 111: I did not see the "functions" mentioned before, so they can't be used with a definite article
```suggestion
Such an object can be then passed between functions (either as an argument or by being returned) instead of raw pointers.
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+Passing such object by value guarantees that the heap memory remains accessible when accessed via the object.
```
necto, line 112: This one, to be honest, is not clear.
As you only mentioned allocating in a constructor and deallocating in a destructor,
what happens to the memory and the pointer to it when you pass the object by value?
Will it be reallocated on each copy? Will it be shared like with a `shared_ptr`?
The first thing I imagined when reading this sentence is that it will segfault when I copy the pointer by passing the object by value, the original deallocates the memory, and the copy dereferences its pointer.

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+
+[source,cpp]
+----
+std::vector<int> intArray; // manages array on heap
```
necto, line 116: You don't have to follow English rules in a comment, but if you can afford it, why not
```suggestion
std::vector<int> intArray; // manages an array on the heap
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,151 @@
...
+std::unique_ptr<Class> objPtr = std::make_unique<Class>(); // manages object on heap
```
necto, line 117: You don't have to follow English rules in a comment, but if you can afford it, why not
```suggestion
std::unique_ptr<Class> objPtr = std::make_unique<Class>(); // manages an object on the heap
```

rules/S3529/cfamily/rule.adoc
necto, line 116: Here again, the code as it stands will result in a crash because you allocate the vector of 0 size and access the 6-th element three lines below

necto (changes requested):

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: "the RAII" sounds good.

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Indeed that is cryptic. I will try to rewrite this and the previous sentence, to provide more context.

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I rewrote this section based on Arseniy suggestion. Could you please check if that address you concerns?

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I rewrote the section based on your suggestion. Could you please take a look.

rules/S3529/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Good catch. Updated, and used a constructor with adjusted comment.

rules/S3529/cfamily/rule.adoc
LoghinVladDev: Yep, is does address it

LoghinVladDev (approved):
Looks good! Before merging, check grammarly again, as it does provide some insightful rewritings for clarity, and one rewrite for the recommended confident tone.

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,155 @@
...
+
+=== Going the extra mile
+
+In {cpp}, manually allocating and deallocating memory is considered a code smell.
+It is recommended to follow the _RAII_ idiom and create a class that manages the memory by allocating it when the object is constructed and freeing it when it is destroyed.
+Furthermore, copy and move operations on such objects are designed such that this object can be passed by value between functions (either as an argument or by being returned)
+in place of raw pointers.
+Depending on the type, passing an _RAII_ object operations may either allocate a new block of memory and copy the elements (`std::vector`),
+transfer ownership of the memory to constructed object (`std::unique_ptr`), or shared ownership and free memory when the last object is destroyed (`std::shared_ptr`).
+This guarantees that accessing a memory managed by such an object is not released as long as such an object is not modified or destroyed (some _RAII_ types provide a stronger guarantee).
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Thanks, for suggestion.

rules/S3529/cfamily/rule.adoc
```diff
@@ -1,27 +1,161 @@
...
+
+In most situations, the use of an uninitialized object is a strong indication of a defect in the code,
+and fixing it requires a review of the object allocation and deallocation strategies.
+Generally, the fix requires adjusting the code, so either:
+
+* Moving accesses to the memory before the deallocation
+* Moving the deallocation so it happens after all the uses
 
+If possible, it is desired to remove manual memory allocation,
+and replace them with stack-allocated objects, or in case of {cpp},
```
necto, line 61: Sorry, when suggesting a fix, I missed "them".
In fact, you could shorten this part:
```suggestion
and use stack-allocated objects, or in the case of {cpp},
```
or more conservatively
```suggestion
and replace it with stack-allocated objects, or in the case of {cpp},
```

necto (commented):
You have a good "going the extra mile" section about RAII.
There are other rules that could share this "going the extra mile" section.
Here is one, merged, with its own version:
https://github.com/SonarSource/rspec/blob/master/rules/S1232/cfamily/rule.adoc#going-the-extra-mile

Here are two still to be reviewed,
one with a very short mention:
https://github.com/SonarSource/rspec/pull/2612/files#diff-db51b8ece43cc28aa3e104886ce2891212e7ddd8788a3c87d9b46d4e4a1208abR73
one with a more expanded proposal:
https://github.com/SonarSource/rspec/pull/2623/files#diff-1ed012e909700770758b20353d520851b756620d46499fa6a5b34acc1773584eR145

Do you think you should share this section with any of the rules above?

tomasz-kaminski-sonarsource:
I wonder about the RAII section. I think that each rule may on a bit different aspect of the RAII.
For example, in the case of my rule the important fact is when memory is released, and how the passing object around handles it.
For S1232 (matching allocation with deallocation), the RAII aspect that is important is, the allocation and deallocation handling is done by single class, that couples the operation together. So this is how it prevents a mismatch on them.

However, my section seems to be a good fit for the new rules. About freeing and deleting.

tomasz-kaminski-sonarsource:
I have extract part of RAII to shared. Let me know what do you think.

rules/S1232/cfamily/rule.adoc
```diff
@@ -107,17 +107,15 @@ free(_pChar); // Compliant: free the memory C-style
...
 
-Manually allocating and deallocating memory is a code smell
-because of the above, and because it is easy to leak memory.
+include::../../../shared_content/cfamily/memory_raii_extra_mile.adoc[]
 
-Whenever possible, prefer the RAII (Resource Acquisition Is Initialization) {cpp} idiom.
-
-If, in the example above, you use the appropriate standard objects,
-you do not need to remember how and when to deallocate them:
+This guarantees that the memory managed by such oobject is automatically deallocated by destructor,
```
necto, line 112:
```suggestion
This guarantees that the memory managed by such object is automatically deallocated by the destructor,
```

rules/S3529/cfamily/rule.adoc
```diff
@@ -58,7 +58,7 @@ Generally, the fix requires adjusting the code, so either:
 * Moving the deallocation so it happens after all the uses
 
 If possible, it is desired to remove manual memory allocation,
-and replace them with stack-allocated objects, or in case of {cpp},
+and replace it with stack-allocated objects, or in the case of {cpp},
 stack objects that manage memory (that uses RAII idiom).
```
necto, line 62: Sorry, another missing place for the fix to be complete ("that uses" can't refer to "stack objects").
```suggestion
stack objects that manage memory (using RAII idiom).
```

necto (approved):
Looking good. Thanks for also improving S1232

PR #2633, opened 2023-07-25
Modify rule S6214: Convert to LaYC format

LoghinVladDev:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6214/cfamily/rule.adoc
+++ b/rules/S6214/cfamily/rule.adoc
@@ -1,4 +1,4 @@
-Compare integers safely using `std::cmp_*` functions to avoid any unexpected results.
+Compare integers of mixed signedness safely using `std::cmp_*` functions to avoid any unexpected results.
 
 == Why is this an issue?
 
@@ -11,7 +11,7 @@ This rule raises an issue when an unsigned integer is compared with a negative v
 
 === What is the potential impact
 
-Integer comparisons are usually used in creating branch conditions.
+Integer comparisons are often used in branch conditions.
 An unexpected result from one of these conditions can lead to hard-to-detect issues.
 
 For example, using container size functions in a comparison can lead to such a problem since these return an unsigned integer value.
@@ -63,7 +63,7 @@ if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns
 == Resources
 
 * S845 - a more generic rule about mixing signed and unsigned values.
-* S6183 - a version of this rule that triggers when signed and unsigned variables are compared.
+* S6183 - a version of this rule that triggers when any signed and unsigned variables are compared.
 
 
 ifdef::env-github,rspecator-view[]
--- a/rules/S6214/cfamily/rule.adoc
+++ b/rules/S6214/cfamily/rule.adoc
@@ -71,6 +71,10 @@ if (0 <= v.size() && v.size() < 100) { // Compliant, even though v.size() return
 
 == Resources
 
+=== Documentation
+
+* {cpp} reference - https://en.cppreference.com/w/cpp/utility/intcmp[std::cmp_* functions]
+
 === Related rules
 
 * S845 - a more generic rule about mixing signed and unsigned values.
--- a/rules/S6214/cfamily/rule.adoc
+++ b/rules/S6214/cfamily/rule.adoc
@@ -26,7 +26,7 @@ For example, `std::cmp_less(2U, -1)` is `false`.
 
 ==== Noncompliant code example
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
 bool less = 2U < -1; // Noncompliant
 
@@ -40,7 +40,7 @@ if (x < y) { // Noncompliant
 
 ==== Compliant solution
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=compliant]
 ----
 bool less = std::cmp_less(2U, -1); // Compliant
 
--- a/rules/S6214/cfamily/rule.adoc
+++ b/rules/S6214/cfamily/rule.adoc
@@ -22,7 +22,7 @@ For example, using container size functions in a comparison can lead to such a p
 These functions correctly handle negative numbers and lossy integer conversion.
 For example, `std::cmp_less(2U, -1)` is `false`.
 
-=== Code Examples
+=== Code examples
 
 ==== Noncompliant code example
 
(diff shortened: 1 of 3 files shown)


rules/S6214/cfamily/rule.adoc
```diff
@@ -1,18 +1,30 @@
+Compare integers safely using `std::cmp_*` functions to avoid any unexpected results.
```
necto, line 1:
```suggestion
Compare integers of mixed signedness safely using `std::cmp_*` functions to avoid any unexpected results.
```

rules/S6214/cfamily/rule.adoc
```diff
@@ -1,18 +1,30 @@
...
+
 == Why is this an issue?
 
-Comparison between `signed` and `unsigned` integers is dangerous because it produces counterintuitive results outside of their common range of values.
+Comparison between `signed` and `unsigned` integers is dangerous because it produces counter-intuitive results due to implicit conversions.
+When a signed integer is compared to an unsigned one, the former might be converted to unsigned.
+The conversion preserves the two's-complement bit pattern of the signed value that often corresponds to a large unsigned result.
```
necto, line 7: Is this guaranteed to be two's-compliment?

rules/S6214/cfamily/rule.adoc
```diff
@@ -1,18 +1,30 @@
...
+For example, `2U < -1` is `true`.
 
+This rule raises an issue when an unsigned integer is compared with a negative value.
 
-When a signed integer is compared to an unsigned one, the former might be converted to unsigned. The conversion preserves the two's-complement bit pattern of the signed value that often corresponds to a large unsigned result. For example, `2U < -1` is `true`.
+=== What is the potential impact
 
+Integer comparisons are usually used in creating branch conditions.
```
necto, line 14:
```suggestion
Integer comparisons are often used in branch conditions.
```

rules/S6214/cfamily/rule.adoc
```diff
@@ -1,18 +1,30 @@
...
+{cpp}20 introduced a remedy to this common pitfall: a family of `std::cmp_*` functions defined in the `<utility>` header.
+These functions correctly handle negative numbers and lossy integer conversion.
+For example, `std::cmp_less(2U, -1)` is `false`.
 
+=== Code Examples
 
-=== Noncompliant code example
+==== Noncompliant code example
 
 [source,cpp]
```
necto, line 29: I think the diff view here will be helpful

rules/S6214/cfamily/rule.adoc
```diff
@@ -51,7 +63,7 @@ if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns
 == Resources
 
 * S845 - a more generic rule about mixing signed and unsigned values.
-* S6183 - a version of this rule that triggers as soon as signed and unsigned variables are compared.
+* S6183 - a version of this rule that triggers when signed and unsigned variables are compared.
```
necto, line 66:
```suggestion
* S6183 - a version of this rule that triggers when any signed and unsigned variables are compared.
```

rules/S6214/cfamily/rule.adoc
```diff
@@ -51,7 +63,7 @@ if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns
 == Resources
```
necto, line 64:
```suggestion

=== Related rules

```

necto (changes requested):

rules/S6214/cfamily/rule.adoc
LoghinVladDev: Yes, part of the cpp20 standard, and the rule is aimed "since-cpp20"

necto: To avoid confusing the reader, I would qualify this statement with "Since C++20, ...". No need to specify what happens in other standards since the rule is applicable only starting from C++20

necto (changes requested):
A minor fix, and could you, please, update the "quickfix" metadata field?
This needs a second thought, but it looks like it is feasible to always provide a quick fix for this rule

necto:
Additionally, I think it is useful to mention the danger of entering an infinite loop in the "Pitfalls" section

rules/S6214/cfamily/rule.adoc
```diff
@@ -26,9 +38,9 @@ if (x < y) { // Noncompliant
 ----
```
necto, line 38: In the compliant solution you have two more examples:
```
bool fun(int x, std::vector<int> const& v) {
  return std::cmp_less(x, v.size()); // Compliant
}
std::vector<int> v = foo();
if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns an unsigned integer
}
```
It would be instructive to also show their noncompliant version

necto (changes requested):

rules/S6214/cfamily/rule.adoc
```diff
@@ -1,20 +1,32 @@
+Compare integers of mixed signedness safely using `std::cmp_*` functions to avoid any unexpected results.
+
 == Why is this an issue?
 
-Comparison between `signed` and `unsigned` integers is dangerous because it produces counterintuitive results outside of their common range of values.
+Comparison between `signed` and `unsigned` integers is dangerous because it produces counter-intuitive results due to implicit conversions.
+When a signed integer is compared to an unsigned one, the former might be converted to unsigned.
+Since {cpp}20, the conversion preserves the two's-complement bit pattern of the signed value that often corresponds to a large unsigned result.
```
necto, line 7: Depending on your definition of large, but given the sign bit, the negative number converted to unsigned will occupy the largest half of the unsigned range, i.e., will always be bigger than any number representable in the signed type of the same width, and that sounds pretty large to me
```suggestion
Since {cpp}20, the conversion preserves the two's-complement bit pattern of the signed value that always corresponds to a large unsigned result.
```

rules/S6214/cfamily/rule.adoc
```diff
@@ -43,15 +64,17 @@ bool fun(int x, std::vector<int> const& v) {
 }
 
 std::vector<int> v = foo();
-if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns an unsigned integer
+if (0 <= v.size() && v.size() < 100) { // Compliant, even though v.size() returns an unsigned integer
 }
 ----
 
 
 == Resources
```
necto, line 72: Perhaps you can also point to the documentation of the cmp_* functions https://en.cppreference.com/w/cpp/utility/intcmp

necto (approved):

PR #2634, opened 2023-07-25
Modify rule S5797: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5797/rule.adoc
+++ b/rules/S5797/rule.adoc
@@ -7,38 +7,33 @@ When a constant is used as a condition, either it has no effect on the execution
 
 === Noncompliant code example
 
-[source,text,diff-id=1,diff-type=noncompliant]
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
-def func(param = None):
-    param = (1,)
-    if param:  # Noncompliant. var is always set to (1,), the first branch of the if will always execute.
-        return sum(param)
+def foo():
+    a = True
+    if a: # Noncompliant: the condition is always true
+      return 1
     else:
-        return None
-
-var2 = 1 if func else 2  # Noncompliant. "func" will always be equivalent to True.
-var3 = func and 1 else 2  # Noncompliant.
+      return 2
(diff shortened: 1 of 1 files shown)


rules/S5797/rule.adoc
```diff
@@ -41,6 +40,6 @@ var3 = func() and 1 else 2
 
 == Resources
```
joke1196, line 42: the  level 3 documentation title is missing

rules/S5797/rule.adoc
```diff
@@ -1,14 +1,13 @@
...
 When a constant is used as a condition, either it has no effect on the execution flow and it can be removed, or some code will never be executed and it is a bug.
 
 
-This rule raises an issue when a constant expression is used as a condition in an ``++if++``, ``++elif++``, a conditional expression or other boolean expressions.
-
-
 === Noncompliant code example
 
-[source,text]
+[source,text,diff-id=1,diff-type=noncompliant]
```
joke1196, line 10: As discussed we can go with the `python` source type

rules/S5797/rule.adoc
```diff
@@ -1,14 +1,13 @@
...
 ----
 def func(param = None):
     param = (1,)
```
joke1196, line 13: I do not think the examples really convey what the rule is supposed to do. Maybe we can simplify those to properly use constants instead of function objects, which would be more in line with the description. 
And small issue: the example refers to a `var` variable that does not exist. 
Maybe for this one, I would drop the fancy parametrization of the function and really have a small but concrete example.  Also, I don't think the last line of the example is syntactically correct (I would guess a `if` is missing).

joke1196 (changes requested):
Looks good! Only a few issues with the examples they deserve a bit more love I think.

joke1196 (approved):
LGTM

jean-jimbo-sonarsource (approved):
LGTM

PR #2638, opened 2023-07-25
Modify rule S1186: Adapt to LaYC

victor-diez-sonarsource:


rules/S1186/javascript/rule.adoc
```diff
@@ -1,36 +1,42 @@
...
 
-[source,javascript]
+[source,javascript,diff-id=1,diff-type=compliant]
 ----
 function foo() {
+    do_something();
 }
+----
+
+Throw an exception if it is done intentionally because some functionality is not supported.
```
yassin-kammoun-sonarsource, line 20:
```suggestion
Throw an exception if it is done intentionally due to some functionality not supported yet.
```

rules/S1186/javascript/rule.adoc
```diff
@@ -1,36 +1,42 @@
 == Why is this an issue?
 
-There are several reasons for a function not to have a function body:
+Empty functions can be a sign of poor code quality and can make the code harder to read and maintain. It usually happens when a developer forgets to add any statements to it, or when a function is no longer needed but is not removed from the codebase. In these cases when it is an unintentional omission, and should be fixed to prevent an unexpected behavior in production.
```
yassin-kammoun-sonarsource, line 3:
```suggestion
Empty functions can be a sign of poor code quality and can make the code harder to read and maintain. It usually happens when a developer forgets to add any statements to it, or when a function is no longer needed but is not removed from the codebase. In these cases when it is an unintentional omission, it should be fixed to prevent an unexpected behavior in production.
```

yassin-kammoun-sonarsource: Not entirely sure about my suggestion, but something is fishy here :)

yassin-kammoun-sonarsource (approved):

rules/S1186/javascript/rule.adoc
victor-diez-sonarsource: I think yours look good. Indeed, I rephrased it multiple times and left something weird

PR #2640, opened 2023-07-25
Modify rule S5704: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5704/python/rule.adoc
+++ b/rules/S5704/python/rule.adoc
@@ -8,11 +8,11 @@ If the bare "raise" statement is in a ``++finally++`` block, it will only have a
 
 
 
+=== Code examples
 
+==== Noncompliant code example
 
-=== Noncompliant code example
-
-[source,python]
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
 def foo(param):
     result = 0
@@ -32,9 +32,9 @@ def foo(param):
 ----
 
 
-=== Compliant solution
+==== Compliant solution
 
-[source,python]
+[source,python,diff-id=1,diff-type=compliant]
 ----
 def foo(param):
     result = 0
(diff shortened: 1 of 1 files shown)


rules/S5704/python/rule.adoc
```diff
@@ -1,16 +1,18 @@
+This rule raises an issue when a bare ``++raise++`` statements is in a ``++finally++`` block.
```
guillaume-dequenne-sonarsource, line 1: statements -> statement

rules/S5704/python/rule.adoc
```diff
@@ -1,16 +1,18 @@
...
+
 == Why is this an issue?
 
 A bare ``++raise++`` statement, i.e. a ``++raise++`` with no exception provided, will re-raise the last active exception in the current scope. If no exception is active a ``++RuntimeError++`` is raised instead.
 
 If the bare "raise" statement is in a ``++finally++`` block, it will only have an active exception to re-raise when an exception from the ``++try++`` block is not caught or when an exception is raised by an ``++except++`` or ``++else++`` block. Thus bare ``++raise++`` statements should not be relied upon in ``++finally++`` blocks. It is simpler to let the exception raise automatically.
```
guillaume-dequenne-sonarsource, line 7: I vaguely have the feeling that we could be a bit more explicit, saying that in other cases, no exception will be active, leading to the `RuntimeError`.
I also have the feeling we could say explicitly that without any `raise` statement in the block, the exception will propagate automatically anyway (as it is implied in the last sentence, but not very explicitly).

And at the same time I think that the current description is very clear and straight to the point, and would want to make it less readable.

Just sharing my thoughts without asking for changes. Wdyt @jean-jimbo-sonarsource?

guillaume-dequenne-sonarsource (approved):
Looking good!
A minor typo to fix, and a more general comment (which I'm not asking to address, but if you do, let's have a 2nd round of review).

rules/S5704/python/rule.adoc
jean-jimbo-sonarsource: I really like how concise it is .

jean-jimbo-sonarsource (approved):

PR #2647, opened 2023-07-26
Modify rule S2107: Expand and adjust for LaYC

necto:
I moved and closed the Java rule description because they do not plan to implement it, but it is nice to preserve the description for posterity.

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2107/cfamily/rule.adoc
+++ b/rules/S2107/cfamily/rule.adoc
@@ -4,12 +4,13 @@ and might lead to hard-to-catch bugs.
 == Why is this an issue?
 
 In a {cpp} class,
-all member variables of a type with an implicit or explicit constructor will be automatically initialized.
+all member variables of a type with an explicit constructor
+or of non-trivial type with a default constructor will be automatically initialized.
 This is not the case for the others:
 if no initialization is explicitly written,
 the variable will be left uninitialized.
 Trivial types with a ``++default++``ed or implicit default constructor follow the aggregate initialization syntax:
-if you omit their initialization, they will not be initialized.
+if you omit them in the initialization list, they will not be initialized.
 
 [source,cpp]
 ----
@@ -17,7 +18,6 @@ struct Trivial {
   int x;
   int y;
 };
-
 struct Container {
   Trivial t;
   int arr[2];
@@ -39,7 +39,6 @@ struct Container {
     // this->arr is initialized to {1, 0}
   }
 };
-
 struct DefaultedContainer {
   Trivial t;
   int arr[2];
--- a/rules/S2107/cfamily/rule.adoc
+++ b/rules/S2107/cfamily/rule.adoc
@@ -18,8 +18,7 @@ This comes with all the risks associated with uninitialized variables,
 and these risks propagate to all the classes using the faulty class as a type
 as a base class or a data member.
 This is all the more surprising
-that most programmers expect a constructor to correctly initialize the members of its class
-(this is its raison d'être after all).
+that most programmers expect a constructor to correctly initialize the members of its class.
 
 include::../../../shared_content/cfamily/garbage_value_impact.adoc[]
 
--- a/rules/S2107/cfamily/rule.adoc
+++ b/rules/S2107/cfamily/rule.adoc
@@ -40,7 +40,7 @@ For this reason, constructors should always initialize all data members of a cla
 While in some cases data members are initialized by their default constructor,
 in others, they are left with garbage.
 
-Trivial types with a ``++default++``ed or implicit default constructor follow the aggregate initialization syntax:
+Types with a ``++default++``ed or implicit trivial default constructor follow the aggregate initialization syntax:
 if you omit them in the initialization list, they will not be initialized.
 
 [source,cpp]
@@ -78,7 +78,7 @@ struct DefaultedContainer {
 };
 ----
 
-The same is true for an explicit ``++default++``ed default constructor.
+The same is true for an ``++default++``ed default constructor.
 [source,cpp]
 ----
 struct Defaulted {
--- a/rules/S2107/cfamily/rule.adoc
+++ b/rules/S2107/cfamily/rule.adoc
@@ -8,6 +8,88 @@ all member variables of a type with an implicit or explicit constructor will be 
 This is not the case for the others:
 if no initialization is explicitly written,
 the variable will be left uninitialized.
+Trivial types with a ``++default++``ed or implicit default constructor follow the aggregate initialization syntax:
+if you omit their initialization, they will not be initialized.
+
+[source,cpp]
+----
+struct Trivial {
+  int x;
+  int y;
+};
+
+struct Container {
+  Trivial t;
+  int arr[2];
+  Container() {
+    // Noncompliant
+    // this->t is not initialized
+    // this->t.x and this->t.y contain garbage
+    // this->arr contains garbage
+  }
+  Container(int) :t{}, arr{} {
+    // Compliant
+    // this->t.x and this->t.y are initialized to 0
+    // this->arr is initialized to {0, 0}
+  }
+  Container(int, int) :t{1}, arr{1} {
+    // Compliant
+    // this->t.x is 1
+    // this->t.y is 0
+    // this->arr is initialized to {1, 0}
+  }
+};
+
+struct DefaultedContainer {
+  Trivial t;
+  int arr[2];
+  DefaultedContainer() = default; // Noncompliant
+  // this->t and this->arr are not initialized
+};
+----
+
+The same is true for an explicit ``++default++``ed default constructor.
+[source,cpp]
+----
+struct Defaulted {
+  int x;
+  Defaulted() = default;
+};
+struct ContainerDefaulted {
+  Defaulted d;
+  ContainerDefaulted() {
+    // Noncompliant this->d.x is not initialized
+  }
+};
+----
+
+
+Even if some of the members have class initializers,
+the other members are still not initialized by default.
+
+[source,cpp]
+----
+struct Partial {
+  int x;
+  int y = 1;
+  int z;
+};
+struct ContainerPartial {
+  Partial p;
+  ContainrePartial() {
+    // Noncompliant
+    // this->p.x is not initialized
+    // this->p.y is initialized to 1
+    // this->p.z is not initialized
+  }
+  ContainrePartial(bool) :p{3} {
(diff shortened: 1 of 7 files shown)


rules/S2107/cfamily/rule.adoc
```diff
@@ -1,22 +1,58 @@
+Partially-initialized objects are surprising to the `class` users
+and might lead to hard-to-catch bugs.
+
 == Why is this an issue?
 
-In a {cpp} class, all member variables of a type with an implicit or explicit constructor will be automatically initialized. This is not the case for the others: if no initialization is explicitly written, the variable will be left uninitialized.
+In a {cpp} class,
+all member variables of a type with an implicit or explicit constructor will be automatically initialized.
```
tomasz-kaminski-sonarsource, line 7: I struggling to understand this sentence. I think we should focus on the default constructor in the first place.
However, if the class has a trivial default constructor, such a class will not be initialized. For example:
```
struct Trivial {
  int x;
  int y;
};

struct Container {
  Container() { /* no initialization of t */ }
  Trivial t;
};
````

File: rules/S2107/cfamily/rule.adoc [hunk 2]
```diff
@@ -1,22 +1,58 @@
...
 
+=== What is the potential impact?
 
-This comes with all the risks associated with uninitialized variables, and these risks propagate to all the classes using  the faulty class as a type. This is all the more surprising that most programmers expect a constructor to correctly initialize the members of its class (this is its raison d'être after all).
+This comes with all the risks associated with uninitialized variables,
+and these risks propagate to all the classes using the faulty class as a type
+as a base class or a data member.
+This is all the more surprising
+that most programmers expect a constructor to correctly initialize the members of its class
+(this is its raison d'être after all).
```
tomasz-kaminski-sonarsource, line 22: I would remove this reference to language I do not recognize. It may be only understandable to people exposed to french.

rules/S2107/cfamily/rule.adoc
```diff
@@ -1,22 +1,58 @@
...
+and might lead to hard-to-catch bugs.
+
 == Why is this an issue?
 
-In a {cpp} class, all member variables of a type with an implicit or explicit constructor will be automatically initialized. This is not the case for the others: if no initialization is explicitly written, the variable will be left uninitialized.
+In a {cpp} class,
+all member variables of a type with an implicit or explicit constructor will be automatically initialized.
+This is not the case for the others:
+if no initialization is explicitly written,
+the variable will be left uninitialized.
```
tomasz-kaminski-sonarsource, line 10: I think we should extend this section, to have code samples with classes and an explanation of which members are not initialized. This rule seems like the place where we should put this deduction.

rules/S2107/cfamily/rule.adoc
```diff
@@ -1,22 +1,58 @@
...
+== How to fix it
+
+To avoid partially-initialized objects,
+all non `+class+`-type fields should always be initialized (in order of preference):
 
 * With an in-class initializer
 * In the initialization list of a constructor
 * In the constructor body
 
 See S3230 for more details about this order.
```
tomasz-kaminski-sonarsource, line 41: One this rule is in related rules I am not sure if we should keep the reference.

rules/S2107/cfamily/rule.adoc
```diff
@@ -1,22 +1,58 @@
...
+This is all the more surprising
+that most programmers expect a constructor to correctly initialize the members of its class
+(this is its raison d'être after all).
 
+include::../../../shared_content/cfamily/garbage_value_impact.adoc[]
+
+=== Exceptions
 
-To avoid such situations, all non class type fields should always be initialized (in order of preference):
+Aggregate classes do not initialize most of their data members
```
tomasz-kaminski-sonarsource, line 28: I would mention in this section, that using `x{}` or `x()` (in class initializers) will value initialize all aggregate members.

rules/S2107/cfamily/rule.adoc
```diff
@@ -1,22 +1,58 @@
...
 * In the constructor body
 
 See S3230 for more details about this order.
 
-=== Noncompliant code example
+If none of the above places are a good fit for a data member,
+ask yourself whether it belongs to this class.
+Instead of a data member, it might be more appropriate to keep the value
+
+- in a delegate class,
```
tomasz-kaminski-sonarsource, line 47: What is a delegate class? This term does not link the bell for me.

tomasz-kaminski-sonarsource (changes requested):
I think this rule would be good place to explain some details on how defaulted trivial constructor works, and we should expand on it in Why this is an issue.

rules/S2107/cfamily/rule.adoc
necto: It is a relatively common [English idiom](https://en.wikipedia.org/wiki/Raison_d%27%C3%AAtre#:~:text=Raison%20d'%C3%AAtre%20%2F%CB%8Cr,or%20%22reason%20to%20be%22.), but it is certainly not part of a "simple English". removed

rules/S2107/cfamily/rule.adoc
necto: I am divided on this one. On one hand, it is duplication, to have the reference both here in the context, and in the Resources section. On another hand, here the reference is most relevant, so it helps both to put the reference in context and to explain the order mentioned here; and the reference is good to mention in the Resources section because that's the first place you would look at when searching for related rules.

It does not seem like too much duplication, though, so if you have no preference, I would leave both references in.

rules/S2107/cfamily/rule.adoc
necto: expanded

rules/S2107/cfamily/rule.adoc
necto: I added a few examples. Additionally, the Noncompliant code example section covers some. Do you think any cases are still missing?

necto:
@tomasz-kaminski-sonarsource 
I added some C++ reference links for the initialization subject. Do you know of a good article/blog post that would introduce the subject?

rules/S2107/cfamily/rule.adoc
```diff
@@ -4,10 +4,91 @@ and might lead to hard-to-catch bugs.
 == Why is this an issue?
 
 In a {cpp} class,
-all member variables of a type with an implicit or explicit constructor will be automatically initialized.
+all member variables of a type with an explicit constructor
+or of non-trivial type with a default constructor will be automatically initialized.
```
tomasz-kaminski-sonarsource, line 8: When I read this sentence, I get a lot of question about what is trying to convey, for example:
* What is the trivial type? There are trivial operations, like cosntruyctor destructors, not trivial type
* What are explicit constructors (does it mean `explicit` keyword). Maybe it is written by user?
```
struct A {
   A() = default;
   int x;
   int y;  
};
```

The examples that are below helps a lot, and I think that maybe just dropping the sentence may be good idea. And starting from the `Trivial types` section, but also would remove Trivial there.

As more concrete suggestions:
* Add a small sentence, before why this is an issue, including a description that when the class type has a constructor, it is expected that it will initialize all members. This rule raises the issue if some members are left initialized.
* Start why this is an issue, with the example of no constructor.

rules/S2107/cfamily/rule.adoc
```diff
@@ -44,7 +125,9 @@ If none of the above places are a good fit for a data member,
 ask yourself whether it belongs to this class.
 Instead of a data member, it might be more appropriate to keep the value
 
-- in a delegate class,
+- in a member of the class of a dats member
```
tomasz-kaminski-sonarsource, line 128: Would this rewrite say the same say thing? If so I will found it cleaner.
```suggestion
- in the constructor of member type, if it is a class
```

tomasz-kaminski-sonarsource (changes requested):
This looks a lot better now, however, I still get some confusion from the first sentence of the introduction section. Added some suggestion to reworking why this is an issue introduction.

rules/S2107/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I think that examples already cover most of the cases I can think of. Good job on that.

rules/S2107/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I see your point now. This makes sense. However, I think that just saying S3230 may be confusing for user, as it does not immediately scream this is rule id. I think saying "See rule S3230 for more details about this order" would be fine.

necto: I added "rule"

rules/S2107/cfamily/rule.adoc
necto: Rewritten as discussed over zoom

rules/S2107/cfamily/rule.adoc
necto: "keep value in the constructor of member type, if it is a class" does not make sense to me, as you can't "keep value in constructor"
But you can put the value into one of the fields of the class of a data member, that's what I meant.
I've rewritten it, see if that is understandable now

rules/S2107/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Looks a lot better now. Thanks!

rules/S2107/cfamily/rule.adoc
```diff
@@ -1,22 +1,172 @@
...
+  int y;
+  PartInit(int n) :y(n) {
+    // this->ai is initialized
+    // this->y is initialized
+    // Noncompliant: this->x is left uninitialized
+  }
+};
+----
 
+This leads to garbage being printed in benign-looking code like this:
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Maybe we should say, this leads to undefined behavior in following bening-looking code. In C++ this is UB.
And comment on the court. // Undefined behavior,

tomasz-kaminski-sonarsource (approved):
Some very minor suggestions, but otherwise look good.

rules/S2107/cfamily/rule.adoc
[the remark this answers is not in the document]
necto: rewritten

tomasz-kaminski-sonarsource:
For the blog, I couldn't think any out of top of my head.

PR #2652, opened 2023-07-26
Modify rule S2251: Adapt to LaYC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2251/javascript/rule.adoc
+++ b/rules/S2251/javascript/rule.adoc
@@ -1,25 +1,24 @@
 == Why is this an issue?
 
-A `for` loop with a stop condition that can never be reached, such as one with a counter that moves in the wrong direction, will run infinitely. While there are occasions when an infinite loop is intended, the convention is to construct such loops as `while` loops. More typically, an infinite `for` loop is a bug. 
+In a `for` loop, the update clause is responsible for modifying the loop counter variable in the appropriate direction to control the loop's iteration. It determines how the loop counter variable changes with each iteration of the loop. The loop counter should move in the right direction to prevent several issues:
 
-== How to fix it
+* If the loop counter does not change, and the condition never becomes false, the loop will continue to run indefinitely, causing an infinite loop. This can crash the program or freeze the browser, making it unresponsive.
+* If the loop counter is updated in the wrong direction, the loop may not iterate the intended number of times, skip some iterations altogether, or provoke integer overflow. This can lead to unexpected and incorrect behavior in the code.
 
-=== Code examples
-
-==== Noncompliant code example
+This rule raises an issue when the loop counter is updated in the wrong direction with respect to the loop termination condition.
 
 [source,javascript,diff-id=1,diff-type=noncompliant]
 ----
-for (var i = 0; i < strings.length; i--) { // Noncompliant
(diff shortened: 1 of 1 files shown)


rules/S2251/javascript/rule.adoc
```diff
@@ -1,25 +1,24 @@
 == Why is this an issue?
 
-A `for` loop with a stop condition that can never be reached, such as one with a counter that moves in the wrong direction, will run infinitely. While there are occasions when an infinite loop is intended, the convention is to construct such loops as `while` loops. More typically, an infinite `for` loop is a bug. 
+In a `for` loop, the update clause is responsible for modifying the loop counter variable in the appropriate direction to control the loop's iteration. It determines how the loop counter variable changes with each iteration of the loop. The loop counter should move in the right direction to prevent several issues:
 
-== How to fix it
+* If the loop counter does not change, and the condition never becomes false, the loop will continue to run indefinitely, causing an infinite loop. This can crash the program or freeze the browser, making it unresponsive.
+* If the loop counter is updated in the wrong direction, the loop may not iterate the intended number of times, skip some iterations altogether, or provoke integer overflow. This can lead to unexpected and incorrect behavior in the code.
```
victor-diez-sonarsource, line 6:
```suggestion
* If the loop counter is updated in the wrong direction, the loop may not iterate the intended number of times, skip some iterations altogether, or provoke integer overflow. This can lead to unexpected and incorrect behavior in the code.
```

Not sure about "skip some iterations altogether, or provoke integer overflow". counter will search `Infinite` and will become also an infinite loop. I'm not sure we can speak of overflows in javascript in the traditional way.

victor-diez-sonarsource (approved):

rules/S2251/javascript/rule.adoc
yassin-kammoun-sonarsource: #### Skip some iterations altogether

You can imagine a loop in which the body executes conditionally some code depending on the value of the counter. An iteration may wrongly be skipped partially or totally. For example:

```js
for (let i = 0; i < 10; i--) {
  if (i >= 0 && i < 5) {
    /* ... */
  }
}
```

In this example, the body of the if-statement will be skipped after the first iteration since the counter will become negative. Maybe I should reformulate "skip some iterations altogether" to "skip some parts of the loop body".

WDYT?

#### Provoke integer overflow

For me, an integer overflow is a broad term that means that we try to store a value that is no longer representable in a data type. Whether the value becomes `Infinite`, goes from positive to negative, or triggers an exception is a consequence of an overflow that is language-specific.

victor-diez-sonarsource: agreed on interget overflow as a broader term. as for skip some iterations, this is specific to that implementation with an if statement, but does not come from the for loop itself. I would just say unexpected behaviour (as it's not what the user expects). I think it's a broader term and not dependant on what's inside the block. the for loop itself will not skip any iterations

yassin-kammoun-sonarsource: 👍 I will remove the bullet points and limit myself to emphasize infinite loops and unexpected behavior. Let's avoid too many details specific to an implementation as you said.

PR #2658, opened 2023-07-26
Modify S125: Set quickfix field for C#

gregory-paidis-sonarsource:
Fix I forgot to do when merging [the PR](https://github.com/SonarSource/sonar-dotnet/pull/6966) that implemented the codefix.


antonioaversa (approved):
LGTM

PR #2659, opened 2023-07-26
Modify rule S5717: LaYC format

guillaume-dequenne-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5717/python/rule.adoc
+++ b/rules/S5717/python/rule.adoc
@@ -12,13 +12,12 @@ def myfunction(param=foo()):  # foo is called only once, when the function is de
     ...
 ----
 
-For the same reason, it is also a bad idea to store mutable default value in another object (ex: as an attribute). Multiple instances will then share the same value and modifying one objet will modify all of them.
+For the same reason, it is also a bad idea to store mutable default values in another object (ex: as an attribute). Multiple instances will then share the same value and modifying one object will modify all of them.
 
 
 This rule raises an issue when:
 
-* a default value is either modified in the function or assigned to anything else than a variable and it has one of the following types:
-** builtins: set, dict, list.
+* a default value is either modified in the function or assigned to anything other than a variable and it has one of the following types:
 ** https://docs.python.org/3/library/collections.html[collections] module: deque, UserList, ChainMap, Counter, OrderedDict, defaultdict, UserDict.
 * an attribute of a default value is assigned.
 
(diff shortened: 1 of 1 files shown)


rules/S5717/python/rule.adoc
```diff
@@ -1,27 +1,49 @@
...
+
+== How to fix it
+
+When a parameter default value is meant to be a mutable object, it is best to keep the parameter optional and instantiate the mutable object in the function's body directly.
+
+=== Code examples
+
+==== Noncompliant code example
 
 In the following example, the parameter "param" has ``++list()++`` as a default value. This list is created only once and then reused in every call. Thus when it appends ``++'a'++`` to this list, the next call will have ``++['a']++`` as a default value.
```
joke1196, line 41: I find the last sentence is bit confusing. `Thus when __it__ appends` feels a bit strange, as I was notit referred to the `This list` of the previous sentence instead of the `call`. Maybe we could go with something like: Thus when appending `a` in the body of the function, its next call will have `[a]` as a default value. Does it make sense?

rules/S5717/python/rule.adoc
```diff
@@ -1,27 +1,49 @@
...
+These default values are expressions which are evalutated when the function is defined, i.e. only once. The same default value will be used every time the function is called. Therefore, modifying it will have an effect on every subsequent call. This can lead to confusing bugs.
 
+[source,python]
+----
+def myfunction(param=foo()):  # foo is called only once, when the function is defined.
+    ...
+----
 
-It is also a bad idea to store mutable default value in another object (ex: as an attribute). Multiple instances will then share the same value and modifying one objet will modify all of them.
+For the same reason, it is also a bad idea to store mutable default value in another object (ex: as an attribute). Multiple instances will then share the same value and modifying one objet will modify all of them.
```
joke1196, line 15: Super optional, maybe we could add a small example here to better illustrate the attribute part?

joke1196 (approved):
LGTM. I have just added some small comments, but they could be ignored, I'll let you decide.

maksim-grebeniuk-sonarsource (approved):
LGTM

rules/S5717/python/rule.adoc
jean-jimbo-sonarsource, line 15: A few grammar/spelling issues here. Please change to "For the same reason, it is also a bad idea to store mutable default values in another object (ex: as an attribute). Multiple instances will then share the same value and modifying one object will modify all of them.



"

rules/S5717/python/rule.adoc
```diff
@@ -1,27 +1,49 @@
...
 
 
 This rule raises an issue when:
 
 * a default value is either modified in the function or assigned to anything else than a variable and it has one of the following types:
```
jean-jimbo-sonarsource, line 20: This can be improved with:
"a default value is either modified in the function or assigned to anything other than a variable and it has one of the following types:"

jean-jimbo-sonarsource (changes requested):
Left some comments on grammar/spelling

rules/S5717/python/rule.adoc
guillaume-dequenne-sonarsource: I must admit I was hesitant to even keep that part, so if you're okay, I'll leave it as it is.

jean-jimbo-sonarsource (approved):
LGTM

PR #2660, opened 2023-07-26
Modiffy rule S2190: Fix link formats

martin-strecker-sonarsource:


cristian-ambrosini-sonarsource (approved):
LGTM

rules/S2190/csharp/rule.adoc
```diff
@@ -95,7 +98,7 @@ public int Sum()
 }
 ----
 
-* As https://rules.sonarsource.com/csharp/RSPEC-907[S907] generally suggests, avoid using `goto` statements. Instead, you can use a loop statement or explicit recursion.
+* As S907 generally suggests, avoid using `goto` statements. Instead, you can use a loop statement or explicit recursion.
```
duncanp-sonar, line 101: This could (should?) be a rule cross reference: `{rule:csharpsquid:S907}`.

SQ, SC and SL will turn it into the type of link appropriate to the particular UI (which might or might not be a link to `rules.sonarsource.com`).

PR #2665, opened 2023-07-26
Modify rule S5706: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5706/python/rule.adoc
+++ b/rules/S5706/python/rule.adoc
@@ -35,6 +35,8 @@ The special method {link-with-uscores1}[``++__exit__++``] should only raise an e
 
 == How to fix it
 
+To fix this issue, make sure to avoid raising the exception provided to an ``++__exit__++`` method.
+
 === Code examples
 
 ==== Noncompliant code example
--- a/rules/S5706/python/rule.adoc
+++ b/rules/S5706/python/rule.adoc
@@ -13,45 +13,61 @@ Raising this exception will make the stack trace difficult to understand.
 
 The ``++__exit__++`` method can filter passed-in exceptions by simply returning True or False.
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 [source,python,diff-id=1,diff-type=noncompliant]
 ----
 class MyContextManager:
     def __enter__(self):
         return self
+
     def __exit__(self, *args):
         raise  # Noncompliant
+
+class MyContextManager:
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
         raise args[2]  # Noncompliant
 
 class MyContextManager:
     def __enter__(self):
         return self
+
     def __exit__(self, exc_type, exc_value, traceback):
         raise exc_value # Noncompliant
 ----
 
 
-=== Compliant solution
+==== Compliant solution
 
 [source,python,diff-id=1,diff-type=compliant]
 ----
 class MyContextManager:
+    def __enter__(self, stop_exceptions):
+        return self
+
+    def __exit__(self, *args):
+        try:
+            ...
+        except:
+            raise  # No issue when raising another exception. The __exit__ method can fail and raise an exception
+
+class MyContextManager:
     def __enter__(self):
         return self
+
     def __exit__(self, exc_type, exc_value, traceback):
-        # by default the function will return None, which is always False, and the exc_value will naturally raise.
-        pass
+        pass # by default the function will return None, which is always False, and the exc_value will naturally raise.
 
 class MyContextManager:
     def __enter__(self, stop_exceptions):
         return self
+
     def __exit__(self, *args):
-        try:
-            print("42")
-        except:
-            print("exception")
-            raise  # No issue when raising another exception. The __exit__ method can fail and raise an exception
         raise MemoryError("No more memory")  # This is ok too.
 ----
 
--- a/rules/S5706/python/rule.adoc
+++ b/rules/S5706/python/rule.adoc
@@ -7,33 +7,33 @@ This rule raises an issue when:
 
 :link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special#object.__exit__
 
-Methods ``++__enter__++`` and ``++__exit__++`` allows you to implement objects which can be used as the expression of a ``++with++`` statement.
+Methods ``++__enter__++`` and ``++__exit__++`` make it possible to implement objects which can be used as the expression of a ``++with++`` statement:
 
 [source,python]
 ----
 with MyContextManager() as c :
-   #do something with c
+    ... # do something with c
 ----
 
-This statement can be rewritten as a ``++try...finally++`` and an explicit call to the ``++__enter__++`` and ``++__exit__++`` methods.
+This statement can be rewritten as a ``++try...finally++`` and an explicit call to the ``++__enter__++`` and ``++__exit__++`` methods:
 
 [source,python]
 ----
 c = MyContextManager()
 c.__enter__()
 try:
-#do something with c
+    ... # do something with c
 finally:
-c.__exit__()
+    c.__exit__()
 ----
 
-The ``++__exit__++`` is the method of a statement context manager which is called on exit the runtime context related to this object.
+The ``++__exit__++`` is the method of a statement context manager which is called when exiting the runtime context related to this object.
 
-If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value. Otherwise, the exception will be processed normally upon exit from this method.
+If an exception is supplied as an argument, its propagation can be suppressed by having the method return a truthy value. Otherwise, the exception will be processed normally upon exit from the method.
 
-The special method {link-with-uscores1}[``++__exit__++``] should only raise an exception when it fails. It should never raise the provided exception, it is the caller's responsibility. The ``++__exit__++`` method can filter passed-in exceptions by simply returning True or False. Raising this exception will make the stack trace difficult to understand.
+The special method {link-with-uscores1}[``++__exit__++``] should only raise an exception when it fails. It should never raise the provided exception, it is the caller's responsibility. The ``++__exit__++`` method can filter provided exceptions by simply returning True or False. Raising this exception will make the stack trace difficult to understand.
 
-
+== How to fix it
 
 === Code examples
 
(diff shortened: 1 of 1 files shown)


rules/S5706/python/rule.adoc
```diff
@@ -6,55 +11,63 @@ The special method {link-with-uscores1}[``++__exit__++``] should only raise an e
 
 Raising this exception will make the stack trace difficult to understand.
 
-
 The ``++__exit__++`` method can filter passed-in exceptions by simply returning True or False.
```
joke1196, line 14: Maybe we could be a bit more context on what is the issue. Like maybe a bit of detail about the `__exit__` method, just so that we can understand better what is meant with the `filter passed-in exception`.  WDYT?

rules/S5706/python/rule.adoc
```diff
@@ -6,55 +11,63 @@ The special method {link-with-uscores1}[``++__exit__++``] should only raise an e
...
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
         raise args[2]  # Noncompliant
 
 class MyContextManager:
     def __enter__(self):
         return self
+
```
joke1196, line 39: Maybe we could add a bit of detail on why each case is Noncompliant.

joke1196 (changes requested):
Looks good. Added small comments, for the examples and maybe a bit more info in the description.

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,77 @@
+This rule raises an issue when:
+
+* an ``++__exit__++`` method has a bare ``++raise++`` outside of an ``++except++`` block.
+* an ``++__exit__++`` method raises the exception provided as parameter.
+
 == Why is this an issue?
 
 :link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special#object.__exit__
 
+The ``++__exit__++`` is the method of a statement context manager which is called on exit the runtime context related to this object.
```
joke1196, line 10: I am not sure I understand properly this part, I think a few words are missing. At the moment it is a bit abstract maybe we could make it more concrete. I would suggest something like: The magic methods `__enter__` and `__exit__` allows you to implement objects which can be used as the expression of a `with` statement. 
```
with MyContextManager() as c :
   #do something with c
```
This statement can be rewritten as a `try...finally` and an explicit call to the `__enter__` and `__exit__` methods.

```
c = MyContextManager()
c.__enter__()
try:
   #do something with c
finally:
  c.__exit__()
```
Then maybe we could say a few words about the role of the exit method with the happy path. We can then introduce the exception path this way will tie nicely with the next paragraph.

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,77 @@
...
+
+If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value. Otherwise, the exception will be processed normally upon exit from this method.
+
 The special method {link-with-uscores1}[``++__exit__++``] should only raise an exception when it fails. It should never raise the provided exception, it is the caller's responsibility.
 
 Raising this exception will make the stack trace difficult to understand.
 
-
 The ``++__exit__++`` method can filter passed-in exceptions by simply returning True or False.
```
joke1196, line 18: I guess the paragraph on line 18 could be merged with the paragraph starting at line 12, as they say basically the same thing.

joke1196 (changes requested):
Looks good, I like the details on the examples. I left a suggestion for the description in order to make it a bit more concrete, with some examples. The goal would be to show why/when you would use the `__exit__` method, so it makes it easier to show why you should not re-raise the exception

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+
+* an ``++__exit__++`` method has a bare ``++raise++`` outside of an ``++except++`` block.
+* an ``++__exit__++`` method raises the exception provided as parameter.
+
 == Why is this an issue?
 
 :link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special#object.__exit__
 
-The special method {link-with-uscores1}[``++__exit__++``] should only raise an exception when it fails. It should never raise the provided exception, it is the caller's responsibility.
+Methods ``++__enter__++`` and ``++__exit__++`` allows you to implement objects which can be used as the expression of a ``++with++`` statement.
```
guillaume-dequenne-sonarsource, line 10: `allow` since it's plural.
I'd suggest "make it possible to implement" as a more neutral alternative.
I'd end with ":" as the snippet that follows directly completes this.

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+        raise args[2]  # Noncompliant: __exit__() methods should not reraise the passed-in exception; this is the caller’s responsibility.
+
+class MyContextManager:
+    def __enter__(self):
+        return self
+
     def __exit__(self, exc_type, exc_value, traceback):
-        raise exc_value # Noncompliant
+        raise exc_value # Noncompliant: __exit__() methods should not reraise the passed-in exception; this is the caller’s responsibility.
 ----
```
guillaume-dequenne-sonarsource, line 66: In this case I think the resulting tab is quite long and we should probably have a dedicated "How to fix it" section.

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+----
+with MyContextManager() as c :
+   #do something with c
+----
 
+This statement can be rewritten as a ``++try...finally++`` and an explicit call to the ``++__enter__++`` and ``++__exit__++`` methods.
 
-The ``++__exit__++`` method can filter passed-in exceptions by simply returning True or False.
+[source,python]
+----
```
guillaume-dequenne-sonarsource, line 21: The indentation on this snippet is missing.

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
 :link-with-uscores1: https://docs.python.org/3/reference/datamodel.html?highlight=__exit__%20special#object.__exit__
 
-The special method {link-with-uscores1}[``++__exit__++``] should only raise an exception when it fails. It should never raise the provided exception, it is the caller's responsibility.
+Methods ``++__enter__++`` and ``++__exit__++`` allows you to implement objects which can be used as the expression of a ``++with++`` statement.
 
-Raising this exception will make the stack trace difficult to understand.
+[source,python]
+----
+with MyContextManager() as c :
+   #do something with c
```
guillaume-dequenne-sonarsource, line 15: You probably can use ellipsis to have syntactically correct code:
```
with MyContextManager() as c :
   ... # do something with c
```

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+----
 
+This statement can be rewritten as a ``++try...finally++`` and an explicit call to the ``++__enter__++`` and ``++__exit__++`` methods.
```
guillaume-dequenne-sonarsource, line 18: If you change it on the previous line, I'd also end this one with ":"

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+    def __exit__(self, *args):
+        raise args[2]  # Noncompliant: __exit__() methods should not reraise the passed-in exception; this is the caller’s responsibility.
+
+class MyContextManager:
+    def __enter__(self):
+        return self
+
     def __exit__(self, exc_type, exc_value, traceback):
-        raise exc_value # Noncompliant
+        raise exc_value # Noncompliant: __exit__() methods should not reraise the passed-in exception; this is the caller’s responsibility.
```
guillaume-dequenne-sonarsource, line 63: I'd suggest rewording "passed-in" to "provided".
(note that this is in multiple places in the examples but I'll just leave this one comment to not bloat things).

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+----
+c = MyContextManager()
+c.__enter__()
+try:
+#do something with c
+finally:
+c.__exit__()
+----
 
+The ``++__exit__++`` is the method of a statement context manager which is called on exit the runtime context related to this object.
```
guillaume-dequenne-sonarsource, line 30: Grammar fix:
...which is called when exiting the runtime context

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
 
-This rule raises an issue when:
+If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value. Otherwise, the exception will be processed normally upon exit from this method.
```
guillaume-dequenne-sonarsource, line 32: I feel the wording is a bit off ("the method wishes" sounds like the method is a living thing).
What about:
"If an exception is supplied as an argument, its propagation can be suppressed by having the method return a truthy value. Otherwise, the exception will be processed normally upon exit from the method."

guillaume-dequenne-sonarsource (changes requested):
I left a few grammar/styling comments.
I'm also under the impression that we could have a dedicated "How to fix it" section here, given that the code snippet is a bit long.
Wdyt @jean-jimbo-sonarsource ?

rules/S5706/python/rule.adoc
jean-jimbo-sonarsource: Sie note: I'm not sure we've been consistent with the use of colons before we show code snippets. I think we should always have them.

jean-jimbo-sonarsource (approved):
LGTM

rules/S5706/python/rule.adoc
```diff
@@ -1,60 +1,95 @@
...
+The ``++__exit__++`` is the method of a statement context manager which is called when exiting the runtime context related to this object.
 
+If an exception is supplied as an argument, its propagation can be suppressed by having the method return a truthy value. Otherwise, the exception will be processed normally upon exit from the method.
 
-This rule raises an issue when:
+The special method {link-with-uscores1}[``++__exit__++``] should only raise an exception when it fails. It should never raise the provided exception, it is the caller's responsibility. The ``++__exit__++`` method can filter provided exceptions by simply returning True or False. Raising this exception will make the stack trace difficult to understand.
 
-* an ``++__exit__++`` method has a bare ``++raise++`` outside of an ``++except++`` block.
-* an ``++__exit__++`` method raises the exception provided as parameter.
+== How to fix it
```
guillaume-dequenne-sonarsource, line 36: Here we should have a small explanation of what is needed to fix the issue, before the code snippets.
Something like the following should be enough:
To fix this issue, make sure to avoid raising the exception provided to an ``++__exit__++`` method.

guillaume-dequenne-sonarsource (approved):
Looks good, left a last comment.

PR #2667, opened 2023-07-26
Modify rule S1542: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1542/python/rule.adoc
+++ b/rules/S1542/python/rule.adoc
@@ -1,4 +1,4 @@
-This rule raises an issue when function name is out of python naming convention (snake case).
+This rule raises an issue when function name doesn't respect the naming convention.
 
 == Why is this an issue?
 
@@ -7,9 +7,11 @@ include::../description.adoc[]
 According to PEP8 function names should be in all lowercase letters, multiple words should be separated by underscores. This convention is known as "snake_case." For example: ``++calculate_area++``, ``++print_hello++``, ``++process_data++``
 
 === How to fix it
-Use the PEP8 standard of the naming strategy for Python.
+Respect the naming convention for the function (by default it is the snake case as recommended by PEP8).
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 With the default provided regular expression: ``++^[a-z_][a-z0-9_]*$++``
 
@@ -19,13 +21,18 @@ def MyFunction(a,b):
     ...
 ----
 
-=== Compliant solution
+==== Compliant solution
 
 [source,python,diff-id=1,diff-type=compliant]
 ----
 def my_function(a,b):
     ...
 ----
+
+== Resources
+
+=== Documentation
+* PEP8 - https://peps.python.org/pep-0008/#function-and-variable-names[Function and Variable Names]
 
 ifdef::env-github,rspecator-view[]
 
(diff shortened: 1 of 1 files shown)


rules/S1542/python/rule.adoc
```diff
@@ -1,25 +1,39 @@
+This rule raises an issue when function name doesn't respect the naming convention.
```
guillaume-dequenne-sonarsource, line 1: when function name => when a function name

rules/S1542/python/rule.adoc
```diff
@@ -1,25 +1,39 @@
...
+
 == Why is this an issue?
 
 include::../description.adoc[]
 
-=== Noncompliant code example
+According to PEP8 function names should be in all lowercase letters, multiple words should be separated by underscores. This convention is known as "snake_case." For example: ``++calculate_area++``, ``++print_hello++``, ``++process_data++``
```
guillaume-dequenne-sonarsource, line 7: Slight rewording suggestion:
According to PEP8, function names should be lowercase, with words separated by underscores as necessary to improve readability. This convention...

rules/S1542/python/rule.adoc
```diff
@@ -1,25 +1,39 @@
...
+
+=== How to fix it
```
guillaume-dequenne-sonarsource, line 9: How to fix it should be a 2nd level title `==`. That said, I think we shouldn't have it at all here, as the rule description is so small (let's keep everything in one single tab).

rules/S1542/python/rule.adoc
```diff
@@ -1,25 +1,39 @@
...
+Respect the naming convention for the function (by default it is the snake case as recommended by PEP8).
```
guillaume-dequenne-sonarsource, line 10: If we remove the header, then we can rephrase this:
To fix this, respect the naming convention for the function. The default naming convention is snake case, as recommended by PEP8. Other naming conventions can be defined through the `format` rule parameter.

guillaume-dequenne-sonarsource (changes requested):
Some minor suggestions.

guillaume-dequenne-sonarsource (approved):
LGTM.

rules/S1542/python/rule.adoc
```diff
@@ -1,25 +1,38 @@
+This rule raises an issue when a function name doesn't respect the naming convention.
+
 == Why is this an issue?
 
 include::../description.adoc[]
```
jean-jimbo-sonarsource, line 5: what's going on here?

jean-jimbo-sonarsource (changes requested):
please see my note

rules/S1542/python/rule.adoc
guillaume-dequenne-sonarsource: This is supposed to include a common description for multiple languages. Is it not working?
To test it, you can clone the project and open a preview of the page in VSCode or IntelliJ, it should resolve these include statements.
Alternatively, to test it in the SQ UI, you can run `rule-api` to generate the actual html in the `sonar-python` project, then build the analyzer, include the jar into a local SonarQube and run it. It's a bit more involved to do it the first time, but makes reviewing a bit easier afterwards, as you can really make sure everything displays properly.

maksim-grebeniuk-sonarsource: You can also use the IntelliJ [AsciiDoc](https://plugins.jetbrains.com/plugin/7391-asciidoc) plugin as well.

jean-jimbo-sonarsource: Ahh thanks, I got it working now

PR #2670, opened 2023-07-26
Modify rule S2836: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2836/python/rule.adoc
+++ b/rules/S2836/python/rule.adoc
@@ -1,5 +1,4 @@
-This rule raises an issue when the loop with ``++else++`` clause doesn't have the ``++break++`` statement in the loop body.
-
+This rule raises an issue when a loop with an ``++else++`` clause doesn't contain any ``++break++`` statement in its body.
 == Why is this an issue?
 
 The ``++else++`` clause of a loop is skipped when a ``++break++`` is executed in this loop. In other words, a loop with an ``++else++`` but no ``++break++`` statement will always execute the ``++else++`` part (unless of course an exception is raised or ``++return++`` is used). If this is what the developer intended, it would be much simpler to have the ``++else++`` statement removed and its body unindented. Thus having a loop with an ``++else++`` and no ``++break++`` is most likely an error.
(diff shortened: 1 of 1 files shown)


rules/S2836/python/rule.adoc
```diff
@@ -1,15 +1,29 @@
+This rule raises an issue when the loop with ``++else++`` clause doesn't have the ``++break++`` statement in the loop body.
```
guillaume-dequenne-sonarsource, line 1: I think the following would read a bit nicer:
This rule raises an issue when a loop with an `else` clause doesn't contain any `break` statement in its body.

rules/S2836/python/rule.adoc
```diff
@@ -48,7 +56,9 @@ def search_first_number_without_else(elements: List[str]):
 
 == Resources
 
-* https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops[Python documentation - break and continue Statements, and else Clauses on Loops]
+=== Documentation
+
+* Python documentation - https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops[ Break and continue Statements, and else Clauses on Loops]
```
guillaume-dequenne-sonarsource, line 61: Is the extra whitespace intended? (in `[ Break`)

rules/S2836/python/rule.adoc
```diff
@@ -1,15 +1,29 @@
...
+[source,python,diff-id=1,diff-type=noncompliant]
 ----
 from typing import List
 
-def search_first_number_without_break(elements: List[str]):
+def foo(elements: List[str]):
+    for elt in elements:
+        if elt.isnumeric():
+            return elt
+    else:  # Noncompliant. This will be executed every time
```
guillaume-dequenne-sonarsource, line 23: If we want to be pedantic, if won't be executed if the `return` statement is triggered.
What about
`# Noncompliant: no break in the loop`
I still think that expressing that the else will always be executed (and could just be unintended) is better, but I can't really formulate it in a clean way.

guillaume-dequenne-sonarsource (changes requested):
Looks good, just some minor suggestions.

guillaume-dequenne-sonarsource (approved):
LGTM.
There is an issue with asciidoc validation and the `== Why is this an issue?` section. I think you need an extra whitespace after top level description?

PR #2683, opened 2023-07-27
Remove unused asciidoctor script

marco-antognini-sonarsource:
The asciidoctor binary we use is installed in the Docker image.
The JS frontend gets it from NPM.


necto (approved):
Well spotted!

PR #2684, opened 2023-07-27
Clean ci/validate_asciidoc.sh

marco-antognini-sonarsource:
This PR is a prerequisite for:
 * #2685

---

 * Regroup readonly variable declaration. `declare -xr` is used for exported readonly variable.
 * Use more accurate log messages.
 * Use consistent indentation of 2 spaces.
 * Fail fast if dependencies cannot be installed.
 * Regroup HTML generation & its validation.
 * Document the main validation points.
 * Simplify some commands and address some ShellCheck warnings.

### DIFF
--- a/ci/validate_asciidoc.sh
+++ b/ci/validate_asciidoc.sh
@@ -49,6 +49,7 @@ fi
 #  * All asciidoc files are used/included.
 
 echo "Testing the following rules: ${affected_rules}"
+supportedLanguages=$(sed 's/ or//' supported_languages.adoc | tr -d '`,')
 for dir in $affected_rules
 do
   if [ ! -d "$dir" ]; then
@@ -83,7 +84,6 @@ do
     fi
     rm -f stuck
 
-    supportedLanguages=$(sed 's/ or//' supported_languages.adoc | tr -d '`,')
     for language in "${dir}"/*/
     do
       language=${language%*/}
--- a/ci/validate_asciidoc.sh
+++ b/ci/validate_asciidoc.sh
@@ -1,11 +1,26 @@
 #!/bin/bash
 set -uo pipefail
 
-readonly ALLOWED_RULE_SUB_FOLDERS=['common'];
-
 # Install script dependencies
+(set -e; cd rspec-tools && pipenv install)
+
+# This script runs all tests; it doesn't exit at the first failure.
+exit_code=0
+
+readonly ALLOWED_RULE_SUB_FOLDERS=['common'];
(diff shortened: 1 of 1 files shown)


ci/validate_asciidoc.sh
```diff
@@ -1,11 +1,26 @@
 #!/bin/bash
 set -uo pipefail
 
+# Install script dependencies
+(set -e; cd rspec-tools && pipenv install)
```
necto, line 5: IIUC, `set -e` will only terminate the subshell on an error. If you are going for the early failure of the entire script, you'd have to do something like
```suggestion
set -e
cd rspec-tools && pipenv install
set +e
```

ci/validate_asciidoc.sh
```diff
@@ -1,11 +1,26 @@
...
+
+# This script runs all tests; it doesn't exit at the first failure.
+exit_code=0
+
 readonly ALLOWED_RULE_SUB_FOLDERS=['common'];
+readonly ROOT=$PWD
+declare -xr PATH_WITH_VARIABLE="$(realpath ./ci/replace_variables_in_path.sh)"
```
necto, line 12: As I had to google and read up on `declare -xr` and I expect many people who have to touch this file are also not fluent in Bash enough to understand it, I would add a comment explaining what this line does (for my selfish convenience).

```suggestion
# Make constant PATH_WITH_VARIABLE available to child processes
declare -xr PATH_WITH_VARIABLE="$(realpath ./ci/replace_variables_in_path.sh)"
```

necto (approved):
Thanks for overhauling the script!

ci/validate_asciidoc.sh
marco-antognini-sonarsource: Indeed, I made a mistake while testing! Thanks for catching this.

ci/validate_asciidoc.sh
```diff
@@ -47,37 +54,34 @@ do
...
-        exit_code=1
+      echo "ERROR: These adoc files contain an include that is stuck to other content."
+      echo "This may result in broken tags and other display issues."
+      echo "Make sure there is an empty line before and after each include:"
+      cat stuck
+      exit_code=1
     fi
     rm -f stuck
 
     supportedLanguages=$(sed 's/ or//' supported_languages.adoc | tr -d '`,')
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: Done in d27ef335f9f70685e920f7218f9769dd4075ed3e

PR #2685, opened 2023-07-27
Run include and cross-references validation on all files

marco-antognini-sonarsource:
The previous validation code was taking about 3 minutes when running on
all rules, which is the case when modifying anything outside rules/,
including shared_content/.

The runtime was marginal when only the content specific to one rule was
modified, e.g. only rules/S100/cfamily/rule.adoc is modified.

The new validation code always takes about 30 seconds on the CI. This is
a slight increase for small PR but a large reduction for PR affecting
the whole system.

Furthermore, the new approach validates includes and cross-references
properly. An issue is now raised when:
 * The last include to shared_content/lang/foo.adoc is removed.
 * A shared_content doc includes a file from another rule, e.g.
   rules/S100/lang/rule.adoc includes shared_content/lang/foo.adoc,
   which includes rules/S200/bar.adoc.

This is achieved with proper Asciidoc extensions written in Ruby.
See https://docs.asciidoctor.org/asciidoctor/latest/extensions/

---

Requires:
 - [x] #2726

---

Later, we can add more validation, like diff-id/type validation for [RULEAPI-788](https://sonarsource.atlassian.net/browse/RULEAPI-788):

```diff
diff --git a/ci/custom-asciidoctor b/ci/custom-asciidoctor
index 2c779c797a..48e0af9b44 100755
--- a/ci/custom-asciidoctor
+++ b/ci/custom-asciidoctor
@@ -72,9 +72,30 @@ class IncludeLogger < Asciidoctor::Extensions::IncludeProcessor
   # Intentionnaly no process function here.
 end
 
+class SourceLogger < Asciidoctor::Extensions::TreeProcessor
+  include Asciidoctor::Logging
+
+  def process document
+    blocks = document.find_by(context: :listing, style: 'source')
+
+    # Find blocks with only diff-id but no diff-type, or vice-versa, or invalid diff-type.
+    blocks.each do |src|
+      if src.attr?('diff-id') ^ src.attr?('diff-type')
+        puts "ERROR #{document.base_dir}"
+      end
+      if (type = src.attr('diff-type')) && !['compliant', 'noncompliant'].include?(type)
+        puts "ERROR diff-type is not valid: #{type}"
+      end
+    end
+
+    # TODO diff-id should have exactly 1 compliant and 1 noncompliant block. Find blocks that break this rule.
+  end
+end
+
 Asciidoctor::Extensions.register do
   preprocessor MainFileLogger
   include_processor IncludeLogger.new @document
+  treeprocessor SourceLogger
 end
 
 invoker = Asciidoctor::Cli::Invoker.new ARGV
```

[RULEAPI-788]: https://sonarsource.atlassian.net/browse/RULEAPI-788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ci/validate_asciidoc.sh
```diff
@@ -38,12 +35,16 @@ if printf '%s\n' "$changeset" | grep -qv '/S[0-9]\+/'; then
 fi
 
 # Validate some properties of the asciidoc:
+#
+# [properties validated only on affected rules]
 #  * Rules should have at least one language specification,
 #    unless they are closed or deprecated.
 #  * The include:: should have an empty line before and after them.
```
necto, line 42: I wonder if this one would be easier to validate from the Ruby extension

ci/validate_asciidoc.sh
```diff
@@ -38,12 +35,16 @@ if printf '%s\n' "$changeset" | grep -qv '/S[0-9]\+/'; then
...
 #  * Only valid languages can be used as subdirectories in rule directories,
 #    with the exception of ALLOWED_RULE_SUB_FOLDERS.
 #  * Asciidoc files are free or errors and warnings.
```
necto, line 45: This looks like like a property that could be also validated when you run asciidoctor with a custom plugin. Any reason, you keep a separate asciidoctor run for this?

ci/validate_asciidoc.sh
```diff
@@ -156,6 +133,48 @@ else
...
+
+git ls-files --cached -- 'rules/**.adoc' 'shared_content/**.adoc' \
+  | xargs realpath \
+  > all_asciidoc_files
+
+cross_references=$(grep -e 'CROSSREFERENCE' asciidoc_introspection | cut -d ':' -f 4 | sort -u)
+if [[ -n "$cross_references" ]]; then
+  echo 'ERROR: Some rule try to include content from unallowed directories.'
+  echo 'To share content between rules, you should use the "shared_content" folder at the root of the repository.'
+  echo "List of errors:"
```
necto, line 167: for consistency with the above
```suggestion
  echo 'List of errors:'
```

necto (changes requested):
I think it is worth the common-case time increase because it is not on a critical path: some other tasks take longer anyway, e.g., `tooling_test`, `validate_links`.
I agree with you that this is a more natural and flexible way to add validation checks, and I am looking forward to how it can be extended for more sophisticated and specialized checks.

Have you tried running it on CI with some failures it detects (smoke tests)?
I'd be curious to see the CI logs if any

ci/custom-asciidoctor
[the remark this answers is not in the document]
marco-antognini-sonarsource: Binary executables usually don't have any extension, like the official `asciidoctor`.

ci/validate_asciidoc.sh
marco-antognini-sonarsource: Sadly, it is not: the API is not designed to read the source surrounding the include.

ci/validate_asciidoc.sh
marco-antognini-sonarsource: Maybe it's _possible_, but not easy not maintainable AFAICT: we'd need to hack into the logging system somehow. It would be easier to fix https://github.com/asciidoctor/asciidoctor/issues/3414.

Also, we need a run with `rspectator-view` and one without it. The first is currently only executed if the second did not fail. We'd need to redesign this and how the plugins are enabled in the custom-asciidoctor (e.g., we don't want them when rspecatator view is disabled).

marco-antognini-sonarsource (commented):
Thanks for the review

marco-antognini-sonarsource:
> Have you tried running it on CI with some failures it detects (smoke tests)? I'd be curious to see the CI logs if any

I tested locally during development and at one point on the CI. I can't find the run again -- too many squashed commits to go through. Here are fresh runs:

* [this one](https://cirrus-ci.com/task/5497328451190784?logs=asciidoc_tests#L86) shows how it fails when there are unused files.

* [this other one](https://cirrus-ci.com/task/6335321147703296?logs=asciidoc_tests#L94) shows how it fails with bad cross refs.

(The good thing is, in rewriting the test I noticed a small issue with the impl. Fixed in 6ace02f.)

ci/custom-asciidoctor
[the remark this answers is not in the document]
necto: Did you mean just "executable"?
The official `asciidoctor` is a text executable.

This is a matter of taste. To me, the common well-known executables have no extension, and that distinguished them from the custom scripts like this one

necto (approved):
LGTM

ci/custom-asciidoctor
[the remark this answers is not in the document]
marco-antognini-sonarsource: yes

PR #2692, opened 2023-07-27
Modify rule S4124: Adapt to LaYC

victor-diez-sonarsource:


rules/S4124/javascript/rule.adoc
```diff
@@ -1,31 +1,43 @@
 == Why is this an issue?
 
-Declare a constructor inside an ``++interface++``, and you will get a simple method with the name "constructor". The same thing will happen if you create a ``++new++`` method inside the ``++interface++``: you'll get a simple method named "new".
+JavaScript and TypeScript classes may define a `constructor` method that is executed when a new instance is created. TypeScript allows interfaces that describe a static class object to define a `new()` method.
```
yassin-kammoun-sonarsource, line 3:
```suggestion
JavaScript and TypeScript classes may define a `constructor` method that is executed when a new instance is created. TypeScript allows interfaces that describe a static class object to define a `new()` method. Using these terms to name methods in other contexts can lead to confusion and make the code unclear and harder to understand.
```

rules/S4124/javascript/rule.adoc
```diff
@@ -1,31 +1,43 @@
...
 
+This rule reports when:
 
-Instead, the intent was probably to specify that the type did not originate from a TypeScript file. In such cases, just use the ``++declare class++`` syntax.
+* A class defines a method named `new`. The `new` keyword is used to create new instances of the class. If a method with the same name is defined, it can be unclear whether the method is intended to create new instances or perform some other action.
+* An interface defines a method named `constructor`. The constructor method is used to define the constructor function for a class that implements the interface. If a method with the same name is defined in the interface, it can be unclear whether the method is intended to define the constructor function or perform some other action.
 
+This rule helps to prevent confusion and ensure that the code is clear and easy to understand.
```
yassin-kammoun-sonarsource, line 10:
```suggestion
```

yassin-kammoun-sonarsource: I moved that part above and reformulated it a bit. You can ignore that suggestion if you prefer to keep things as they are.

yassin-kammoun-sonarsource (approved):

PR #2707, opened 2023-07-28
Modify rule S6464: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6464/python/rule.adoc
+++ b/rules/S6464/python/rule.adoc
@@ -1,27 +1,32 @@
+This rule raises an issue when trying to access or remove a dictionary key that does not exists.
+
 == Why is this an issue?
 
-Trying to access a dictionary key that does not exist will raise a `KeyError`.
+Trying to access or remove a dictionary key that does not exist will raise a `KeyError` exception.
 
 When trying to access or remove a key that may not be there, several solutions are possible:
 
 * Use the `get()` method instead of a subscription. It will return `None` for a missing key instead of raising a `KeyError`.
 * Use the `setdefault()` method to provide keys that may be missing with a default value.
 * Check that the key is present in the dictionary with the `if key in dict:` construct.
-* Use a `try/except` block and handle the potential `KeyError`.
-* Use a `defaultdict` instead of a regular `dict` object. 
+* Use a `try/except` block and handle the potential `KeyError` exception.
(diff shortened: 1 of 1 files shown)


rules/S6464/python/rule.adoc
```diff
@@ -1,27 +1,32 @@
+This rule raises an issue when trying to access or remove a dictionary key that does not exists.
```
guillaume-dequenne-sonarsource, line 1: Does the rule actually raise on removing keys?
Looking at the DBD tests, I'm not sure it does (at least it doesn't seem to raise on `del` statements). Even though it might make sense, I guess we'd better keep the description aligned with the implementation.

rules/S6464/python/rule.adoc
```diff
@@ -1,27 +1,32 @@
...
+
 == Why is this an issue?
 
-Trying to access a dictionary key that does not exist will raise a `KeyError`.
+Trying to access or remove a dictionary key that does not exist will raise a `KeyError` exception.
```
guillaume-dequenne-sonarsource, line 5: Same comment here.

guillaume-dequenne-sonarsource (approved):
Looks good.
Just a small comment about the fact that we probably should keep the description aligned with how the implementation currently behaves.

rules/S6464/python/rule.adoc
joke1196: Good call! I think I added this because of the paragraph on line 7. Not sure if I should keep the `remove` there or not. It felt a bit weird to propose a solution for something the rule will not raise. WDYT?

guillaume-dequenne-sonarsource: I guess we can have general recommendations for cases that the rule doesn't raise.
I don't think we should keep the "remove" on the high level description though, as it really imply we expect the rule to raise in cases it does not. 
That is, except if we actually intend to raise this soon. @andrea-guarino-sonarsource would you have an idea about that?

guillaume-dequenne-sonarsource (approved):
LGTM, I took the liberty to add a commit to fix a small typo.

maksim-grebeniuk-sonarsource (approved):

jean-jimbo-sonarsource (approved):
LGTM

PR #2709, opened 2023-07-28
Modify rule S6466: LaYC format

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6466/python/rule.adoc
+++ b/rules/S6466/python/rule.adoc
@@ -1,4 +1,4 @@
-This rule raises an issue when trying to access a list or a tuple index that is out of bound.
+This rule raises an issue when trying to access a list or a tuple index that is out of bounds.
 
 == Why is this an issue?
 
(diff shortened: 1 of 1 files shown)


rules/S6466/python/rule.adoc
```diff
@@ -1,27 +1,37 @@
...
     ls = [1, 2, 3]
     foo(ls[2])
 
 ----
 
+== Resources
+
+=== Documentation
+
+* Python Documentation - https://docs.python.org/3/library/exceptions.html#IndexError[IndexError]
```
joke1196, line 33: Was not 100% sure about the documentation. Is it really needed?

guillaume-dequenne-sonarsource: I don't think it really is, but now that it's there, I don't think we need to remove it 😄

guillaume-dequenne-sonarsource (approved):
LGTM

rules/S6466/python/rule.adoc
```diff
@@ -1,27 +1,37 @@
+This rule raises an issue when trying to access a list or a tuple index that is out of bound.
```
jean-jimbo-sonarsource, line 1: "...out of bounds"
is the correct phrase

PR #2713, opened 2023-07-31
Modify rule S3925: Adopt exception description to the new behavior

martin-strecker-sonarsource:
Add the new behavior to the RSpec introduced by
https://github.com/SonarSource/sonar-dotnet/pull/7673

## Review

- [ ] text style and tone


rules/S3925/csharp/rule.adoc
```diff
@@ -18,9 +18,34 @@ Specifically, this rule checks for these problems:
...
 
+This rule only raises an issue on classes that indicate that they are interested in serialization (see the _Exceptions_ section). That is to reduce noise because a https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2[lot] https://learn.microsoft.com/en-us/dotnet/api/system.exception[of] https://learn.microsoft.com/en-us/dotnet/api/system.uri[classes] https://learn.microsoft.com/en-us/dotnet/api/system.data.dataset[in] https://learn.microsoft.com/en-us/dotnet/api/system.collections.hashtable[the] https://learn.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest[base] https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex[class] https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.treenode[library] are implementing `ISerializable`.
```
cristian-ambrosini-sonarsource, line 21: id this chained link sentence on purpose? :satisfied: 
I propose to have the text followed by a brief bullet-point list to increase the readability, or a list within brackets (System.Exception, Regex...).

rules/S3925/csharp/rule.adoc
```diff
@@ -18,9 +18,34 @@ Specifically, this rule checks for these problems:
...
+
 === Exceptions
 
 * Classes in test projects are not checked.
+* Classes need to indicate that they are interested in serialization support by either
+  1. Having a `[Serializable]` attribute applied
```
cristian-ambrosini-sonarsource, line 27: Information: do you know if the nested bullet point list is displayed correctly? I wonder if it needs an empty line before this

rules/S3925/csharp/rule.adoc
```diff
@@ -18,9 +18,34 @@ Specifically, this rule checks for these problems:
...
+  3. Declaring an https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/serialization#supporting-runtime-serialization[serialization constructor]
+
+[source,csharp]
+----
+[Serializable]                                                                                 // 1.
+public class SerializationOptIn_Attribute
+{
+}
+
+public class SerializationOptIn_Interface: ISerilalizable                                      // 2.
```
cristian-ambrosini-sonarsource, line 38:
```suggestion
public class SerializationOptIn_Interface : ISerilalizable                                      // 2.
```

cristian-ambrosini-sonarsource (changes requested):

rules/S3925/csharp/rule.adoc
martin-strecker-sonarsource: The list is not complete and has more full illustration purposes. It is a bit unusual but does the job. If I do a bullet list, it will have to end with `* and more`. I don't think that this aspect is important enough to warrant such a big paragraph, but I wouldn't mind changing it. What do you think?

rules/S3925/csharp/rule.adoc
martin-strecker-sonarsource: According to the documentation, it is possible and a newline is optional:
https://docs.asciidoctor.org/asciidoc/latest/lists/ordered/#nested-ordered-list

rules/S3925/csharp/rule.adoc
cristian-ambrosini-sonarsource: I think for this kind of stuff it's better to ask @cynthiabethea.
Can you please give us your opinion on this sentence? :pray:

cristian-ambrosini-sonarsource (approved):
LGTM, please wait until we have an answer from Cynthia before merging 👍

rules/S3925/csharp/rule.adoc
cynthiabethea: My concern is, I don't think it's clear that each word is a distinct link. I would assume it's taking me to a page with a list of such classes. 

What about doing list as part of a setence.

That is to reduce noise because a lot of classes in the base class library are implementing ISerializable, including the following classes: System.Exception, Regex...

So it takes up less space than a bulleted list

cristian-ambrosini-sonarsource (approved):
I prefer this one a lot. LGTM!

PR #2714, opened 2023-07-31
Modify rule S5912: LaYC format

marco-antognini-sonarsource:
Given the nature of the bug is easy to understand, I didn't feel it was worth explicitly mentioning this is makes the application unreliable. Let me know if you disagree.

## Review

- [x] text style and tone

### DIFF
--- a/rules/S5912/cfamily/rule.adoc
+++ b/rules/S5912/cfamily/rule.adoc
@@ -10,7 +10,6 @@ The following example illustrates the unintended loss of information.
 struct PartData {
   int uuid;
   std::string manufacturer;
-  std::string description;
 };
 
 // Use inheritance to share common data definitions.
--- a/rules/S5912/cfamily/rule.adoc
+++ b/rules/S5912/cfamily/rule.adoc
@@ -31,11 +31,10 @@ void orderBike(TireData tire, ...) {
 
 == How to fix it
 
-This code defect is usually the consequence of not using polymorphism correctly.
-
-Instead of working with values and incomplete copies, polymorphism is achieved by casting a reference or a pointer to an object of the derived class to a reference or a pointer to an object of a base class.
+This code defect usually results from using values instead of references or pointers to pass polymorphic objects to functions.
 
 It is usually a good idea to design a base class so that slicing cannot happen: it can be abstract or non-copiable.
+The standard library follows this practice and prevents copying, for example, `std::ostream` objects.
 
 === Code examples
 
@@ -84,7 +83,6 @@ void application(int userId) {
 
 To prevent slicing from happening, the base class can be made non-copyable.
 This implies passing a reference instead of a copy to `writeAll`.
-The same strategy is applied for `std::ostream` and co.
 
 [source,cpp,diff-id=1,diff-type=compliant]
 ----
@@ -123,6 +121,11 @@ void application(int userId) {
 }
 ----
 
+=== Going the extra mile
+
+When slicing is actually required, it is best to make it explicit to avoid the element of surprise.
+You can create a dedicated member function for this purpose, with its own documentation.
+This goes well in hand with non-copyable classes.
 
 == Resources
 
--- a/rules/S5912/cfamily/rule.adoc
+++ b/rules/S5912/cfamily/rule.adoc
@@ -130,5 +130,6 @@ This goes well in hand with non-copyable classes.
 
 === External coding guidelines
 
-* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/efbc482f0d10555c6a8827906af71430a3023bd2/CppCoreGuidelines.md#es63-dont-slice[ES.63 Don't slice]
+* {cpp} Core Guidelines - 
+https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#es63-dont-slice[ES.63 Don't slice]
 
--- a/rules/S5912/cfamily/rule.adoc
+++ b/rules/S5912/cfamily/rule.adoc
@@ -39,7 +39,9 @@ Instead of working with values and incomplete copies, polymorphism is achieved b
 
 It is usually a good idea to design a base class so that slicing cannot happen: it can be abstract or non-copiable.
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 This example illustrates the problem with a `FileStream` concrete class and a derived buffered implementation, `BufferedFileStream`.
 
@@ -79,7 +81,7 @@ void application(int userId) {
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
 In the noncompliant example above, the `stream` gets sliced when calling `writeAll`, and its `buffer` may be lost or written out-of-sequence.
 
--- a/rules/S5912/cfamily/rule.adoc
+++ b/rules/S5912/cfamily/rule.adoc
@@ -19,15 +19,13 @@ struct TireData : PartData {
   TireType type;
 };
 
-std::vector<PartData> orderBike(TireData tire, ...) {
+void orderBike(TireData tire, ...) {
   std::vector<PartData> parts;
 
   // Noncompliant: the vector does not store the tire color and type.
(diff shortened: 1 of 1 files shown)


rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
+}
+----
+
+== How to fix it
+
+This code defect is usually the consequence of not using polymorphism correctly.
+
+Instead of working with values and incomplete copies, polymorphism is achieved by casting a reference or a pointer to an object of the derived class to a reference or a pointer to an object of a base class.
+
+It is usually a good idea to design a base class so that slicing cannot happen: it can be abstract or non-copiable.
```
amelie-renard-sonarsource, line 40: How about putting it into "Going the extra mile"?
And also mention what to do when you do the slicing on purpose (like what is explained in C++ Core Guidelines)

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
+  std::vector<int> data = getData();
+
+  writeAll(stream, data); // Noncompliant: stream is sliced.
 }
 ----
 
+==== Compliant solution
 
-=== Compliant solution
+In the noncompliant example above, the `stream` gets sliced when calling `writeAll`, and its `buffer` may be lost or written out-of-sequence.
```
amelie-renard-sonarsource, line 86: I would remove this sentence and update the "Noncompliant" comment above (as in my suggestion)

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
-  myShapes.push_back(c); // Noncompliant. What will be in the vector is a Shape, not a Circle, and won't contain any radius
+void writeAll(FileStream stream, std::vector<int> const& ints);
+
+void application(int userId) {
+  BufferedFileStream stream;
+  stream.write(userId);
+
+  std::vector<int> data = getData();
+
+  writeAll(stream, data); // Noncompliant: stream is sliced.
```
amelie-renard-sonarsource, line 80:
```suggestion
  writeAll(stream, data); // Noncompliant: stream is sliced, and its buffer may be lost or written out-of-sequence
```

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
 }
 ----
 
+==== Compliant solution
 
-=== Compliant solution
+In the noncompliant example above, the `stream` gets sliced when calling `writeAll`, and its `buffer` may be lost or written out-of-sequence.
 
-[source,cpp]
+To prevent this from happening, the base class can be made non-copyable.
```
amelie-renard-sonarsource, line 88:
```suggestion
To prevent slicing from happening, the base class can be made non-copyable.
```

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
+This implies passing a reference instead of a copy to `writeAll`.
+The same strategy is applied for `std::ostream` and co.
```
amelie-renard-sonarsource, line 90: I would put this last sentence when we explain to the user that slicing can be prevented by making a class non-copyable or abstract (so in the description of "How to fix it" or "Going the extra mile"

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
+
+  // ...
+
+  return parts;
+}
+----
+
+== How to fix it
+
+This code defect is usually the consequence of not using polymorphism correctly.
```
amelie-renard-sonarsource, line 36: I am uncomfortable with this sentence: for me, it's more a problem of (for example) passing by copy instead of reference. The general polymorphism architecture may be correct.
When I read incorrect uses of polymorphism, I think more about not using inheritance, redefining functions instead of overriding etc.
Does that make sense?

amelie-renard-sonarsource (changes requested):
I agree that there is no need to talk more about the software qualities, as it is simple.

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
 
+The following example illustrates the unintended loss of information.
 
-Note that it's usually a good idea to design a base class so that slicing cannot happen (it can be _abstract_, or _non-copiable_).
+[source,cpp]
+----
+struct PartData {
+  int uuid;
+  std::string manufacturer;
+  std::string description;
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: I feel that only one field makes the example less obvious why we are using inheritance to reuse definitions.

Maybe keeping two would be a good compromise, WDYT?

```suggestion
  int uuid;
  std::string manufacturer;
```

rules/S5912/cfamily/rule.adoc
marco-antognini-sonarsource: It totally does. I'll think about how to improve this. Thanks.

rules/S5912/cfamily/rule.adoc
marco-antognini-sonarsource: Makes sense indeed. It would be closer to the defect and easier to understand/follow.

marco-antognini-sonarsource:
```suggestion
```

rules/S5912/cfamily/rule.adoc
```diff
@@ -1,44 +1,135 @@
...
 }
 ----
 
 
 == Resources
 
-* https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es63-dont-slice[{cpp} Core Guidelines ES.63] - Don’t slice
+=== External coding guidelines
+
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/efbc482f0d10555c6a8827906af71430a3023bd2/CppCoreGuidelines.md#es63-dont-slice[ES.63 Don't slice]
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: The revision you suggest to use is 4 years old. Mine is only a month old. Ideally, we should update all links, but it requires manually verifying that the anchors are valid.

rules/S5912/cfamily/rule.adoc
marco-antognini-sonarsource: Makes sense. I moved it to How to fix it, and rephrased it a bit.

rules/S5912/cfamily/rule.adoc
marco-antognini-sonarsource: I thought a bit about this one.

I feel the "non-copyable" part should be part of the main "how to fix it" -- it's also used in an example below. I tried to move it to "going the extra mile" section but it felt strange when reading the description.

I do agree on the "slicing on purpose", it belongs there.

rules/S5912/cfamily/rule.adoc
marco-antognini-sonarsource: I rephrased a bit this part of the description. Let me know if it's better or can be further improved.

rules/S5912/cfamily/rule.adoc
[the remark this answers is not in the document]
amelie-renard-sonarsource: Yes, I'm fine with that

rules/S5912/cfamily/rule.adoc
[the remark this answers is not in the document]
amelie-renard-sonarsource: We had discussions about it and chose to keep the links consistent with one another (see ticket [CPP-4080](https://sonarsource.atlassian.net/browse/CPP-4080)). Here the two versions are identical so we can keep the consistency

amelie-renard-sonarsource (approved):
I really like this new version, good work!

rules/S5912/cfamily/rule.adoc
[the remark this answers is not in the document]
marco-antognini-sonarsource: Fair. I recorded my thoughts on the ticket. I'll apply the suggestion for consistency and we can revisit when the ticket is worked on.

PR #2747, opened 2023-07-31
Modify rule S1913: LaYC format

marco-antognini-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1913/cfamily/rule.adoc
+++ b/rules/S1913/cfamily/rule.adoc
@@ -1,13 +1,11 @@
 == Why is this an issue?
 
-Given an expression `e` of type `T`, `sizeof(e)` returns the size in bytes of `T`, without evaluating `e`.
-Furthermore, the expression `sizeof(e)` has type `size_t`.
-This means that `sizeof(sizeof(e))` is equivalent to `sizeof(size_t)`.
+Given an expression `e` of type `T`, `sizeof(e)` returns the size in bytes of `T`.
+The `sizeof` operator results in a value of `size_t`.
+Also, `sizeof(e)` has no side effects because `e` is not evaluated.
+Therefore, `sizeof(sizeof(e))` is equivalent to `sizeof(size_t)`.
 
-In other words:
-
-* `sizeof(sizeof(e))` always returns the same result and does not depend on `e`.
-* `sizeof` does not evaluate the expression, meaning it has no side effect.
+In other words, `sizeof(sizeof(e))` always gives the same result and does not depend on `e`, which is unlikely what was expected.
 
 == How to fix it
 
--- a/rules/S1913/cfamily/rule.adoc
+++ b/rules/S1913/cfamily/rule.adoc
@@ -8,7 +8,7 @@ In other words, its value does not depend on `e` nor has any side effects.
 
 == How to fix it
 
-This defect is usually the consequence of a typo, and the fix is to remove on level of `sizeof`.
+This defect is usually the consequence of a typo, and the fix is to remove one level of `sizeof`.
 
 === Code examples
 
@@ -16,7 +16,6 @@ This defect is usually the consequence of a typo, and the fix is to remove on le
 
 [source,c,diff-id=1,diff-type=noncompliant]
 ----
-#include <string.h>
 
 int main(int argc, char* argv[])
 {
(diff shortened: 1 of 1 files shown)


rules/S1913/cfamily/rule.adoc
```diff
@@ -1,11 +1,20 @@
 == Why is this an issue?
 
-A call to ``++sizeof(sizeof(...))++`` is equivalent to ``++sizeof(size_t)++``, and indicates a misuse or misunderstanding of the ``++sizeof++`` construct.
+Given an expression `e` of type `T`, `sizeof(e)` returns the size in bytes of `T`, without evaluating `e`.
+Furthermore, the expression `sizeof(e)` has type `size_t`.
 
+This means that `sizeof(sizeof(e))` is equivalent to `sizeof(size_t)`.
+In other words, its value does not depend on `e` nor has any side effects.
```
amelie-renard-sonarsource, line 7: "nor has any side effects": I find this information necessary and think we should emphasize it more.
The way I see it is that the user likely made a mistake because:
- sizeof(sizeof()) always returns the same result and does not depend on `e`
- sizeof does not evaluate the expression, which means that there is no side effect

I would separate these two things to make the second one more explicit than it currently is. WDYT?

amelie-renard-sonarsource (changes requested):

rules/S1913/cfamily/rule.adoc
marco-antognini-sonarsource: Great idea. I rephased the description in bdd67d6213. Let me know what you think.

rules/S1913/cfamily/rule.adoc
```diff
@@ -1,45 +1,52 @@
 == Why is this an issue?
 
-A call to ``++sizeof(sizeof(...))++`` is equivalent to ``++sizeof(size_t)++``, and indicates a misuse or misunderstanding of the ``++sizeof++`` construct.
+Given an expression `e` of type `T`, `sizeof(e)` returns the size in bytes of `T`, without evaluating `e`.
+Furthermore, the expression `sizeof(e)` has type `size_t`.
+This means that `sizeof(sizeof(e))` is equivalent to `sizeof(size_t)`.
```
amelie-renard-sonarsource, line 5: What about:
```suggestion
Given an expression `e` of type `T`, `sizeof(e)` returns the size in bytes of `T`. The `sizeof` function does not evaluate the expression `e`, and the result of `sizeof(e)` is of type `size_t`. Therefore, `sizeof(sizeof(e))` is equivalent to `sizeof(size_t)`.
```

rules/S1913/cfamily/rule.adoc
```diff
@@ -1,45 +1,52 @@
...
 
+In other words:
```
amelie-renard-sonarsource, line 7: And then, I would keep:
```suggestion
In other words, `sizeof(sizeof(e))` always returns the same result and does not depend on `e`, which is unlikely what was expected.
```

rules/S1913/cfamily/rule.adoc
amelie-renard-sonarsource: I find the result better but still a bit hard to follow. I wrote down suggestions, let me know what you think.

rules/S1913/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks for the suggestion.

Pedantic nit: `sizeof` is an operator, not a function. Although the difference does not really matter here, I've adapted your suggestion.

I also changed it slightly in 8771cba278 to have one statement/fact per sentence. I've also put more emphasis on the "no side effect" parts.

rules/S1913/cfamily/rule.adoc
```diff
@@ -1,45 +1,50 @@
 == Why is this an issue?
 
-A call to ``++sizeof(sizeof(...))++`` is equivalent to ``++sizeof(size_t)++``, and indicates a misuse or misunderstanding of the ``++sizeof++`` construct.
+Given an expression `e` of type `T`, `sizeof(e)` returns the size in bytes of `T`.
+The `sizeof` operator results in a value of `size_t`.
```
amelie-renard-sonarsource, line 4: I am not used to this wording, so it seems strange to me.
What about:
```suggestion
The `sizeof` operator results in a value of type `size_t`.
```
Or
```suggestion
The return type of the `sizeof` operator is `size_t`.
```

amelie-renard-sonarsource (approved):
Small suggestion, otherwise LGTM

PR #2748, opened 2023-07-31
Modify rule S5314: Expand and adjust for LaYC

balazs-benics-sonarsource:
I was thinking of these words as "glossary":
  * blocking
  * critical-section
  * data race
  * mutual exclusion

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5314/cfamily/rule.adoc
+++ b/rules/S5314/cfamily/rule.adoc
@@ -1,20 +1,229 @@
+Blocking functions should not be called inside critical sections
+
 == Why is this an issue?
 
-When entering a critical section the program holds a lock that can be hold by a single entity at a time. It is then recommended that the critical section is reduced to the minimum required and executed without any interruption.
+Concurrent accesses to shared resources are guarded by synchronization
+primitives such as mutexes to prevent data races. The section of code
+where a mutex is held is called the critical section. Critical sections
+are generally designed to be as small as possible, allowing concurrent
+threads to progress.
 
-
-=== Noncompliant code example
+It's usually unintentional to perform blocking operations inside a
+critical section because the operation might block for long or even
+indefinitely, degrading performance or causing a deadlock.
 
 [source,cpp]
 ----
-void f() {
-  std::mutex m;
-  // ...
-  m.lock();
-  // ...
-  sleep(1); // Noncompliant
-  // ...
-  m.unlock();
+#include <cstdio>    // printf()
+#include <cstdlib>   // atoi()
+#include <mutex>
+#include <unistd.h>  // sleep()
+
+std::mutex m;
+int load_shared_resource(); // Guarded by mutex 'm'
+
+// Some time-intensive computation.
+void do_expensive_work(int value, FILE *fd) {
+  char buf[4] = "";
+  std::fgets(buf, sizeof(buf), fd);
+  int sum = value + std::atoi(buf);
+  std::printf("value + line: %d\n", sum);
+}
+
+void worker_thread(FILE *fd) {
+  std::scoped_lock guard(m);
+  int value = load_shared_resource();
+  // Mutex 'm' could have been released here.
+  do_expensive_work(value, fd);
+} // Mutex 'm' only released here, after 'do_expensive_work' is returned.
+----
+
+Usually, blocking operations involve I/O operations, such as reading or
+writing a file or socket or sleeping for some specified time.
+
+=== What is the potential impact?
+
+Doing time-intensive operations while holding one or multiple locks will
+prevent concurrent threads from making progress updating the shared resource.
+This can lead to "bottlenecks" and the under-utilization of the hardware
+capabilities.
+
+== How to fix it
+
+Usually, one can move blocking or expensive computations from critical sections.
+This leads to smaller and cleaner critical sections in general,
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+#include <unistd.h>  // sleep()
+#include <mutex>
+#include <algorithm>
+
+int magic_number;
+std::mutex m;
+
+void consumer_thread() {
+  int current;
+  for (int items_processed = 0; items_processed < 100; ++items_processed) {
+    std::scoped_lock guard(m);
+    current = magic_number;
+
+    sleep(std::clamp(current, 0, 10)); // Noncompliant: 'sleep' blocks while holding mutext 'm'
+  }
 }
 ----
 
+[source,c,diff-id=2,diff-type=noncompliant]
+----
+#include <unistd.h>  // sleep()
+#include <pthread.h> // pthread_mutex_*()
+
+int magic_number;
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+void consumer_thread() {
+  int current;
+  for (int items_processed = 0; items_processed < 100; ++items_processed) {
+    pthread_mutex_lock(&m);
+    current = magic_number;
+    // Could have 'unlocked' mutex 'm' here.
+    sleep(current); // Noncompliant: 'sleep' blocks while holding mutex 'm'
+    pthread_mutex_unlock(&m);
+  }
+}
+----
+
+==== Compliant solution
+
+[source,cpp,diff-id=1,diff-type=compliant]
+----
+#include <unistd.h>  // sleep()
+#include <mutex>
+#include <algorithm>
+
+int magic_number;
+std::mutex m;
+
+void consumer_thread() {
+  int current;
+  for (int items_processed = 0; items_processed < 100; ++items_processed) {
+    {
+      std::scoped_lock guard(m);
+      current = magic_number;
+    }
+
+    sleep(std::clamp(current, 0, 10)); // Ok, blocking without holding any mutexes.
+  }
+}
+----
+
+[source,c,diff-id=2,diff-type=compliant]
+----
+#include <unistd.h>  // sleep()
+#include <pthread.h> // pthread_mutex_*()
+
+int magic_number;
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+void consumer_thread() {
+  int current;
+  for (int items_processed = 0; items_processed < 100; ++items_processed) {
+    pthread_mutex_lock(&m);
+    current = magic_number;
+    pthread_mutex_unlock(&m);
+    sleep(current); // Ok, blocking without holding any mutexes.
+  }
+}
+----
+
+=== Pitfalls
+
+It's important to note that not all operations can be moved out from the
+critical section without redesigning the algorithm.
+It's tempting to read, compute and commit back the result to some shared
+resource, but that coding pattern is vulnerable to
+Time-Of-Check-To-Time-Of-Use (TOCTOU) bugs.
+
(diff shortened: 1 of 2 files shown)


rules/S5314/cfamily/rule.adoc
```diff
@@ -1,20 +1,116 @@
+Blocking functions should not be called inside critical sections
+
 == Why is this an issue?
 
-When entering a critical section the program holds a lock that can be hold by a single entity at a time. It is then recommended that the critical section is reduced to the minimum required and executed without any interruption.
+Concurrent accesses to shared resources are guarded by synchronization
```
tomasz-kaminski-sonarsource, line 5: I think you should include an example code illustrating deadlock here. I.e. Two mutexes trying to be locked.
In general, I think having some code samples, explaining what we are talking about, in why this is an issues helps.

rules/S5314/cfamily/rule.adoc
```diff
@@ -1,20 +1,116 @@
...
+Doing time-intensive operations while holding one or multiple locks will
+prevent concurrent threads from making progress updating the shared resource.
+This can lead to "bottlenecks" and the under-utilization of the hardware
+capabilities.
 
-=== Noncompliant code example
+== How to fix it
 
-[source,cpp]
+Usually, blocking operations are separated from the critical section along
```
tomasz-kaminski-sonarsource, line 27: This sounds like a statement of the fact, and not an instruction, guide how to fix it. I think it should say something like, that  usually code can be changed to move such operation out of critical section.

rules/S5314/cfamily/rule.adoc
```diff
@@ -1,20 +1,116 @@
...
 ----
-void f() {
-  std::mutex m;
-  // ...
-  m.lock();
-  // ...
-  sleep(1); // Noncompliant
-  // ...
-  m.unlock();
+#include <unistd.h>  // sleep()
```
tomasz-kaminski-sonarsource, line 37: This rule applies for C, but examples are only for cpp. I would check what rule detect, and add examples for C here, and to lit, if they are not present.

rules/S5314/cfamily/rule.adoc
```diff
@@ -1,20 +1,116 @@
...
+}
+----
+
+=== Pitfalls
+
+It's important to note that not all operations can be moved out from the
+critical section without redesigning the algorithm.
+It's tempting to read, compute and commit back the result to some shared
+resource, but that coding pattern is vulnerable to
+Time-Of-Check-To-Time-Of-Use (TOCTOU) bugs.
```
tomasz-kaminski-sonarsource, line 85: Code example, if possible would be great.

tomasz-kaminski-sonarsource (changes requested):
Looks good. However, I think we need some more code examples, and adjust how to fix it.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: The rule fires for C++ as well.
Here is the non-compliant example on CompilerExplorer: https://godbolt.org/z/r1Ee6j7cq
I added a C example as well, using `pthreads`, which also passes: https://godbolt.org/z/5Gc76M8vY

tomasz-kaminski-sonarsource: If we add such example for rspec, we should make sure that we actually detect them. What we have done up to this point is to create an sonar-cpp PR, that contains both updated rspec and additional lit tests.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: Added, however, it ended up quite complicated.
And I also decided to mention the ABA problem.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: I'll add these tests in a separate PR to the sonar-cpp.

rules/S5314/cfamily/rule.adoc
```diff
@@ -71,7 +142,26 @@ void consumer_thread() {
       current = magic_number;
     }
 
-    sleep(std::clamp(current, 0, 10));
+    sleep(std::clamp(current, 0, 10)); // Ok, blocking without holding any mutexes.
+  }
+}
+----
+
+[source,c]
```
tomasz-kaminski-sonarsource, line 150: You can have second diff mark.

rules/S5314/cfamily/rule.adoc
```diff
@@ -8,10 +8,62 @@ where a mutex is held is called the critical section. Critical sections
...
+std::mutex m;
+int load_shared_resource(); // Guarded by mutex 'm'
+
+void do_expensive_work(int value); // Time-intensive computation.
+
+void worker_thread() {
+  std::scoped_lock guard(m);
+  int value = load_shared_resource();
+  // Mutex 'm' could have been released here.
+  do_expensive_work(value);
```
tomasz-kaminski-sonarsource, line 63: Would add `uncomplaint` marker here.

rules/S5314/cfamily/rule.adoc
```diff
@@ -8,10 +8,62 @@ where a mutex is held is called the critical section. Critical sections
 are generally designed to be as small as possible, allowing concurrent
 threads to progress.
 
+[source,cpp]
```
tomasz-kaminski-sonarsource, line 11: I know that I requested it, but I am not sure about this example. It is long, and I do not think add much. So maybe remove it.

rules/S5314/cfamily/rule.adoc
```diff
@@ -8,10 +8,62 @@ where a mutex is held is called the critical section. Critical sections
...
+}
+----
+
 It's usually unintentional to perform blocking operations inside a
 critical section because the operation might block for long or even
 indefinitely, degrading performance or causing a deadlock.
 
+[source,cpp]
+----
+#include <mutex>
```
tomasz-kaminski-sonarsource, line 52: This one is great.

rules/S5314/cfamily/rule.adoc
```diff
@@ -82,7 +172,53 @@ It's important to note that not all operations can be moved out from the
...
 resource, but that coding pattern is vulnerable to
-Time-Of-Check-To-Time-Of-Use (TOCTOU) bugs.
+Time-Of-Check-To-Time-Of-Use (TOCTOU) bugs and the ABA problem.
+
+[source,cpp]
+----
+#include <unistd.h>  // sleep()
+#include <mutex>
+
+int input;
```
tomasz-kaminski-sonarsource, line 182: Would declare some classes for that. Even as `class Input { ... }`.

tomasz-kaminski-sonarsource (changes requested):

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: Do you mean "non-compliant"?

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: Reverted!

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: I'm not sure I share this opinion.
To me, the example is already convoluted and bloated.
By using a custom UDT here, I would need to have a "default" input basically a sentinel, which is currently represented by `-1`.
As well as an equality operator to check if the input has changed or not.
Additionally, if I have a class, it might be actually a better design to encapsulate the mutex in the object itself, depending on the scenario.

Given all this, I'd prefer not to have a class here.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: I decided not to put `non-compliant` in this example, as we cannot see a blocking call, thus we won't diagnose it either.
I'd rather keep the `non-compliant` comments for coding patterns for which we raise an issue.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: I've heard that only the first diff will be displayed, thus it's not advised to have multiple diff markers.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: I split the ABA problem from the example, and mentioned after the example.

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: Here is the PR: https://github.com/SonarSource/sonar-cpp/pull/2916

rules/S5314/cfamily/rule.adoc
balazs-benics-sonarsource: I added the diff markers.

rules/S5314/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I would change this example then to something we can actually detect. We may otherwise put false expectation on the user. Same would apply to pitfall.

balazs-benics-sonarsource: I'm not exactly sure how to achieve that. Given how limited this rule is. I actually mentioned this in the spreadsheet.
Look at [here](https://github.com/SonarSource/llvm-project/blob/v16/clang/lib/StaticAnalyzer/Checkers/BlockInCriticalSectionChecker.cpp#L96), how the list of "blocking" functions is defined.
```C++
bool BlockInCriticalSectionChecker::isBlockingFunction(const CallEvent &Call) const {
  return matchesAny(Call, SleepFn, GetcFn, FgetsFn, ReadFn, RecvFn);
}
```
I don't really want to put a `sleep()` there as I don't think would be representative.
Calling `getc`, `fgets`, `read`, and `recv` is also quite inappropriate in this context. They you at least need some setup code to get a file-descriptor to do anything.

And in general, the rule won't be able to catch all of these issues. So, I had a mindset to rather educate the user with a sensible, but not too concrete example.

balazs-benics-sonarsource: Fixed. Have a look.

tomasz-kaminski-sonarsource (approved):

rules/S5314/cfamily/rule.adoc
```diff
@@ -1,20 +1,226 @@
...
+      if (input == previous_input) {
+        std::scoped_lock out_guard(output_mutex);
+        output = previous_result; // Commit the result.
+      }
+    }
+  }
 }
 ----
 
+Note that in the example, it might not be always correct to commit the result even if the `input` didn't change since we last checked. It could be that it was changed, but it was restored as we are checking it again. This is commonly known as the ABA problem, where `A` and `B` refers to the values of the resource.
```
tomasz-kaminski-sonarsource, line 194: The `in example` for me sugget above example code. I would just drop it.
```suggestion
Note that, it might not be always correct to commit the result even if the `input` didn't change since we last checked. It could be that it was changed, but it was restored as we are checking it again. This is commonly known as the ABA problem, where `A` and `B` refers to the values of the resource.
```

tomasz-kaminski-sonarsource:
Just single edit, and you can merge it.

PR #2750, opened 2023-08-01
Modify rule S3807: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone


rules/S3807/cfamily/rule.adoc
```diff
@@ -1,9 +1,23 @@
...
+The standard C library includes a variety of functions for string and general memory manipulations.
+These functions take pointers and respective lengths as parameters to specify chunks of memory to be processed.
+Passing ``++NULL++`` for the pointer will do nothing at best, but may result in undefined behavior and program crashes.
+
+
+== What is the potential impact?
+
+If an argument to a C library function has an invalid value, a pointer outside the address space of the program or a ``++NULL++`` pointer, for instance, the behavior is undefined.
+
+Undefined behavior is oftentimes hard to detect and may result in program crashes.
```
tomasz-kaminski-sonarsource, line 14: I would look on the other rules, that refer to undefined behavior like `S5553`, `S6655` and `S836` and try to harmonize the text here.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,9 +1,23 @@
...
+In the worst case, a program that comprises undefined behavior appears to execute correctly, while silently losing data or producing incorrect results.
 
 
-This rule raises an issue when the pointer passed to any of the following functions is ``++NULL++``:
+== How to fix it
+
+Ensure that any pointer passed to any of the following functions _is not_ ``++NULL++``:
```
tomasz-kaminski-sonarsource, line 20: I think that the list of functions should be in what is an issue, otherwise, for me it does not look complete.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,9 +1,23 @@
...
 
 * ``++int aio_suspend(const struct aiocb* const aiocb_list[], int nitems, const struct timespec* timeout);++``
 * ``++void* bsearch(const void* key, const void* base, size_t nmemb, size_t size, int (*compar)(const void *, const void *));++``
```
tomasz-kaminski-sonarsource, line 23: I would use the occasion to validate if this list is actual, i.e. matches current LLVM.

rules/S3807/cfamily/rule.adoc
```diff
@@ -117,13 +131,44 @@ This rule raises an issue when the pointer passed to any of the following functi
...
+  memcpy(&target, buffer, sizeof(buffer)); // Ok, 1st parameter denotes a valid target.
+}
 ----
 
+
+== Resources
+
+=== Standards
+
+* International Standard ISO/IEC 9899:201x (Committee Draft) https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf[Programming Languages C] cf. 7.1.4 "Use of library functions"
```
tomasz-kaminski-sonarsource, line 168: We do not link to standards usually, as the are not much helpful, for usually programmer. Instead I would suggest:
* checking the excel here for mapping: https://docs.google.com/spreadsheets/d/1HtofM6Gezv2VbGnMVn7jksrbNTRTHPy1XbjElywPL_8/edit#gid=0
* maybe linking the null poitner derefence rule as related.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,9 +1,23 @@
+Passing ``++NULL++`` as a parameter may result in a crash of the application.
```
tomasz-kaminski-sonarsource, line 1: I think we would need a bit more context here. Passing null in general to a function accepting a pointer may be fine. I would make it precise, and refer to undefined behavior. Example:
Passing ``++NULL++` to C function library functions that require non-null arguments leads to undefined behavior.

rules/S3807/cfamily/rule.adoc
```diff
@@ -117,13 +131,44 @@ This rule raises an issue when the pointer passed to any of the following functi
...
+
+==== Noncompliant code example
+
+[source,c,diff-id=1,diff-type=noncompliant]
+----
+#include <string.h>
+
+void string_copy() {
+  char buffer[] = "Hello, World!";
+  memcpy(NULL, buffer, sizeof(buffer)); // Noncompliant: 1st parameter is expected to be 'nonnull'.
```
tomasz-kaminski-sonarsource, line 144: This example looks very syntetic. Maybe having one, when we would get a path when pointer remained null would be better.

tomasz-kaminski-sonarsource (changes requested):

rules/S3807/cfamily/rule.adoc
pdschbrt: Thanks for the pointers. I've checked for the other rules and they just mention that undefined behavior is caused, they don't aim at describing undefined behavior (except one rule that provides a short sentence).

I'll adjust the description accordingly.

rules/S3807/cfamily/rule.adoc
pdschbrt: I've checked and updated the list according to the two checkers that are responsible for this rule 👍🏻

rules/S3807/cfamily/rule.adoc
pdschbrt: I've updated the link(s) now.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,129 +1,237 @@
...
 
-This rule raises an issue when the pointer passed to any of the following functions is ``++NULL++``:
+The standard C library includes a variety of functions for string and general memory manipulations.
+These functions typically take pointers and respective lengths as parameters to specify the chunks of memory to be processed.
+Many of these functions require valid parameters and when given values outside a function's domain, the behavior is undefined.
+In that case, the application may crash or, even worse, silently lose data or produce incorrect results.
+The C library functions listed in what follows require non-``++NULL++`` pointer parameters.
+Failing to provide non-``++NULL++`` parameters introduces undefined behavior.
 
 * ``++int aio_suspend(const struct aiocb* const aiocb_list[], int nitems, const struct timespec* timeout);++``
```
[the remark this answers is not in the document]
pdschbrt: I've tried to address your points. Please have a look at the updated version.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,129 +1,298 @@
+Passing ``++NULL++`` to a C library function that requires non-null pointer arguments results in undefined behavior and may crash the application.
```
tomasz-kaminski-sonarsource, line 1:
```suggestion
Passing ``++NULL++`` to a C library function that requires non-null pointer arguments results in undefined behavior
```

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,129 +1,298 @@
...
+
 == Why is this an issue?
 
-The standard C library includes a number of functions for string and memory manipulation. They take pointers and lengths as parameters. Passing ``++NULL++`` for the pointers will at best do nothing and at worst crash the application.
+The standard C library includes a variety of functions for string and general memory manipulations.
+These functions typically take pointers and respective lengths as parameters to specify the chunks of memory to be processed.
+Many of these functions require valid parameters and when given values outside a function's domain, the behavior is undefined.
+In that case, the application may crash or, even worse, silently lose data or produce incorrect results.
+The C library functions listed in what follows require non-``++NULL++`` pointer parameters.
```
tomasz-kaminski-sonarsource, line 9: I would go for something like:
Some C library functions, like `malloc` (pick a few). requires a non-NULL pointer parameters, and passing a null value introduces an undefined function.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,129 +1,298 @@
...
+Failing to provide non-``++NULL++`` parameters introduces undefined behavior.
+
+[source,text]
+----
+aio_suspend, bcmp, bcopy, bsearch, bzero, confstr, encrypt, erand48,
```
tomasz-kaminski-sonarsource, line 14: And then put the list in the separate section, without signatures. But consult someone else, as this one is big.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,129 +1,298 @@
...
+Hence, users have to ensure that the C standard library funcions' preconditions are met and valid parameters are passed.
+
 
+== What is the potential impact?
 
-This rule raises an issue when the pointer passed to any of the following functions is ``++NULL++``:
+If an argument to one of the C library functions mentioned by this rule is a ``++NULL++`` pointer, the behavior of the application is undefined.
+
+When a program comprises undefined behavior, the compiler no longer needs to adhere to the language standard, and the program has no meaning assigned to it.
+
```
tomasz-kaminski-sonarsource, line 74: I would add some more concrete examples here. Can this lead to voluneability, if some explain how, e.t.c.

rules/S3807/cfamily/rule.adoc
```diff
@@ -1,129 +1,298 @@
...
+==== Noncompliant code example
+
+[source,c,diff-id=1,diff-type=noncompliant]
+----
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+void string_copy() {
```
tomasz-kaminski-sonarsource, line 242: This function does not return a copy, so I would suggest renaming it and then using return type to indicate an error.
```suggestion
int process_string() {
```

Then have `return 1` instead of `error(1)`, and `return 0` at the end.

rules/S3807/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Program crashes.

tomasz-kaminski-sonarsource (changes requested):

pdschbrt:
I've addressed your comments, @tomasz-kaminski-sonarsource 
I also had a discussion with Arseniy concerning a good place for the list of C library functions. We've decided to only keep the condensed list and put it in the Resource section.

tomasz-kaminski-sonarsource (changes requested):
This looks good, it is great improvement compared to previous version.
For the next steps, it would be:
* Prepare a sonar-cpp PR that updates the HTML file, and includes new lit examples (if you have one)
* Merge the rspec to master
* Update the sonar-cpp rule HTML to match rspec master
* Merget the sonar-cpp branch

For generating HTML for rspec use
`sonar-cpp/sonar-cfamily-plugin$ gradle ruleMetadata`
The arguments are:
```
  Please specify property "rule":
    to update single rule: -Prule=S12345
    to update single rule from a branch: -Prule=S12345 -Pbranch=my-rspec-branch
```

PR #2751, opened 2023-08-01
Modify rule S5486: Adapt to LaYC

balazs-benics-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5486/cfamily/rule.adoc
+++ b/rules/S5486/cfamily/rule.adoc
@@ -1,51 +1,184 @@
+`pthread_mutex_t` should not be locked when already locked, or unlocked when already unlocked.
+
 == Why is this an issue?
 
 _Mutexes_ are synchronization primitives that allow to manage concurrency.
+This is the most fundamental building block for creating _safe_ concurrent applications.
+By using a _mutex_, one can ensure that a block of code is executed by a single thread concurrently.
+Data structures are designed to maintain their invariants between member-function calls.
+If a data structure is accessed concurrently, and one of the accesses is a write operation, then it has a _data race_.
+Having _data races_ is _undefined behavior_.
 
-* _non recursive mutexes_ are targeted by this rule. They can be locked/unlocked only once. Any locking/unlocking sequence that contains two consecutive identical operations leads to an undefined behaviour.
-* _recursive mutexes_ are not target by this rule.  They can be locked several times and unlocked several times as long as the number of locks/unlocks is the same.
+Adversaries actively exploit _data races_ to take over systems, but _data races_
+are also a common source of _data corruption_ in concurrent applications
+resulting in dormant and hard-to-find bugs.
 
-This rule raises an issue when a ``++pthread_mutex_t++`` is locked or unlocked several times in a row. We assume that all ``++pthread_mutex_t++`` are non-recursive (this is the most common case).
+To prevent _data races_, the shared resource (usually memory) must be protected
+by obtaining mutual access to the data during both reading and writing.
+Such mutual exclusion is generally achieved by using a _mutex_, which is
+frequently referred to as a _lock_.
 
+A _mutex_ has two states: _released_ - which is the initial state, or _acquired_.
+These two states are frequently called _unlocked_ and _locked_ as well.
 
-=== Noncompliant code example
+To effectively protect the shared resource from concurrent accesses, all such accesses should be guarded by the same _mutex_.
+They need to _lock_ the _mutex_ to gain _safe_ exclusive access to the resource and _unlock_ it after they are done mutating or reading it.
 
-[source,cpp]
+You can abstract away the concurrent threads sharing the mutex and think of it as owned by the current thread.
+It never spontaneously changes between _acquired_ and _released_.
+
+In this view, these are the possible transitions when calling ``lock`` or ``unlock`` on a _mutex_ in a given state:
+
+. _released_ + ``lock()``   => _acquired_
+. _acquired_ + ``unlock()`` => _released_
+. _acquired_ + ``lock()``   => _deadlock_
+. _released_ + ``unlock()`` => _undefined behavior_
+
+When a thread locks a mutex, another thread trying to _acquire_ the same mutex will be _blocked_ and have to wait for the first thread to _release_ it.
+This waiting period can take some time.
+If a thread attempts to lock a mutex it has already acquired, it will _deadlock_ because it would need to release it to lock it again.
+
+=== What is the potential impact?
+
+Locking an _acquired_ mutex leads to a _deadlock_, as a mutex can only be obtained once.
+Unlocking a _released_ mutex is _undefined behavior_.
+Removing synchronization can cause _data races_, leading to _data corruption_,
+which adversaries might leverage to take over the system.
+
+=== Exceptions
+
+There are _recursive mutexes_ that can be _acquired_ multiple times by the same
+thread, given that just as many times we also _release_ the _mutex_.
+They are rare in practice and usually signal design problems in the code.
+Thus we assume ``++pthread_mutex_t++`` refers to _non-recursive_ mutexes.
+
+== How to fix it
+
+Never lock an _acquired_ mutex.
+Lock the mutexes in a strict order that is followed throughout the project.
+Unlock operations should be done the same way but in reversed order.
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-pthread_mutex_t mtx1;
+#include <pthread.h>
 
-void bad1(void)
-{
-  pthread_mutex_lock(&mtx1);
-  pthread_mutex_lock(&mtx1);
+pthread_mutex_t m1 = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t m2 = PTHREAD_MUTEX_INITIALIZER;
+
+void locks() {
+  pthread_mutex_lock(&m1);
+  pthread_mutex_lock(&m1); // Noncompliant: 'm1' is already acquired
 }
 
-void bad2(void)
-{
-  pthread_mutex_unlock(&mtx1);
-  pthread_mutex_unlock(&mtx1);
+void unlocks() {
+  pthread_mutex_unlock(&m1);
+  pthread_mutex_unlock(&m1); // Noncompliant: 'm1' is already released
 }
 ----
 
+==== Compliant solution
 
-=== Compliant solution
-
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=compliant]
 ----
-pthread_mutex_t mtx1;
+#include <pthread.h>
 
-void ok(void)
-{
-  pthread_mutex_lock(&mtx1);
-  pthread_mutex_unlock(&mtx1);
+pthread_mutex_t m1 = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t m2 = PTHREAD_MUTEX_INITIALIZER;
+
+void locks() {
+  pthread_mutex_lock(&m1);
+  pthread_mutex_lock(&m2); // Compliant: we acquired both 'm1' and 'm2' for the first time.
+}
+
+void unlocks() {
+  pthread_mutex_unlock(&m2);
+  pthread_mutex_unlock(&m1); // Compliant: we released both 'm1' and 'm2' for the first time.
 }
 ----
 
+=== Pitfalls
+
+Calling arbitrary functions while holding a lock should be avoided,
+as the function might also want to lock the resource we already acquired,
+causing a deadlock. One particular example of such functions is callbacks.
+
+[source,c]
+----
+#include <pthread.h>
+
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+void handle_callback(void (*callback)(void)) {
+  pthread_mutex_lock(&m);
+  callback(); // If tries to lock mutex 'm', then we have a deadlock.
+  pthread_mutex_unlock(&m);
+}
+----
+
+=== Going the extra mile
+
+The section of code for which the _mutex_ is _acquired_ is called the _critical section_.
+Inside the _critical section_, we are the only ones with access to the shared resource.
+Thus we are free to mutate or read it without considering what other threads are doing
+concurrently. However, having large _critical sections_ can prevent other threads
+from making progress if they want to also _acquire_ the same _mutex_ and access
+the shared resource.
+Consequently, _critical sections_ are supposed to be as small as possible.
+
+[source,c]
+----
+#include <pthread.h>
+#include <stdbool.h>
+
+int input;
+int result;
+bool isFib;
+
+// Guards both 'num1' and 'num2'.
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+int fibonacci(int n);
+int factorial(int n);
+
+void locks(bool calcFib, int n) {
+  // Do the calculations without taking the lock.
+  int res;
+  if (calcFib) {
+    res = fibonacci(n);
+  } else {
+    res = factorial(n);
+  }
+
+  pthread_mutex_lock(&m);
+  // Critical section starts
+  input = n;
+  result = res;
+  isFib = calcFib;
+  // Critical section ends.
+  pthread_mutex_unlock(&m);
+}
+----
 
 == Resources
 
-* https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_destroy.html[The Open Group] pthread_mutex_init, pthread_mutex_destroy
+=== Standards
 
+* CWE - https://cwe.mitre.org/data/definitions/764[764 Multiple Locks of a Critical Resource]
+* CWE - https://cwe.mitre.org/data/definitions/362[362 Multiple Concurrent Execution using Shared Resource with Improper Synchronization ('Race Condition')]
+
+=== Related rules
+
+* S5487 enforces proper initialization and destruction of ``pthread`` mutexes.
+* S5489 enforces unlocking held ``pthread`` mutexes in reverse order.
+
+=== Documentation
+
+* The Open Group - https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_init.html[``++pthread_mutex_init++``, ``++pthread_mutex_destroy++``]
+* The Open Group - https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_lock.html[``++pthread_mutex_lock++``, ``++pthread_mutex_unlock++``]
 
 ifdef::env-github,rspecator-view[]
 '''
@@ -59,8 +192,5 @@ ifdef::env-github,rspecator-view[]
 === is related to: S5487
 
 === is related to: S5489
-
-=== on 6 Nov 2019, 23:32:48 Loïc Joly wrote:
-\[~geoffray.adde]Can you please review my changes?
 
 endif::env-github,rspecator-view[]
--- a/rules/S5486/cfamily/metadata.json
+++ b/rules/S5486/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "\"pthread_mutex_t\" should not be consecutively locked or unlocked twice",
+  "title": "\"pthread_mutex_t\" should not be locked when already locked, or unlocked when already unlocked",
   "type": "BUG",
   "status": "ready",
   "remediation": {
@@ -7,6 +7,7 @@
     "constantCost": "20min"
   },
   "tags": [
+    "cwe",
     "symbolic-execution",
     "multi-threading"
   ],
@@ -22,9 +23,14 @@
   "ruleSpecification": "RSPEC-5486",
   "sqKey": "S5486",
   "scope": "All",
+  "securityStandards": {
+    "CWE": [
+      764
+    ]
+  },
   "defaultQualityProfiles": [
     "Sonar way",
     "MISRA C++ 2008 recommended"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
 }


rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,108 @@
+"pthread_mutex_t" should not be consecutively locked or unlocked twice
```
tomasz-kaminski-sonarsource, line 1: This for me suggest that this rule works only if two locks are placed one after the another. Which is not the case.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,108 @@
...
+
 == Why is this an issue?
 
 _Mutexes_ are synchronization primitives that allow to manage concurrency.
 
-* _non recursive mutexes_ are targeted by this rule. They can be locked/unlocked only once. Any locking/unlocking sequence that contains two consecutive identical operations leads to an undefined behaviour.
-* _recursive mutexes_ are not target by this rule.  They can be locked several times and unlocked several times as long as the number of locks/unlocks is the same.
+* _non recursive mutexes_ are targeted by this rule.
```
tomasz-kaminski-sonarsource, line 7: I think to make that statement, you should explain what recursive on non-recursive mutexes are. 
I think I would suggest going with a description, starting with explaining critical section, and how mutexes are used to create it.

For example, something like, to be reworded
1. When a program run multiple threads, multiple threads may at the same time access (concurrently) the same data or other resources. Such concurrent access needs to be avoided...., and here come mutext.
2. Then how mutexes are used to create critical section (with code examples).
3. The go into locking.


In general, when writing rspec, I would think what kind of programmer would actually read 
"why this is an issue?" section. Somoene who understands concepts like recursive mutext, will not bother to checking why locking twice is an problem.

tomasz-kaminski-sonarsource (changes requested):

rules/S5486/cfamily/rule.adoc
balazs-benics-sonarsource: I think it makes sense. Have a look at how I incorporated this suggestion.

rules/S5486/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I think all synonyms of consecutively will have the same issues. Maybe something like more descriptive:
> locked twice, without being unlocked in between, or unlocked twice, without being locked in between.

tomasz-kaminski-sonarsource (approved):
This looks good to me. I would recommend having a second pair of eyes to check it.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+  isFib = calcFib;
+  // Critical section ends.
+  pthread_mutex_unlock(&m);
+}
+----
+
+=== What is the potential impact?
+
+Locking an _acquired_ mutex leads to a _deadlock_, as a mutex can only be obtained once.
+Unlocking a _released_ mutex is _undefined behavior_.
```
necto, line 82: In other rules, we add a sentence about what is _undefined behavior_ really about. like

> This means the compiler is not bound by the language standard anymore and your program has no meaning assigned to it.

e.g. [S3529](https://github.com/SonarSource/rspec/blob/cfb0d089bbf1141761208c5908d80486489b3620/rules/S3529/cfamily/rule.adoc#what-is-the-potential-impact)

And add possible concrete consequences if any. Concrete examples of undefined behavior have more impact on the reader

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+  pthread_mutex_lock(&m2);
+}
 
-void ok(void)
-{
-  pthread_mutex_lock(&mtx1);
-  pthread_mutex_unlock(&mtx1);
+void unlocks() {
+  pthread_mutex_unlock(&m1);
+  pthread_mutex_unlock(&m2);
```
necto, line 137: Didn't you write above that it should be in reverse order? This might be confusing

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+pthread_mutex_t m1 = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t m2 = PTHREAD_MUTEX_INITIALIZER;
 
-void bad1(void)
-{
-  pthread_mutex_lock(&mtx1);
-  pthread_mutex_lock(&mtx1);
+void locks() {
+  pthread_mutex_lock(&m1);
+  pthread_mutex_lock(&m1);
```
necto, line 112:
```suggestion
  pthread_mutex_lock(&m1); // Noncompliant: m1 is already locked
```

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
 }
 
-void bad2(void)
-{
-  pthread_mutex_unlock(&mtx1);
-  pthread_mutex_unlock(&mtx1);
+void unlocks() {
+  pthread_mutex_unlock(&m1);
+  pthread_mutex_unlock(&m1);
```
necto, line 117:
```suggestion
  pthread_mutex_unlock(&m1); // Noncompliant: m1 is already unlocked
```

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
 ----
-pthread_mutex_t mtx1;
+#include <pthread.h>
+
+pthread_mutex_t m1 = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t m2 = PTHREAD_MUTEX_INITIALIZER;
+
+void locks() {
+  pthread_mutex_lock(&m1);
+  pthread_mutex_lock(&m2);
```
necto, line 132:
```suggestion
  pthread_mutex_lock(&m2); // Compliant: m1 is locked, but m2 is not
```

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+
+[source,c]
+----
+#include <pthread.h>
+
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+void handle_callback(void (*callback)(void)) {
+  pthread_mutex_lock(&m);
+  callback(); // If locks mutex 'm', then we have a deadlock.
```
necto, line 155:
```suggestion
  callback(); // If tries to lock mutex 'm', then we have a deadlock.
```

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+  pthread_mutex_unlock(&m);
+}
+----
 
 == Resources
```
necto, line 160: Please format the links according to https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc

necto (changes requested):

rules/S5486/cfamily/rule.adoc
necto: I was primarily referring to the hyperlink being on the title, and not the source. Here is an [example](https://github.com/SonarSource/rspec/blob/a6984ff38766deefba7a7fe7fb38ce27ab22814f/rules/S1232/cfamily/rule.adoc#resources)

`grep` is not effective at the moment because most of our rules follow an outdated format.

For CWE, we do not need to mention MITRE, just CWE as a source, and the naked number as the subject title prefix, see [this example](https://github.com/SonarSource/rspec/pull/2596/files#diff-399a5f611cc941a3a4072b888d23eeba6341ccf6fc42e0850c2435d3f321f73fR273)

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+corruption or even corrupt the control-flow.
+
+Adversaries actively exploit data races to take over systems, but _data races_
+are also a common source of _data corruption_ in concurrent applications
+resulting in dormant and hard-to-find bugs.
+
+To prevent _data races_, the shared resource (usually memory) must be protected
+by obtaining mutual access to the data during both reading and writing.
+Such mutual exclusion is generally achieved by using a _mutex_, which is
+frequently referred to as a _lock_.
```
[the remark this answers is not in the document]
necto: That's fine. I can only suggest a shorter expression
```suggestion
Such mutual exclusion is generally achieved by using a _mutex_, also known as _"lock"_.
```
But I just wanted to make sure the "which is ... referred to" is indeed applied to the "mutex" and the "action of achieving mutual exclusion".

rules/S5486/cfamily/rule.adoc
```diff
@@ -60,7 +188,4 @@ ifdef::env-github,rspecator-view[]
 
 === is related to: S5489
 
-=== on 6 Nov 2019, 23:32:48 Loïc Joly wrote:
```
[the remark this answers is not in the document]
necto: "is related to XXXX" is an artifact of the migration. I agree that it is better placed in the metadata, but that's an improvement for another day

rules/S5486/cfamily/rule.adoc
necto: I do not know any specifics, so I would just add the sentence cited above (as the reader might not know what "undefined behavior" means and hos bad it is)

rules/S5486/cfamily/rule.adoc
necto: since `m` is already locked, `callback` cannot lock it. It can call `lock` which will block until it can lock the mutex (which will never happen). I call that "try to lock" because it will try, but it can't succeed, and it is the trying that leads to deadlock

balazs-benics-sonarsource: Indeed, I thought I can abbreviate things and I meant by "locks" here that it "calls lock".
I feel it's bikeshedding, but whatever.

rules/S5486/cfamily/rule.adoc
balazs-benics-sonarsource: Thanks for the clarification. Now it makes sense.

balazs-benics-sonarsource:
I start to get lost in the review. Did I address everything?

balazs-benics-sonarsource:
I reworded the state transitions paragraph to be more "declarative".
I experimented with using tables, but I couldn't figure out how to make it pretty.
A major problem was: how to give labels to the "state" and "action" dimensions in a nice way.

I considered representing the result of the combination in the cell for a given row (state) and column (action).
In the end, I ended up having a pattern matching-like construct, that I believe is still pretty good.

rules/S5486/cfamily/rule.adoc
```diff
@@ -23,19 +18,23 @@ by obtaining mutual access to the data during both reading and writing.
...
+A _mutex_ has two states: _released_ - which is the initial state, or _acquired_. +
+These are the possible transitions when calling ``lock`` or ``unlock`` on a _mutex_ in a given state:
+
+. _released_ + ``lock()``   => _acquired_ ^(*)^
+. _acquired_ + ``unlock()`` => _released_
+. _acquired_ + ``lock()``   => _deadlock_
+. _released_ + ``unlock()`` => _undefined behavior_
+
+(*): Some thread might have _locked_ the mutex, thus our thread needs to _block_, and wait for that thread to _unlock_ the _mutex_. This might take a while.
+A consequence of this blocking behavior is why calling ``lock`` on a _mutex_ that is already _acquired_ by our thread will result in _deadlock_ because it would block until the _mutex_ becomes _available_, which will never happen.
```
necto, line 31: This sentence is too long, it overflows my reading buffer

rules/S5486/cfamily/rule.adoc
```diff
@@ -23,19 +18,23 @@ by obtaining mutual access to the data during both reading and writing.
...
-Calling ``++unlock++`` on a mutex that is already _released_ is _undefined behavior_
-in most implementations, thus, should be avoided.
+
+A _mutex_ has two states: _released_ - which is the initial state, or _acquired_. +
+These are the possible transitions when calling ``lock`` or ``unlock`` on a _mutex_ in a given state:
+
+. _released_ + ``lock()``   => _acquired_ ^(*)^
+. _acquired_ + ``unlock()`` => _released_
+. _acquired_ + ``lock()``   => _deadlock_
+. _released_ + ``unlock()`` => _undefined behavior_
```
necto, line 28: This is much more clear, thanks!

Since the case you are describing in (*) corresponds to "acquired + lock" (it was acquired by another thread before it had a chance to finish lock), I would move the (*) to the third line "acquired + lock", in this case, it will block and wait.
in the (*) you can note that if no one releases the mutex during this waiting, it is called a "deadlock".

So, if the mutex was locked by a different thread, it is fine, the thread that locked it will likely unlock it eventually.
However, if the mutex was locked by this same thread, it has no one to rescue it.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
+"pthread_mutex_t" should not be locked when already locked, or unlocked when already unlocked
```
necto, line 1:
```suggestion
"pthread_mutex_t" should not be locked when already locked, or unlocked when already unlocked.
```

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
...
+
 == Why is this an issue?
 
 _Mutexes_ are synchronization primitives that allow to manage concurrency.
+This is the most fundamental building block for creating _safe_ concurrent applications.
+By using a _mutex_, one can ensure that a block of code is executed by a single thread concurrently.
+Data structures are designed to maintain their invariants in member-function calls while modifying their internal representation.
```
necto, line 8: This still sounds odd to me, because you usually can't maintain an invariant while modifying internal representation. That's why member-functions are necessary: to maintain invariant between the calls, and encapsulate the code where the invariant is broken. How about dropping the second part here:
```suggestion
Data structures are designed to maintain their invariants between member-function calls.
```

necto (changes requested):
> I start to get lost in the review. Did I address everything?

I believe you have not addressed [this thread](https://github.com/SonarSource/rspec/pull/2751#discussion_r1281946570)
> - To me, this together with the code snippet belongs to "going the extra mile". You've fixed your double lock as the rule told you, now if you want your code to be better and less prone to such an error, read on about "critical sections" and why to make them small
> - On second thought, I agree with you. It should be in the "extra mile" section.

Personally, to avoid getting lost, I usually abstain from commenting on threads without making sure to address them. For example, "I'll come back to that", is nice communication, but then it puts your comment as the last in the thread. And that is what I rely on when I am looking for the threads I have not addressed: they are the ones that do not end with my own comment.

> I ended up having a pattern matching-like construct, that I believe is still pretty good.

I find it also a nice and concise enough, no need for a table

Another overall comment, subjectively I feel you are overusing _emphasis_. Also sometimes you use it instead of double quotes `"`, but not always.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,50 +1,178 @@
...
+Calling ``++unlock++`` on a mutex that is already _released_ is _undefined behavior_
+in most implementations, thus, should be avoided.
+
+The section of code for which the _mutex_ is _acquired_ is called the _critical section_.
+Inside the _critical section_, we are the only one who has access to the shared resource.
+Thus we are free to mutate or read it without considering what other threads are doing
+concurrently. However, having large _critical sections_ can prevent other threads
+from making progress if they want to also _acquire_ the same _mutex_ and access
+the shared resource.
+Consequently, _critical sections_ are supposed to be as small as possible.
```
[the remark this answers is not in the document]
balazs-benics-sonarsource: Move to the "extra mile" section, as-is.

rules/S5486/cfamily/rule.adoc
necto: Or maybe it is malformed:
"A consequence of ... is why calling ... because it would ..., which will ..." - something is amiss.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,177 @@
...
+resulting in dormant and hard-to-find bugs.
+
+To prevent _data races_, the shared resource (usually memory) must be protected
+by obtaining mutual access to the data during both reading and writing.
+Such mutual exclusion is generally achieved by using a _mutex_, which is
+frequently referred to as a _lock_.
+
+
+A _mutex_ has two states: _released_ - which is the initial state, or _acquired_. +
+These are the possible transitions when calling ``lock`` or ``unlock`` on a _mutex_ in a given state:
```
necto, line 23: Here is my summary of your description of the mental model for dealing with mutexes.
With this preamble, the table below looks correct to me.
Let me know if this summary matches your thinking.
```suggestion
A _mutex_ has two states: _released_ - which is the initial state, or _acquired_.
You can abstract away the concurrent thread sharing the mutex and think of it as owned by the current thread:
it never spontaneously changes between acquired and released.
In this view, these are the possible transitions when calling ``lock`` or ``unlock`` on a _mutex_ in a given state:
```

rules/S5486/cfamily/rule.adoc
balazs-benics-sonarsource: I think I addressed this one. Let me know if you like it.

balazs-benics-sonarsource (commented):
Fixed everything, except rewriting the state transitions.

balazs-benics-sonarsource:
@arseniy-sonar Sorry about the spurious review request about 5 minutes ago.
I didn't know you already wrote your suggestion to the "transition" section.
I liked it.

Now, it's in the final from.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,184 @@
...
+
+To prevent _data races_, the shared resource (usually memory) must be protected
+by obtaining mutual access to the data during both reading and writing.
+Such mutual exclusion is generally achieved by using a _mutex_, which is
+frequently referred to as a _lock_.
+
+
+A _mutex_ has two states: _released_ - which is the initial state, or _acquired_.
+
+You can abstract away the concurrent thread sharing the mutex and think of it as owned by the current thread.
```
necto, line 24: Is there any value in restricting the number of concurrent threads? If not, just use plural
```suggestion
You can abstract away the concurrent threads sharing the mutex and think of it as owned by the current thread.
```

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,184 @@
...
+In this view, these are the possible transitions when calling ``lock`` or ``unlock`` on a _mutex_ in a given state:
+
+. _released_ + ``lock()``   => _acquired_ ^(*)^
+. _acquired_ + ``unlock()`` => _released_
+. _acquired_ + ``lock()``   => _deadlock_
+. _released_ + ``unlock()`` => _undefined behavior_
+
+(*): When a thread locks a mutex, another thread trying to _acquire_ the same mutex will be _blocked_ and have to wait for the first thread to _release_ it.
+This waiting period can take some time.
+If a thread attempts to lock a mutex it has already acquired, it will _deadlock_ because it would need to release it to lock it again.
```
necto, line 36: It is confusing to put into the footnote for the first row the situation describing the third row.

What do you think of
- having two footnotes - for first and third rows, or
- adding it as a standalone sentence in a separate pararaph?

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,184 @@
...
 
-* _non recursive mutexes_ are targeted by this rule. They can be locked/unlocked only once. Any locking/unlocking sequence that contains two consecutive identical operations leads to an undefined behaviour.
-* _recursive mutexes_ are not target by this rule.  They can be locked several times and unlocked several times as long as the number of locks/unlocks is the same.
+To effectively protect the shared resource from concurrent accesses, all such accesses should be guarded by the same _mutex_.
+They need to _lock_ the _mutex_ to gain _safe_ exclusive access to the resource and _unlock_ it after they are done mutating or reading it.
```
necto, line 39: It looks like the "punchline" of this section - "why this is an issue" - is the end of the footnote `(*)`, and then it is followed by a generic statement of why mutexes are useful.
Yet, this paragraph is (more) relevant where you introduce the mutex states.

If you agree that this is an issue, I propose to rearrange the paragraphs:
1. two states of the mutex
2. use of mutexes for protection of resource access
3. mental model of mutex state evolution
4. table of state transitions
5. footnote
6. > If a thread attempts to lock a mutex it has already acquired, it will _deadlock_ because it would need to release it to lock it again.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,184 @@
...
-void bad1(void)
-{
-  pthread_mutex_lock(&mtx1);
-  pthread_mutex_lock(&mtx1);
+pthread_mutex_t m1 = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t m2 = PTHREAD_MUTEX_INITIALIZER;
+
+void locks() {
+  pthread_mutex_lock(&m1);
+  pthread_mutex_lock(&m1); // Noncompliant: m1 is already locked
```
necto, line 74: Other than the rule summary and title (which are identical), this is the first time you use "locked". In the "Why this is an issue" you've used the term "acquired" instead. Should we speak about "acquired" instead of locked here as well?
Or perhaps, use "locked" everywhere, and maybe mention that it is also called "acquired"?

necto (changes requested):

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,184 @@
...
+  isFib = calcFib;
+  // Critical section ends.
+  pthread_mutex_unlock(&m);
+}
+----
 
 == Resources
 
-* https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_destroy.html[The Open Group] pthread_mutex_init, pthread_mutex_destroy
+=== External coding guidelines
```
tomasz-kaminski-sonarsource, line 168: I think this should be the standards section.

rules/S5486/cfamily/rule.adoc
balazs-benics-sonarsource: Added the sentence "These two states are frequently called _unlocked_ and _locked_ as well." after introducing the term "acquired" to clarify this.
Also now using "acquired" and "released" in the "Noncompliant" comments.

rules/S5486/cfamily/rule.adoc
balazs-benics-sonarsource: I followed what is present in S836, where CWE was part of the "External coding guidelines" section.
I've moved my CWE to the "Standards" section, as suggested. Thanks.

Speaking of sibling sections, does the order matter?
For example, I mention "Standards" and then, "Related rules" and then "Documentation".
While in the rspec reference they are mentioned in this order:
> Resources (optional)
>  - Documentation   <---
>  - Articles & blog posts
>  - Conference presentations
>  - Standards   <---
>  - External coding guidelines
>  - Benchmarks
>  - Related rules   <---

rules/S5486/cfamily/rule.adoc
balazs-benics-sonarsource: I agree that the paragraph "To effectively protect..." is at a weird place.

It took me a while to understand that all you want is to move the paragraph "To effectively protect..." in between the "A _mutex_ has two states:..." and "You can abstract away the concurrent threads...".

I've moved the paragraph.

balazs-benics-sonarsource (commented):
Resolved the review comments.

tomasz-kaminski-sonarsource (approved):
This looks great! The section on the data races, critical section, and mutexes is spot on for me, it explains the concepts in a simple manner.

rules/S5486/cfamily/rule.adoc
```diff
@@ -1,51 +1,185 @@
+"pthread_mutex_t" should not be locked when already locked, or unlocked when already unlocked.
```
necto, line 1: I believe we use double quotes for code only in the title and the messages. In the rule description you can enjoy the proper formatting
```suggestion
`pthread_mutex_t` should not be locked when already locked, or unlocked when already unlocked.
```

rules/S5486/cfamily/rule.adoc
necto: Looks good

necto (approved):
Agree, the rule is really solid now

PR #2752, opened 2023-08-01
Modify rule S5272: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5272/cfamily/rule.adoc
+++ b/rules/S5272/cfamily/rule.adoc
@@ -15,7 +15,7 @@ Move-assignment operators behave analogously, except that they are used once the
 In contrast to copy-assignment operators, a move-assignment operator too "steals" the moved-from object's resources without the need for making a potentially expensive copy.
 
 The ``++DynamicIntArray++`` type defined in the following manages a resource, i.e., memory, and shall serve as an example that showcases how move operations "steal" another objects resources and how they differ from making copies.
-While the copy constructor does make a full copy by allocating memory and then copying the ``++other++`` object's array values into the freshly allocated memory, the move constructor only assignes the pointer to point to ``++other++``'s dynamically allocated memory.
+While the copy constructor does make a full copy by allocating memory and then copying the ``++other++`` object's array values into the freshly allocated memory, the move constructor only assigns the pointer to point to ``++other++``'s dynamically allocated memory.
 It then sets the pointer of the ``++other++`` object to ``++nullptr++`` to allow its correct cleanup by the destructor of the ``++DynamicIntArray++`` type.
 The implementations for the copy- and move-assignment operators are similar with the main difference being that the objects have already been constructed.
 
--- a/rules/S5272/cfamily/rule.adoc
+++ b/rules/S5272/cfamily/rule.adoc
@@ -158,7 +158,7 @@ int foo() {
 }
 ----
 
-[source,cpp,diff-id=1,diff-type=compliant]
+[source,cpp]
 ----
 int foo() {
   DynamicIntArray a{128, 0};
--- a/rules/S5272/cfamily/rule.adoc
+++ b/rules/S5272/cfamily/rule.adoc
@@ -2,13 +2,41 @@ Relying on an object that has been moved-from may lead to undefined behavior.
 
 == Why is this an issue?
 
-In accordance with the specification, objects supporting move operations will be left in a valid _but_ unspecified state after the move took place.
+In accordance with the C++ language specification, objects supporting move operations will be left in a valid _but_ unspecified state after the move took place.
 Even if in a valid state, the fact of being in an unspecified state leads to undefined behavior.
+
+Move construction and its respective move semantics has been introduced in C++11.
+Moving objects becomes interesting if one wishes to get an object into a different scope, while no longer requiring the original object.
+While one would previously need to make a potentially expensive copy to get an object into another scope and then destroy the original, move constructors allow one to _move_ objects without performing a copy.
+Move constructors are typically implemented by "stealing" the resources held by another object specified as the move constructor's parameter, rather than making a copy.
+"Stealing" resources (e.g. memory) from another object is oftentimes much more efficient than making a copy and destroying the original, and can frequently be implemented by reassigning a few pointer variables.
+
+The ``++ObjectTy++`` defined in the following offers a default constructor, a destructor, a copy constructor and a move constructor as well as a copy-assignment and move-assignment operation.
+
+[source,cpp]
+----
+struct ObjectTy {
+  ObjectTy();
+  ~ObjectTy();
+  ObjectTy(const ObjectTy &other);
+  ObjectTy(ObjectTy &&other);
+
+  ObjectTy& operator=(const ObjectTy &other);
+  ObjectTy& operator=(ObjectTy &&other);
+};
+----
+
+=== Exceptions
+
+There are some C++ standard template library types such as `std::unique_ptr` for which the moved-from state is fully specified.
 
 
 == What is the potential impact?
 
 Using an object after it has been moved-from leads to *undefined behavior*.
+
+For programs that comprise undefined behavior, the compiler is no longer bound to the language specification.
+The application may crash or, even worse, the application appears to execute correctly while losing data or producing incorrect results, for instance.
 
 
 == How to fix it
@@ -22,7 +50,7 @@ Do not use an object after it has been moved-from, i.e., has been used to move-c
 
 [source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-#include <algorithm> // std::fill
+#include <algorithm> // std::copy, std::fill
 #include <memory> // std::move
 
 class DynamicIntArray {
@@ -31,15 +59,25 @@ private:
   int *data;
 
 public:
+  // Constructor
   explicit DynamicIntArray(size_t size) : size(size), data(new int[size]) {
     std::fill(data, &data[size], 42);
   }
+  // Destructor
   ~DynamicIntArray() {
     delete[] data;
     size = 0;
   }
+  // Copy constructor (copies object)
+  DynamicIntArray(const DynamicIntArray &other)
+      : size(other.size), data(new int[other.size]) {
+    std::copy(other.data, &other.data[size], data);
+  }
+  // Move constructor ("steals" data, no allocation or copy necessary)
   DynamicIntArray(DynamicIntArray &&other) noexcept
       : size(other.size), data(other.data) {
+    // Ensure that the moved-from object `other` can be safely destructed (using
+    // delete[]).
     other.data = nullptr;
     other.size = 0;
   }
@@ -61,7 +99,7 @@ int foo() {
 
 [source,cpp,diff-id=1,diff-type=compliant]
 ----
-#include <algorithm> // std::fill
+#include <algorithm> // std::copy, std::fill
 #include <memory> // std::move
 
 class DynamicIntArray {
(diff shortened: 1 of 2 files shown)


rules/S5272/cfamily/rule.adoc
```diff
@@ -1,38 +1,160 @@
...
+};
+
+int foo() {
+  DynamicIntArray a{32};
+
+  DynamicIntArray a2 = std::move(a);
+  int x = a2.getValueAt(0); // Ok, the correctly move-constructed variable `a2` can be used, of course.
+  return x;
 }
 ----
```
necto, line 145: You gave one way to deal with the problem. To increase the rule value, you can provide other ways to fix it. Here are some ideas:
- copy instead of move
- take a regular (lvalue) reference instead of move-constructing another object (although, be careful not to suggest creating a dangling reference)
- remove the use
- remove the move, and use the same object instead of it's "move-constructed" successor

Also, another example of non-compliant code could use a move constructor, and yet another, could pass `std::move(x)` to a function taking an rvalue reference and use it after the call.

As you can only use diff view on a pair of code snippets, it might be more comprehensible if for different non-compliant  examples, you give different compliant solutions

necto (changes requested):

pdschbrt:
Thanks for your feedback, @necto 

I'll try to address your points.

I was also thinking of discussing the move-assignement operator, but then thought that the description could become too lengthy. But I'll try to give it a shot. We should be able to make it shorter by putting the `DynamicIntArray` implementation in the _Why is this an issue?_ section and then referring back to it.

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,38 +1,160 @@
+Relying on an object that has been moved-from may lead to undefined behavior.
```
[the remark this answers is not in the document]
pdschbrt: There are a few selected STL types such as `std::unique_ptr` for which the moved-from state is fully specified. I mention that later in the `Exceptions` section.

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,38 +1,160 @@
...
+In accordance with the C++ language specification, objects supporting move operations will be left in a valid _but_ unspecified state after the move took place.
+Even if in a valid state, the fact of being in an unspecified state leads to undefined behavior.
 
+Move construction and its respective move semantics has been introduced in C++11.
+Moving objects becomes interesting if one wishes to get an object into a different scope, while no longer requiring the original object.
+While one would previously need to make a potentially expensive copy to get an object into another scope and then destroy the original, move constructors allow one to _move_ objects without performing a copy.
+Move constructors are typically implemented by "stealing" the resources held by another object specified as the move constructor's parameter, rather than making a copy.
+"Stealing" resources (e.g. memory) from another object is oftentimes much more efficient than making a copy and destroying the original, and can frequently be implemented by reassigning a few pointer variables.
```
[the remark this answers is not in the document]
pdschbrt: I've added another brief paragraph that comments on the assignment operators. It's also picked up once more in the example.

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,38 +1,160 @@
...
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+#include <algorithm> // std::copy, std::fill
+#include <memory> // std::move
+
+class DynamicIntArray {
```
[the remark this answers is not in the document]
pdschbrt: This is now factored out.

rules/S5272/cfamily/rule.adoc
pdschbrt: I've tried to address these points in additional code examples.

pdschbrt:
I've addressed all of your comments. Please give it another shot, @necto.

Thanks!

rules/S5272/cfamily/rule.adoc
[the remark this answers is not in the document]
necto: I see. How about making it a little stronger, like
```suggestion
Relying on an object that has been moved-from leads to undefined behavior in most cases.
```
Because I imagine, even though `std::unique_ptr` moved-from state is well specified, it is not relied upon often

rules/S5272/cfamily/rule.adoc
```diff
@@ -99,56 +141,90 @@ int foo() {
...
+
+==== Compliant solution
+
+[source,cpp,diff-id=2,diff-type=compliant]
+----
+void array_user(DynamicIntArray &&array);
+
+int bar() {
+  DynamicIntArray a{512, 0};
+  array_user(a);
```
necto, line 208: This fails to compile

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,37 +1,235 @@
...
+While one would previously need to make a potentially expensive copy to get an object into another scope and then destroy the original, move constructors allow one to _move_ objects without performing a copy.
+Move constructors are typically implemented by "stealing" the resources held by another object specified as the move constructor's parameter, rather than making a copy.
+"Stealing" resources (e.g. memory) from another object is oftentimes much more efficient than making a copy and destroying the original, and can frequently be implemented by reassigning a few pointer variables.
 
+Move-assignment operators behave analogously, except that they are used once the object that is moved-to has already been constructed.
+In contrast to copy-assignment operators, a move-assignment operator too "steals" the moved-from object's resources without the need for making a potentially expensive copy.
 
-=== Noncompliant code example
+The ``++DynamicIntArray++`` type defined in the following manages a resource, i.e., memory, and shall serve as an example that showcases how move operations "steal" another objects resources and how they differ from making copies.
```
necto, line 17:
```suggestion
The ``++DynamicIntArray++`` type defined in the following manages a resource, i.e., memory, and shall serve as an example that showcases how move operations "steal" another object's resources and how they differ from making copies.
```

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,37 +1,235 @@
...
-  A a;
+#include <algorithm> // std::copy, std::fill
+#include <memory>    // std::move
+
+class DynamicIntArray {
+  size_t size;
+  int *data;
+
+public:
+  // Constructor
```
necto, line 32: IMO, it is obvious that this is a constructor
```suggestion
```

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,37 +1,235 @@
...
+  explicit DynamicIntArray(size_t size, int initial_value)
+      : size(size), data(new int[size]) {
+    std::fill(data, &data[size], initial_value);
+  }
+  // Destructor
```
necto, line 37: IMO, it is obvious that this is a destructor
```suggestion
```

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,37 +1,235 @@
...
+int foo() {
+  DynamicIntArray a{128, 0};
+
+  DynamicIntArray a2 = std::move(a);
+  int x = a2.getValueAt(0); // Compliant: the correctly move-constructed variable `a2` can be used, of course.
+  return x;
+}
+----
+
+[source,cpp,diff-id=1,diff-type=compliant]
```
necto, line 153: You cannot reuse same `diff-id` more or less than in 2 code snippets. This is a limitation, so in the alternative fix proposal you have to abstain from using diff view
```suggestion
[source,cpp]
```

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,37 +1,235 @@
...
+  DynamicIntArray a{128, 0};
 
-  A a2 = std::move(a);
-  a2.fun();
+  int x = a.getValueAt(0); // Compliant: remove the move and use original object.
+  return x;
+}
+----
+
+[source,cpp,diff-id=1,diff-type=compliant]
```
necto, line 163:
```suggestion
[source,cpp]
```

rules/S5272/cfamily/rule.adoc
```diff
@@ -1,37 +1,235 @@
...
+int foo() {
+  DynamicIntArray a{128, 0};
+
+  DynamicIntArray a2 = std::move(a);
+  // Compliant: use of `a` removed.
+  return 9001;
+}
+----
+
+[source,cpp,diff-id=1,diff-type=compliant]
```
necto, line 174:
```suggestion
[source,cpp]
```

necto (changes requested):
Note that you don't need to use the heavy \`\`++`` syntax if the simple \`\` works.
Please update the "quickfix" status also

rules/S5272/cfamily/rule.adoc
```diff
@@ -2,74 +2,45 @@ Relying on an object that has been moved-from may lead to undefined behavior.
...
-== What is the potential impact?
-
-Using an object after it has been moved-from leads to *undefined behavior*.
-
-For programs that comprise undefined behavior, the compiler is no longer bound to the language specification.
-The application may crash or, even worse, the application appears to execute correctly while losing data or producing incorrect results, for instance.
+Move-assignment operators behave analogously, except that they are used once the object that is moved-to has already been constructed.
+In contrast to copy-assignment operators, a move-assignment operator too "steals" the moved-from object's resources without the need for making a potentially expensive copy.
 
+The ``++DynamicIntArray++`` type defined in the following manages a resource, i.e., memory, and shall serve as an example that showcases how move operations "steal" another objects resources and how they differ from making copies.
```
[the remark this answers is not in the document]
pdschbrt: That is a good idea---sold!

pdschbrt:
Just addressed the next round of feedback, @necto.

rules/S5272/cfamily/rule.adoc
```diff
@@ -96,29 +114,21 @@ int main() {
...
-There are some C++ standard template library types such as `std::unique_ptr` for which the moved-from state is fully specified.
-
-
 == How to fix it
 
-Do not use objects after they have been moved-from such as objects that have been used to move construct others or objects that have been move assigned to other objects.
+There are several potential ways to fix this issue, depending on the concrete situation.
+Potential fixes include:
+
+* Use the object _before_ moving it
```
necto, line 129:
```suggestion
* Use the object _only before_ moving it
```

rules/S5272/cfamily/rule.adoc
```diff
@@ -127,6 +137,15 @@ Do not use objects after they have been moved-from such as objects that have bee
 
 [source,cpp,diff-id=1,diff-type=noncompliant]
 ----
+#include <memory>
```
necto, line 140: Wasn't the whole point of hoisting the definition of `DynamicIntArray` into the section above to shorten these examples?
I presume you've inserted it here to make the code compile, but that is not a requirement. I think it is fine if the code compiles when you concatenate the code snippet with the definition of `DynamicIntArray` and a (non)compliant example

rules/S5272/cfamily/rule.adoc
necto: I see that you added a shortened definition of `DynamicIntArray`, but that wasn't the only reason why this fails to compile. The other reason is more important: you can't pass an lvalue to `array_user` that takes an rvalue reference. So you have to fix this example differently. For example, create a temporary copy `array_user(DynamicArray{a})`.

rules/S5272/cfamily/rule.adoc
pdschbrt: I've reverted the change(s).

pdschbrt:
I've just addressed your comments, @necto.

necto (approved):

PR #2757, opened 2023-08-02
Modify rule S5489: modernize example

amelie-renard-sonarsource:


balazs-benics-sonarsource (approved):
LGTM;
Do you plan to sync this with sonar-cpp?

PR #2769, opened 2023-08-03
Modify rule S1700: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone


guillaume-dequenne-sonarsource (changes requested):
I am under the impression that the code examples are written in idiomatic Java and are not really accurate for many of the languages that this rule cover. I think they should be updated as part of the migration.

Given the number of languages involved, I'd suggest dropping this from the current sprint and let it be tackled by a future cross-squad effort to migrate shared rules.

PR #2776, opened 2023-08-03
Modify rule S1265: LaYC format

amelie-renard-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1265/cfamily/rule.adoc
+++ b/rules/S1265/cfamily/rule.adoc
@@ -12,9 +12,9 @@ By defining the `operator delete` along with the `operator new`, the memory is d
 
 === What is the potential impact?
 
-Overriding both operators will ensure that your program does not consume more and more memory over time, eventually leading to a crash. 
+Overriding only one of the two operators may result in your program consuming more and more memory over time, eventually leading to a crash. 
 
-Deallocating memory that was not allocated results in undefined behavior.
+Deallocating memory that was not allocated with the corresponding strategy results in undefined behavior.
 
 == How to fix it
 
--- a/rules/S1265/cfamily/rule.adoc
+++ b/rules/S1265/cfamily/rule.adoc
@@ -82,7 +82,6 @@ private:
 void f() {
     MyClass* obj = new MyClass();
     delete obj; // It will call MyClass::delete to correctly deallocate the memory
-    return 0;
 }
 ----
 
--- a/rules/S1265/cfamily/rule.adoc
+++ b/rules/S1265/cfamily/rule.adoc
@@ -1,20 +1,95 @@
 == Why is this an issue?
 
-A class can override the `operator new` when it needs to customize memory allocation.
-Not overriding the `operator delete` method means relying on the program to release the allocated memory automatically. It is unlikely that the program will know how to do it correctly. To ensure it does, the class must override the `operator delete` method in a complementary manner.
+The `operator new` allocates memory for objects, and the `operator delete` frees the memory allocated by the matching `operator new`. When a class needs to customize memory allocation, it can override the `operator new` to use a custom memory allocation strategy.
 
-[source,cpp]
+However, it is important to override the `operator delete` when the `operator new` is overridden because the compiler's default implementation of the `operator delete` provided may not be suitable for the custom memory allocation strategy used by the `operator new`. By defining the `operator delete` along with the `operator new`, the user can ensure that the memory is deallocated in a way consistent with the custom allocation strategy used by the `operator new`.
+
+For instance, if the `operator new` function draws memory from a preallocated buffer instead of allocating memory, the `operator delete` function should not call the `free` function to release the memory.
+
+On the other hand, if the `operator delete` is overridden without overriding the `operator new`, it is suspicious because it may not correctly release the memory allocated by the default `operator new`.
+
+=== What is the potential impact?
+
+Overriding both operators will ensure that your program does not consume more and more memory over time, eventually leading to a crash. Otherwise, memory leaks or memory corruption will result.
+Deallocating memory that was not allocated results in undefined behavior.
+
+== How to fix it
+
+Each override of the `operator new` should have a matching overridden `operator delete` and vice versa.
+
+=== Code examples
+
+Let's imagine there is a custom allocator `MyAllocator` which is defined like this:
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-class AirPlane {
+class MyAllocator {
 public:
-  void* operator new(size_t size);
-  void operator delete(void* deadObject, size_t size); // Compliant: it releases the memory allocated in operator new()
+    void* allocate(size_t size) {
+        void* p = malloc(size);
+        return p;
+    }
+
+    void deallocate(void* p) {
+        free(p);
+    }
 };
 ----
 
-Overriding both operators will ensure that your program does not consume more and more memory over time, eventually leading to a crash. Otherwise, memory leaks or memory corruption will result.
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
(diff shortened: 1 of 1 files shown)


rules/S1265/cfamily/rule.adoc
marco-antognini-sonarsource, line 1: Your rewrite is already an improvement.
But the implementation is wrong (CPP-2429) for the same reason the description is lacking precision: there are many possible `operator new` and `operator delete`, and each should have the corresponding one.

To be fair, neither the C++ CG nor CERT articles describe what the corresponding operator is, and this is not trivial. C++ reference is a bit better in that area, but the wording is not easy to follow because the elements are on different pages. See https://en.cppreference.com/w/cpp/language/delete and https://en.cppreference.com/w/cpp/memory/new/operator_delete for example.

There are essentially two things I would improve in the description:
 * Why is it an issue to rely on the default new/delete when one of them is defined? You give a hint (the program will not do it correctly). Being more explicit would dissipate misunderstanding in this complex topic. E.g., if the impl of `new` uses `malloc`, the `free` must be used in `delete`; if `new` draw memory from a preallocated buffer, `delete` should not call `free`, ...
 * Which `delete` overload corresponds to a given `new` overload?

---

I wonder if we should put this on hold until the FPs are fixed (CPP-2429).
WDYT?

marco-antognini-sonarsource (changes requested):

rules/S1265/cfamily/rule.adoc
amelie-renard-sonarsource: If I understand correctly, the rule does not raise FP but has FN when the new operator and the delete operator it finds do not correspond. Or as soon as one new operator and one delete operator are defined.
I don't think this should prevent the rule from being improved. It would help the user to define the new or delete operator missing correctly.
I'll work on it (it seems so easy... 😅 )

tomasz-kaminski-sonarsource:
I think it may be worth to link the updated symbolic-execution rule as related to this one: https://sonarsource.github.io/rspec/#/rspec/S1232. Or take same inspiration from it.

rules/S1265/cfamily/rule.adoc
```diff
@@ -1,21 +1,99 @@
 == Why is this an issue?
 
-A class can override the `operator new` when it needs to customize memory allocation.
-Not overriding the `operator delete` method means relying on the program to release the allocated memory automatically. It is unlikely that the program will know how to do it correctly. To ensure it does, the class must override the `operator delete` method in a complementary manner.
+The `operator new` allocates memory for objects, and the `operator delete` frees the memory allocated by the matching `operator new`. When a class needs to customize memory allocation, it can override the `operator new` to use a custom memory allocation strategy and override the `operator delete` accordingly.
 
-[source,cpp]
+If the `operator delete` is not overridden alongside the `operator new`, the program will call its default implementation. And this latter may not be suitable for the custom memory allocation strategy used by the overridden `operator new`.
```
marco-antognini-sonarsource, line 5: While grammatically correct, I found "this latter" hard to read for some reason.
A simple suggestion, but feel free to dismiss it.
```suggestion
If the `operator delete` is not overridden alongside the `operator new`, the program will call its default implementation and may not be suitable for the custom memory allocation strategy used by the overridden `operator new`.
```

rules/S1265/cfamily/rule.adoc
```diff
@@ -1,21 +1,99 @@
...
+
+For instance, if the `operator new` draws memory from a preallocated buffer instead of allocating memory, the `operator delete` should not call the `free` function to release the memory. On the contrary, if the `operator new` allocate memory with `malloc`, the `operator delete` must call `free`.
```
marco-antognini-sonarsource, line 7: I think "on the contrary" does not 100% fit here. What about "reciprocally"?

```suggestion
For instance, if the `operator new` draws memory from a preallocated buffer instead of allocating memory, the `operator delete` should not call the `free` function to release the memory. Reciprocally, if the `operator new` allocate memory with `malloc`, the `operator delete` must call `free`.
```

rules/S1265/cfamily/rule.adoc
```diff
@@ -1,21 +1,99 @@
...
+
+On the other hand, if the `operator delete` is overridden without overriding the `operator new`, it is suspicious because it may not correctly release the memory allocated by the default `operator new`.
+
+By defining the `operator delete` along with the `operator new`, the memory is deallocated in a way consistent with the custom allocation strategy used by the `operator new`.
+
+=== What is the potential impact?
+
+Overriding both operators will ensure that your program does not consume more and more memory over time, eventually leading to a crash. 
+
+Deallocating memory that was not allocated results in undefined behavior.
```
marco-antognini-sonarsource, line 17: The first sentence is "positive" (this should be done), while the second is "negative" (don't do that). This makes it slightly hard to follow.

Given we are talking about the defect, I wonder if we should prefer the negative style.

Also, small nit in the second sentence: in a way, all memory is allocated in some way (heap, stack, buffer, ...). What matters is the two new/delete match.

I tried to make a suggestion to address these points.

```suggestion
Overriding only one of the two operators may result in your program consuming more and more memory over time, eventually leading to a crash. 

Deallocating memory that was not allocated with the corresponding strategy results in undefined behavior.
```

rules/S1265/cfamily/rule.adoc
```diff
@@ -1,21 +1,99 @@
...
+
+private:
+    static MyAllocator allocator;
+};
+
+void f() {
+    MyClass* obj = new MyClass();
+    delete obj; // It will call the default implementation of the operator delete
+    // and this latter will not call the MyAllocator::deallocate function to correctly release the memory
+    return 0;
```
marco-antognini-sonarsource, line 61: Function is void.

```suggestion
```

rules/S1265/cfamily/rule.adoc
```diff
@@ -1,21 +1,99 @@
...
+    }
+
+private:
+    static MyAllocator allocator;
+};
+
+void f() {
+    MyClass* obj = new MyClass();
+    delete obj; // It will call MyClass::delete to correctly deallocate the memory
+    return 0;
```
marco-antognini-sonarsource, line 86:
```suggestion
```

rules/S1265/cfamily/rule.adoc
```diff
@@ -1,21 +1,99 @@
...
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-class AirPlane {
+class MyAllocator {
 public:
-  void* operator new(size_t size);
-  void operator delete(void* deadObject, size_t size); // Compliant: it releases the memory allocated in operator new()
+    void* allocate(size_t size) {
+        void* p = malloc(size);
+        return p;
```
marco-antognini-sonarsource, line 33: The operators new below are not standard compliant because they should throw on failure.

This should fix the issue:
```suggestion
        if (!p) {
            throw std::bad_alloc();
        }
        return p;
```

rules/S1265/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks, looks much better already! Let's leave the second point to CPP-2429.

marco-antognini-sonarsource (approved):
Looks good, I have a couple of small points. No need for another round unless you want to.

rules/S1265/cfamily/rule.adoc
amelie-renard-sonarsource: Good point, thanks!

rules/S1265/cfamily/rule.adoc
amelie-renard-sonarsource:
```suggestion
If the `operator delete` is not overridden alongside the `operator new`, the program will call its default implementation, which may not be suitable for the custom memory allocation strategy used by the overridden `operator new`.
```

PR #2777, opened 2023-08-03
Modify rule S6476: LaYC format

nils-werner-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6476/docker/rule.adoc
+++ b/rules/S6476/docker/rule.adoc
@@ -1,17 +1,45 @@
 == Why is this an issue?
 
-The Dockerfile instructions are not case-sensitive. However, they should be in upper case to distinguish them from arguments more easily. Shared coding conventions allow teams to collaborate effectively.
+While Dockerfile instructions are not case-sensitive, adhering to uppercase conventions for instructions helps enhance clarity and collaboration within development teams.
+This ensures that instructions are more easily distinguishable from arguments and contributes to effective collaboration.
 
-=== Noncompliant code example
+=== What is the potential impact?
 
-[source,docker]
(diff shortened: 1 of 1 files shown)


rules/S6476/docker/rule.adoc
```diff
@@ -1,17 +1,45 @@
...
+Inconsistencies in letter casing can lead to errors that are difficult to diagnose.
+Imagine spending hours trying to locate a bug only to realize that it resulted from a misinterpretation of an instruction due to inconsistent casing.
+Such situations can be frustrating and can slow down the development process.
+
+Additionally, modern IDEs provide syntax highlighting to help programmers navigate and comprehend code.
+However, IDEs can sometimes fail to highlight lowercase or mixed-case Dockerfile keywords properly, as they typically expect instructions to be in uppercase.
+
+In essence, the inconsistent use of uppercase letters in Dockerfile instructions might seem like a minor concern, but its impact reaches beyond aesthetics.
+It affects comprehension, teamwork, error detection, and the overall development lifecycle.
+
```
rudy-regazzoni-sonarsource, line 24: Another argument that we could add, is that it's also a convention that is encouraged by the [official documentation](https://docs.docker.com/engine/reference/builder/):

> The instruction is not case-sensitive. However, convention is for them to be UPPERCASE to distinguish them from arguments more easily.

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my comment.

rules/S6476/docker/rule.adoc
nils-werner-sonarsource: WDYT about adding the documentation as a Resource?

PR #2778, opened 2023-08-03
Modify rule S1845: LaYC format

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1845/python/rule.adoc
+++ b/rules/S1845/python/rule.adoc
@@ -1,4 +1,4 @@
-This rule raises an issue when there is a method and a field in a ``++class++`` that differ only by capitalization.
+This rule raises an issue when there is a method and a field in a class with names that differ only by capitalization.
 
 == Why is this an issue?
 
--- a/rules/S1845/python/rule.adoc
+++ b/rules/S1845/python/rule.adoc
@@ -1,13 +1,16 @@
+This rule raises an issue when there is a method and a field in a ``++class++`` that differ only by capitalization.
+
 == Why is this an issue?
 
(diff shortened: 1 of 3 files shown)


rules/S1845/python/rule.adoc
```diff
@@ -1,23 +1,26 @@
+This rule raises an issue when there is a method and a field in a ``++class++`` that differ only by capitalization.
```
guillaume-dequenne-sonarsource, line 1: I have a feeling we are missing to mention we're referring about the name of the field/method. 
What about:
This rule raises an issue when there is a method and a field in a class with names that differ only by capitalization.

I'm also unsure we need to highlight "class" here, since we don't highlight other nouns like "method" or "field".

rules/S1845/swift/rule.adoc
```diff
@@ -1,10 +1,14 @@
+This rule raises an issue when there is a method and a field in a ``++class++``, ``++struct++``, ``++enum++``, or ``++extension++`` that differ only by capitalization.
```
guillaume-dequenne-sonarsource, line 1: Here the highlighting makes sense though I think, since you're referring to multiple structures of the language.
I'd still add the "with names" part, though.

guillaume-dequenne-sonarsource (changes requested):
Looks good, some minor rewording suggestions.

guillaume-dequenne-sonarsource (approved):
LGTM

jean-jimbo-sonarsource (approved):

PR #2781, opened 2023-08-03
Modify rule S6321: Add language AzureResourceManager (Bicep)

petertrr:


rules/S6321/azureresourcemanager/rule.adoc
```diff
@@ -58,6 +72,20 @@ include::../common/how-to-fix-it/intro.adoc[]
 }
 ----
 
+[source,bicep,diff-id=2,diff-type=compliant]
+----
+resource securityRules 'Microsoft.Network/networkSecurityGroups/securityRules@2022-11-01' = {
+  name: 'noncimpliantSecurityRules'
```
mstachniuk, line 78: Name is misleading. I would suggest using something "neutral" here like `securityRule`,  `rule`, `rule1`, etc.
It should not be marked in diff as a diff because it can mislead users. Please also use the same name in line 41.

mstachniuk (approved):
LGTM, and please see my comment

PR #2787, opened 2023-08-04
Modify rule S5283: Adapt to LaYC

tomasz-kaminski-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S5283/cfamily/metadata.json
+++ b/rules/S5283/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "Size of variable length arrays should be positive",
+  "title": "Size of variable length arrays should be greater than zero",
   "type": "CODE_SMELL",
   "status": "ready",
   "remediation": {
@@ -22,9 +22,15 @@
   "ruleSpecification": "RSPEC-5283",
   "sqKey": "S5283",
   "scope": "Main",
+  "securityStandards": {
+    "CERT": [
+      "ARR32-C."
+    ]
+  },
+
   "defaultQualityProfiles": [
     "Sonar way",
     "MISRA C++ 2008 recommended"
   ],
-  "quickfix": "unknown"
+  "quickfix": "infeasible"
 }
--- a/rules/S5283/cfamily/rule.adoc
+++ b/rules/S5283/cfamily/rule.adoc
@@ -21,7 +21,7 @@ void merge(int* tab, int size, int first) {
 The syntax follows the one used by an array with static size (where size is constant),
 and the value placed in the square brackets (`[size]` in the above example) determines the size of the array.
 For both static and runtime-sized arrays, the size is required to be greater than zero.
-However, in the case of VLAs, this cannot be verified at compile-time, 
+However, in the case of VLAs, this cannot be checked by compiler, 
 and any program that creates such an array with a size that has a value of zero or is negative, 
 has undefined behavior.
 
@@ -87,12 +87,12 @@ before using it as size for VLA.
 
 Zero size VLA is created due of by-one error.
 
-[source,cpp,diff-id=1,diff-type=noncompliant]
+[source,c,diff-id=1,diff-type=noncompliant]
 ----
 void zeroSize() {
   int array[5];
   for (int i = 0; i <= 5; ++i) {
-    int buffer[5 - i]; // Noncompliant: buffer size is zero for i == 5
+    int buffer[5 - i]; // Noncompliant: buffer size is zero for last iteration
     /* ... */
   }
 }
@@ -100,7 +100,7 @@ void zeroSize() {
 
 ==== Compliant solution
 
-[source,cpp,diff-id=1,diff-type=noncompliant]
+[source,c,diff-id=1,diff-type=compliant]
 ----
 void zeroSize() {
   int array[5];
@@ -114,7 +114,7 @@ void zeroSize() {
 
 ==== Noncompliant code example
 
-[source,cpp,diff-id=1,diff-type=noncompliant]
+[source,c,diff-id=3,diff-type=noncompliant]
 ----
 int taintedSize() {
   int size = 0;
@@ -129,7 +129,7 @@ int taintedSize() {
 
 Validate value of `size` before using as size of array.
 
-[source,cpp,diff-id=1,diff-type=noncompliant]
+[source,c,diff-id=3,diff-type=compliant]
 ----
 int taintedSize() {
   int size = 0;
@@ -145,17 +145,35 @@ int taintedSize() {
 ----
 
 
-=== Pitfalls
+== Going the extra mile
 
-// Review question: mentions that large VLA may lead to stack overflow, and there is no way for a program to reject
-// allocation. In this case, using a heap may be better. However, this is not related to a problem in the rule.
-// and does not avoid it.
+Variable lenght arrays are allocated on stack, so in situation when a large value of the size is used,
+creating such array may lead to stack overflow and undefined behavior:
+ 
+[source,c]
+----
+void largeVLA() {
+  usinged int s = INT_MAX;
+  int vla[s][100]; // requires allocation of the INT_MAX * 100
+}
+----
+
+In addition, the language does not provide a way to query available stack space, nor possiblity of reporting failure in creation of such array.
+
+When applicable, it is recommended to replace the VLA with heap allocated memory. 
+In constrast to VLA, heap allocation functions reports in situation whe sufficient memory cannot be provided, by returning `NULL` or throwing an exception (in {cpp}).
+Furthermore, {cpp} standard library provides container like `std::vector`, that handles managment of such heap allocated memory.
+
+Moreover, the C language standared makes suport for declaring VLA variable on stack optional (with `__STDC_NO_VLA__`),
+and C++ standard does not support them, however they are commonly supported as extension).
+
 
 == Resources
 
 === Documentation
 
 * {cpp} reference - https://en.cppreference.com/w/c/language/array#Variable-length_arrays[Variable length arrays]
+* {cpp} reference - https://en.cppreference.com/w/cpp/container/vector[``std::vector``]
 
 === Standards
 
--- a/rules/S5283/cfamily/rule.adoc
+++ b/rules/S5283/cfamily/rule.adoc
@@ -21,7 +21,7 @@ void merge(int* tab, int size, int first) {
 The syntax follows the one used by an array with static size (where size is constant),
 and the value placed in the square brackets (`[size]` in the above example) determines the size of the array.
 For both static and runtime-sized arrays, the size is required to be greater than zero.
-However, in the case of VLAs, this cannot be checked by compiler, 
+However, in the case of VLAs, this cannot be checked by the compiler, 
 and any program that creates such an array with a size that has a value of zero or is negative, 
 has undefined behavior.
 
@@ -34,8 +34,8 @@ void example() {
 }
 ----
 
-This defect may also manifest, when variable used as size is not inintialized, and it's value is not determinstic.
-This may result in negative or zero value being read from it in case of some runs of the program.
+This defect may also manifest, when the variable used as size is not initialized, and its value is not deterministic.
+This may result in a negative or zero value being read from it in case of some runs of the program.
 
 [source,c]
 ----
@@ -147,8 +147,8 @@ int taintedSize() {
 
 == Going the extra mile
 
-Variable lenght arrays are allocated on stack, so in situation when a large value of the size is used,
-creating such array may lead to stack overflow and undefined behavior:
+Variable length arrays are allocated on the stack, so in situations when a large value of the size is used,
+creating such an array may lead to stack overflow and undefined behavior:
  
 [source,c]
 ----
@@ -158,14 +158,14 @@ void largeVLA() {
 }
 ----
 
-In addition, the language does not provide a way to query available stack space, nor possiblity of reporting failure in creation of such array.
+In addition, the language does not provide a way to query available stack space, nor the possibility of reporting failure in the creation of such an array.
 
-When applicable, it is recommended to replace the VLA with heap allocated memory. 
-In constrast to VLA, heap allocation functions reports in situation whe sufficient memory cannot be provided, by returning `NULL` or throwing an exception (in {cpp}).
-Furthermore, {cpp} standard library provides container like `std::vector`, that handles managment of such heap allocated memory.
+When applicable, it is recommended to replace the VLA with heap-allocated memory. 
+In contrast to VLA, heap allocation functions report in a situation when sufficient memory cannot be provided, by returning `NULL` or throwing an exception (in {cpp}).
+Furthermore, {cpp} standard library provides a container like `std::vector`, that handles management of such heap allocated memory.
 
-Moreover, the C language standared makes suport for declaring VLA variable on stack optional (with `__STDC_NO_VLA__`),
-and C++ standard does not support them, however they are commonly supported as extension).
+Moreover, the C11 language standard made support for declaring VLA on stack optional (with `__STDC_NO_VLA__`),
+and C++ standard does not support them, however, they are commonly supported as extensions.
 
 
 == Resources
--- a/rules/S5283/cfamily/rule.adoc
+++ b/rules/S5283/cfamily/rule.adoc
@@ -1,8 +1,8 @@
-Creating a Variable Length Array (VLA) with the size that is not greater than zero leads to undefined behavior.
+Creating a Variable Length Array (VLA) with a size that is not greater than zero leads to undefined behavior.
 
 == Why is this an issue?
 
-Variable length arrays are used to allocate an stack size for the number of elements that is know at the runtime,
+Variable length arrays are used to allocate a stack size for the number of elements that are known at the runtime,
 for example: 
 
 [source,c]
@@ -14,11 +14,11 @@ void merge(int* tab, int size, int first) {
 }
 ----
 
-The syntax follows the one used by array with static size (where size is constant),
-and the value placed in the square brackets (`[size]` in above example) determine the size of the array.
-For both static and runtime sized array, a size is required to be greater than zero,
-however in case of VLAs this cannot be verified at compile-time, 
-and any program that creates such array with size that has a value of zero or is negative, 
+The syntax follows the one used by an array with static size (where size is constant),
+and the value placed in the square brackets (`[size]` in the above example) determines the size of the array.
+For both static and runtime-sized arrays, the size is required to be greater than zero,
+however, in the case of VLAs, this cannot be verified at compile-time, 
+and any program that creates such an array with a size that has a value of zero or is negative, 
 has undefined behavior.
 
 [source,c]
@@ -30,21 +30,20 @@ void example() {
 }
 ----
 
-This defect may also manifest, when variable used as size is not inintialized, and it's value is not determinstic.
-This may result in negative or zero value being read from it in case of some runs of the program.
+This defect may also manifest when a variable used as size is not initialized and its value is not deterministic.
+This may result in a negative or zero value being read from it in case of some program runs.
  
 [source,c]
 ----
 void uninitialized() {
   int s;      // uninitialized, the value is not determined
-  int vla[s]; // program compiles, have undefined behavior if value read from s is negative or zero
+  int vla[s]; // program compiles, have undefined behavior if the value read from s is negative or zero
 }
 ----
 
 
-A non-positive size of the VLAs may also be result of using a value that was loaded by the program from file,
-or other external resource, is used as a size, without previous validation.
-Such values as usually referred as being _tainted_.
+A non-positive size of the VLAs may also result from using a value loaded by the program from a file or other external resources used as size without previous validation.
+Such values as usually referred to as being _tainted_.
 
 [source,c]
 ----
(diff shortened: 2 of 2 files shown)


rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
-  int n = 0;
-  int a[n]; // Noncompliant; array of zero size
-}
 ----
 
+=== Pitfalls
+
+// Review question: mentions that large VLA may lead to stack overflow, and there is no way for a program to reject
+// allocation. In this case, using a heap may be better. However, this is not related to a problem in the rule.
+// and does not avoid it.
```
balazs-benics-sonarsource, line 100: Maybe then this should be in "extra mile"?
I definitely think it should be mentioned.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+
+== Resources
+
+=== Documentation
+
+* C++ reference - https://en.cppreference.com/w/c/language/array#Variable-length_arrays[Variable length arrays]
```
balazs-benics-sonarsource, line 106:
```suggestion
* {cpp} reference - https://en.cppreference.com/w/c/language/array#Variable-length_arrays[Variable length arrays]
```
I think I've seen it written this way.

However, according to grep, it's not entirely consistent.
We have:
```
{cpp} reference - https  <--- most frequent
CPP reference - https
C++ reference - https
CPP reference/C++ reference - https
C{plus}{plus} reference - https
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
+Creating a Variable Length Array (VLA) with a size that is not greater than zero leads to undefined behavior.
```
balazs-benics-sonarsource, line 1: Shouldn't this be the same as the metadata title?

rules/S5283/cfamily/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Size of variable length arrays should be positive",
+  "title": "Size of variable length arrays should be greater than zero",
```
balazs-benics-sonarsource, line 2: Why did you decide to change this?

balazs-benics-sonarsource (changes requested):
minor remarks

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+Variable length arrays are used to allocate a stack size for the number of elements that are known at the runtime,
+for example: 
+
+[source,c]
+----
+void merge(int* tab, int size, int first) {
+  int res[size]; // allocate buffer for merging on stack
+  // merge the ranges [tab, tab + first) and [tab + first, tab + last) into res
+  memcpy(tab, res, size * sizeof(int));
+}
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Ah, I see. I mean the comment to say that there is code here that perform merging. I will try to reformat that.

rules/S5283/cfamily/metadata.json
tomasz-kaminski-sonarsource: To use simpler, more understandable English when possible.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This first setence in GUI is displayed just under the title. So usually this should be a quick explanation of the issue, and for someone who understands the topic should be enough to know what should be fixed.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+* memory corruption and data losses, when the program overwrites bytes responsible for storing data or executable code,
+* stack overflow, when negative size is interpreted as a large positive number.
+
+Furthermore, in a situation when VLA size is dependent on the user input, it can lead to vulnerabilities in the program. 
+
+== How to fix it
+
+Ensure that the size of the variable length array is always greater than zero.
+In case of value that depends on user input (is _tainted_), include appropriate checks to ensure that value is positive
+before using it as size for VLA.
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: VLAs avoid memory allocations, and thus can be used in embedded. Also they avoid any allocations. I would be not comfortable with saying that if you decide to use them, vector is viable alternative.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+
+
+A non-positive size of the VLAs may also result from using a value loaded by the program from a file or other external resources used as size without previous validation.
+Such values as usually referred to as being _tainted_.
+
+[source,c]
+----
+void loadFromInput() {
+  int size;
+  scanf("%d", &size); // loads size from input
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Good catch.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+  char bytes[size]; // size may be negative
+  /* ... */  
+}
+----
 
-=== Noncompliant code example
+The above code will lead to undefined behavior when the value of the `size` load from the input is not greater than zero.
+This may happen due to source data corruption, accidental user mistake, or malicious action.
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Reading corrupted data, may be caused by both source data corruption (you are reading the right files, but they got accidentally corrupted), or user mistake, where the wrong file was given. I was using source data corruption, to indicate that you program may stop working, because the file got accidentally changed.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Hmm, I think I see now where you were going with using the extra mile here, will try to do that.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I will change it to {cpp} reference.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+void loadFromInput() {
+  int size;
+  scanf("%d", &size); // loads size from input
+  char bytes[size]; // size may be negative
+  /* ... */  
+}
+----
 
-=== Noncompliant code example
+The above code will lead to undefined behavior when the value of the `size` load from the input is not greater than zero.
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: I think the ``++xx++`` syntax was used during automatic migration, as it is more bullet proof, but we prefer to use `xx` when this just works.

rules/S5283/cfamily/rule.adoc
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Updated

rules/S5283/cfamily/rule.adoc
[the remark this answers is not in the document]
balazs-benics-sonarsource: I was referring to double backticks instead of single backticks. I was not referring to double backticks and double plus signs.
AFAIK the double plus sign makes the text emphasized. That is in combination with the inline code marker, results in emphasized inline-code AFAIK.

According to the [ref](https://docs.asciidoctor.org/asciidoc/latest/text/monospace/#unconstrained), it seems like single and double backticks both result in _monoscaped_ font, but the single backtick version only works if it's not in the middle of the word.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,24 +1,166 @@
...
+==== Noncompliant code example
+
+Zero size VLA is created due of by-one error.
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+void zeroSize() {
+  int array[5];
+  for (int i = 0; i <= 5; ++i) {
+    int buffer[5 - i]; // Noncompliant: buffer size is zero for i == 5
```
balazs-benics-sonarsource, line 95:
```suggestion
    int buffer[5 - i]; // Noncompliant: buffer size is zero for the 5th iteration
```
Maybe even "last" iteration, because that is an even stronger statement.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,24 +1,166 @@
...
+----
+
+==== Compliant solution
 
-void f2() {
-  int n = 0;
-  int a[n]; // Noncompliant; array of zero size
+Validate value of `size` before using as size of array.
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
```
balazs-benics-sonarsource, line 132:
```suggestion
[source,cpp,diff-id=2,diff-type=compliant]
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,24 +1,166 @@
...
+  for (int i = 0; i <= 5; ++i) {
+    int buffer[5 - i]; // Noncompliant: buffer size is zero for i == 5
+    /* ... */
+  }
+}
+----
+
+==== Compliant solution
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
```
balazs-benics-sonarsource, line 103:
```suggestion
[source,cpp,diff-id=1,diff-type=compliant]
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,24 +1,166 @@
...
+    int buffer[5 - i]; // Complaint
+    /* ... */
+  }
+}
+----
+
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
```
balazs-benics-sonarsource, line 117:
```suggestion
[source,cpp,diff-id=2,diff-type=noncompliant]
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,24 +1,166 @@
...
+  int array[5];
+  for (int i = 0; i < 5; ++i) {
+    int buffer[5 - i]; // Complaint
+    /* ... */
+  }
+}
+----
+
+
+==== Noncompliant code example
```
balazs-benics-sonarsource, line 115: Is it deliberate that this section title is mentioned for the second time?
I thought we should have a single noncompliant section, where we could have multiple code snippets, and a single "compliant" section, where the correct code snippets are.

balazs-benics-sonarsource (changes requested):

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Yes. I understood that this is the case now, when we pair examples by repeating complaints and non-compliant solutions. I.e. each orthogonal example can have separate pair.
We already done so here: https://sonarsource.github.io/rspec/#/S5553

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,113 @@
...
+  int res[size]; // allocate buffer for merging on stack
+  // merge the ranges [tab, tab + first) and [tab + first, tab + last) into res
+  memcpy(tab, res, size * sizeof(int));
+}
+----
+
+The syntax follows the one used by an array with static size (where size is constant),
+and the value placed in the square brackets (`[size]` in the above example) determines the size of the array.
+For both static and runtime-sized arrays, the size is required to be greater than zero,
+however, in the case of VLAs, this cannot be verified at compile-time,
```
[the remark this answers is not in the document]
tomasz-kaminski-sonarsource: Resolved a bit.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This is 6th iteration, not 5th, because of off by one error. I will use last.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Done. Let me know what you think.

tomasz-kaminski-sonarsource:
I think I have addressed everything. Now also all examples are in-place.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+[source,c,diff-id=2,diff-type=compliant]
+----
+int countGrapheneClusters(int codeUnits, int bytesInCodeUnit) {
+  int size;
+  if (bytesInCodeUnit > 0) {
+    size = codeUnits * bytesInCodeUnit;
+  } else {
+    return -1;
+  }
+  char buffer[size]; // Noncomplaint: size may not be initialized
```
balazs-benics-sonarsource, line 142: The comments suggest it's noncompliant, while the section should contain only compliant solutions.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+----
+
+==== Noncompliant code example
+
+[source,c,diff-id=3,diff-type=noncompliant]
+----
+int taintedSize() {
+  int size = 0;
+  scanf("%d", &size); // size value is loaded from input, value is tainted here
+  char bytes[size]; // Noncomplaint: size may be negative
```
balazs-benics-sonarsource, line 155:
```suggestion
  scanf("%d", &size); // size value is loaded from input, value is tainted here
  char bytes[size];   // Noncomplaint: size may be negative
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+
+Variable length arrays are allocated on the stack, so in situations when a large value of the size is used,
+creating such an array may lead to stack overflow and undefined behavior:
+ 
+[source,c]
+----
+void largeVLA() {
+  usinged int s = INT_MAX;
+  int vla[s][100]; // requires allocation of the INT_MAX * 100
+}
```
balazs-benics-sonarsource: I'm not sure about this code.
I'd suggest to use `int` for `s`. It would make it clear, that the subsequent multiplication for calculating the size of the array is not expected to wrap-around.
I'm not sure about the semantics, but seeing `unsigned` definitely caught my attention; which shouldn't be the point.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+----
+
+In addition, the language does not provide a way to query available stack space, nor the possibility of reporting failure in the creation of such an array.
+
+When applicable, it is recommended to replace the VLA with heap-allocated memory. 
+In contrast to VLA, heap allocation functions report in a situation when sufficient memory cannot be provided, by returning `NULL` or throwing an exception (in {cpp}).
+Furthermore, {cpp} standard library provides a container like `std::vector`, that handles management of such heap allocated memory.
```
balazs-benics-sonarsource, line 197:
```suggestion
Furthermore, the {cpp} standard library provides containers like `std::vector`, that manage the heap-allocated memory.
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+
+Moreover, the C11 language standard made support for declaring VLA on stack optional (with `__STDC_NO_VLA__`),
+and C++ standard does not support them, however, they are commonly supported as extensions.
```
balazs-benics-sonarsource, line 200:
```suggestion
Moreover, the C11 language standard and above only optionally supports VLAs (with `__STDC_NO_VLA__`),
and the {cpp} standard never supported it, however, they are commonly accepted as extensions.
```

It felt a bit too "passive" to me.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+
+
+== Resources
+
+=== Documentation
+
+* {cpp} reference - https://en.cppreference.com/w/c/language/array#Variable-length_arrays[Variable length arrays]
```
balazs-benics-sonarsource, line 207: Why is "Variable" in upper case?

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+Validate value of `size` before using as size of array.
+
+[source,c,diff-id=3,diff-type=compliant]
+----
+int taintedSize() {
+  int size = 0;
+  scanf("%d", &size);
+  if (size <= 0) {
+    return -1;
+  }
```
balazs-benics-sonarsource, line 172:
```suggestion
  if (size <= 0 || size > 1000) {
    return -1;
  }
```
For tainted values, it's recommended to check both lower and upper bounds.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,216 @@
...
+* memory corruption and data losses, when the program overwrites bytes responsible for storing data or executable code,
+* stack overflow, when negative size is interpreted as a large positive number.
+
+Furthermore, in a situation when VLA size is dependent on the user input, it can lead to vulnerabilities in the program. 
+
+== How to fix it
+
+Ensure that the size of the variable length array is always greater than zero.
+In case of value that depends on user input (is _tainted_), include appropriate checks to ensure that value is positive
+before using it as size for VLA.
```
balazs-benics-sonarsource, line 82: Maybe also add that it's not enough to ensure that it's greater than zero, but also that it shouldn't be too large. This is even more so important for tainted sizes.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,215 @@
+Creating a Variable Length Array (VLA) with a size that is not greater than zero leads to undefined behavior.
```
balazs-benics-sonarsource, line 1:
```suggestion
Creating a Variable Length Array (VLA) with a size that is not greater than zero is to undefined behavior.
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,215 @@
...
+  /* Code that merges two sorted ranges [tab, tab + first) and [tab + first, tab + last)
+   * into res buffer.
+   * ....
+   */
+  // copy merged data into input
+  memcpy(tab, res, size * sizeof(int));
+}
+----
+
+The syntax follows the one used by an array with static size (where size is constant),
```
balazs-benics-sonarsource, line 21:
```suggestion
The syntax follows the one used by an array with static size (where size is a constant),
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,215 @@
...
+[source,c]
+----
+void example() {
+  int s = -1;
+  int vla[s];  // program compiles, and have undefined behavior
+  int arr[-1]; // program is ill-formed
+}
+----
+
+This defect may also manifest, when the variable used as size is not initialized, and its value is not deterministic.
```
balazs-benics-sonarsource, line 37:
```suggestion
This defect may also manifest when the variable used as size is not initialized.
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,215 @@
...
+This may result in a negative or zero value being read from it in case of some runs of the program.
```
balazs-benics-sonarsource, line 38:
```suggestion
Uninitialized variables might have zero, negative, or any value at each time the program executes.
```

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,215 @@
...
+  /* ... */
+}
+----
+
+The above code will lead to undefined behavior when the value of the `size` load from the input is not greater than zero.
+This may happen due to source data corruption, accidental user mistake, or malicious action. 
+
+=== What is the potential impact?
+
+Creating a variable length array with a size smaller or equal to zero causes undefined behavior.
```
balazs-benics-sonarsource, line 67: This sentence is really similar to the very first line.
However, it uses the verb "causes", while in the headline it uses "leads to".
And I proposed already "is" for this.
We should consider harmonizing these.

rules/S5283/cfamily/rule.adoc
```diff
@@ -1,23 +1,215 @@
...
+=== Code examples
+
+==== Noncompliant code example
+
+Zero size VLA is created due of by-one error.
+
+[source,c,diff-id=1,diff-type=noncompliant]
+----
+void zeroSize() {
+  int array[5];
```
balazs-benics-sonarsource, line 93: Why do we need the "array" object in these examples?
I don't see them referred anywhere.

balazs-benics-sonarsource (changes requested):
Here is my in-depth review.
I'm sorry I didn't notice a lot of stuff in the previous round.
This time, it should be exhaustive.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Good catch. `unsigned` was leftover from previous attempt when I used `-1` to initialize.

tomasz-kaminski-sonarsource: Fixed

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This is title of the section on the page. I will add hypen for exact match,.

rules/S5283/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Removed

balazs-benics-sonarsource (approved):
I think it's really good now.
Please check each example to confirm we raise when we should, and we don't raise when we shouldn't.
Feel free to merge this.

PR #2789, opened 2023-08-04
Modify rule S100: Revert accidental change

martin-strecker-sonarsource:
Revert an accidental change introduced in #2713.


csaba-sagi-sonarsource (approved):
LGTM!

PR #2790, opened 2023-08-04
RULEAPI-788 Validate use of diff-id and diff-type

marco-antognini-sonarsource:
It seems we have quite a few rules (78) [misusing](https://cirrus-ci.com/task/5975346382635008?logs=asciidoc_tests#L90-L238) diff-id and diff-type examples.

```
#rule lang
   5 azureresourcemanager
   8 cfamily
   1 cloudformation
   7 csharp
   3 docker
  23 java
  25 javascript
   3 kotlin
  15 php
  15 python
   1 swift
   2 terraform
   1 vb6
   5 vbnet
```


ci/asciidoc_validation/custom-asciidoctor
```diff
@@ -73,9 +79,86 @@ class IncludeLogger < Asciidoctor::Extensions::IncludeProcessor
...
+  def process document
+    rule = get_rule(document)
+    source_blocks = document.find_by(context: :listing, style: 'source')
+
+    # Collect individually valid blocks per diff-id.
+    blocks_per_id = Hash.new { |hash, key| hash[key] = Array.new }
+
+    # Find blocks with only diff-id but no diff-type, or vice-versa, or invalid diff-type.
+    source_blocks.each do |block|
+      loc = get_source_location(block)
```
necto, line 106: As it is not used in the next 7 lines, you can put it after the first `if` block

ci_tests/asciidoc_validation/run_tests.sh
```diff
@@ -77,4 +77,22 @@ run_test "test_bad_cross_ref" \
   "S100 cross-references .*rules/S1000/bad.adoc" \
   "S1000 cross-references .*rules/S100/java/bad.adoc"
 
+run_test "test_diff_source" \
+  "ERROR: Diff sources are incorrectly used:" \
```
necto, line 81: "diff sources" is something new. I've heard the term "diff view" and "diff highlighting"
Also, I would inverse "incorrectly used", that sounds better subjectively, but I don't know if there is any English rule about it.
```suggestion
  "ERROR: Diff highlighting is used incorrectly:" \
```

necto (approved):
Nice work!

ci/asciidoc_validation/custom-asciidoctor
marco-antognini-sonarsource: Makes sense. Done.

PR #2804, opened 2023-08-04
Diff blocks: fix some incorrect use for php

marco-antognini-sonarsource:
Improvement identified in #2790.

Add a prefix to the diff-id when it is used multiple times in different "how to fix it in XYZ" sections to avoid ambiguity and pedantically follow the spec:

>  A single and unique diff-id should be used only once for each type of code example as shown in the description of a rule.

Obvious typos around `diff-type` were fixed. 

---

NB: this does not fix _all_ incorrect use of diff blocks, cf. CI log (once #2790 is merged).

### DIFF
--- a/rules/S2076/php/how-to-fix-it/symfony.adoc
+++ b/rules/S2076/php/how-to-fix-it/symfony.adoc
@@ -6,7 +6,7 @@ include::../../common/fix/code-rationale.adoc[]
 
 ==== Noncompliant code example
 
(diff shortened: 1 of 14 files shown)


jonas-wielage-sonarsource (approved):
LGTM. Regarding the diff-id: I'm not sure the prefix is needed, I believe the unique id refers to one rule and not the whole language / framework.

PR #2805, opened 2023-08-04
Diff blocks: fix incorrect use for swift

marco-antognini-sonarsource:
Improvement identified in #2790.

Add a prefix to the diff-id when it is used multiple times in different "how to fix it in XYZ" sections to avoid ambiguity and pedantically follow the spec:

>  A single and unique diff-id should be used only once for each type of code example as shown in the description of a rule.


hendrik-buchwald-sonarsource (approved):
LGTM, thanks for the fix!

PR #2813, opened 2023-08-07
SONARJAVA-4495 Update S2175 in line with LayC.

johann-beleites-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2175/java/rule.adoc
+++ b/rules/S2175/java/rule.adoc
@@ -1,51 +1,75 @@
 == Why is this an issue?
 
-The ``++java.util.Collection++`` API has methods that accept ``++Object++`` parameters such as ``++Collection.remove(Object o)++``, and ``++Collection.contains(Object o)++``. When the actual type of the object provided to these methods is not consistent with the type declared on the ``++Collection++`` instantiation, those methods will always return ``++false++`` or ``++null++``. This is most likely unintended and hides a design problem.
+The `java.util.Collection` type and its subtypes provide methods to access and modify collections such as
+`Collection.remove(Object o)` and `Collection.contains(Object o)`. Some of these methods accept arguments
+of type `java.lang.Object` and will compare said argument with objects already in the collection.
 
+If the actual type of the argument is unrelated to the type of object contained in the collection,
+these methods will always return `false`, `null`, or `-1`. This behavior is most likely unintended and
+can be indicative of a design issue.
 
-This rule raises an issue when the type of the argument of the following APIs is unrelated to the type used for the ``++Collection++`` declaration:
+This rule raises an issue when the type of the argument provided to one of the following methods is
+unrelated to the type used for the collection declaration:
 
-* ``++Collection.remove(Object o)++``
-* ``++Collection.removeAll(Collection<?>)++``
-* ``++Collection.contains(Object o)++``
-* ``++List.indexOf(Object o)++``
-* ``++List.lastIndexOf(Object o)++``
-* ``++Map.containsKey(Object key)++``
-* ``++Map.containsValue(Object value)++``
-* ``++Map.get(Object key)++``
-* ``++Map.getOrDefault(Object key, V defaultValue)++``
-* ``++Map.remove(Object key)++``
-* ``++Map.remove(Object key, Object value)++``
+* `Collection.remove(Object o)`
+* `Collection.removeAll(Collection<?>)`
+* `Collection.contains(Object o)`
+* `List.indexOf(Object o)`
+* `List.lastIndexOf(Object o)`
+* `Map.containsKey(Object key)`
+* `Map.containsValue(Object value)`
+* `Map.get(Object key)`
+* `Map.getOrDefault(Object key, V defaultValue)`
+* `Map.remove(Object key)`
+* `Map.remove(Object key, Object value)`
 
+== How to fix it
 
-=== Noncompliant code example
+Ask yourself what the purpose of this method call is. Check whether the provided argument and collection are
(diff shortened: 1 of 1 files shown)


rules/S2175/java/rule.adoc
```diff
@@ -1,51 +1,75 @@
 == Why is this an issue?
 
-The ``++java.util.Collection++`` API has methods that accept ``++Object++`` parameters such as ``++Collection.remove(Object o)++``, and ``++Collection.contains(Object o)++``. When the actual type of the object provided to these methods is not consistent with the type declared on the ``++Collection++`` instantiation, those methods will always return ``++false++`` or ``++null++``. This is most likely unintended and hides a design problem.
+The `java.util.Collection` type and its subtypes provide methods to access and modify collections such as
+`Collection.remove(Object o)` and `Collection.contains(Object o)`. Some of these methods accept arguments
+of type `java.lang.Object` and will compare said argument with objects already in the collection.
```
leonardo-pilastri-sonarsource, line 5:
```suggestion
The `java.util.Collection` type and its subtypes provide methods to access and modify collections such as `Collection.remove(Object o)` and `Collection.contains(Object o)`.
Some of these methods accept arguments of type `java.lang.Object` and will compare said argument with objects already in the collection.
```

rules/S2175/java/rule.adoc
```diff
@@ -1,51 +1,75 @@
...
 
+If the actual type of the argument is unrelated to the type of object contained in the collection,
+these methods will always return `false`, `null`, or `-1`. This behavior is most likely unintended and
+can be indicative of a design issue.
```
leonardo-pilastri-sonarsource, line 9:
```suggestion
If the actual type of the argument is unrelated to the type of object contained in the collection, these methods will always return `false`, `null`, or `-1`.
This behavior is most likely unintended and can be indicative of a design issue.
```

rules/S2175/java/rule.adoc
```diff
@@ -1,51 +1,75 @@
...
 
-This rule raises an issue when the type of the argument of the following APIs is unrelated to the type used for the ``++Collection++`` declaration:
+This rule raises an issue when the type of the argument provided to one of the following methods is
+unrelated to the type used for the collection declaration:
```
leonardo-pilastri-sonarsource, line 12:
```suggestion
This rule raises an issue when the type of the argument provided to one of the following methods is unrelated to the type used for the collection declaration:
```

rules/S2175/java/rule.adoc
```diff
@@ -1,51 +1,75 @@
...
+* `Map.getOrDefault(Object key, V defaultValue)`
+* `Map.remove(Object key)`
+* `Map.remove(Object key, Object value)`
 
+== How to fix it
 
-=== Noncompliant code example
+Ask yourself what the purpose of this method call is. Check whether the provided argument and collection are
+correct in this context and for the desired purpose. Remove unnecessary calls and otherwise provide an
+argument of which the type is compatible with the list content's type.
```
leonardo-pilastri-sonarsource, line 30:
```suggestion
Ask yourself what the purpose of this method call is.
Check whether the provided argument and collection are correct in this context and for the desired purpose.
Remove unnecessary calls and otherwise provide an argument of which type is compatible with the list content's type.
```

rules/S2175/java/rule.adoc
```diff
@@ -1,51 +1,75 @@
...
 ----
 
+[source,java,diff-id=2,diff-type=compliant]
+----
+void main(List<String> list, Integer integer) {
+  if (list.contains(integer.toString())) { // Compliant, 'integer' is converted to a String before checking if the list contains it.
+    // ...
+  }
+}
+----
```
leonardo-pilastri-sonarsource, line 67: It was mentioned in the document produced by the .NET squad that multiple code blocks might have issues rendering, and if possible we should prefer squashing code example into one `noncompliant` block and one `compliant` block. Something like:

```
[source,java,diff-id=1,diff-type=noncompliant]
----
void removeFromMap(Map<Integer, Object> map, String strKey) {
  map.remove(strKey); // Noncompliant, this call will remove nothing and always return 'null' because 'map' is handling only Integer keys and String cannot be cast to Integer.
}

void ifContains(List<String> list, Integer integer) {
  if (list.contains(integer)) { // Noncompliant; always false as the list only contains Strings, not integers.
    // ...
  }
}
----

==== Compliant solution
[source,java,diff-id=1,diff-type=compliant]
----
void removeFromMap(Map<Integer, Object> map, String strKey) {
  map.remove(Integer.parseInt(strKey)); // Compliant, strKey is parsed into an Integer before trying to remove it from the map.
}

void ifContains(List<String> list, Integer integer) {
  if (list.contains(integer.toString())) { // Compliant, 'integer' is converted to a String before checking if the list contains it.
    // ...
  }
}
----
```

leonardo-pilastri-sonarsource (changes requested):
Just some line breaks adjustment to facilitate visualizing future changes, and more compact code examples, everything else looks good!

rules/S2175/java/rule.adoc
johann-beleites-sonarsource: Oh really? Didn't know that, thanks. Sounds like a bug that should be fixed on product side as well.

leonardo-pilastri-sonarsource (approved):
Looks good!

necto:
@johann-beleites-sonarsource, for the future PRs, please make sure 

@leonardo-pilastri-sonarsource, I'll appreciate if you can also watch out for the violation of the guidelines in the PRs you review

PR #2817, opened 2023-08-07
SONARJAVA-4495 Update S2168 in line with LayC.

johann-beleites-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2168/java/rule.adoc
+++ b/rules/S2168/java/rule.adoc
@@ -1,27 +1,19 @@
 == Why is this an issue?
 
-Double-checked locking is the practice of checking a lazy-initialized object's state both before
-and after a `synchronized` block is entered to determine whether to initialize the object.
-In early JVM versions, synchronizing entire methods was not performant, which sometimes
-caused this practice to be used in its place.
+Double-checked locking is the practice of checking a lazy-initialized object's state both before and after a `synchronized` block is entered to determine whether to initialize the object.
+In early JVM versions, synchronizing entire methods was not performant, which sometimes caused this practice to be used in its place.
 
 
-Apart from `float` and `int` types, this practice does not work reliably in a platform-independent
-manner without additional synchronization of mutable instances.
-Using double-checked locking for the lazy initialization of any other type of primitive or mutable
-object risks a second thread using an uninitialized or partially initialized member while the first
-thread is still creating it.
+Apart from `float` and `int` types, this practice does not work reliably in a platform-independent manner without additional synchronization of mutable instances.
+Using double-checked locking for the lazy initialization of any other type of primitive or mutable object risks a second thread using an uninitialized or partially initialized member while the first thread is still creating it.
 The results can be unexpected, potentially even causing the application to crash.
 
 
 == How to fix it
 
-More recent JVM versions have significantly improved the performance of `synchronized` methods, so
-prefer using it over double-checked locking.
-Synchronized methods are now the preferred solution and are significantly more robust.
+Given significant performance improvements of `synchronized` methods in recent JVM versions, `synchronized` methods are now preferred over the less robust double-checked locking.
 
-If marking the entire method as `synchronized` is not an option, consider using an inner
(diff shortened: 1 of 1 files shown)


rules/S2168/java/rule.adoc
```diff
@@ -1,101 +1,84 @@
 == Why is this an issue?
 
-Double-checked locking is the practice of checking a lazy-initialized object's state both before and after a ``++synchronized++`` block is entered to determine whether or not to initialize the object.
+Double-checked locking is the practice of checking a lazy-initialized object's state both before
+and after a `synchronized` block is entered to determine whether to initialize the object.
+In early JVM versions, synchronizing entire methods was not performant, which sometimes
+caused this practice to be used in its place.
```
leonardo-pilastri-sonarsource, line 6:
```suggestion
Double-checked locking is the practice of checking a lazy-initialized object's state both before and after a `synchronized` block is entered to determine whether to initialize the object.
In early JVM versions, synchronizing entire methods was not performant, which sometimes caused this practice to be used in its place.
```

rules/S2168/java/rule.adoc
```diff
@@ -1,101 +1,84 @@
...
 
 
-It does not work reliably in a platform-independent manner without additional synchronization for mutable instances of anything other than ``++float++`` or ``++int++``. Using double-checked locking for the lazy initialization of any other type of primitive or mutable object risks a second thread using an uninitialized or partially initialized member while the first thread is still creating it, and crashing the program.
+Apart from `float` and `int` types, this practice does not work reliably in a platform-independent
+manner without additional synchronization of mutable instances.
+Using double-checked locking for the lazy initialization of any other type of primitive or mutable
+object risks a second thread using an uninitialized or partially initialized member while the first
+thread is still creating it.
+The results can be unexpected, potentially even causing the application to crash.
```
leonardo-pilastri-sonarsource, line 14:
```suggestion
Apart from `float` and `int` types, this practice does not work reliably in a platform-independent manner without additional synchronization of mutable instances.
Using double-checked locking for the lazy initialization of any other type of primitive or mutable object risks a second thread using an uninitialized or partially initialized member while the first thread is still creating it.
The results can be unexpected, potentially even causing the application to crash.
```

rules/S2168/java/rule.adoc
```diff
@@ -1,101 +1,84 @@
...
 
 
-There are multiple ways to fix this. The simplest one is to simply not use double checked locking at all, and synchronize the whole method instead. With early versions of the JVM, synchronizing the whole method was generally advised against for performance reasons. But ``++synchronized++`` performance has improved a lot in newer JVMs, so this is now a preferred solution. If you prefer to avoid using ``++synchronized++`` altogether, you can use an inner ``++static class++`` to hold the reference instead. Inner static classes are guaranteed to load lazily.
+== How to fix it
 
+More recent JVM versions have significantly improved the performance of `synchronized` methods, so
+prefer using it over double-checked locking.
+Synchronized methods are now the preferred solution and are significantly more robust.
```
leonardo-pilastri-sonarsource, line 21:
```suggestion
Given the significant performance and robustness improvements of `synchronized` methods in more recent JVM versions, using them over double-checked locking is recommended.
```

rules/S2168/java/rule.adoc
```diff
@@ -1,101 +1,84 @@
...
 
-=== Noncompliant code example
+If marking the entire method as `synchronized` is not an option, consider using an inner
+`static class` to hold the reference instead.
```
leonardo-pilastri-sonarsource, line 24:
```suggestion
If marking the entire method as `synchronized` is not an option, consider using an inner `static class` to hold the reference instead.
```

leonardo-pilastri-sonarsource (changes requested):
Just some line breaks adjustments, and a small rephrasing. Looks good to me for the rest!

leonardo-pilastri-sonarsource (approved):

PR #2822, opened 2023-08-08
Modify rule S3020: LaYC format

leonardo-pilastri-sonarsource:


rules/S3020/java/rule.adoc
```diff
@@ -1,27 +1,33 @@
 == Why is this an issue?
 
-Given no arguments, the ``++Collections.toArray++`` method returns an ``++Object []++``, which will cause a ``++ClassCastException++`` at runtime if you try to cast it to an array of the proper class. Instead, pass an array of the correct type in to the call.
+The `Collection.toArray()` method returns an `Object[]` when no arguments are provided to it.
+This can lead to a `ClassCastException` at runtime if you try to cast the returned array to an array of a specific class.
```
johann-beleites-sonarsource, line 4:
```suggestion
This can lead to a `ClassCastException` at runtime if you try to cast the returned array to an array of a specific type.
```

rules/S3020/java/rule.adoc
```diff
@@ -1,27 +1,33 @@
...
+Instead, you should use this method by also providing an array of the desired type as the parameter.
```
johann-beleites-sonarsource, line 5:
```suggestion
Instead, use this method by providing an array of the desired type as the argument.
```

johann-beleites-sonarsource, line 5: Maybe add a sentence about the fact that the array should have length 0?

rules/S3020/java/rule.adoc
```diff
@@ -1,27 +1,33 @@
...
 
+==== Compliant solution
 
-=== Compliant solution
-
-[source,java]
+[source,java,diff-id=1,diff-type=compliant]
 ----
 public String [] getStringArray(List<String> strings) {
   return strings.toArray(new String[0]);
```
johann-beleites-sonarsource, line 23: Do we need a `// Compliant` comment here?

rules/S3020/java/rule.adoc
```diff
@@ -1,27 +1,33 @@
...
 }
 ----
 
+== Resources
+
+* https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#toArray-T:A-[Collection.toArray()]
```
johann-beleites-sonarsource, line 29: The source is missing (see https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc)

johann-beleites-sonarsource (approved):

PR #2829, opened 2023-08-08
Modify rule S6643: Remove diff-tag, noncompliant and how to fix it sections

yassin-kammoun-sonarsource:


alexander-kamushkin-sonarsource (approved):
LGTM

PR #2834, opened 2023-08-08
CPP-3691 Remove "MISRA C++ 2008 recommended" and "MISRA C++ 2008" quality profiles from all RSPECs

amelie-renard-sonarsource:


frontend/src/deployment/__tests__/resources/metadata/S1000/cfamily-metadata.json
```diff
@@ -29,8 +29,7 @@
     ]
   },
   "defaultQualityProfiles": [
-    "Sonar way",
-    "MISRA C++ 2008 recommended"
+    "Sonar way"
```
michael-jabbour-sonarsource, line 32: If I understand correctly, the modified test used to verify that choosing multiple profiles gives the intersection of rules from these profiles. Do you think that it is worth introducing a "Test profile" into test data to replace "MISRA C++ 2008 recommended" here?
Note: This also applies to the `'collects all quality profiles'` test below...

rules/S6013/cfamily/metadata.json
```diff
@@ -29,7 +29,6 @@
   "sqKey": "S6013",
   "scope": "All",
   "defaultQualityProfiles": [
-    "MISRA C++ 2008",
```
michael-jabbour-sonarsource, line 32: Happy to see this being removed, because the rule doesn't seem related to MISRA as far as I understand :smile:

michael-jabbour-sonarsource (commented):
LGTM! I added a minor comment, but I'll leave it up to you to decide whether it is worth working on it or not. :smile:

PR #2838, opened 2023-08-09
Modify rule S2583: Add VB.NET version

mary-georgiou-sonarsource:


mary-georgiou-sonarsource:
Following up here https://github.com/SonarSource/rspec/pull/2857

PR #2843, opened 2023-08-09
Modify rule S3034: LaYC format

leonardo-pilastri-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3034/java/rule.adoc
+++ b/rules/S3034/java/rule.adoc
@@ -5,7 +5,7 @@ More specifically, narrower operands get promoted to the type of wider operands.
 For instance, an operation between a byte and an `int`, will trigger a promotion of the byte operand, converting it into an `int`.
 
 When this happens, the sequence of 8 bits that represents the `byte` will need to be extended to match the 32-bit long sequence that represents the `int` operand.
-Since Java uses two's complement representation, the promotion will fill the missing bits with zeros or ones, depending on the sign of the represented number.
+Since Java uses two's complement notation for signed number types, the promotion will fill the missing leading bits with zeros or ones, depending on the sign of the value.
 For instance, the byte `0b1000_0000` (that represents -128 in two's complement) when promoted to `int` will become `0b1111_1111_1111_1111_1111_1111_1000_0000`.
 
 If we perform shifting or bitwise operations without considering that we are dealing with signed bytes, the bits that were added during the promotion will have unexpected effects on the final result of the operations.
--- a/rules/S3034/java/rule.adoc
+++ b/rules/S3034/java/rule.adoc
@@ -15,8 +15,9 @@ Masking can be achieved with the bitwise AND `&` symbol and the `0xff` hexadecim
 
 This rule raises an issue any time a `byte` value is used as an operand combined with shifts without being masked.
 
+=== Code examples
 
-=== Noncompliant code example
+==== Noncompliant code example
 
 [source,java]
 ----
@@ -46,12 +47,12 @@ This rule raises an issue any time a `byte` value is used as an operand combined
 ----
 
 
-=== Compliant solution
+==== Compliant solution
 
 [source,java]
 ----
   public static void main(String[] args) {
-    byte[] bytes12 = BigInteger.valueOf(12).toByteArray(); 
+    byte[] bytes12 = BigInteger.valueOf(12).toByteArray();
     System.out.println(intFromBuffer(bytes12)); 
 
     byte[] bytes640 = BigInteger.valueOf(640).toByteArray();
@@ -78,7 +79,7 @@ This rule raises an issue any time a `byte` value is used as an operand combined
 ----
 
 
-== Resources
+== References
 
 * https://wiki.sei.cmu.edu/confluence/x/kDZGBQ[CERT, NUM52-J.] - Be aware of numeric promotion behavior 
 * https://en.wikipedia.org/wiki/Signed_number_representations#Two.27s_complement[Wikipedia] - Two's complement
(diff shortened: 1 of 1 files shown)


rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
 == Why is this an issue?
 
-When reading bytes in order to build other primitive values such as ``++int++``s or ``++long++``s, the ``++byte++`` values are automatically promoted, but that promotion can have unexpected results.
+In Java, numeric promotions happen when two operands of an arithmetic expression have different sizes.
+More specifically, narrower operands get promoted to the type of wider operands.
+For instance, an operation between a byte and an `int`, will trigger a promotion of the byte operand, converting it into an `int`.
 
+When this happens, the binary sequence of length 8 that represents the `byte` will need to match the length of the binary sequence of length 32 that represents the `int` operand.
```
johann-beleites-sonarsource, line 7:
```suggestion
When this happens, the sequence of 8 bits that represents the `byte` will need to be extended to match the 32-bit long sequence that represents the `int` operand.
```

rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
...
+Since Java uses two's complement representation, the promotion will fill the missing bits with zeros or ones, depending on the sign of the represented number.
```
johann-beleites-sonarsource, line 8:
```suggestion
Since Java uses two's complement notation for signed number types, the promotion will fill the missing leading bits with zeros or ones, depending on the sign of the value.
```

rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
...
+For instance, the byte `0b1000_0000` (that represents -128 in two's complement) when promoted to `int` will become `0b1111_1111_1111_1111_1111_1111_1000_0000`.
```
johann-beleites-sonarsource, line 9:
```suggestion
For instance, the byte `0b1000_0000` (equal to `-128` in decimal notation), when promoted to `int`, will become `0b1111_1111_1111_1111_1111_1111_1000_0000`.
```

rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
...
 
-For instance, the binary representation of the integer 640 is ``++0b0000_0010_1000_0000++``, which can also be written with the array of (unsigned) bytes ``++[2, 128]++``. However, since Java uses two's complement, the representation of the integer in signed bytes will be ``++[2, -128]++``  (because the ``++byte++`` ``++0b1000_0000++`` is promoted to the ``++int++`` ``++0b1111_1111_1111_1111_1111_1111_1000_0000++``). Consequently, trying to reconstruct the initial integer by shifting and adding the values of the bytes without taking care of the sign will not produce the expected result.  
+If we perform shifting or bitwise operations without considering that we are dealing with signed bytes, the bits that were added during the promotion will have unexpected effects on the final result of the operations.
```
johann-beleites-sonarsource, line 11:
```suggestion
When performing shifting or bitwise operations without considering that bytes are signed, the bits added during the promotion may have unexpected effects on the final result of the operations.
```

rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
...
 
+To prevent such accidental value conversions, we can mask promoted bytes to only consider the least significant 8 bits.
```
johann-beleites-sonarsource, line 13: Above this line looks like a natural place for the `== How to fix it` section to me. Perhaps this would require the "This rule raises an issue any time..." sentence to be moved up. Given the now much better explanation of the problem, I think a "How to fix it" section would make sense.

johann-beleites-sonarsource, line 13: We don't use 1st person pronouns in writing, instead use the 2nd person (i.e. `you`). (I was wondering about this topic myself, so I asked in guild-education about it. You can [read the thread here](https://sonarsource.slack.com/archives/C04P0SB2KJP/p1691591715236009), if you're interested.)

rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
...
+Masking can be achieved with the bitwise AND `&` symbol and the `0xff` hexadecimal symbol for 255 (0b1111_1111 in binary) or, since Java 8, with the more convenient `Byte.toUnsignedInt(byte b)` or `Byte.toUnsignedLong(byte b)`.
```
johann-beleites-sonarsource, line 14:
```suggestion
Masking can be achieved with the bitwise AND operator `&` and the appropriate mask of `0xff` (255 in decimal and 0b1111_1111 in binary) or, since Java 8, with the more convenient `Byte.toUnsignedInt(byte b)` or `Byte.toUnsignedLong(byte b)`.
```

rules/S3034/java/rule.adoc
```diff
@@ -1,48 +1,88 @@
...
+  }
+
+  static int intFromBufferAlternative(byte[] bytes) {
+    int originalInt = 0;
+    for (int i = 0; i < bytes.length; i++) {
+      originalInt = (originalInt << 8) | Byte.toUnsignedInt(bytes[i]); // Compliant, only the relevant 8 least significant bits will affect the bitwise OR
     }
-    return result;
+    return originalInt;
   }
```
johann-beleites-sonarsource, line 78: I think for a better UX, given the size of the examples, we should make use of the diff feature. I would:
* remove the compliant solution that uses `0xff` explicitly and only leave the `toUnsingedInt` one. People with enough expertise to have a good reason to safely use `0xff` instead should not need us to provide that example. We should assume by default people are using at least Java 8, so this method is available.
* Put the same comments in both the noncompliant and compliant snippets. Thanks to the diff highlighting, it will still be clear where the difference is, even if the comments make the examples a bit longer.

johann-beleites-sonarsource (changes requested):

johann-beleites-sonarsource (approved):
LGTM

PR #2845, opened 2023-08-10
Use new link format for secret resources

hendrik-buchwald-sonarsource:
To comply with https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc.


gaetan-ferry-sonarsource (approved):
LGTM! Thanks for taking care of this.

PR #2873, opened 2023-08-14
Modify rule S4348: reworked rule description for LaYC format

kaufco:


rules/S4348/java/rule.adoc
```diff
@@ -56,6 +66,11 @@ class FooSequence implements Iterable<Foo> {
 }
 ----
 
+== Resources
+
+* https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html[Java SE 8 API Specification: java.lang.Iterable]
```
irina-batinic-sonarsource, line 71:
```
Since: 1.5
See The Java™ Language Specification:14.14.2 The enhanced for statement
```

PR #2880, opened 2023-08-15
Modify rule S2189: Update rule description according to the LaYC

irina-batinic-sonarsource:


rules/S2189/java/rule.adoc
```diff
@@ -1,13 +1,59 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+An infinite loop will never end while the program runs, meaning you have to kill the program to get out of the loop.
+Every loop should have an end condition, whether by meeting the loop's end condition or via a `break`.
```
kaufco, line 4: I suggest a change here, because:

1. The sentence explains the thing with itself: a loop can have an end condition by meeting the loop's end condition 
2. "End condition" could be misleading for `for` and `while` loops, because the boolean expression there is the negated end condition (unlike for `repeat .. until` loops e.g. in Pascal, where it is the end condition)

```suggestion
Every loop should have an end condition, whether by meeting the loop's termination condition or via a `break` statement.
```

kaufco (changes requested):
Just one suggestion, and I think we should have a surrounding function for the code examples. Rest looks good!

kaufco (approved):
Looks good to me!

PR #2881, opened 2023-08-15
Modify rule S1751: Update rule description according to the LaYC

irina-batinic-sonarsource:


rules/S1751/java/rule.adoc
```diff
@@ -1,17 +1,86 @@
-include::../rule.adoc[]
+== Why is this an issue?
+
+A `loop` statement with at most one iteration is equivalent to an `If` statement; the following block is executed only once.
```
kaufco, line 3: - either "loop statement" (without backticks), "loop", or "`for`, `while` or `do`-`while` statement", because `loop` is not a keyword for a loop in Java
- "the following block is executed only once" should be mentioned directly over the block, i.e., in the "How to fix it" section.

Suggestion:

```suggestion
A loop with at most one iteration is equivalent to an `If` statement. This can confuse developers and make the code less readable since loops are not meant to replace `if` statements.
```

rules/S1751/java/rule.adoc
```diff
@@ -1,17 +1,86 @@
...
+
+If the initial intention was to conditionally execute the block only once, an `If` statement should be used instead.
```
kaufco, line 5:
```suggestion
If the intention was to conditionally execute the block only once, an `If` statement should be used instead.
```

rules/S1751/java/rule.adoc
```diff
@@ -1,17 +1,86 @@
...
+If that was not the initial intention, the loop block should be fixed so the block is executed multiple times.
```
kaufco, line 6:
```suggestion
Otherwise, the loop should be fixed so the loop block can be executed multiple times.
```

rules/S1751/java/rule.adoc
```diff
@@ -1,17 +1,86 @@
...
+
+A loop statement with at most one iteration can happen when a statement unconditionally transfers control, such as a jump or throw statement, is misplaced inside the loop block.
```
kaufco, line 8: Or when there is a mistake in the loop condition, but I don't know if the rule checks for such things:

```
for (boolean cond = true; cond; cond = false { ... }
```

```
do { ... } while (false)
```

kaufco, line 8: Gramatical mistake; should be "such as ... that is misplaced" instead?

rules/S1751/java/rule.adoc
```diff
@@ -1,17 +1,86 @@
...
+
+* `break`
+* `return`
+* `throw`
+
+== How to fix it
+
+=== Code examples
+
+==== Noncompliant code example
```
kaufco, line 20: Two things about the code examples:

1. I think they should have a surrounding function.
2. Since we have two examples anyway, maybe we should use different types of loops (e.g., `while` for the first one, `for` for the second one)

kaufco (changes requested):
Some remarks about the text and the examples.

rules/S1751/java/rule.adoc
irina-batinic-sonarsource: I doesn't check.

irina-batinic-sonarsource: Not really.
If you look at the sentence like this:
_A loop statement with at most one iteration can happen when a statement unconditionally transfers control is misplaced inside the loop block._

rules/S1751/java/rule.adoc
irina-batinic-sonarsource: Why do you think examples should have a surrounding function?
I feel like it is an extra useless code.

rules/S1751/java/rule.adoc
irina-batinic-sonarsource: We agreed to:
_A loop statement with at most one iteration can happen when a statement that unconditionally transfers control, such as a jump or throw statement, is misplaced inside the loop block._

kaufco (changes requested):

PR #2888, opened 2023-08-16
Modify rule S135: Update rule description according to the LaYC

irina-batinic-sonarsource:


kaufco (changes requested):
I think we should change the wording of the description a bit.

PR #2889, opened 2023-08-16
Modify rule S1994: reworked rule description for LaYC format

kaufco:


irina-batinic-sonarsource (changes requested):
The same thing about noncompliant and compliant examples as in the previous [PR](https://github.com/SonarSource/rspec/pull/2886/files)

irina-batinic-sonarsource (approved):

PR #2892, opened 2023-08-16
Modify rule S6073: SONARJAVA-4573 Update rule description; Add LaYC f…

irina-batinic-sonarsource:


rules/S6073/java/rule.adoc
```diff
@@ -1,49 +1,59 @@
...
-----
+    // Setting up mock responses
+    given(foo.bar(anyInt(), i1, i2)).willReturn(null); // Noncompliant
+    when(foo.baz(eq(val1), val2)).thenReturn("hi"); // Noncompliant
+
+    // Simulating exceptions
+    doThrow(new RuntimeException()).when(foo).quux(intThat(x -> x >= 42), -1); // Noncompliant
 
+    // Verifying method invocations
+    verify(foo).bar(i1, anyInt(), i2); // Noncompliant
```
kaufco, line 28: The comments increase the readability a lot. I would add some details to "Noncompliant" here though. I think with "it’s super hard to see where is the problem" the user means more to understand why the noncompliant lines are noncompliant. Suggestion:

```suggestion
    // Setting up mock responses
    given(foo.bar(anyInt(), i1, i2)).willReturn(null); // Noncompliant, no macthers for "i1" and "i2"
    when(foo.baz(eq(val1), val2)).thenReturn("hi"); // Noncompliant, no matcher for "val2"

    // Simulating exceptions
    doThrow(new RuntimeException()).when(foo).quux(intThat(x -> x >= 42), -1); // Noncompliant, no matcher for "-1"

    // Verifying method invocations
    verify(foo).bar(i1, anyInt(), i2); // Noncompliant, no matchers for "i1" and "i2"
```

rules/S6073/java/rule.adoc
```diff
@@ -1,49 +1,59 @@
...
 
-=== Compliant solution
+    // Capturing arguments for verification
+    ArgumentCaptor<Integer> captor = ArgumentCaptor.forClass(Integer.class);
+    verify(foo).bar(captor.capture(), i1, any()); // Noncompliant
```
kaufco, line 32: Same as above:

```suggestion
    verify(foo).bar(captor.capture(), i1, any()); // Noncompliant, no matchers for "i1"
```

rules/S6073/java/rule.adoc
```diff
@@ -1,49 +1,59 @@
...
+
+    // Simulating exceptions
+    doThrow(new RuntimeException()).when(foo).quux(intThat(x -> x >= 42), eq(-1));
+
+    // Verifying method invocations
+    verify(foo).bar(eq(i1), anyInt(), eq(i2));
+
+    // Capturing arguments for verification
+    ArgumentCaptor<Integer> captor = ArgumentCaptor.forClass(Integer.class);
+    verify(foo).bar(captor.capture(), any(), any());
```
kaufco, line 53: Add a comment here also for compliant, because there are two different reasons why they are now compliant: either all arguments have matchers now, or all matchers have been removed. Suggestion:

```suggestion
    // Setting up mock responses
    given(foo.bar(anyInt(), eq(i1), eq(i2))).willReturn(null); // Compliant, matchers also for "i1" and "i2"
    when(foo.baz(val1, val2)).thenReturn("hi"); // Compliant, no argument has matchers

    // Simulating exceptions
    doThrow(new RuntimeException()).when(foo).quux(intThat(x -> x >= 42), eq(-1)); // Compliant, matcher also for "-1"

    // Verifying method invocations
    verify(foo).bar(eq(i1), anyInt(), eq(i2)); // Compliant, matchers also for "i1" and "i2"

    // Capturing arguments for verification
    ArgumentCaptor<Integer> captor = ArgumentCaptor.forClass(Integer.class);
    verify(foo).bar(captor.capture(), any(), any()); // Compliant, all arguments have matchers
```

kaufco (changes requested):
Looks good! I would add some more details in the comments of the compliant / non-compliant code lines. Also, maybe we could shorten the whole example by a few lines in general? I think it's not required to demonstrate it with five different function calls to get the point of the rule, but cut some lines only if the example as a whole still makes sense.

kaufco (approved):

PR #2893, opened 2023-08-16
Fix rule coverage on a rule page: include all rule keys to metadata

necto:
For some rules, rule-specific metadata.json overridden the `sqKey` for all languages, so when the predeployment script collected all keys, it did not list the actual, most canonical key "S****".
This leads to the incorrect display of coverage information on a rule page that relies on the `allKeys` precomputed property to retrieve the version of the analyzer that implements the rule.

The fix includes the generic rule metadata.json file into the aggregator that computes `allKeys`, thus propagating it to every language-specific metadata.

### DIFF
--- a/frontend/src/deployment/__tests__/metadata.test.ts
+++ b/frontend/src/deployment/__tests__/metadata.test.ts
@@ -278,4 +278,34 @@ describe('metadata generation', () => {
       });
     });
   });
+
+  test('allKeys include the sqKey overridden in all languages', () => {
+    return withTestDir((srcPath) => {
+      createFiles(srcPath, {
+        'S100/metadata.json': JSON.stringify({
+          sqKey: 'S100',
+        }),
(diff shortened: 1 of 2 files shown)


necto:
Hi, @frederic-tingaud-sonarsource,
Sorry for the impromptu PR review request. I think you are the most familiar with the code here.
@costin-zaharia-sonarsource has spotted that some rule pages do not reflect the rule coverage correctly.
For example, for PL-SQL, [S999](http://localhost:3000/rspec#/rspec/S999/plsql) shows as covered, but [S1559](http://localhost:3000/rspec#/rspec/S1559/plsql) - not, even though on the search page, both of them show as covered.

necto:
I think this PR is not urgent, so feel free to review it at your leisure, or redirect it to someone else

frederic-tingaud-sonarsource (approved):

PR #2894, opened 2023-08-16
Modify rule S127: reworked rule description for LaYC format

kaufco:


irina-batinic-sonarsource (changes requested):
The same thing about noncompliant and compliant examples as in the previous [PR](https://github.com/SonarSource/rspec/pull/2886/files)

rules/S127/java/rule.adoc
```diff
@@ -1,23 +1,62 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+A `for` loop termination condition should test the loop counter against an invariant value which does not change during the execution of the loop.
```
irina-batinic-sonarsource, line 3: Let's change a word here:
A `for` loop termination condition should test the loop counter against an invariant value that does not change during the execution of the loop.

rules/S127/java/rule.adoc
```diff
@@ -1,23 +1,62 @@
...
+Invariant termination conditions make the program logic easier to understand and maintain.
 
-=== Noncompliant code example
+This rule tracks three types of non-invariant termination conditions:
 
-[source,java]
+* When the loop counters are updated in the body of the `for` loop
+* When the stop condition depends on a method call
+* When the stop condition depends on an object property, since such properties could change during the execution of the loop.
```
irina-batinic-sonarsource, line 10: You can remove comma here:
When the stop condition depends on an object property since such properties could change during the execution of the loop.

rules/S127/java/rule.adoc
```diff
@@ -1,23 +1,62 @@
...
+
+== How to fix it
+
+=== Code examples
+
+==== Noncompliant code example
+
+Make the termination condition invariant by using a constant or a local variable instead of an expression that could possibly change during execution of the loop.
```
irina-batinic-sonarsource, line 18: WDYT about this:
Make the termination condition invariant by using a constant or a local variable instead of an expression that could change during the execution of the loop.

kaufco: Yes, the "possibly" is redundant here.

irina-batinic-sonarsource (approved):

PR #2900, opened 2023-08-17
SONARJAVA-4463 add significant example to rule S2589

erwan-serandour-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
We could use a few tweaks for the solution to leverage all features of diff-highlighting

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #2905, opened 2023-08-18
Modify rule S5413: reworked rule description for LaYC format

kaufco:


irina-batinic-sonarsource (approved):
LGTM!
It seems like we made great progress this sprint in clarification of "pain points"

PR #2909, opened 2023-08-18
Modify rule S1488: add code example when returning awaited promise

ilia-kebets-sonarsource:
Fixes https://github.com/SonarSource/eslint-plugin-sonarjs/issues/413

### DIFF
--- a/rules/S1488/javascript/rule.adoc
+++ b/rules/S1488/javascript/rule.adoc
@@ -4,7 +4,7 @@ include::../description.adoc[]
 
 === Noncompliant code example
 
-[source,javascript]
+[source,js,diff-id=1,diff-type=noncompliant]
 ----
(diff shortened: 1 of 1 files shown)


yassin-kammoun-sonarsource (commented):
I am really not convinced by the need for this change. I mean it's just the same thing as the initial code snippet but applied with async-await.

Do you really think this additional snippet bring any value?

ilia-kebets-sonarsource:
I would because most of the time, you either await a promise (1) if it's not the value you wish to return or return the promise (2), so returning the awaited value (3) might seem counter-intuitive, as seen by the raised issue.

The numbers refer to `fct1`, `fct2`, `fct3` in https://github.com/SonarSource/eslint-plugin-sonarjs/issues/413

yassin-kammoun-sonarsource (approved):
Still not convinced, but if it can help...

PR #2914, opened 2023-08-21
Modify rule S5488: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5488/cfamily/rule.adoc
+++ b/rules/S5488/cfamily/rule.adoc
@@ -3,7 +3,7 @@ Passing invalid arguments to UNIX/POSIX functions may result in undefined behavi
 == Why is this an issue?
 
 Many UNIX/POSIX functions put certain constraints on the values of their parameters.
-The behavior for some of those UNIX/POSIX functions is unspecified, i.e., implementation-defined, if one passes incorrectly constrained parameters.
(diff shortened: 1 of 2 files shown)


rules/S5488/cfamily/rule.adoc
```diff
@@ -1,27 +1,189 @@
+Passing invalid arguments to UNIX/POSIX functions may result in undefined behavior.
+
 == Why is this an issue?
 
-UNIX/POSIX functions can have undefined behavior if they are not called correctly. More specifically:
+Many UNIX/POSIX functions put certain constraints on the values of their parameters.
+The behavior for some of those UNIX/POSIX functions is unspecified, i.e., implementation-defined, if one passes incorrectly constrained parameters.
```
tomasz-kaminski-sonarsource, line 6: Unspecified is not the same as implementation defined. Unspecified means that each execution could produce different results, even with the same arguments, while implementation defined need to be document. Maybe say `is not defined` instead of referring to world, that is world of power.

tomasz-kaminski-sonarsource (changes requested):
Looks good.

rules/S5488/cfamily/rule.adoc
pdschbrt: I rephrased it a bit to avoid the term `unspecified`.

tomasz-kaminski-sonarsource (changes requested):
Looks really good. Please prepare PR for the sonar-cpp.

PR #2916, opened 2023-08-21
Modify rule S5308: Expand and adjust for LaYC

pdschbrt:
## Review

- [x] text style and tone


rules/S5308/cfamily/rule.adoc
```diff
@@ -1,30 +1,159 @@
...
-``++setuid++`` family of functions are used to change the user identity of the caller process in order to change privileges for the following actions. If the function call returns with an error, the rest of the program will execute with unexpected privileges leading to unexpected behaviour.
+== Why is this an issue?
 
+Functions from the ``++setuid++``-family including ``++setuid++`` and ``++setgid++`` are used to change the identity of the caller process.
+They are used to change privileges for the subsequent actions to be executed.
+If a call to these functions returns an error that is not checked and handled appropriately, the subsequent parts of the program will execute with unexpected privileges.
+This, in turn, leads to unexpected program behavior and poses a _serious_ security risk.
 
-=== Noncompliant code example
+A brief example on how ``++setuid++`` can be used and how to correctly check its return value is given in what follows:
```
tomasz-kaminski-sonarsource, line 10: This seems to be useful, but it does not match why this is an issue. The code as far I understand, contains no issue. I think that is really nice example, and maybe we could rework it to show, how you create or remove file with root privileges if set_uid failed. I would put that in separate section, after what is potential impact.

tomasz-kaminski-sonarsource (changes requested):
This looks really nice. I have added some suggestion how to rewrite the example from how to use correctly, to one that illustrates potential impact.

rules/S5308/cfamily/rule.adoc
pdschbrt: I agree that we should put the example into a separate section; currently, it's not in the right place.

The code deliberately does not contain an issue. Since this is really security-critical stuff, I didn't want to show the user a broken example. There are enough bad examples on StackOverflow and on the internet in general.

I'll put it in a separate section and try to modify it slightly to showcase what could go wrong without correct error handling, however, I'll try to refrain from making it like really bad.

pdschbrt: It's now updated. Please check the most recent version.

tomasz-kaminski-sonarsource (approved):
Good read. Made me late to the standup.

pdschbrt:
Excellent. Thanks!

PR #2918, opened 2023-08-22
Create rule S6707

maksim-grebeniuk-sonarsource:
## Review

- [ ] text style and tone


rules/S6707/python/rule.adoc
```diff
@@ -0,0 +1,55 @@
+This rule raises an issue when the comparison variable compared with ``++numpy.nan++``.
```
guillaume-dequenne-sonarsource, line 1: This rule raises an issue when an equality check is made against ``++numpy.nan++``.

rules/S6707/python/rule.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+== Why is this an issue?
+
+Comparing variables with ``++numpy.nan++`` in NumPy can lead to unexpected and incorrect results due to the special nature of ``++numpy.nan++`` as a "Not a Number" value. NumPy treats ``++numpy.nan++`` as a placeholder for undefined or missing values in numerical computations. Because of its unique properties, comparing variables with ``++numpy.nan++`` using standard comparison operators can produce indeterminate and counterintuitive outcomes.
```
guillaume-dequenne-sonarsource, line 5: I think the description is quite vague ("can produce counterintuitive outcome") and should probably be more specific knowing the result will always be `False`.

I think we should mention explicitly what the `np.isnan` API does and why it should be the only way to check for NaN.

rules/S6707/python/metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "title": "The comparison with \"numpy.nan\" should not be used",
```
guillaume-dequenne-sonarsource, line 2: Maybe:
Equality checks should not be made against "numpy.nan"
?
Not 100% convinced though, but I feel we could have a more explicit title.

guillaume-dequenne-sonarsource (changes requested):
Some small improvements suggestions.

rules/S6707/python/rule.adoc
```diff
@@ -1,8 +1,12 @@
-This rule raises an issue when the comparison variable compared with ``++numpy.nan++``.
+This rule raises an issue when an equality check is made against ``++numpy.nan++``.
 
 == Why is this an issue?
 
-Comparing variables with ``++numpy.nan++`` in NumPy can lead to unexpected and incorrect results due to the special nature of ``++numpy.nan++`` as a "Not a Number" value. NumPy treats ``++numpy.nan++`` as a placeholder for undefined or missing values in numerical computations. Because of its unique properties, comparing variables with ``++numpy.nan++`` using standard comparison operators can produce indeterminate and counterintuitive outcomes.
+The ``++numpy.nan++`` is a floating point representation of Not a Number (NaN) used as a placeholder for undefined or missing values in numerical computations.
+
+Equality check of variables against ``++numpy.nan++`` in NumPy will always be ``++False++`` due to the special nature of ``++numpy.nan++``. This can lead to unexpected and incorrect results.
```
guillaume-dequenne-sonarsource, line 7: I think it should be plural: Equality checks ...

guillaume-dequenne-sonarsource (changes requested):
Looks good, very minor changes still required

rules/S6707/python/metadata.json
```diff
@@ -5,7 +5,7 @@
     "impacts": {
       "MAINTAINABILITY": "HIGH"
     },
-    "attribute": "CONVENTIONAL"
+    "attribute": "COMPLETE"
```
guillaume-dequenne-sonarsource, line 8: Shouldn't this be "LOGICAL"?
i.e from the decision tree:
```
Code will never do what was intended (or will do something undefined).
Statement that are always true or run but do nothing (e.g. a=a)
```

I don't think the comparison ever works as intended so it fits the definition, wdyt?

guillaume-dequenne-sonarsource (changes requested):
I think attribute is not correct, LGTM otherwise.

PR #2933, opened 2023-08-23
Modify rule S3588: Expand and adjust for LaYC

pdschbrt:
Removed code example which contains a FN. A corresponding Jira ticket has been created at https://sonarsource.atlassian.net/browse/CPP-4655

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3588/cfamily/rule.adoc
+++ b/rules/S3588/cfamily/rule.adoc
@@ -60,8 +60,8 @@ An exemplary class that manages a pointer to a file is shown in what follows.
 #include <iostream>
 #include <string>
 
-// Although `std::fstream` should be preferred, if available, this `File` class
-// does not and cannot suffer from "double-close" issues.
+// Although `std::fstream` should be preferred, if available, a file stream
+// managed by this `File` class cannot suffer from "double-close" issues.
 class File {
   FILE *f;
 
@@ -86,25 +86,17 @@ public:
   // created in and hence disallow copying or moving it.
   File(const File &other) = delete;
   File(File &&other) = delete;
-  size_t fread(void *ptr, size_t size, size_t nmemb); // Forward to C's `fread`.
-  size_t fwrite(const void *ptr, size_t size,
-                size_t nmemb); // Forward to C's `fwrite`.
-  size_t size();               // Determine file size in bytes.
+  // Allow access to underlying file via `f`.
+  FILE *handle() { return f; }
 };
 
-int main() {
-  File f{"example.txt", "r"};
-  size_t bytes = f.size();
-  std::string buffer(bytes + 1, '\0');
-  size_t items_read = f.fread(buffer.data(), bytes, 1);
-  if (items_read < 1) {
-    printf("Could not read items!\n");
-    return 1;
-  }
-  std::cout << "file contents:\n" << buffer << '\n';
-  // The file stream managed by `f` will be automatically closed when it goes
+void file_user() {
+  File fh{"example.txt", "r"};
+  FILE *f = fh.handle();
+  // Use `f` for the desired file operation(s).
+  //
+  // The file stream managed by `fh` will be automatically closed when `fh` goes
   // out of scope at the end of this function.
-  return 0;
 }
 ----
 
--- a/rules/S3588/cfamily/rule.adoc
+++ b/rules/S3588/cfamily/rule.adoc
@@ -75,7 +75,11 @@ public:
     }
   }
   // Will close the file stream upon destruction.
-  ~File() { close(); }
+  ~File() {
+    // Here we are fine with `std::terminate` being called here in case `close`
+    // throws and exception.
+    close();
+  }
   // Allow only one owner of a file, disallow copy operations.
   File(const File &other) = delete;
   File &operator=(const File &other) = delete;
@@ -93,9 +97,7 @@ public:
   // Allow file to be closed explicitly.
   void close() {
     if (f != nullptr && fclose(std::exchange(f, nullptr)) == EOF) {
-      // We made the choice to exit immediately, if a file cannot be closed.
-      perror("fclose() failed");
-      exit(1);
+      throw std::ios_base::failure("fclose() failed");
     }
   }
   // Allow access to underlying file via `f`.
--- a/rules/S3588/cfamily/rule.adoc
+++ b/rules/S3588/cfamily/rule.adoc
@@ -57,7 +57,6 @@ An exemplary class that manages a pointer to a file is shown in what follows.
 [source,cpp]
 ----
 #include <cstdio>
-#include <exception>
 #include <fstream>
 #include <string>
 #include <utility>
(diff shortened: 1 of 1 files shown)


tomasz-kaminski-sonarsource (changes requested):
Could you rebase this on master, so I would see actual diff?

tomasz-kaminski-sonarsource (changes requested):
I think we should mention RAII in going extra mile in this section when representing file as object in C++ would remove the closing and avoid double closing. similarly what we have to use after free or double free.

pdschbrt:
Please give it another read.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,12 @@ The application might just crash, but in the worst case, the application may app
 
 Do not use the value of a pointer to a ``++FILE++`` object after the associated file has been closed.
 
+=== Going the extra mile
+
+Using {cpp}'s _RAII_ idiom can mitigate these "double-close" issues.
```
tomasz-kaminski-sonarsource, line 51: I think I would add example of such class here. It should be that hard. We have some shared entry for RAII idiom, and maybe you could include it here.

pdschbrt: I've sketched an exemplary class that manages a variable of type `FILE *`.

I've also seen the shared description for RAII but felt that it does not quite match the situation and hence chose not to include it. Instead, I've added a few lines on RAII that match this rule's context.

tomasz-kaminski-sonarsource (changes requested):
Looks a lot as the think I was looking for. Just few comments.

rules/S3588/cfamily/rule.adoc
```diff
@@ -86,43 +86,24 @@ public:
...
-    }
-    ssize_t bytes = ftell(f);
-    if (bytes == -1) {
-      perror("ftell() failed");
-      exit(-1);
-    }
-    rewind(f);
-    return bytes;
-  }
+  size_t fread(void *ptr, size_t size, size_t nmemb); // Forward to C's `fread`.
```
tomasz-kaminski-sonarsource, line 89: I would add `handle()` method to extract `FILE*`, so it show that you do not need to wrap every API to gets the benefits.

pdschbrt:
The problem with returning the pointer using `handle()` is that some madman could then pass that pointer to `fclose`, again. I wish to eliminate misuse with such a wrapper class.

I've cleaned-up the code a bit, which should do the trick.

rules/S3588/cfamily/rule.adoc
pdschbrt: I've got a concern with that. See my comment above.

tomasz-kaminski-sonarsource: Yes, but it is same for `unique_ptr`, you can call explicit delete. You are protecting against `Machieveli`, and making user believe that you need wrap everything. I would add a comment, that they could also wrap each call.

pdschbrt: Fair enough. I've updated the class accordingly. It should be less distracting now.

tomasz-kaminski-sonarsource: I would still add a comment at the end mention that providing a functions that wrap underlying call may be better. But this is sufficient to address double free. Or something, like:
> With the above design it is still possible double close file handle, by calling passing `f.handle()` to function that closes stream (for example `close(f.handle())`. But this design reduces the risk of such occurrence, but eliminating risk of manual closing.

Maybe we should also add `release()` function, that will return handle, but disamr file by setting internal one to nullptr.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,62 @@ The application might just crash, but in the worst case, the application may app
...
+class File {
+  FILE *f;
+
+public:
+  // Opens file stream on construction.
+  File(std::string const &path, std::string const &modes) {
+    f = fopen(path.c_str(), modes.c_str());
+    if (!f) {
+      perror("fopen() failed");
+      exit(1);
```
tomasz-kaminski-sonarsource, line 74: We should throw something here. Or just make file empty.

rules/S3588/cfamily/rule.adoc
tomasz-kaminski-sonarsource: OK. I would skip release, but rename this class to `ScopedFile` or would add move constructor and operator bool.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,74 @@ The application might just crash, but in the worst case, the application may app
...
+
+// Although `std::fstream` should be preferred, if available, a file stream
+// managed by this `File` class cannot suffer from "double-close" issues.
+class File {
+  FILE *f;
+
+public:
+  // Opens file stream on construction.
+  File(std::string const &path, std::string const &modes) {
+    f = fopen(path.c_str(), modes.c_str());
```
tomasz-kaminski-sonarsource: Use ctor initializer.

pdschbrt:
Slowly getting there. Let me know what you think.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,91 @@ The application might just crash, but in the worst case, the application may app
...
+  // Opens file stream on construction.
+  File(std::string const &path, std::string const &modes)
+      : f(fopen(path.c_str(), modes.c_str())) {
+    if (!f) {
+      throw std::ios_base::failure("fopen() failed");
+    }
+  }
+  // Will close the file stream upon destruction.
+  ~File() {
+    if (!f) {
```
tomasz-kaminski-sonarsource, line 80: Just call close inside, and add a comment regarding termination,.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,91 @@ The application might just crash, but in the worst case, the application may app
...
+      return *this;
+    }
+    // Close the currently managed file and "steal" the other's file.
+    close();
+    f = std::exchange(other.f, nullptr);
+    return *this;
+  }
+  // Allow file to be closed explicitly.
+  void close() {
+    int status = fclose(std::exchange(f, nullptr));
```
tomasz-kaminski-sonarsource, line 107: You should not close nullptr. So this need a check if f is not null.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,91 @@ The application might just crash, but in the worst case, the application may app
...
+  // Allow only one owner of a file, disallow copy operations.
+  File(const File &other) = delete;
+  File &operator=(const File &other) = delete;
+  // Moving a file to a different scope shall be allowed.
+  File(File &&other) { f = std::exchange(other.f, nullptr); }
+  File &operator=(File &&other) {
+    if (this == &other) { // Handle self-assignment.
+      return *this;
+    }
+    // Close the currently managed file and "steal" the other's file.
```
tomasz-kaminski-sonarsource, line 100: I would put that inside if and reverse condition.

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,91 @@ The application might just crash, but in the worst case, the application may app
...
+      // file cannot be closed.
+      perror("fclose() failed");
+      exit(1);
+    }
+  }
+  // Allow only one owner of a file, disallow copy operations.
+  File(const File &other) = delete;
+  File &operator=(const File &other) = delete;
+  // Moving a file to a different scope shall be allowed.
+  File(File &&other) { f = std::exchange(other.f, nullptr); }
```
tomasz-kaminski-sonarsource, line 95: Again, use ctor initializer.

tomasz-kaminski-sonarsource (changes requested):

rules/S3588/cfamily/rule.adoc
```diff
@@ -46,6 +46,91 @@ The application might just crash, but in the worst case, the application may app
...
+      throw std::ios_base::failure("fopen() failed");
+    }
+  }
+  // Will close the file stream upon destruction.
+  ~File() {
+    if (!f) {
+      return;
+    }
+    int status = fclose(f);
+    if (status == EOF) {
```
tomasz-kaminski-sonarsource, line 84: Why not use if with intialier, of just inline call of close into condition:
```
if (close(f) == EOF) {
  throw....;
}
```

rules/S3588/cfamily/rule.adoc
```diff
@@ -76,37 +75,27 @@ public:
...
     return *this;
   }
   // Allow file to be closed explicitly.
   void close() {
-    int status = fclose(std::exchange(f, nullptr));
-    if (status == EOF) {
-      throw std::ios_base::failure("fclose() failed");
+    if (f != nullptr && fclose(std::exchange(f, nullptr)) == EOF) {
+      // We made the choice to exit immediately, if a file cannot be closed.
+      perror("fclose() failed");
```
tomasz-kaminski-sonarsource, line 97: I would throw here, and let destructor terminate or hitting noexcept.

tomasz-kaminski-sonarsource: Just comment, that destructor will terminate in case of exceptions.

pdschbrt: Done.

tomasz-kaminski-sonarsource (approved):

PR #2937, opened 2023-08-23
Modify rule S5280: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5280/cfamily/rule.adoc
+++ b/rules/S5280/cfamily/rule.adoc
@@ -1,20 +1,76 @@
+The ``++template++`` parameter for creating temporary files or directories must contain at least six trailing "X"s to ensure uniqueness.
+
 == Why is this an issue?
 
-``++mktemp++``, ``++mkstemp++``, ``++mkstemps++`` or ``++mkdtemp++`` are useful functions to create temporary files and directories. They accept a ``++template++`` parameter which defines the path where the file should be created and which should contain at least six trailing "X"s to ensure the uniqueness of the filename.
+The standard C library provides several functions to create temporary files and directories.
+These functions include ``++mktemp++``, ``++mkstemp++``, ``++mkstemps++`` and ``++mkdtemp++``, and they accept as a parameter a ``++template++`` that defines the path where the file (or directory) should be created.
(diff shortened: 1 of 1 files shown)


rules/S5280/cfamily/rule.adoc
```diff
@@ -1,20 +1,76 @@
-== Why is this an issue?
+The ``++template++`` parameter for creating temporary files or directories must contain at least six trailing "X"s to ensure uniqueness.
 
-``++mktemp++``, ``++mkstemp++``, ``++mkstemps++`` or ``++mkdtemp++`` are useful functions to create temporary files and directories. They accept a ``++template++`` parameter which defines the path where the file should be created and which should contain at least six trailing "X"s to ensure the uniqueness of the filename.
+== Why is this an issue?
 
+The standard C library provides several functions to create temporary files and directories.
+These functions include ``++mktemp++``, ``++mkstemp++``, ``++mkstemps++`` and ``++mkdtemp++``, and they accept as a parameter a ``++template++`` that defines the path where the file (or directory) should be created.
+The filename part of the ``++template++`` must contain at least six trailing "X"s to ensure the uniqueness of the filename, since these "X"'s are replaced with a string that makes the filename unique.
```
tomasz-kaminski-sonarsource, line 7: Using the word `must` suggest that it would be ill-formed otherwise.
```suggestion
The filename part of the ``++template++`` should contain at least six trailing "X"s to ensure the uniqueness of the filename, since these "X"'s are replaced with a string that makes the filename unique.
```

rules/S5280/cfamily/rule.adoc
```diff
@@ -1,20 +1,76 @@
-== Why is this an issue?
+The ``++template++`` parameter for creating temporary files or directories must contain at least six trailing "X"s to ensure uniqueness.
```
tomasz-kaminski-sonarsource, line 1: Could we use different word than template? Like pattern, or this is the usual convection for this function.

tomasz-kaminski-sonarsource (approved):
One question regarding the use of a template, but otherwise looks good.

rules/S5280/cfamily/rule.adoc
pdschbrt: I kept calling it `template` here since the parameter is called that way in libc
```
char *mktemp(char *template);
```
Is that OK with you, if we keep it like that?

PR #2960, opened 2023-08-24
Modify rule S117: Update regular expression for PHP

petertrr:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S117/php/rule.adoc
+++ b/rules/S117/php/rule.adoc
@@ -4,11 +4,16 @@ include::../description.adoc[]
 
 === Noncompliant code example
 
-With the default regular expression ``++^[a-z][a-zA-Z0-9]*$++``:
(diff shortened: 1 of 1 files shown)


rules/S117/php/rule.adoc
```diff
@@ -4,7 +4,7 @@ include::../description.adoc[]
 
 === Noncompliant code example
 
-With the default regular expression ``++^[a-z][a-zA-Z0-9]*$++``:
+With the default regular expression ``++^[a-z_][a-zA-Z0-9_]*$++``:
```
jonas-wielage-sonarsource, line 7: The compliant code example needs to be changed, as `$my_param` is now an allowed name.

petertrr: I would still say that it's advisable not to mix different styles in the same scope, even though we don't check it explicitly. I can add a couple of examples then

PR #3006, opened 2023-09-04
Modify S5876(common): Add an illustrative example

loris-s-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5876/rationale.adoc
+++ b/rules/S5876/rationale.adoc
@@ -5,3 +5,8 @@ Session fixation attacks take advantage of the way web applications manage sessi
 * In a session fixation attack, an attacker tricks a user into using a predetermined session identifier controlled by the attacker. For example, the attacker sends the victim an email containing a link with this predetermined session identifier.
 * When the victim clicks on the link, the web application does not create a new session identifier but uses this identifier known to the attacker.
 * At this point, the attacker can hijack and impersonate the victim's session.
+
+
+Session fixation attacks also occur when sessions are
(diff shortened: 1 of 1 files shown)


rules/S5876/rationale.adoc
```diff
@@ -5,3 +5,8 @@ Session fixation attacks take advantage of the way web applications manage sessi
 * In a session fixation attack, an attacker tricks a user into using a predetermined session identifier controlled by the attacker. For example, the attacker sends the victim an email containing a link with this predetermined session identifier.
 * When the victim clicks on the link, the web application does not create a new session identifier but uses this identifier known to the attacker.
 * At this point, the attacker can hijack and impersonate the victim's session.
+
+
+Session fixation attacks also occur when sessions are
+https://hackerone.com/reports/2121960[not invalidated after users reset their password].
```
hendrik-buchwald-sonarsource, line 11: I don't think it is a great resource. At least, to me, it is not clear after reading it how it is related to session fixation.

hendrik-buchwald-sonarsource (changes requested):
As I wrote in the other comment, I don't think the resource is very illustrative.

rules/S5876/rationale.adoc
loris-s-sonarsource: Makes sense ! After a clear re-reading, it's true that there's a lot of info to process and most importantly lots of info related to the target more than the vuln. Thanks for your feedback 👍

PR #3019, opened 2023-09-06
APPSEC-1053 Validate S6648 ARM default value for secret

daniel-teuchert-sonarsource:
## Review

- [ ] text style and tone


loris-s-sonarsource (commented):
Hey, I don't understand why secureString is better than secureStringWithDefaultValue. What is the why of this change ?

daniel-teuchert-sonarsource:
> Hey, I don't understand why secureString is better than secureStringWithDefaultValue. What is the why of this change ?

Because in this example it is a secureString without a default value, so the name does not make sense.

loris-s-sonarsource (approved):
LGTM ! Makes sense then

PR #3028, opened 2023-09-08
Modify rule S6597: "WORKDIR" instruction should be used instead of "cd" commands

mstachniuk:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6597/docker/rule.adoc
+++ b/rules/S6597/docker/rule.adoc
@@ -18,6 +18,17 @@ RUN cd /tmp && \
 
 In this example, the first `cd /tmp` command can be replaced by `WORKDIR cd /tmp` put above `RUN`.
 The last `cd /app/bin` command can be replaced too.
+The result will be following:
+
+[source,docker]
+----
+WORKDIR /tmp
+RUN git clone myrepository.com/MyOrganization/project && \
+  cd project && \
+  make
+WORKDIR /app/bin
+----
+
 Those actions will reduce the length of the `RUN` instruction, which makes it easier to read and understand.
 Sometimes, it is hard to avoid the usage of `cd` command, especially in the middle of a long script.
 Removing them from the beginning and end of a multi-line is an easy improvement.
@@ -33,7 +44,7 @@ Following this recommendation provides a clear structure for Dockerfiles, making
 == How to fix it
 
 Where possible, ensure that all usages of `cd` commands are replaced by a `WORKDIR` instruction.
-`cd` commands at the beginning or end of multiline commands are a reliable sign, that they can be replaced.
+The `cd` commands at the beginning or end of multiline commands are a reliable sign, that they can be replaced.
 Also, using absolute paths can be considered for commands that accept them.
 
 === Code examples
--- a/rules/S6597/docker/rule.adoc
+++ b/rules/S6597/docker/rule.adoc
@@ -1,12 +1,38 @@
-The `cd` command should not be used to change the working directory.
-Instead `WORKDIR` should be used.
-
 == Why is this an issue?
 
-The `cd` commands in `RUN` (and other) instructions make them longer, harder to read, troubleshoot, and maintain.
-It is better to use the `WORKDIR` instruction to set a working directory for the next Docker instructions.
+In single Dockerfile instructions like `RUN`, `CMD`, and `ENTRYPOINT` multiple commands can be run, including changing directory `cd` command.
+Using `WORKDIR` instead reduces the complexity of the above instructions and makes them easy to read, understand, troubleshoot, and maintain.
+
+=== What is the potential impact?
+
+The Dockerfile instructions like `RUN` allow users to execute longer scripts. See the following example:
+
+[source,docker]
+----
+RUN cd /tmp && \
+  git clone myrepository.com/MyOrganization/project && \
+  cd project && \
+  make && \
+  cd /app/bin
+----
+
+In this example, the first `cd /tmp` command can be replaced by `WORKDIR cd /tmp` put above `RUN`.
+The last `cd /app/bin` command can be replaced too.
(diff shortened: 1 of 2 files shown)


rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
-The `cd` command should not be used to change the working directory.
-Instead `WORKDIR` should be used.
-
 == Why is this an issue?
 
-The `cd` commands in `RUN` (and other) instructions make them longer, harder to read, troubleshoot, and maintain.
-It is better to use the `WORKDIR` instruction to set a working directory for the next Docker instructions.
+In single Dockerfile instructions like `RUN`, `CMD`, and `ENTRYPOINT` multiple commands can be run, including changing directory `cd` command.
```
jonas-wielage-sonarsource, line 3:
```suggestion
In single Dockerfile instructions like `RUN`, `CMD`, and `ENTRYPOINT` multiple commands can be run, including the `cd` command for changing directories.
```

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+----
+RUN cd /tmp && \
+  git clone myrepository.com/MyOrganization/project && \
+  cd project && \
+  make && \
+  cd /app/bin
+----
+
+In this example, the first `cd /tmp` command can be replaced by `WORKDIR cd /tmp` put above `RUN`.
+The last `cd /app/bin` command can be replaced too.
```
jonas-wielage-sonarsource, line 20: I would add the resulting command block for this example as well, that the user can see the "compliant" solution

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+Those actions will reduce the length of the `RUN` instruction, which makes it easier to read and understand.
+Sometimes, it is hard to avoid the usage of `cd` command, especially in the middle of a long script, but removing them from the beginning and end is straightforward.
```
jonas-wielage-sonarsource, line 22:
```suggestion
Sometimes, it is hard to avoid the usage of `cd` command, especially in the middle of a long script.
Removing them from the beginning and end of a multi-line is an easy improvement.
```

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+Additionally, most commands work well with absolute paths, so changing directories can be avoided very often.
```
jonas-wielage-sonarsource, line 23:
```suggestion
Additionally, a lot of commands work well with absolute paths, so changing directories can be avoided in most cases.
```

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+
+The `WORKDIR` instruction can be used multiple times in Dockerfile.
```
jonas-wielage-sonarsource, line 25:
```suggestion
The `WORKDIR` instruction can be used multiple times in a Dockerfile.
```

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+It changes the current directory for next instructions (until the following change).
```
jonas-wielage-sonarsource, line 26:
```suggestion
It changes the current directory for the next instructions and until there is a following change.
```

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+This approach simplifies understanding of what is a current directory.
+
+The same principles apply to `CMD` and `ENTRYPOINT` instructions.
+Following this recommendation provides a clear structure for Dockerfiles, making it easier to maintain.
 
 == How to fix it
 
+Ensure that all usages of `cd` command are replaced by `WORKDIR` instruction, where possible.
+Also, using absolute paths can be considered for commands that accept them.
```
jonas-wielage-sonarsource, line 35: Why include that in this rule? Isn't that unrelated advice?

rules/S6597/docker/rule.adoc
```diff
@@ -1,13 +1,39 @@
...
+The `WORKDIR` instruction can be used multiple times in Dockerfile.
+It changes the current directory for next instructions (until the following change).
+This approach simplifies understanding of what is a current directory.
+
+The same principles apply to `CMD` and `ENTRYPOINT` instructions.
+Following this recommendation provides a clear structure for Dockerfiles, making it easier to maintain.
 
 == How to fix it
 
+Ensure that all usages of `cd` command are replaced by `WORKDIR` instruction, where possible.
```
jonas-wielage-sonarsource, line 34:
```suggestion
Where possible, ensure that all usages of `cd` commands are replaced by a `WORKDIR` instruction.
`cd` commands at the beginning or end of multiline commands are a reliable sign, that they can be replaced.
```

jonas-wielage-sonarsource (approved):
Please change the title of the PR to be in the required format:
`Modify rule S6597: cd command should not be replaced in the middle of a multiline command`

rules/S6597/docker/rule.adoc
mstachniuk: It's related, it's another way to avoid `cd` for example in the middle of the script.

mstachniuk:
> Also please change the title of the PR to be in the required format:
> `Modify rule S6597: cd command should not be replaced in the middle of a multiline command`

The rule name is different, but a good point of the title of PR.

jonas-wielage-sonarsource:
> The rule name is different, but a good point of the title of PR.

I believe we don't need to add the complete rule title here in rspec, as this isn't the PR creating the rule. We are only modifying it, and I understood the [rspec guidelines](The rule name is different, but a good point of the title of PR.) that the title should contain the summary of the PR. In this case the changes.

PR #3054, opened 2023-09-12
Modify rule S5523: Correct spec and update metadata

michael-jabbour-sonarsource:
- Metadata is updated so that it is consistent with the metadata of the new MISRA rule M23_321. See SonarSource/sonar-cpp#3006.
- Correct the spec since the implementation has been updated later so that it raises only on built-in and pointer types. See SonarSource/sonar-cpp#684.

I considered updating the rule to LaYC format, but I wanted to keep this PR simple for now because the rule is not in "Sonar Way"...

## Review

- [x] text style and tone

### DIFF
--- a/rules/S5523/cfamily/metadata.json
+++ b/rules/S5523/cfamily/metadata.json
@@ -3,9 +3,9 @@
   "type": "CODE_SMELL",
   "code": {
     "impacts": {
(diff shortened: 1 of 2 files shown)


marco-antognini-sonarsource (approved):
LGTM

PR #3092, opened 2023-09-18
APPSEC-1088: S6385 Improve RSPEC

daniel-teuchert-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6385/description.adoc
+++ b/rules/S6385/description.adoc
@@ -1,5 +1,2 @@
 Defining a custom role for a Subscription or a Management group that allows all actions will give them the same capabilities as the built-in Owner role.
 It's recommended to limit the number of subscription owners in order to mitigate the risk of being breached by a compromised owner.
-
-This rule raises an issue when a custom role has an assignable scope set to a Subscription or a Management Group and allows all actions (``++*++``)
-¨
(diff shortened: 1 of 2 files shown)


rules/S6385/description.adoc
```diff
@@ -1,5 +1,2 @@
 Defining a custom role for a Subscription or a Management group that allows all actions will give them the same capabilities as the built-in Owner role.
 It's recommended to limit the number of subscription owners in order to mitigate the risk of being breached by a compromised owner.
```
loris-s-sonarsource, line 2:
```suggestion

To reduce the risk of intrusion of a compromised owner, it is recommended to limit the number of subscription owners.
```

Contractions are not compliant, "it's" and equivalents should be uncontracted to "it is". After fixing it, Grammarly suggested something and I got struck by inspiration

Important: This should not be in description.adoc, but should be in the "how can I fix" section

rules/S6385/description.adoc
```diff
@@ -1,5 +1,2 @@
...
-
-This rule raises an issue when a custom role has an assignable scope set to a Subscription or a Management Group and allows all actions (``++*++``)
-¨
```
loris-s-sonarsource, line 5: You can keep this, it is still interesting to understand the rule

loris-s-sonarsource (changes requested):
The change I request is mostly for the mitigation part

loris-s-sonarsource (approved):
LGTM! This rule is not LayC compliant but it should be tackled during a sprint in the future

PR #3102, opened 2023-09-19
Modify rule S6759: Restrict the scope to functional components

yassin-kammoun-sonarsource:
I am removing the JavaScript part of the initial RSPEC about Flow.js. Although we do support Flow.js to some extent, we limit ourselves to parsing its specific syntax. However, we don't have rule implementations that visit AST node types specific to Flow.js.


rules/S6759/javascript/rule.adoc
```diff
@@ -1,82 +1,51 @@
...
-----
-type Props = {
-  +name: string,
-}
-class Welcome extends React.Component<Props> {
-  render () {
-    return <div>Hello {this.props.name}</div>;
-  }
+function Welcome(props: Props) { // Noncompliant: The component props are not read-only
+  return <div>Hello {this.props.name}</div>;
```
alexander-kamushkin-sonarsource, line 16:
```suggestion
  return <div>Hello {props.name}</div>;
```

rules/S6759/javascript/rule.adoc
```diff
@@ -1,82 +1,51 @@
...
+interface Props {
+  name: string;
 }
-class Welcome extends React.Component<Props> {
-  render () {
-    return <div>Hello {this.props.name}</div>;
-  }
+
+function Welcome(props: Readonly<Props>) {
+  return <div>Hello {this.props.name}</div>;
```
alexander-kamushkin-sonarsource, line 29:
```suggestion
  return <div>Hello {props.name}</div>;
```

rules/S6759/javascript/rule.adoc
```diff
@@ -1,82 +1,51 @@
...
+interface Props {
   readonly name: string;
 }
-class Welcome extends React.Component<Props> {
-  render () {
-    return <div>Hello {this.props.name}</div>;
-  }
+
+function Welcome(props: Props) {
+  return <div>Hello {this.props.name}</div>;
```
alexander-kamushkin-sonarsource, line 42:
```suggestion
  return <div>Hello {props.name}</div>;
```

alexander-kamushkin-sonarsource (changes requested):

rules/S6759/javascript/rule.adoc
yassin-kammoun-sonarsource: Thank you so much for catching this and the rest!

alexander-kamushkin-sonarsource (approved):

PR #3111, opened 2023-09-20
Modify rule S1301: LaYC - switch replacable by if

frederic-tingaud-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1301/cfamily/metadata.json
+++ b/rules/S1301/cfamily/metadata.json
@@ -1,4 +1,5 @@
 {
+  "title": "\"switch\" statements shouldn't be used when \"if\" would be simpler",
   "tags": [
     "based-on-misra",
     "bad-practice"
--- a/rules/S1301/cfamily/rule.adoc
+++ b/rules/S1301/cfamily/rule.adoc
@@ -1,20 +1,60 @@
 == Why is this an issue?
 
-``++switch++`` statements are useful when there are many different cases depending on the value of the same expression. For just one or two cases however, the code will be more readable with ``++if++`` statements.
+`switch` statements are useful when there are many different cases depending on the value of the same expression. For just one or two cases, however, the code will be more readable with `if` statements.
 
 
-Moreover, ``++if++`` statements are obviously more suitable when the condition of the ``++switch++`` is boolean.
+In particular, `if` statements are more suitable when the condition of the `switch` is boolean.
 
 
-Here are the rules to count the cases:
+This rule detects statements that could be simplified with a simple `if / else`. That is when there is no more than one `case`, not counting those that fall through to a `default`.
 
-* ``++default++`` is counted as a case.
-* If there is no ``++default++`` clause, the ``++case++`` count is incremented by one (to account for the ``++else++`` branch of an equivalent ``++if++``).
-* All the cases falling through to ``++default++`` are not counted (they would all be the ``++else++`` branch of the equivalent ``++if++``).
+The following code:
 
-include::../noncompliant.adoc[]
+[source,cpp]
+----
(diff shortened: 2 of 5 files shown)


rules/S1301/cfamily/metadata.json
```diff
@@ -1,4 +1,5 @@
 {
+  "title": "\"switch\" statements shouldn't be used when \"if\" would be simpler",
```
marco-antognini-sonarsource, line 2: It's already an improvement over the previous title. I wonder though, should we try phrasing it in a positive way? "if statement should be preferred over equivalent switch statement" (I'm not 100% happy with this suggestion...)

WDYT?

Regardless, the PR's title should probably make this change more visible, no?

rules/S1301/cfamily/rule.adoc
```diff
@@ -1,20 +1,60 @@
...
-include::../noncompliant.adoc[]
-
-include::../compliant.adoc[]
+`switch` statements are useful when there are many different cases depending on the value of the same expression. For just one or two cases, however, the code will be more readable with `if` statements.
+
+
+In particular, `if` statements are more suitable when the condition of the `switch` is boolean.
+
+
+This rule detects statements that could be simplified with a simple `if / else`. That is when there is no more than one `case`, not counting those that fall through to a `default`.
```
marco-antognini-sonarsource, line 9:
```suggestion
This rule detects statements that could be simplified with a simple `if / else`. That is when there is at most one `case`, not counting those that fall through to a `default`.
```

rules/S1301/cfamily/rule.adoc
```diff
@@ -1,20 +1,60 @@
...
+[source,cpp]
+----
+if (variable == 0) {
+  doSomething();
+} else {
+  doSomethingElse();
+}
+----
+
+While the following codes don't trigger the rule, because using `if` would not improve their readability:
```
marco-antognini-sonarsource, line 38: "Code" is a mass noun in this context. Snippets may be better. There is also an extra comma.

```suggestion
While the following snippets don't trigger the rule because using `if` would not improve their readability:
```

rules/S1301/compliant.adoc
```diff
@@ -1,6 +1,6 @@
 === Compliant solution
 
-[source,text]
+[source,java]
```
marco-antognini-sonarsource, line 3: `compliant.adoc` and `noncompliant.adoc` are used in flex too. I do not know that language. I have no idea if `java` is appropriate here... Was this a conscious decision? It may be fine, I just want to double-check.

marco-antognini-sonarsource (changes requested):
a couple of nits -- the new wording is much better I think!

rules/S1301/compliant.adoc
frederic-tingaud-sonarsource: I don't know Flex either, but I guess if they use this sample, it means it looks rather the same and all of the languages would benefit from java coloring of the keywords rather than no coloring at all.

rules/S1301/cfamily/metadata.json
frederic-tingaud-sonarsource: what do you think about this new version?

marco-antognini-sonarsource: Looks good! 👍

rules/S1301/compliant.adoc
marco-antognini-sonarsource: Sounds reasonable. They can improve it later if they want to.

marco-antognini-sonarsource (approved):
LGTM

PR #3118, opened 2023-09-21
Fix Sonar warnings

frederic-tingaud-sonarsource:
Fix a bunch of Sonar warnings that somehow appear as "New warnings" although they are a few years old.


frontend/src/utils/useRuleCoverage.ts
```diff
@@ -43,7 +42,7 @@ export function useRuleCoverage() {
   const coveredRulesUrl = `${process.env.PUBLIC_URL}/covered_rules.json`;
   const [coveredRules, coveredRulesError, coveredRulesIsLoading] = useFetch<RuleCoverage>(coveredRulesUrl);
 
-  function ruleCoverageForSonarpediaKeys(languageKeys: string[], ruleKeys: string[], mapper: any) {
+  function ruleCoverageForSonarpediaKeys(languageKeys: string[], ruleKeys: string[], mapper: ((key: string, value: Version) => JSX.Element)): string | JSX.Element[] {
```
marco-antognini-sonarsource, line 45: A type alias `Mapper` (or whatever the name) would be more readable I think, especially since it's used multiple times.

rspec-tools/rspec_tools/checklinks.py
```diff
@@ -15,12 +15,7 @@
 link_probes_history = {}
 
 # These links consistently fail in CI, but work-on-my-machine
-EXCEPTIONS = ['https://blogs.oracle.com/java-platform-group/diagnosing-tls,-ssl,-and-https',
-              'https://blogs.oracle.com/oraclemagazine/oracle-10g-adds-more-to-forall',
-              # FIXME: RULEAPI-763 - investigate why these live links respond with 403
-              'https://medium.com/@cowtowncoder/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062',
-              'https://medium.com/@nyomanpradipta120/ssti-in-flask-jinja2-20b068fdaeee',
-              'https://elizarov.medium.com/coroutine-context-and-scope-c8b255d59055']
+EXCEPTIONS = ['https://blogs.oracle.com/java-platform-group/diagnosing-tls,-ssl,-and-https']
```
marco-antognini-sonarsource, line 18: It seems the link is now `https://blogs.oracle.com/java/post/diagnosing-tls-ssl-and-https` (no commas, different directory). Maybe it works with the CI? I'm fine if this is not done in this PR though.

Otherwise, I'd suggest writing the list with one item per line to simplify future diffs.
```py
EXCEPTIONS = [
  'https://blogs.oracle.com/java-platform-group/diagnosing-tls,-ssl,-and-https',
]
```

marco-antognini-sonarsource (approved):
LGTM, two small nits.

rspec-tools/rspec_tools/utils.py
```diff
@@ -147,14 +146,14 @@ def get_label_for_language(language: str) -> str:
   return LANG_TO_LABEL[language]
 
 def resolve_rule(rule_id: str) -> int:
-  m = re.search('^S([0-9]{3,4})$', rule_id)
+  m = re.search(r'^S(\d{3,4})$', rule_id)
```
necto, line 149: Should this support the MISRA rule IDs (`M##_###`)?
Sure we don't publish MISRA rules in this repo, but may be we will refer to them in our rules?

marco-antognini-sonarsource: Given how this function is used, modifying to support MISRA IDs would be significantly involved. I don't think we need the feature for now at least.

frederic-tingaud-sonarsource: Same as Marco, I don't think we should address MISRA at the moment.

PR #3123, opened 2023-09-21
Modify rule S6587 Add more detailed explanation, split code examples

jonas-wielage-sonarsource:
## Review

- [ ] text style and tone


rules/S6587/docker/rule.adoc
```diff
@@ -14,52 +14,97 @@ It should be reduced to speed up deployments and reduce storage and bandwidth.
 
 ==== Noncompliant code example
 
+For apk:
```
mstachniuk, line 17: For consideration: Maybe it makes sense to have:
```
== How to fix it in apk
== How to fix it in apt-get
...
```
But I'm not sure

mstachniuk (approved):
LGTM and consider my feedback

rules/S6587/docker/rule.adoc
jonas-wielage-sonarsource: I wouldn't do this. If we would do this consistently we would have a lot of frameworks for docker. Also this doesn't fit exactly as a framework IMO

PR #3127, opened 2023-09-22
Modify rule S6570: Convert to LaYC format

petertrr:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6570/docker/rule.adoc
+++ b/rules/S6570/docker/rule.adoc
@@ -12,9 +12,9 @@ This issue can lead to bugs if the variable contains sensitive characters, which
 
 == How can I fix it?
 
-=== Code examples
-
 Surround variable reference with double quotes.
+
+=== Code examples
 
 ==== Noncompliant code example
 
--- a/rules/S6570/docker/rule.adoc
+++ b/rules/S6570/docker/rule.adoc
@@ -8,46 +8,43 @@ This causes issues if the variable contains whitespaces or shell pathname expans
 
 === What is the potential impact?
 
-This issue can lead to bugs if the variable contains sensitive characters, which may not be interpreted correctly.
+This issue can lead to bugs if the variable contains sensitive characters, which may be interpreted incorrectly and thus lead to undesired behavior.
 
-== How to fix it
-Add the missing double quotes to this variable.
+== How can I fix it?
 
 === Code examples
 
+Surround variable reference with double quotes.
+
 ==== Noncompliant code example
 
-Demonstration of pathname expansion using the `echo` command:
+This example demonstrates pathname expansion using the `echo` command:
 [source,docker,diff-id=1,diff-type=noncompliant]
 ----
-RUN test="command t*.sh"
-RUN echo $test
+RUN test="command t*.sh" && echo $test
 ----
-This will print `"command temp1.sh temp2.sh"`, as `*` is substituted to matching files in the current folder.
-In this example the folder contains the files `temp1.sh` and `temp2.sh`.
+Suppose this code is executed in a directory that contains two files: `temp1.sh` and `temp2.sh`. This code will print `"command temp1.sh temp2.sh"`,
+as `*` is substituted with matching files in the current folder.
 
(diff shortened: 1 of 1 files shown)


rules/S6570/docker/rule.adoc
```diff
@@ -8,46 +8,40 @@ This causes issues if the variable contains whitespaces or shell pathname expans
...
 
-== How to fix it
-Add the missing double quotes to this variable.
+== How can I fix it?
+Surround variable reference with double quotes.
 
-=== Code examples
+=== Noncompliant code example
 
-==== Noncompliant code example
```
rudy-regazzoni-sonarsource, line 18: From this [LaYC presentation](https://docs.google.com/presentation/d/1KuDxBxK77v8dhLnn5BWqSDd9K0ezNmDmNgeKQ1Xb7ag/edit#slide=id.g21a51453c43_0_133), it seems that the previous format was the right one.
Did I miss something? :sweat_smile:

rules/S6570/docker/rule.adoc
```diff
@@ -8,46 +8,40 @@ This causes issues if the variable contains whitespaces or shell pathname expans
...
-
-Demonstration of pathname expansion using the `echo` command:
+This example demonstrates pathname expansion using the `echo` command:
 [source,docker,diff-id=1,diff-type=noncompliant]
 ----
-RUN test="command t*.sh"
-RUN echo $test
+RUN test="command t*.sh" && echo $test
 ----
```
rudy-regazzoni-sonarsource, line 22: I couldn't test this because my docker engine does not want to start anymore... But I'm expecting this to not be working as expected: the variable test would be defined at shell level, but the $test in the echo command would be resolved first to empty string at docker level.
I may be wrong on this but I would recommend checking this.
If that's confirmed, I would recommend using an example with `ARG` instead of setting a variable through RUN.

rudy-regazzoni-sonarsource (commented):
Incomplete review, please still have a look at my comments.

rules/S6570/docker/rule.adoc
petertrr: Exactly this was discussed in my previous PR: https://github.com/SonarSource/rspec/pull/3071#discussion_r1325841207 :thinking: I believe I saw in one of the documents format without `=== Code examples`

petertrr: [Here](https://docs.sonarsource.com/sonarqube/latest/extension-guide/adding-coding-rules/#guidelines-for-bug-vulnerability-and-code-smell-rules) for example, there is directly code subsections inside `How can I fix it` section

rules/S6570/docker/rule.adoc
petertrr: I think, that Docker realizes that it doesn't have a variable with this name and doesn't attempt the substitution. The example works as expected:
```bash
$ cat Dockerfile
...
RUN touch test.sh tost.sh && test="command t*.sh" && echo $test
...

$ docker build --build-arg=FOO=foo --no-cache --progress=plain .
...
#5 [2/2] RUN touch test.sh tost.sh && test="command t*.sh" && echo $test
#5 0.284 command test.sh tost.sh
#5 DONE 0.3s
...
```

rules/S6570/docker/rule.adoc
```diff
@@ -8,46 +8,43 @@ This causes issues if the variable contains whitespaces or shell pathname expans
...
-This issue can lead to bugs if the variable contains sensitive characters, which may not be interpreted correctly.
+This issue can lead to bugs if the variable contains sensitive characters, which may be interpreted incorrectly and thus lead to undesired behavior.
 
-== How to fix it
-Add the missing double quotes to this variable.
+== How can I fix it?
 
 === Code examples
 
+Surround variable reference with double quotes.
```
jonas-wielage-sonarsource, line 17: I prefer this to be directly under `How to fix it` instead of `Code examples`. Nevertheless, I like the wording better than what it was before

jonas-wielage-sonarsource (approved):
LGTM, please see my suggestion

PR #3145, opened 2023-09-25
Modify rule S2637: Expand and adjust for LaYC

pdschbrt:
## Review

- [x] text style and tone


rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,138 @@
...
+  x = 0; // Noncompliant: `x` is marked "nonnull" but is set to null
+  foo(0); // Noncompliant: null is passed as an argument marked as "nonnull"
+  return 0; // Noncompliant: return value is marked "nonnull" but null is returned
 }
 ----
 
+The ``++nonnull++`` attribute is meant for other developers and values marked as ``++nonnull++`` are assumed to have non-null values.
+In particular, the compiler does not enforce that values marked as ``++nonnull++`` are indeed non-null at runtime; it is the developers' responsibility to honor the attribute.
+These values are typically _not_ null-checked before use.
+Setting a value marked as ``++nonnull++`` to null (i.e., ``++NULL++``, ``++0++`` or ``++nullptr++``) is hence likely to cause null pointer dereferences at runtime.
```
necto, line 22:
```suggestion
Setting a value marked as ``++nonnull++`` to null (i.e., ``++NULL++``, ``++0++`` or ``++nullptr++``) is hence likely to cause null-pointer dereferences at runtime.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,138 @@
...
+----
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
+void process(int *p);
+
+void foo(__attribute__((nonnull)) int *p) {
+  p = nullptr; // Noncompliant: `p` is marked "nonnull" but is set to null
```
necto, line 102: To avoid confusion, I would move the declaration of `p` out of the parameter list
```suggestion
void foo() {
  __attribute__((nonnull)) int *p = bar();
  p = nullptr; // Noncompliant: `p` is marked "nonnull" but is set to null
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,138 @@
...
+Setting a value marked as ``++nonnull++`` to null (i.e., ``++NULL++``, ``++0++`` or ``++nullptr++``) is hence likely to cause null pointer dereferences at runtime.
+
+
+== What is the potential impact?
+
+The behavior of a program that dereferences a null pointer is undefined. In that case, the compiler no longer needs to adhere to the language standard and the program has no meaning assigned to it.
+
+In practice, dereferencing a null pointer may lead to program crashes, or the application may appear to execute correctly while losing data or producing incorrect results.
+
+Besides affecting the application’s availability, null pointer dereferences may lead to code execution, in rare circumstances. If null is equivalent to the 0x0 memory address that can be accessed by privileged code, writing, and reading memory is possible, which compromises the integrity and confidentiality of the application.
```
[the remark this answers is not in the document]
necto: Have you heard of specific examples of this?

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,138 @@
+Pointers marked as "nonnull" may not be set to null.
+
 == Why is this an issue?
```
necto, line 3: Can you summarize in 1-2 sentences why it is an issue, i.e., what's wrong with it?

necto (changes requested):
In "Going the extra mile" you can also recommend applying the attribute to your code to propagate the property as far as possible.

rules/S2637/cfamily/rule.adoc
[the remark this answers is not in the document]
pdschbrt: Yes, there is a CWE or CERT-related article about that. I can look it up, in case you're interested.

This is only relevant for special platforms on which `0x0` is actually a valid memory address, and not for our everyday architectures. However, I thought it might be good to mention it here since it serves as a reminder not to make too many assumptions.

rules/S2637/cfamily/rule.adoc
pdschbrt: As far as I know, the 'nonnull' attribute only applies to functions, methods, and parameters, so we need to keep it part of the parameter list.

rules/S2637/cfamily/rule.adoc
pdschbrt: Done.

pdschbrt:
> In "Going the extra mile" you can also recommend applying the attribute to your code to propagate the property as far as possible.

Since one cannot apply the attribute to local pointer variables, I don't think this recommendation applies. We also don't wish to promote the idea of using the "nonnull" attribute blindly all over the place.

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
+Pointers marked as "nonnull" may not be set to null, since they are typically not null-check before use.
```
necto, line 1:
```suggestion
Pointers marked as "nonnull" may not be set to null, since they are typically not null-checked before use.
```

rules/S2637/cfamily/rule.adoc
necto: Can you, please, point me to it?

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
...
+int* foo(__attribute__((nonnull)) int* x) {
+  x = 0; // Noncompliant: `x` is marked "nonnull" but is set to null
+  foo(0); // Noncompliant: null is passed as an argument marked as "nonnull"
+  return 0; // Noncompliant: return value is marked "nonnull" but null is returned
 }
 ----
 
+The ``++nonnull++`` attribute is meant for other developers and values marked as ``++nonnull++`` are assumed to have non-null values.
+In particular, the compiler does not enforce that values marked as ``++nonnull++`` are indeed non-null at runtime; it is the developers' responsibility to honor the attribute.
+These values are typically _not_ null-checked before use.
```
necto, line 21: I would also mention, that the compiler applies optimizations based on this attribute, and might, for example, *remove* a null-check if the parameter is declared as `nonnull` (even in the code outside of the function with the attribute)

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
...
+Setting a value marked as ``++nonnull++`` to null (i.e., ``++NULL++``, ``++0++`` or ``++nullptr++``) is hence likely to cause null-pointer dereferences at runtime.
```
necto, line 22: Usually there can't be more than one null-pointer dereference before the application crashes
```suggestion
Setting a value marked as ``++nonnull++`` to null (i.e., ``++NULL++``, ``++0++`` or ``++nullptr++``) is hence likely to cause null-pointer a dereference.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
...
+
+
+== What is the potential impact?
```
necto, line 25: This section deals only with the impact of a null-pointer dereference. However, I believe the rule checks a violation of a contract with the compiler.
As the compiler applies optimizations based on this attribute, there may be no null-pointer dereference but other scary things if the developer does not respect the attribute.
For example, if `arg` is declared `nonnull`, the compiler might completely remove the code after the loop, leading to a file-descriptor leak:
```
void f(__attribute__((nonnull)) Node* arg) {
  while(arg) {
    arg = arg->next;
  }
  fclose(file);
}
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
...
+
+In practice, dereferencing a null pointer may lead to program crashes, or the application may appear to execute correctly while losing data or producing incorrect results.
+
+Besides affecting the application’s availability, null pointer dereferences may lead to malicious code execution, in rare circumstances.
+If null is equivalent to the 0x0 memory address that can be accessed by privileged code, writing, and reading memory is possible, which compromises the integrity and confidentiality of the application.
+
+
+== How to fix it
+
+Ensure not to pass null values when non-null arguments are expected.
```
necto, line 37: This is the most common fix (as the attribute will likely be in a library function).
However, sometimes it is more appropriate to remove the attribute.

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
+Pointers marked as "nonnull" may not be set to null, since they are typically not null-check before use.
+
 == Why is this an issue?
 
-Functions return values and parameters values marked ``++nonnull++`` are assumed to have non-null values and are not typically null-checked before use. Therefore setting one of these values to ``++null++``, could cause null pointer dereferences at runtime.
+A function's return value and parameters may be decorated with attributes to convey additional information to the compiler and/or other developers.
 
-=== Noncompliant code example
+A commonly used attribute is ``++nonnull++`` which can be used to mark a function's return value and parameters as shown in the following:
```
necto, line 7: In the following example, I expect to see the canonical use of `nonnull`, but I see immediately a broken use.
I suggest first showing how `nonnull` is expected to be used and then adding this code snippet demonstrating how it can be broken.

shared_content/cfamily/reference_over_nonnull_pointer.adoc
```diff
@@ -0,0 +1,14 @@
+In {cpp}, it is preferred to use reference parameters (Type const& or Type&), or pass objects by value (Type), instead of a pointer (Type const* or Type*), if the argument is expected to never be null.
```
necto, line 1: You can link [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-inout) here or in the "Resources" section:

- F.17: For "in-out" parameters, pass by reference to non-const
- F.16: For "in" parameters, pass cheaply-copied types by value and others by reference to const

necto (changes requested):
> Since one cannot apply the attribute to local pointer variables, I don't think this recommendation applies. We also don't wish to promote the idea of using the "nonnull" attribute blindly all over the place.

I meant, not to apply it blindly, but propagate it to the parameter declarations of the function that passes its parameters to a function taking `nonnull` arguments.

Also, I think it is worth mentioning that this `nonnull` attribute is a non-standard (either C or C++) GNU extension.

Please add a link to the GNU documentation of this attribute:
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute
and 
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-returns_005fnonnull-function-attribute

rules/S2637/cfamily/rule.adoc
pdschbrt: Good point.

shared_content/cfamily/reference_over_nonnull_pointer.adoc
pdschbrt: Oh yes. That's always a good resource.

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,139 @@
+Pointers marked as "nonnull" may not be set to null, since they are typically not null-checked before use.
```
pdschbrt, line 1: @necto here is the extended sentence. I'm not sure why github swallowed it somehow.

rules/S2637/cfamily/rule.adoc
pdschbrt: I've commented on that line further down the conversation. I'm not sure why github swallowed this change.

shared_content/cfamily/reference_over_nonnull_pointer.adoc
```diff
@@ -0,0 +1,16 @@
...
+---- 
+
+The preferred signature would be:
+
+[source,cpp]
+----
+bool isDAG(Graph const& graph);
+---- 
+
+For more details refer to {cpp} Code Guidelines https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.16: For "in" parameters, pass cheaply-copied types by value and others by reference to const] and https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.17: For "in-out" parameters, pass by reference to non-const].
```
necto, line 16: I think this part is better left for the "Resources" section. That means it cannot be in the same shared file, unfortunately, but you can create another shared_content file with the pointers, and include them into Resources of both S2637 and S2259. I leave it to your judgement if that will yield more readable rules.

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,188 @@
...
+  }
+  return l;
+}
+----
+
+
+== How to fix it
+
+Ensure not to pass null values when non-null arguments are expected.
+This especially holds for library functions such as the C standard library, for instance, which frequently requires `nonnull` pointer parameters.
```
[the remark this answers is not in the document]
pdschbrt: We could speak of GCC's implementation of the C standard library. But we shouldn't get lost in the details. Let's make it an easier read.

shared_content/cfamily/reference_over_nonnull_pointer.adoc
pdschbrt: I've duplicated them for now to avoid creating yet another shared file.

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,191 @@
...
+}
+----
+
+
+=== Going the extra mile
+
+include::../../../shared_content/cfamily/reference_over_nonnull_pointer.adoc[]
+
+
+== Resources
```
necto, line 174: It is surprisingto have half of the "Resources" section here and half in the "see.adoc". I would put it either all here inline of rule.adoc, or all in the "see.adoc" (which can be renamed to "resources.adoc").
Granted, the "../see.adoc" is shared with Java, so you can either duplicate these 3 links there (btw, it should not be directly under "Resources", but in a subsection), or call the file "Standards" and include the entire "Standards" section there

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,191 @@
...
+
+=== Standards
+
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.16: For "in" parameters, pass cheaply-copied types by value and others by reference to const]
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.17: For "in-out" parameters, pass by reference to non-const]
```
necto, line 179: I'd say these are "External coding guidelines" rather than "Standards"

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,191 @@
...
+=== Going the extra mile
+
+include::../../../shared_content/cfamily/reference_over_nonnull_pointer.adoc[]
+
+
+== Resources
+
+=== Standards
+
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.16: For "in" parameters, pass cheaply-copied types by value and others by reference to const]
```
necto, line 178: Please add these two links also to S2259, IMO they are as relevant there

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,191 @@
...
+}
+----
+
+Failing to adhere to the attribute may introduce serious program errors.
+In particular, the compiler does not enforce that values marked as `nonnull` are indeed non-null at runtime; it is the developers' responsibility to honor the attribute.
+These values are typically _not_ null-checked before use.
+Setting a value marked as `nonnull` to null (i.e., `NULL`, `0` or `nullptr`) is hence likely to cause a null-pointer dereference.
+Compilers may even apply optimizations based on this attribute and might, for instance, _remove_ an explicit null-check if the parameter is declared as `nonnull`---even in code outside of the function with the attribute.
+
+Note that the `nonnull` attribute is a GNU extension (cf. https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute[nonnull] and https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-returns_005fnonnull-function-attribute[returns_nonnull]) which many compiler vendors have implemented.
```
necto, line 48: I did not see "cf." before. According to [wiki](https://en.wikipedia.org/wiki/Cf.) it is used to invite the reader to compare the written with the external reference. IMO, this is not an entirely accurate use here: you are pointing to the documentation of the attributes, how can one compare it with the statement "`nonnull` attribute is a GNU extension"?
So I would either remove "cf." or replaced it with, IMHO, more appropriate "see" (which is also more commonly used).
```suggestion
Note that the `nonnull` attribute is a GNU extension (https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute[nonnull] and https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-returns_005fnonnull-function-attribute[returns_nonnull]) which many compiler vendors have implemented.
```
or
```suggestion
Note that the `nonnull` attribute is a GNU extension (see https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute[nonnull] and https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-returns_005fnonnull-function-attribute[returns_nonnull]) which many compiler vendors have implemented.
```

necto (changes requested):

rules/S2637/cfamily/rule.adoc
pdschbrt: Yes, that's indeed a much better place for these two links.

rules/S2637/cfamily/rule.adoc
pdschbrt: This becomes obsolete when addressing the next point.

rules/S2637/cfamily/rule.adoc
pdschbrt: They are already there, but I'll move them to an `External coding guidelines` section.

rules/S2637/cfamily/rule.adoc
pdschbrt: I'm fine with using "see" instead. Will change that.

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+
+=== Going the extra mile
+
+include::../../../shared_content/cfamily/reference_over_nonnull_pointer.adoc[]
+
+
+== Resources
+
+=== Standards
+
```
necto, line 177: Note that this subtitle is now missing in the Java rule.
I propose to put it directly into the see.adoc (which would be more appropriately named "standards.adoc")

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+  x = 0; // Noncompliant: `x` is marked "nonnull" but is set to null
+  foo(0); // Noncompliant: null is passed as an argument marked as "nonnull"
+  return 0; // Noncompliant: return value is marked "nonnull" but null is returned
 }
 ----
 
+Failing to adhere to the attribute may introduce serious program errors.
+In particular, the compiler does not enforce that values marked as `nonnull` are indeed non-null at runtime; it is the developers' responsibility to honor the attribute.
+These values are typically _not_ null-checked before use.
+Setting a value marked as `nonnull` to null (i.e., `NULL`, `0` or `nullptr`) is hence likely to cause a null-pointer dereference.
```
necto, line 45: You are using two phrases to, as I understand it, refer to the same thing: "adhere to the attribute" = "honor the attribute".
I think this is not necessary and detrimental in technical writing because it implies there is a subtle difference where there's none

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+Compilers may even apply optimizations based on this attribute and might, for instance, _remove_ an explicit null-check if the parameter is declared as `nonnull`---even in code outside of the function with the attribute.
```
necto, line 46: triple-dash strikes again (— = ` -- ` - surrounded by spaces)
```suggestion
Compilers may even apply optimizations based on this attribute and might, for instance, _remove_ an explicit null-check if the parameter is declared as `nonnull` -- even in code outside of the function with the attribute.
```
Also, it is convenient to break long sentences into multiple lines at the point of natural pauses (punctuation). For example:
```suggestion
Compilers may even apply optimizations based on this attribute and might,
for instance, _remove_ an explicit null-check if the parameter is declared as `nonnull` --
even in code outside of the function with the attribute.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+
+
+== What is the potential impact?
+
+In case a program dereferences a null pointer, it's behavior is undefined.
+For programs that exercise undefined behavior the compiler no longer needs to adhere to the language standard and the program has no meaning assigned to it.
+
+In practice, dereferencing a null pointer may lead to program crashes, or the application may appear to execute correctly while losing data or producing incorrect results.
+
+Besides affecting the application's availability, null pointer dereferences may lead to malicious code execution, in rare circumstances.
```
necto, line 58:
```suggestion
Besides affecting the application's availability, null-pointer dereferences may lead to malicious code execution, in rare circumstances.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
 
+=== External coding guidelines
+
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.16: For "in" parameters, pass cheaply-copied types by value and others by reference to const]
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md[F.17: For "in-out" parameters, pass by reference to non-const]
+
+
+=== Related rules
+
+* S2259 detects null pointer dereferences
```
necto, line 188:
```suggestion
* S2259 detects null-pointer dereferences
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+
+== What is the potential impact?
+
+In case a program dereferences a null pointer, it's behavior is undefined.
+For programs that exercise undefined behavior the compiler no longer needs to adhere to the language standard and the program has no meaning assigned to it.
+
+In practice, dereferencing a null pointer may lead to program crashes, or the application may appear to execute correctly while losing data or producing incorrect results.
+
+Besides affecting the application's availability, null pointer dereferences may lead to malicious code execution, in rare circumstances.
+If null is equivalent to the 0x0 memory address that can be accessed by privileged code, writing, and reading memory is possible, which compromises the integrity and confidentiality of the application.
```
necto, line 59: Do you have a reference for this kind of exploit?
I find it hard to believe

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+
+Because compilers may apply optimizations based on the `nonnull` attribute, not respecting `nonnull` can also introduce more complex bugs such as resource leaks or infinite loops as indicated in the following code snippet:
+
+[source,cpp]
+----
+struct Node {
+  int data;
+  Node *next;
```
necto, line 67:
```suggestion
  Node *next; // NULL for a tail node.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
+  }
+  return l;
+}
+----
+
+
+== How to fix it
+
+Ensure not to pass null values when non-null arguments are expected.
+This especially holds for library functions, which frequently require `nonnull` pointer parameters.
```
necto, line 84: In the "Why this is an issue" you also mentioned another fault "assigning null to a nonnull pointer", and it is listed in the code examples. Should you also mention it here (note that it is different from "passing null values to non-null arguments")?

shared_content/cfamily/reference_over_nonnull_pointer.adoc
necto, line 1:
```suggestion
In {cpp}, it is preferred to use reference parameters (`Type const&` or `Type&`), or pass objects by value (`Type`), instead of a pointer (`Type const*` or `Type*`), if the argument is expected to never be null.
```

necto (changes requested):

rules/S2637/cfamily/rule.adoc
```diff
@@ -1,21 +1,194 @@
...
-  return 0; // Noncompliant - This function's return value is marked "nonnull" but null is returned.
+int* foo(__attribute__((nonnull)) int* x) {
+  x = 0; // Noncompliant: `x` is marked "nonnull" but is set to null
+  foo(0); // Noncompliant: null is passed as an argument marked as "nonnull"
+  return 0; // Noncompliant: return value is marked "nonnull" but null is returned
 }
 ----
 
+Failing to adhere to the attribute may introduce serious program errors.
+In particular, the compiler does not enforce that values marked as `nonnull` are indeed non-null at runtime; it is the developers' responsibility to adhere the attribute.
```
necto, line 43:
```suggestion
In particular, the compiler does not enforce that values marked as `nonnull` are indeed non-null at runtime; it is the developers' responsibility to adhere to the attribute.
```

rules/S2637/cfamily/rule.adoc
pdschbrt: You can find more info on that at https://wiki.sei.cmu.edu/confluence/display/c/EXP34-C.+Do+not+dereference+null+pointers or https://cwe.mitre.org/data/definitions/476

These two links are referenced in the "Resources --> Standards" section.

rules/S2637/cfamily/rule.adoc
necto: Sorry for going back and forth on it, and thank you for reconciling me with myself :D

rules/S2637/cfamily/rule.adoc
necto: Nice! thank you, I've learned something new

necto (approved):
Nicely done!

PR #3146, opened 2023-09-25
Update C++ rules

amelie-renard-sonarsource:


rules/S6494/cfamily/rule.adoc
```diff
@@ -60,6 +57,6 @@ char const* localizedFormatString(unsigned id);
 snprintf(buffer, localizedFormatString(123), 10, 20)
 ----
 
-While `std::vformat` may be used in such cases, it will require an change of the format string,
+While `std::vformat` may be used in such cases, it will require a change of the format string,
```
frederic-tingaud-sonarsource, line 60:
```suggestion
While `std::vformat` may be used in such cases, it would require a change of the format string,
```

rules/S1155/cfamily/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Do not check emptiness with a size method when a dedicated function exists",
+  "title": "\"empty()\" or \"is_empty()\" should be used to test for emptiness",
```
frederic-tingaud-sonarsource, line 2: What about "Dedicated method should be used to test for emptiness"?

rules/S6494/cfamily/rule.adoc
```diff
@@ -50,7 +47,7 @@ void printFunc(FILE* f, char* out, size_t n) {
```
frederic-tingaud-sonarsource, line 47: We could take the "Compliant because the rule doesn't apply" snippets outside and remove the `FILE* f` from the signature.

```suggestion
Other printf-like functions are not concerned by this rule
[source,cpp]
----
void printFunc(FILE* f) {
  printf("%i", 10);  // Compliant, no direct remplacment
  std::fprintf(f, "%f", 10.0);  // Compliant, no direct remplacment
}
----
```

frederic-tingaud-sonarsource (approved):
A few suggestions but I'm good with it.

rules/S1155/cfamily/metadata.json
amelie-renard-sonarsource: I prefer being more explicit on the methods to use, as only two exist.

rules/S6494/cfamily/rule.adoc
amelie-renard-sonarsource: I agree, fixed in https://github.com/SonarSource/rspec/pull/3146/commits/36d97f0498482a292fc280a0a6225dabdbde68da

PR #3147, opened 2023-09-25
Modify rule S3518: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone


rules/S3518/cfamily/rule.adoc
```diff
@@ -1,16 +1,87 @@
...
+#include <optional>
+
+std::optional<int> unsafe_division(int a, int b) {
+  if ((a == std::numeric_limits<int>::min()) && (b == -1)) {
+    return std::nullopt;
+  }
+  // Noncompliant: While the above check correctly prevents signed integer overflows,
+  // it fails to prevent divide-by-zero errors. If `b` is equal to `0`, the application
+  // emits undefined behavior.
+  return a / b;
```
necto, line 25:
```suggestion
  return a / b; // Noncompliant: While the above check correctly prevents signed integer overflows,
                // it fails to prevent divide-by-zero errors. If `b` is equal to `0`, the application
                // emits undefined behavior.
```
It probably does not matter now, but I would still put the "Noncompliant" marker on the line with the bug, or immediately before it. The reason is a potential later use by automated tools as a specification of a TP, like we do with the Julient benchmark.

rules/S3518/cfamily/rule.adoc
```diff
@@ -1,16 +1,87 @@
...
+
 == Why is this an issue?
 
-If the denominator to a division or modulo operation is zero it would result in a fatal error.
+The result of the ``++/++`` operator is the quotient from the division of the first arithmetic operand by the second arithmetic operand and the result of the ``++%++`` operator is the remainder from the division of the first arithmetic operand by the second arithmetic operand.
+Division operations are susceptible to divide-by-zero (and signed integer overflow) errors.
+
+If the denominator to a division or modulo operation is zero, the behavior of the application is undefined.
+
+Furthermore, when using signed integer division and the dividend is the minimum value for the signed integer type and the divisor is equal to ``++-1++`` an overflow is provoked due to two's complement representation.
```
necto, line 10: This sounds unrelated to the div-by-0 rule. Is it also detected by our rule implementation?

rules/S3518/cfamily/rule.adoc
```diff
@@ -1,16 +1,87 @@
...
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+#include <limits>
+#include <optional>
+
+std::optional<int> unsafe_division(int a, int b) {
```
necto, line 52: This is odd name of the function. To me it looks rather like `safe_division`. Did you try to highlight that it misses the `/0`? The compliant snipped should be called `safe` then, but that would introduce irrelevant diff in the diff-view, so I would call it `safe_division` from the beginning.

rules/S3518/cfamily/rule.adoc
```diff
@@ -1,16 +1,87 @@
...
+  if ((a == std::numeric_limits<int>::min()) && (b == -1)) {
+    return std::nullopt;
+  }
+  return a / b; // Noncompliant: causes undefined behavior if `b` is zero
+}
```
necto, line 57: Add a caller code that exercises this div-by-0 possibility. Otherwise, our analyzer will not report, and admittedly, not every division must be checked against a `0` divisor because often it is known to be non-zero and the check introduces only a branch overhead.
Should be something non-obvious like
```
int foo(int a, int b) {
  if (b == 0) {
    a = 1;
  }
  return unsafe_division(a, b);
}
```
An alternative fix also should also be in the caller code, to avoid suggesting that you must check for 0 before *every* division.

necto (changes requested):
Note that in most cases single "`" ticks are enough to highlight a snippet as code. Reach out to heavier double-tick "``" and the `++` when the symbols in the snippet would break the formatting.
e.g.,
```
``++/++`` = `/`
```

Overall, I think your example `unsafe_division` is good but too advanced for the introduction. I would put it into "Going the extra mile", where you also note that you do not need to guard every division with a check for 0 divisor.
IMO, the primary motivating example should be the actual bug that exercises division by zero, not merely fails to protect against it.

rules/S3518/cfamily/rule.adoc
pdschbrt: That makes a lot of sense. I'll adjust it accordingly.

rules/S3518/cfamily/rule.adoc
pdschbrt: I'll restructure this slightly and also rename it.

rules/S3518/cfamily/rule.adoc
pdschbrt: That's a good suggestion. I've added this alternative.

rules/S3518/cfamily/rule.adoc
pdschbrt: I don't think it's implemented by this rule, unfortunately.

I've added this remark as div-by-`-1` is equally dangerous, but oftentimes overlooked. The materials I did go through also all mention that risk, including the relevant CERT pages. I strongly feel that this belongs here.

rules/S3518/cfamily/rule.adoc
```diff
@@ -1,16 +1,117 @@
...
+
 == Why is this an issue?
 
-If the denominator to a division or modulo operation is zero it would result in a fatal error.
+If the denominator to a division or modulo operation is zero, the behavior of the application is undefined.
+
+The result of the `/` operator is the quotient from the division of the first arithmetic operand by the second arithmetic operand and the result of the `%` operator is the remainder from the division of the first arithmetic operand by the second arithmetic operand.
+Division operations are susceptible to divide-by-zero (and signed integer overflow) errors.
+
+Furthermore, when using signed integer division and the dividend is the minimum value for the signed integer type and the divisor is equal to `-1` an overflow is provoked due to two's complement representation.
```
necto, line 10: This sentence now does not correspond to the code snippet.
I would put this, together with the code that correctly implements both checks (one of your compliant examples) into the "Going the extra mile" section.

I wonder if it is worth drawing the reader's attention to the integer-overflow issue before that

necto (changes requested):

necto:
You can also link [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md):

> ES.105: Don't divide by integer zero

BTW, they also mention another possible fix: replace integer division with a floating-point division, where the operation is defined. I think it makes sense to add here too

rules/S3518/cfamily/rule.adoc
pdschbrt: I've tried to move the detailed description on overflows to `Going the extra mile` section and I think that makes it a good read.

pdschbrt:
> You can also link [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md):
> 
> > ES.105: Don't divide by integer zero
> 
> BTW, they also mention another possible fix: replace integer division with a floating-point division, where the operation is defined. I think it makes sense to add here too

Yes, indeed. I've added a description for this alternative fix.

necto (approved):

PR #3150, opened 2023-09-25
Modify rule S1854: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone


pdschbrt:
Hey @necto 

I'm not yet happy with the updated rule description. I'm especially unhappy with `What is the potential impact?` section---any feedback is highly appreciated.

Cheers

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,147 @@
+Unused assignments should be removed.
+
 == Why is this an issue?
 
-include::../description.adoc[]
+Assigning a value to a local variable that is not read by any subsequent instruction is called a _dead store_.
```
necto, line 5: In the first paragraph, I would summarize in 1 or 2 sentences *why* this is an issue.
Something like "The only utility of assigning a value to a variable is to use this value later in the code. If it is never used, there is no need for the assignment. The presence of a useless assignment indicates misaligned intentions and is most likely an error". But preferably shorter.

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,147 @@
...
+
+The following code snippet depicts a dead store.
```
necto, line 7: IMO, this snippet contains too much code irrelevant to the issue.
I would limit it to ~5 lines. I would also add multiple snippets:
- assignment
  - the variable is never mentioned again
  - the variable is immediately overwritten
- increment (that you have)
- initialization

Perhaps most of them should be moved to the "noncompliant example"s

rules/S1854/cfamily/rule.adoc
necto: Your sentence after the snippet does the job well:

> Computing or retrieving a value only to then immediately overwrite it or throw it away indicates serious logic errors in the code.

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,147 @@
...
+  long f = 1;
+  for (long i = 1; i <= n; ++i) {
+    f *= i;
+  }
+  return f;
+}
+
+long foo(long n) {
+  // Noncompliant: `factorial`'s return value is unused and `factorial` has no side effects
+  factorial(n);
```
necto, line 102: Does our rule detect this too?

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,147 @@
...
+
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
+long factorial(const long n) {
+  long f = 1;
+  for (long i = 1; i <= n; ++i) {
+    f *= i;
+  }
+  return f;
+}
```
necto, line 98: Given that `factorial` is not the point of the example, I would choose a shorter function, e.g. 
```
long twice(long n) {
  return n*2;
}
```

necto (changes requested):
I think it is important to mention the side effects of the RHS. (In the "Pitfalls" section?)
For example, here the `fwrite` call is probably made for its side effect:
```
size_t written = fwrite(str, sizeof(*str), strlen(str), fd);
```

I would also mention the `(void)` trick to ignore the `[[nodiscard]]` return values:
```
[[nodiscard]] int f();

void g() {
  (void)f();
}
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,147 @@
...
+  int x = 0;
+  if (b) {
+    x = bar();
+    return x;
+  }
+  // Compliant: no more dead stores and superfluous code
+  return x;
+}
+----
+
```
necto, line 89: It looks like you are missing a bunch of "Noncompliant code example" headers

pdschbrt:
> I think it is important to mention the side effects of the RHS. (In the "Pitfalls" section?) For example, here the `fwrite` call is probably made for its side effect:
> 
> ```
> size_t written = fwrite(str, sizeof(*str), strlen(str), fd);
> ```
> 
> I would also mention the `(void)` trick to ignore the `[[nodiscard]]` return values:
> 
> ```
> [[nodiscard]] int f();
> 
> void g() {
>   (void)f();
> }
> ```

Sold! Thanks for the suggestions.

rules/S1854/cfamily/rule.adoc
pdschbrt: Sadly, it doesn't. I've simplified this example.

rules/S1854/cfamily/rule.adoc
pdschbrt: I've removed the function altogether since our rule does not try to reason about side effects.

rules/S1854/cfamily/rule.adoc
pdschbrt: Ups, yes I do.

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,190 @@
...
+int buz();
 
-=== Exceptions
+int foo(bool b) {
+  int x = 0;
+  if (b) {
+    x = bar();
+    return x;
+  }
+  // Compliant: no more dead stores and superfluous code
```
necto, line 99:
```suggestion
  // Compliant: no more dead stores and superfluous code
  // Assuming call to buz() had no important side effects
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,190 @@
...
-include::../description.adoc[]
+Computing or retrieving a value only to then immediately overwrite it or throw it away indicates serious logic errors in the code.
+
+Assigning a value to a local variable that is not read by any subsequent instruction is called a _dead store_.
+The following code snippet depicts a dead store.
+
+[source,cpp]
+----
+int foo() {
+  int x = 0;
```
necto, line 13:
```suggestion
  int x = 0; // Noncompliant: dead store, next line overwrites x
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,190 @@
...
 
-This rule ignores:
 
-* variable declarations initializers
-* prefix and postfix increments and decrements ``x{plus}{plus};``
-* null pointer assignments ``++x = NULL;++``
-* self assignments (i.e. ``++x = x;++``)
+=== Pitfalls
+
+When removing unused values and superfluous code, make sure that the right-hand side of a given assignment has no side effects that update the program's state.
```
necto, line 107: This is either a tautology or it is not complete: if you consider files on disk and the network packets as part of program state, than any side effect update program's state. If you don't, you should still consider these side effects even though they do not update program's state
```suggestion
When removing unused values and superfluous code, make sure that the right-hand side of a given assignment has no side effects.
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,190 @@
...
+#include <string.h>
+
+int square(int n) {
+  return n * n;
+}
+
+int foo(int i) {
+  int sq = square(i); // Noncompliant: dead store, assignment can be removed
+  const char* const str = "Hello, World!\n";
+  size_t n_items = fwrite(str, sizeof(char), strlen(str), stdout); // Noncompliant: `n_items` is never read, but the call has side effects
```
necto, line 125: this line is long.
I am not sure whether you can place the `Noncompliant` comment on the previous one, but you certainly can put the explanation in a separate comment on another line.

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,190 @@
...
+* CERT - https://wiki.sei.cmu.edu/confluence/x/39UxBQ[MSC13-C. Detect and remove unused values]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/9DZGBQ[MSC56-J. Detect and remove superfluous code and values]
+* CWE - https://cwe.mitre.org/data/definitions/563[563 - Assignment to Variable without Use ('Unused Variable')]
+
+=== Related rules
+
+* S1763 detects unreachable code
+* S2583 ensures that conditionally executed code is reachable
+* S2589 detects gratuitous boolean expressions
+* S3516 ensures that function returns are not invariant
```
necto, line 185: IMO, "function returns" as a noun, while grammatically correct, is pretty surprising and "function returns are" makes most readers stumble
```suggestion
* S3516 ensures that a function returns non-invariant
```

necto (changes requested):
Looks pretty solid. Only minor nits left

rules/S1854/cfamily/rule.adoc
pdschbrt: Fair point.

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,192 @@
+Unused assignments should be removed.
+
 == Why is this an issue?
 
-include::../description.adoc[]
+Computing or retrieving a value only to then immediately overwrite it or throw it away indicates a serious logic error in the code.
+
+Assigning a value to a local variable that is not read by any subsequent instruction is called a _dead store_.
+The following code snippet depicts a dead store.
```
necto, line 8:
```suggestion
The following code snippet depicts a few dead stores.
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,192 @@
...
+  int y = 0;
+  y += 9001; // Noncompliant: dead store, y is never used
+  int z = 300; // Noncompliant: dead store, next line overwrites z
+  z = 400;
+  return x + z * 2;
+}
+----
+
+Even if the unnecessary operations do not do any harm in terms of the program's correctness, they are--at best--a waste of computing resources.
+Therefore, unused values and superfluous code should be removed to prevent logic errors.
```
necto, line 25: The second sentence does not really follow from the first (which does not mention logic errors, and even seem to argue otherwise).
```suggestion
Even if the unnecessary operations do not do any harm in terms of the program's correctness, they are--at best--a waste of computing resources.
In most cases, the operations have their intended use but it is not expressed in the code correctly.
Therefore, unused values and superfluous code should be removed to prevent logic errors.
```
Or something similar.

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,192 @@
...
+
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+int foo(int y) {
+  int x = 0;
```
necto, line 48: Does the rule flag zero-initialization?
if it does, this marker is missing:
```suggestion
  int x = 0; // Noncompliant: dead store
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,192 @@
...
-* variable declarations initializers
-* prefix and postfix increments and decrements ``x{plus}{plus};``
-* null pointer assignments ``++x = NULL;++``
-* self assignments (i.e. ``++x = x;++``)
+=== Pitfalls
+
+When removing unused values and superfluous code, make sure that the right-hand side of a given assignment has no side effects.
+
+While it is safe to remove the call to `square` in the following code since it has no side effects, removing the call to `fwrite` changes the program's behavior.
+Still, values that are never read such as `n_items` indicate code smells that should be mitigated.
```
necto, line 111:
```suggestion
Still, values that are never read such as `n` indicate code smells that should be mitigated.
```

rules/S1854/cfamily/rule.adoc
```diff
@@ -1,25 +1,192 @@
...
+}
+----
+
+
+=== Going the extra mile
+
+In {cpp}17, the `nodiscard` attribute has been introduced which can be used to annotate functions, enumerations and classes.
+
+The attribute serves as a hint to the compiler and to other developers that a function's return value should not be ignored.
+A function that is marked `nodiscard` or that returns an enumeration or class declared `nodiscard` by value which is used in a discarded-value expression encourages the compiler to issue a warning.
```
necto, line 138: This sentence is pretty complex.
I propose removing some information from it, given it is already pretty remote to the topic of the rule:
```suggestion
A function that is marked `nodiscard`  whose return value is ignored encourages the compiler to issue a warning.
```
Otherwise, please simplify / break it down into simpler sentences

necto (approved):
LGTM! See the minor nits in line comments

rules/S1854/cfamily/rule.adoc
pdschbrt: I just checked, it doesn't.

PR #3152, opened 2023-09-26
Modify rule S5344: Change text to education framework format (APPSEC-1106)

sebastien-andrivet-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5344/java/rule.adoc
+++ b/rules/S5344/java/rule.adoc
@@ -63,13 +63,7 @@ public void configureGlobal(AuthenticationManagerBuilder auth, DataSource dataSo
 
 === How does this work?
 
-The ``BCryptPasswordEncoder`` is a password hashing function in Java that is designed to be secure and resistant to various types of attacks, including brute-force and rainbow table attacks. It is based on the Blowfish encryption algorithm and uses a technique called key stretching to make it computationally expensive to hash passwords. Here's how ``BCryptPasswordEncoder`` works:
-
-* Salt Generation: When a password is hashed using ``BCryptPasswordEncoder``, a random salt is generated. The salt is a random value that is added to the password before hashing, making it unique for each password. The salt is then stored along with the hashed password.
-
-* Password Hashing: The password, along with the salt, is then passed through the BCrypt hashing function. The function applies a series of iterations to the password and salt, creating a hash value that is unique to that combination.
-
-* Hash Comparison: When a user attempts to authenticate, the entered password is hashed using the same salt and number of iterations as the stored password. The generated hash is then compared with the stored hash. If they match, the password is considered valid.
+The ``BCryptPasswordEncoder`` is a password hashing function in Java that is designed to be secure and resistant to various types of attacks, including brute-force and rainbow table attacks. It is slow, adaptative, and automatically implements a salt.
 
 == Resources
 
--- a/rules/S5344/java/rule.adoc
+++ b/rules/S5344/java/rule.adoc
@@ -1,10 +1,43 @@
+The improper storage of passwords poses a significant security risk to software applications. This vulnerability arises when passwords are stored in plain-text or with a fast hashing algorithm. To exploit this vulnerability, an attacker would typically require access to the stored passwords or the ability to intercept them during transmission. 
+
 == Why is this an issue?
 
-include::../description.adoc[]
+By obtaining these passwords, the attacker can gain unauthorized access to user accounts, potentially leading to various malicious activities.
 
-=== Noncompliant code example
+=== What is the potential impact?
 
-[source,java]
+Plain-text or weakly hashed password storage poses a significant security risk to software applications.
+
+==== Unauthorized Access
+
+When passwords are stored in plain-text or with weak hashing algorithms, an attacker who gains access to the password database can easily retrieve and use the passwords to gain unauthorized access to user accounts. This can lead to various malicious activities, such as unauthorized data access, identity theft, or even financial fraud.
+
+==== Credential Reuse
+
+Many users tend to reuse passwords across multiple platforms. If an attacker obtains plain-text passwords or weakly hashed passwords, they can potentially use these credentials to gain unauthorized access to other accounts held by the same user. This can have far-reaching consequences, as sensitive personal information or critical systems may be compromised.
+
+==== Regulatory Compliance
+
+Many industries and jurisdictions have specific regulations and standards in place to protect user data and ensure its confidentiality. Storing passwords in plain-text or with weak hashing algorithms can lead to non-compliance with these regulations, potentially resulting in legal consequences, financial penalties, and damage to the reputation of the software application and its developers.
+
+
+== How to fix it in Java EE
+
+A user password should never be stored in clear-text, instead a hash should be produced from it using a secure algorithm:
+
+* not vulnerable to brute force attacks.
+* not vulnerable to collision attacks (see rule S4790).
+* and a salt should be added to the password to lower the risk of rainbow table attacks (see rule S2053).
+
+Some algorithms, like MD5 or SHA-family functions are fast to compute hashes and therefore brute force attacks are possible especially with hardware like GPU, FPGA or ASIC. Modern password hashing algorithms such as bcrypt, PBKDF2 or argon2 are recommended.
+
+=== Code examples
+
+==== Noncompliant code example
+
+The following code is vulnerable because it uses a legacy digest-based password encoding that is not considered secure.
+
+[source,java,diff-id=1,diff-type=noncompliant]
 ----
 @Autowired
 public void configureGlobal(AuthenticationManagerBuilder auth, DataSource dataSource) throws Exception {
@@ -12,22 +45,12 @@ public void configureGlobal(AuthenticationManagerBuilder auth, DataSource dataSo
     .dataSource(dataSource)
     .usersByUsernameQuery("SELECT * FROM users WHERE username = ?")
     .passwordEncoder(new StandardPasswordEncoder()); // Noncompliant
-
-  // OR
-  auth.jdbcAuthentication()
-    .dataSource(dataSource)
-    .usersByUsernameQuery("SELECT * FROM users WHERE username = ?"); // Noncompliant; default uses plain-text
-
-  // OR 
-  auth.userDetailsService(...); // Noncompliant; default uses plain-text
-  // OR 
-  auth.userDetailsService(...).passwordEncoder(new StandardPasswordEncoder()); // Noncompliant
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,java]
+[source,java,diff-id=1,diff-type=compliant]
 ----
 @Autowired
 public void configureGlobal(AuthenticationManagerBuilder auth, DataSource dataSource) throws Exception {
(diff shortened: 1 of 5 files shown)


rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
+The improper storage of passwords poses a significant security risk to software applications. This vulnerability arises when passwords are stored in plain-text or with a fast hashing algorithm. To exploit this vulnerability, an attacker would typically require access to the stored passwords or the ability to intercept them during transmission.
```
gaetan-ferry-sonarsource, line 1: I would not speak about transmission here, as in most cases passwords are transmitted in clear-text and THEN hashed. Moreover, the rule does not cover the security of passwords in transit.
```suggestion
The improper storage of passwords poses a significant security risk to software applications. This vulnerability arises when passwords are stored in plain-text or with a fast hashing algorithm. To exploit this vulnerability, an attacker typically requires access to the stored passwords. 
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
+The improper storage of passwords poses a significant security risk to software applications. This vulnerability arises when passwords are stored in plain-text or with a fast hashing algorithm. To exploit this vulnerability, an attacker would typically require access to the stored passwords or the ability to intercept them during transmission. 
+
 == Why is this an issue?
 
-include::../description.adoc[]
+By obtaining these passwords, the attacker can gain unauthorized access to user accounts, potentially leading to various malicious activities.
```
gaetan-ferry-sonarsource, line 5: I think we should emphasize that the attacker first needs to get access to the cold-stored passwords.

```suggestion
Attackers who would get access to the stored passwords could reuse them without further attacks or with little additional effort. Obtaining the clear-text passwords, they could then gain unauthorized access to user accounts, potentially leading to various malicious activities.
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+
+Plain-text or weakly hashed password storage poses a significant security risk to software applications.
+
+==== Unauthorized Access
+
+When passwords are stored in plain-text or with weak hashing algorithms, an attacker who gains access to the password database can easily retrieve and use the passwords to gain unauthorized access to user accounts. This can lead to various malicious activities, such as unauthorized data access, identity theft, or even financial fraud.
+
+==== Credential Reuse
+
+Many users tend to reuse passwords across multiple platforms. If an attacker obtains plain-text passwords or weakly hashed passwords, they can potentially use these credentials to gain unauthorized access to other accounts held by the same user. This can have far-reaching consequences, as sensitive personal information or critical systems may be compromised.
```
gaetan-ferry-sonarsource, line 17:
```suggestion
Many users tend to reuse passwords across multiple platforms. If an attacker obtains plain-text or weakly hashed passwords, they can potentially use these credentials to gain unauthorized access to other accounts held by the same user. This can have far-reaching consequences, as sensitive personal information or critical systems may be compromised.
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+
+==== Regulatory Compliance
+
+Many industries and jurisdictions have specific regulations and standards in place to protect user data and ensure its confidentiality. Storing passwords in plain-text or with weak hashing algorithms can lead to non-compliance with these regulations, potentially resulting in legal consequences, financial penalties, and damage to the reputation of the software application and its developers.
```
gaetan-ferry-sonarsource, line 21:
```suggestion
Many industries and jurisdictions have specific regulations and standards to protect user data and ensure its confidentiality. Storing passwords in plain-text or with weak hashing algorithms can lead to non-compliance with these regulations, potentially resulting in legal consequences, financial penalties, and damage to the reputation of the software application and its developers.
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+
+
+== How to fix it in Java EE
+
+A user password should never be stored in clear-text, instead a hash should be produced from it using a secure algorithm:
```
gaetan-ferry-sonarsource, line 26:
```suggestion
A user password should never be stored in clear text. Instead, a hash should be produced from it using a secure algorithm:
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+
+* not vulnerable to brute force attacks.
+* not vulnerable to collision attacks (see rule S4790).
```
gaetan-ferry-sonarsource, line 29: Collision resistance is not a pre-requisite when it comes to password storage. Only first and second pre-image attacks make sense here.

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+* and a salt should be added to the password to lower the risk of rainbow table attacks (see rule S2053).
```
gaetan-ferry-sonarsource, line 30: I would rephrase this to emphasize what type of algorithm is considered secure here:

```suggestion
A user password should never be stored in clear text. Instead, a hash should be produced from it using a secure algorithm. When dealing with password storage security, best practices recommend relying on a slow hashing algorithm, that will make brute force attacks more difficult. Using a hashing function with adaptable computation and memory complexity also is recommended to be able to increase the security level with time.

Adding a salt to the digest computation is also recommended to prevent pre-computed table attacks (see rule S2053).

In general, relying on an algorithm with no known weaknesses is also a requirement. This prevents the use of the MD5 or SHA-1 algorithms.
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+
+Some algorithms, like MD5 or SHA-family functions are fast to compute hashes and therefore brute force attacks are possible especially with hardware like GPU, FPGA or ASIC. Modern password hashing algorithms such as bcrypt, PBKDF2 or argon2 are recommended.
```
gaetan-ferry-sonarsource, line 32:
```suggestion
While considered strong for some use cases, some algorithms, like SHA-family functions, are too fast to compute and therefore susceptible to brute force attacks, especially with attack-dedicated hardware. Modern, slow, password hashing algorithms such as bcrypt, PBKDF2 or argon2 are recommended.
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
-  auth.jdbcAuthentication()
-    .dataSource(dataSource)
-    .usersByUsernameQuery("SELECT * FROM users WHERE username = ?"); // Noncompliant; default uses plain-text
-
-  // OR 
-  auth.userDetailsService(...); // Noncompliant; default uses plain-text
-  // OR 
-  auth.userDetailsService(...).passwordEncoder(new StandardPasswordEncoder()); // Noncompliant
 }
 ----
```
gaetan-ferry-sonarsource, line 49: The section title is "How to fix it in **Java EE** but this code example is using the Spring Framework.

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
 public void configureGlobal(AuthenticationManagerBuilder auth, DataSource dataSource) throws Exception {
   auth.jdbcAuthentication()
     .dataSource(dataSource)
     .usersByUsernameQuery("Select * from users where username=?")
     .passwordEncoder(new BCryptPasswordEncoder());
-
-  // or 
-  auth.userDetailsService(null).passwordEncoder(new BCryptPasswordEncoder());
 }
 ----
```
gaetan-ferry-sonarsource, line 62: Same here, `BCryptPasswordEncoder` is a Spring class.

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+=== How does this work?
+
+The ``BCryptPasswordEncoder`` is a password hashing function in Java that is designed to be secure and resistant to various types of attacks, including brute-force and rainbow table attacks. It is based on the Blowfish encryption algorithm and uses a technique called key stretching to make it computationally expensive to hash passwords. Here's how ``BCryptPasswordEncoder`` works:
+
+* Salt Generation: When a password is hashed using ``BCryptPasswordEncoder``, a random salt is generated. The salt is a random value that is added to the password before hashing, making it unique for each password. The salt is then stored along with the hashed password.
+
+* Password Hashing: The password, along with the salt, is then passed through the BCrypt hashing function. The function applies a series of iterations to the password and salt, creating a hash value that is unique to that combination.
+
+* Hash Comparison: When a user attempts to authenticate, the entered password is hashed using the same salt and number of iterations as the stored password. The generated hash is then compared with the stored hash. If they match, the password is considered valid.
```
gaetan-ferry-sonarsource, line 72: I think this is over-technical.
```suggestion
The ``BCryptPasswordEncoder`` is a password hashing function in Java that is designed to be secure and resistant to various types of attacks, including brute-force and rainbow table attacks. It is slow, adaptative, and automatically implements a salt.
```

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,102 @@
...
+
+=== Documentation
+
+* Spring Framework Security Documentation - https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.html[Class BCryptPasswordEncoder]
+
+=== Standards
+
+* https://owasp.org/Top10/A02_2021-Cryptographic_Failures/[OWASP Top 10 2021 Category A2] - Cryptographic Failures
+* https://owasp.org/Top10/A04_2021-Insecure_Design/[OWASP Top 10 2021 Category A4] - Insecure Design
+* https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html[OWASP CheatSheet] - Password Storage Cheat Sheet
```
gaetan-ferry-sonarsource, line 84: I would put that in the documentation as it does not reference an actual standard entry.

gaetan-ferry-sonarsource (changes requested):
LGTM in general. I still propose various fixes and improvements.

rules/S5344/java/rule.adoc
```diff
@@ -1,55 +1,96 @@
...
+Many industries and jurisdictions have specific regulations and standards to protect user data and ensure its confidentiality. Storing passwords in plain-text or with weak hashing algorithms can lead to non-compliance with these regulations, potentially resulting in legal consequences, financial penalties, and damage to the reputation of the software application and its developers.
+
+
+== How to fix it in Spring
+
+A user password should never be stored in clear text. Instead, a hash should be produced from it using a secure algorithm:
+
+* not vulnerable to brute force attacks.
+* not vulnerable to first preimage or second preimage attacks: If a hash is leaked, it should be difficut to find a password produding this same hash. It should also be difficult to find a password that has the same hash than another password.
+* and a salt should be added to the password to lower the risk of rainbow table attacks (see rule S2053).
```
gaetan-ferry-sonarsource, line 30: I think there are still some issues here I copy my first suggestion for reference:
```suggestion
A user password should never be stored in clear text. Instead, a hash should be produced from it using a secure algorithm. When dealing with password storage security, best practices recommend relying on a slow hashing algorithm, that will make brute force attacks more difficult. Using a hashing function with adaptable computation and memory complexity also is recommended to be able to increase the security level with time.

Adding a salt to the digest computation is also recommended to prevent pre-computed table attacks (see rule S2053).

In general, relying on an algorithm with no known weaknesses is also a requirement. This prevents the use of the MD5 or SHA-1 algorithms.
```
Some key points:

> not vulnerable to brute force attacks

All algorithms are. Here we limit the capabilities of an attacker by using slow and adaptative functions.
 
> not vulnerable to first preimage or second preimage attacks

I don't think this sentence is necessary. We can generally assume that serious hashing functions are resistant to pre-image computation of all forms. Plus it is probably too technical for our audience.

Also, I think we should add something discouraging against using vulnerable algorithms: MD5 and similar.

gaetan-ferry-sonarsource (changes requested):
Some more suggestions about the strong hashing algorithm definition.

PR #3155, opened 2023-09-26
Modify rule S6183: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6183/cfamily/rule.adoc
+++ b/rules/S6183/cfamily/rule.adoc
@@ -1,48 +1,116 @@
+Functions from the ``++std::cmp_*++`` family should be used to compare signed and unsigned values.
+
 == Why is this an issue?
 
-Comparison between `signed` and `unsigned` integers is dangerous because it produces counterintuitive results outside of their common range of values.
+Comparisons between ``++signed++`` and ``++unsigned++`` integers are dangerous because they produce counterintuitive results outside of their common value range.
+
+When a signed integer is compared to an unsigned one, the former might be converted to unsigned.
+The conversion preserves the two's-complement bit pattern of the signed value that often corresponds to a large unsigned result.
+The expression ``++2U < -1++`` evaluates to ``++true++``, for instance.
+
+{cpp}20 introduced remedy to this common pitfall: a family of ``++std::cmp_*++`` functions defined in the ``++<utility>++`` header:
+
+* ``++std::cmp_equal++``
+* ``++std::cmp_not_equal++``
+* ``++std::cmp_less++``
+* ``++std::cmp_greater++``
+* ``++std::cmp_less_equal++``
+* ``++std::cmp_greater_equal++``
+
+These functions correctly handle negative numbers and are safe against lossy integer conversion.
+For example, the comparison of ``++2U++`` and ``++-1++`` using ``++std::cmp_less(2U, -1)++`` evaluates to ``++false++`` and matches common intuition.
+
+=== Associated rule S6214
+
+This rule (S6183) starts by detecting comparisons between signed and unsigned integers.
+If the signed value can be proven negative, rule S6214 will raise an issue.
+Otherwise, this rule will raise an issue.
+Therefore, if rule S6183 is enabled, S6214 should be enabled, too.
 
 
-When a signed integer is compared to an unsigned one, the former might be converted to unsigned. The conversion preserves the two's-complement bit pattern of the signed value that often corresponds to a large unsigned result. For example, `2U < -1` is `true`.
+== What is the potential impact?
+
+Comparisons between ``++signed++`` and ``++unsigned++`` integer types produce counterintuitive results.
+
+Failing to understand integer conversion rules can lead to tricky bugs and to exploitable vulnerabilities.
+The major integer conversion risks include narrowing types, converting from unsigned to signed and from negative to unsigned.
 
 
-{cpp}20 introduced remedy to this common pitfall: a family of `std::cmp_*` functions defined in the `<utility>` header. These functions correctly handle negative numbers and lossy integer conversion. For example, `std::cmp_less(2U, -1)` is `false`.
+== How to fix it
 
-This rule starts by detecting comparisons between signed and unsigned integers. Then, if the signed value can be proven to be negative, the rule S6214 will raise an issue (it is a bug). Otherwise, this rule will raise an issue. Therefore, if this rule is enabled, S6214 should be enabled too.
+Use the appropriate function from the ``++std::cmp_*++`` family to conduct comparisons between signed _and_ unsigned integer types.
 
 
-=== Noncompliant code example
+=== Code examples
 
-[source,cpp]
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-bool less = 2U < -1; // Compliant, raises S6214
+#include <iostream>
 
-bool foo(unsigned x, signed y) {
+void foo() {
+  if (2U < -1) { // Compliant: raises on associated rule S6214
+    std::cout << "2 is less than -1\n";
+  } else {
+    std::cout << "2 is not less than -1\n";
+  }
+}
+----
+
+==== Compliant solution
+
+[source,cpp,diff-id=1,diff-type=compliant]
+----
+#include <iostream>
+
+void foo() {
+  if (std::cmp_less(2U, -1)) { // Compliant: for this rule (S6183) and associated rule S6214
+    std::cout << "2 is less than -1\n";
+  } else {
+    std::cout << "2 is not less than -1\n";
+  }
+}
+----
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
+bool foo(unsigned x, int y) {
   return x < y; // Noncompliant: y might be negative
 }
+----
 
+==== Compliant solution
+
+[source,cpp,diff-id=2,diff-type=compliant]
+----
+bool foo(unsigned x, int y) {
+  return std::cmp_less(x, y); // Compliant
+}
+----
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
 bool fun(int x, std::vector<int> const& v) {
   return x < v.size(); // Noncompliant: x might be negative
 }
 ----
 
+==== Compliant solution
 
-=== Compliant solution
-
-[source,cpp]
+[source,cpp,diff-id=3,diff-type=compliant]
 ----
-bool less = std::cmp_less(2U, -1); // Compliant for this rule and S6214
-
-bool foo(unsigned x, signed y) {
-  return std::cmp_less(x, y); // Compliant
-}
-
 bool fun(int x, std::vector<int> const& v) {
   return std::cmp_less(x, v.size()); // Compliant
 }
 
 void compute(std::vector<int> const &v) {
-  if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns an unsigned integer
+  // Compliant: v.size() returns an unsigned integer, but only positive signed integer values are involved
+  if (0 < v.size() && v.size() < 100) {
   }
 }
 ----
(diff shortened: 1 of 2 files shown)


rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,133 @@
+Functions from the ``++std::cmp_*++`` family should be used to compare signed and unsigned values.
+
 == Why is this an issue?
 
-Comparison between `signed` and `unsigned` integers is dangerous because it produces counterintuitive results outside of their common range of values.
+Comparisons between ``++signed++`` and ``++unsigned++`` integers are dangerous because they produce counterintuitive results outside of their common value range.
```
necto, line 5: When I read "common value range" I think of the commonly-used range of values for signed or unsigned int (independently).
Here, I think the intended meaning is the "intersection of value ranges"
```suggestion
Comparisons between ``++signed++`` and ``++unsigned++`` integers are dangerous because they produce counterintuitive results outside of their shared value range.
```

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,133 @@
...
+
+These functions correctly handle negative numbers and are safe against lossy integer conversion.
+For example, the comparison of ``++2U++`` and ``++-1++`` using ``++std::cmp_less(2U, -1)++`` evaluates to ``++false++`` and matches common intuition.
+
+=== Associated rule S6214
+
+This rule (S6183) starts by detecting comparisons between signed and unsigned integers.
+If the signed value can be proven negative, rule S6214 will raise an issue.
+Otherwise, this rule will raise an issue.
+Therefore, if rule S6183 is enabled, S6214 should be enabled, too.
```
necto, line 28: This looks like a minor point. It will be relevant only to people discovering the rules and people setting the quality profiles, and they should be used to look into the "Resources" section for related rules.
The most important audience - the dev trying to understand the issue wouldn't find this info useful.
I would put it in the "Resources" section when it references S6214.

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,133 @@
...
+
+
+== What is the potential impact?
+
+Comparisons between ``++signed++`` and ``++unsigned++`` integer types produce counterintuitive results.
 
+Failing to understand integer conversion rules can lead to tricky bugs and to exploitable vulnerabilities.
+The major integer conversion risks include narrowing types, converting from unsigned to signed and from negative to unsigned.
```
necto, line 36: I would appreciate an example of an exploit. E.g., some attacker sends a negative user input used as a loop bound and suddenly we read past the intended memory and leak a secret, or copy past the intended size and deploy shell code.

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,133 @@
...
+If the signed value can be proven negative, rule S6214 will raise an issue.
+Otherwise, this rule will raise an issue.
+Therefore, if rule S6183 is enabled, S6214 should be enabled, too.
+
+
+== What is the potential impact?
+
+Comparisons between ``++signed++`` and ``++unsigned++`` integer types produce counterintuitive results.
 
+Failing to understand integer conversion rules can lead to tricky bugs and to exploitable vulnerabilities.
```
necto, line 35: Did you check that we shouldn't replace the "exploitable" word for inclusive language?

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,133 @@
...
+The major integer conversion risks include narrowing types, converting from unsigned to signed and from negative to unsigned.
 
-When a signed integer is compared to an unsigned one, the former might be converted to unsigned. The conversion preserves the two's-complement bit pattern of the signed value that often corresponds to a large unsigned result. For example, `2U < -1` is `true`.
 
+== How to fix it
 
-{cpp}20 introduced remedy to this common pitfall: a family of `std::cmp_*` functions defined in the `<utility>` header. These functions correctly handle negative numbers and lossy integer conversion. For example, `std::cmp_less(2U, -1)` is `false`.
+Use the appropriate function from the ``++std::cmp_*++`` family to conduct comparisons between signed _and_ unsigned integer types.
```
necto, line 41: Why do you emphasize "_and_"?

rules/S6183/cfamily/rule.adoc
necto: Reading on, I see that the code examples look odd if you don't know how these two rules relate. However, it is not relevant before that.
How about moving this to Resources and adding a note in the "How to fix it" section saying that S6183 deliberately avoids intersection with S6214?

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,133 @@
...
 bool fun(int x, std::vector<int> const& v) {
   return std::cmp_less(x, v.size()); // Compliant
 }
 
 void compute(std::vector<int> const &v) {
-  if (0 < v.size() && v.size() < 100) { // Compliant, even though v.size() returns an unsigned integer
+  // Compliant: v.size() returns an unsigned integer, but only positive signed integer values are involved
+  if (0 < v.size() && v.size() < 100) {
   }
 }
```
necto, line 115: I would split this out into a separate code snippet because it will be surprising in the diff view.
Maybe put it into the "Why this is an issue" section? (I am not sure products can handle multiple snippets in the "Compliant solution" section.)

necto (changes requested):

rules/S6183/cfamily/rule.adoc
pdschbrt: Thanks for pointing this out. "Shared" is the intended meaning.

rules/S6183/cfamily/rule.adoc
pdschbrt: I put a note in the "How to fix it" section and added another sentence to the description of related rules. This should reduce the noise for developers trying to understand the rule.

rules/S6183/cfamily/rule.adoc
pdschbrt: It's not really necessary, I guess.

rules/S6183/cfamily/rule.adoc
pdschbrt: I've removed the `compute` function entirely. It doesn't have a "noncompliant counterpart" if that makes sense.

rules/S6183/cfamily/rule.adoc
pdschbrt: I've made an attempt. Please give it a read.

rules/S6183/cfamily/rule.adoc
pdschbrt: I've rephrased that a bit. I'm not aware of any inclusive language guide(s).

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,171 @@
...
+
+[source,cpp,diff-id=3,diff-type=compliant]
+----
+bool fun(int x, std::vector<int> const& v) {
+  return std::cmp_less(x, v.size()); // Compliant
 }
 ----
 
 
 == Resources
```
necto, line 157: you can add https://en.cppreference.com/w/cpp/utility/intcmp to the "Documentation" subsection

necto (changes requested):

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,171 @@
...
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+#include <iostream>
+
+void foo() {
+  if (2U < -1) { // Compliant: raises on associated rule S6214
```
necto, line 93:
```suggestion
  if (2U < -1) { // Compliant: the comparison is incorrect but S6214 raises an issue instead of S6183
```

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,171 @@
...
+== How to fix it
+
+Use the appropriate function from the ``++std::cmp_*++`` family to conduct comparisons between signed and unsigned integer types.
+
+Note that this rule (S6183) deliberately avoids intersection with S6214.
+
+
+=== Code examples
+
+==== Noncompliant code example
```
necto, line 86: This example might be confusing: it is noncompliant but marked as compliant for this rule only due to detail in how SQ rules work together.
I would move it later in the list of examples. And I think this would be a better place for

> Note that this rule (S6183) deliberately avoids intersection with S6214.

Instead of directly in the "How to fix it" section

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,171 @@
...
+Before copying the string into its internal buffer it checks whether the user-provided string fits into the buffer.
+However, the respective check only validates if the provided string length is smaller or equal to the buffer's size.
+Since the `atoi` function returns a signed integer, a user may provide a negative number to withstand that check.
+The result of `sizeof(*)` on the other hand returns an unsigned integer which causes the expression `user_input * sizeof(char)` to be evaluated by (i) converting both operands to unsigned integers, (ii) performing the multiplication, and (iii) returning the result as an unsigned integer type.
+A malicious user is hence able to provide a carefully crafted negative integer and string to bypass the size check while still providing a valid size argument for the call to `memcpy`.
+This, in turn, enables the malicious user to overflow the buffer variable `buf` to override the `user_input` variable which allows the second `if` statement to be evaluated to true, eventually opening a new shell with the target program's possibly elevated permissions.
```
necto, line 74: This is a long paragraph, and that makes it difficult to read. Yet, it is important to understand the tricky code above.

Here is one possible way to break it down.
- 1 paragraph only "description of what it does" - the first two sentences.
- 1 paragraph "what can go wrong"
- 1 paragraph "how a malicious user can abuse this"

I.e. (although now that the three paragraphs stands on their own, they might need a bit of reformulation):

> The program takes as arguments a string and its size, and uses these arguments to copy the string argument into an internal buffer.
Before copying the string into its internal buffer it checks whether the user-provided string fits into the buffer.

> However, the respective check only validates if the provided string length is smaller or equal to the buffer's size.
Since the `atoi` function returns a signed integer, a user may provide a negative number to withstand that check.
The result of `sizeof(*)` on the other hand returns an unsigned integer which causes the expression `user_input * sizeof(char)` to be evaluated by
>  1. converting both operands to unsigned integers,
>  2. performing the multiplication, and
>  3. returning the result as an unsigned integer type.

> A malicious user is hence able to provide a carefully crafted negative integer and string to bypass the size check while still providing a valid size argument for the call to `memcpy`.
This, in turn, enables the malicious user to overflow the buffer variable `buf` to override the `user_input` variable which allows the second `if` statement to be evaluated to true, eventually opening a new shell with the target program's possibly elevated permissions.

necto (changes requested):

rules/S6183/cfamily/rule.adoc
pdschbrt: I've made an attempt and introduced a separate section to discuss this.

rules/S6183/cfamily/rule.adoc
pdschbrt: I think I could improve it to make the example easier to follow.

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,186 @@
...
+
+The check for the buffer size only validates that the provided string length (`user_input`) is smaller or equal to the buffer's size.
+Since the `atoi` function returns a signed integer, a user may provide a negative number to withstand that check.
+The result of `sizeof(*)` on the other hand returns an unsigned integer which causes the expression `user_input * sizeof(char)` to be evaluated by
+
+  . converting both operands to unsigned integers,
+  . performing the multiplication, and
+  . returning the result as an unsigned integer type.
+
+A malicious user is hence able to provide a carefully crafted negative integer and string to bypass the size check while still providing a valid size argument for the call to `memcpy`.
```
necto, line 82: Article "a" vs. plural nouns
```suggestion
A malicious user is hence able to provide carefully crafted negative integer and string to bypass the size check while still providing a valid size argument for the call to `memcpy`.
```
or
```suggestion
A malicious user is hence able to provide a carefully crafted negative integer and a string to bypass the size check while still providing a valid size argument for the call to `memcpy`.
```

necto, line 82: Also, I am not sure what you mean by "a valid size argument for the call to `memcpy`".
I guess you mean the memcpy won't crash trying to copy gazillion of bytes. If so, how about this wording?
> ... while still arriving at the appropriate size argument to not crash `memcpy`.

rules/S6183/cfamily/rule.adoc
```diff
@@ -1,57 +1,186 @@
...
+}
+----
 
-[source,cpp]
+
+== Interactions with associated rule S6214
+
+Note that this rule (S6183) deliberately avoids intersection with S6214.
+
+While S6183 triggers when detecting comparisons between signed and unsigned integers, S6214 will raise an issue if the signed value can be proven to be negative -- in which case it is definitely a bug.
```
necto, line 134: This does not say that S6183 does not trigger for cases that are provably a bug.

How about this reformulation:
```suggestion
While S6214 raises an issue if the signed value can be proven to be negative (in which case it is definitely a bug), S6281 will flag all *other* comparisons between signed and unsigned integers.
```

necto (changes requested):
I like how it's taking shape. Only a few minor tweaks remain

rules/S6183/cfamily/rule.adoc
pdschbrt: That reads well!

rules/S6183/cfamily/rule.adoc
pdschbrt: That reads well, too. Sold!

pdschbrt:
> I like how it's taking shape. Only a few minor tweaks remain

Sounds good 🎉

PR #3158, opened 2023-09-26
Modify rule S6320: Adapt to LaYC format

Swalkyn:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6320/java/rule.adoc
+++ b/rules/S6320/java/rule.adoc
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-A cast operation allows an object to be converted from one type to another.
+A cast operation allows an object to be "converted" from one type to another.
 The compiler raises an error if it can determine that the target type is incompatible with the declared type of the object, otherwise it accepts the cast.
 However, depending on the actual runtime type of the object, a cast operation may fail at runtime.
 When a cast operation fails, a `ClassCastException` is thrown.
@@ -34,6 +34,7 @@ String hexString(Integer i) {
 ----
 
 Another one is to return a default value for types that are not `Integer`.
+Here, the `if` statement with the condition relying on `instanceof` prevents references to objects with an incompatible type from making it to the cast operation.
 
 [source,java,diff-id=1,diff-type=compliant]
 ----
@@ -45,7 +46,7 @@ String hexString(Object o) {
 }
 ----
 
-It depends on which part of the code should be responsible to handle non-Integer types.
+The solution to adopt depends on which part of the code should be responsible to handle non-Integer types.
 
 === Going the extra mile
 
--- a/rules/S6320/java/rule.adoc
+++ b/rules/S6320/java/rule.adoc
@@ -1,9 +1,17 @@
 == Why is this an issue?
 
+A cast operation allows an object to be converted from one type to another.
+The compiler raises an error if it can determine that the target type is incompatible with the declared type of the object, otherwise it accepts the cast.
+However, depending on the actual runtime type of the object, a cast operation may fail at runtime.
+When a cast operation fails, a `ClassCastException` is thrown.
+This type of exception is usually unexpected, so it is not caught and the application crashes.
 
-A cast operation makes it possible to convert an object from one type to another one. The compiler raises an error when it can determine that the target type is incompatible with the declared type of the object. Other cases are accepted by the compiler. However, depending on the actual runtime type of the object, a cast operation may fail at runtime with a `ClassCastException` which can crash the program. This kind of failure happens when a piece of code makes an invalid assumption about the runtime type of an object.
+== How can I fix it?
 
-The appropriate fix for those cast operations highly depends on the context: adding a condition before the cast or changing the runtime type of the object are some of the possible solutions.
+When an object is cast, the code makes assumptions about the type of the object.
+A `ClassCastException` indicates that this assumption has been broken.
+If the assumption is reasonable, then some prior logic should ensure that only objects of a compatible type can be cast.
+You should try to identify the code responsible for these checks and fix it.
 
 === Noncompliant code example
 
@@ -47,4 +55,11 @@ or
 LinkedList<String> list = new LinkedList<>();
 list.addLast("abc");
 ----
+
+=== Going the extra mile
+
+==== Consider polymorphism
+
+Casting is considered an anti-pattern in object-oriented programming.
+It is sometimes necessary, but you may want to evaluate whether your solution can use polymorphism instead.
 
--- a/rules/S6320/java/rule.adoc
+++ b/rules/S6320/java/rule.adoc
@@ -17,49 +17,97 @@ You should try to identify the code responsible for these checks and fix it.
 
 [source,java,diff-id=1,diff-type=noncompliant]
 ----
-List<String> list = new LinkedList<>();
-if (someCondition) {
-  list = new ArrayList<>();
+String hexString(Object o) {
+  return Integer.toHexString((Integer) o); // Noncompliant if hexString is called with a String for example 
 }
-((LinkedList<String>) list).addLast("abc"); // Noncompliant, throws a ClassCastException when someCondition is true
 ----
 
 === Compliant solution
 
-[source,java,diff-id=1,diff-type=compliant]
-----
-List<String> list = new LinkedList<>();
-if (someCondition) {
-  list = new ArrayList<>();
-}
-if (list instanceof LinkedList) {
-  ((LinkedList<String>) list).addLast("abc");
-}
-----
-
-or
+One possible solution is to change `hexString` to only accept integers and adapt call sites.
 
 [source,java,diff-id=1,diff-type=compliant]
 ----
-List<String> list = new LinkedList<>();
-if (someCondition) {
-  list = new ArrayList<>();
+String hexString(Integer i) {
+  return Integer.toHexString(i);
 }
-list.add("abc");
 ----
 
-or
+Another one is to return a default value for types that are not `Integer`.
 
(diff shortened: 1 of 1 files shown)


rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
 == Why is this an issue?
 
+A cast operation allows an object to be converted from one type to another.
+The compiler raises an error if it can determine that the target type is incompatible with the declared type of the object, otherwise it accepts the cast.
+However, depending on the actual runtime type of the object, a cast operation may fail at runtime.
+When a cast operation fails, a `ClassCastException` is thrown.
+This type of exception is usually unexpected, so it is not caught and the application crashes.
 
-A cast operation makes it possible to convert an object from one type to another one. The compiler raises an error when it can determine that the target type is incompatible with the declared type of the object. Other cases are accepted by the compiler. However, depending on the actual runtime type of the object, a cast operation may fail at runtime with a `ClassCastException` which can crash the program. This kind of failure happens when a piece of code makes an invalid assumption about the runtime type of an object.
+== How can I fix it?
```
anton-haubner-sonarsource, line 9: This section has to be named exactly `How to fix it`, otherwise no tab will be created for it in the SonarQube UI

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
 == Why is this an issue?
```
anton-haubner-sonarsource, line 1: For level 1 LaYC format, there needs to be an introductory section without a heading before the `Why is this an issue?` section. This introduction section is also displayed separately in SonarQube.

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
 
+A cast operation allows an object to be converted from one type to another.
+The compiler raises an error if it can determine that the target type is incompatible with the declared type of the object, otherwise it accepts the cast.
```
anton-haubner-sonarsource, line 4: Nitpick:
```suggestion
The compiler raises an error if it can determine that the declared target type is incompatible with the actual runtime type of the object, otherwise it accepts the cast.
```

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
+However, depending on the actual runtime type of the object, a cast operation may fail at runtime.
+When a cast operation fails, a `ClassCastException` is thrown.
+This type of exception is usually unexpected, so it is not caught and the application crashes.
```
anton-haubner-sonarsource, line 8: Regarding a `What is the Potential Impact?` section for this rule:

The impact of rules about exceptions being raised, is usually the same, [see the potential impact section for S6466](https://github.com/SonarSource/rspec/blob/fa6a5beb638fe7e4502104cdd899ae2e083bcea4/rules/S6466/impact.adoc).

Apparently there is a `shared_content` folder for sections that are shared across rules, see [here](https://github.com/SonarSource/rspec/blob/fa6a5beb638fe7e4502104cdd899ae2e083bcea4/docs/description.adoc#share-content-between-rules).
Maybe we should move the impact description of S6466 there and also import it here?

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
```
anton-haubner-sonarsource, line 7: Regarding a `What is the Potential Impact?` section for this rule:

The impact of rules about exceptions being raised, is usually the same, see for example [the potential impact section for S6466](https://github.com/SonarSource/rspec/blob/fa6a5beb638fe7e4502104cdd899ae2e083bcea4/rules/S6466/impact.adoc).

Apparently there is a `shared_content` folder for sections that are shared across rules, see [here](https://github.com/SonarSource/rspec/blob/fa6a5beb638fe7e4502104cdd899ae2e083bcea4/docs/description.adoc#share-content-between-rules).
Maybe we should move the impact description of S6466 there and also import it in this rule?

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
+String hexString(Object o) {
+  if (o instanceof Integer) {
+    return Integer.toHexString((Integer) o);
+  }
+  return "0x0";
+}
+----
+
+It depends on which part of the code should be responsible to handle non-Integer types.
+
```
anton-haubner-sonarsource, line 49: This is not something we need to worry about right now, since we are not aiming for level 2, but I still wanted to leave a comment about the `How does it work?` section that is technically part of `How to fix it`.
This comment is not really a part of the review, but maybe something to discuss in our feedback meeting with John/Loris.

We are given an example of a `How to fix it` section, and a `How does it work?` section [here](https://next.sonarqube.com/sonarqube/coding_rules?languages=java&q=xss&open=javasecurity%3AS5131).
However, the difference between the two really is not clear to me.

The [LaYC specification](https://xtranet-sonarsource.atlassian.net/wiki/spaces/PM/pages/2715320326/LaYC+Rule+Content+Specification#How-can-I-fix-it?) gives me the impression that examples go directly into the `How to fix it` section, while general advise and best practices should go into `How does it work?`.
And that is also roughly what we see in the XSS example linked above.

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
+Casting is considered an anti-pattern in object-oriented programming.
+It is sometimes necessary, but there often is a better alternative.
+Consider:
+
+* Polymorphism
+* The visitor design pattern
+* Pattern matching
+
+They come with some garantees from the compiler, which makes your code more reliable.
```
anton-haubner-sonarsource, line 60:
```suggestion
They come with some guarantees from the compiler, which make your code more reliable.
```

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
-if (list instanceof LinkedList) {
-  ((LinkedList<String>) list).addLast("abc");
+----
+
+==== Compliant solution
+
+[source,java,diff-id=2,diff-type=compliant]
+----
+int foo(Shape shape) {
+  return shape.fooValue(...); // Code was moved in subclasses
```
anton-haubner-sonarsource, line 83: Nitpick
```suggestion
  return shape.fooValue(...); // Code was moved into subclasses

File: rules/S6320/java/rule.adoc [hunk 9]
```diff
@@ -1,50 +1,113 @@
...
 or
 
-[source,java]
+[source,java,diff-id=2,diff-type=compliant]
 ----
-List<String> list = new LinkedList<>();
-if (someCondition) {
-  list = new ArrayList<>();
+int foo(Shape shape) {
+  return shape.accept(new FooVisitor()); // Code was moved in FooVisitor
```
anton-haubner-sonarsource, line 92: Nitpick:
```suggestion
  return shape.accept(new FooVisitor()); // Code was moved into FooVisitor
```

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
+int foo(Shape shape) {
+  if (shape instanceof Circle c) {
+    // Code for objects of type Circle
+  } else if (shape instanceof Square s) {
+    // Code for objects of type Square
+  }
+}
 ----
 
+== More info
```
anton-haubner-sonarsource, line 110: Unfortunately, the LaYC spec is again not very accurate when it comes to the heading names to use in practice.
To have this section displayed in the `More Info` tab in SonarQube, it apparently needs to be named "Resources" :/

```suggestion
== Resources
```

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
+
```
anton-haubner-sonarsource, line 111: Nitpick: I am not entirely sure about this, but I believe the links in "More Info" need to be organized into category headings. I.e. "Documentation" or "Articles & Blog Posts"

```suggestion

=== Documentation

* https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ClassCastException.html[ClassCastException]

=== Articles & blog posts

```

rules/S6320/java/rule.adoc
```diff
@@ -1,50 +1,113 @@
...
+* https://refactoring.guru/design-patterns/visitor[Refactoring Guru: Visitor]
+* https://openjdk.org/jeps/394[JEP 394: Pattern Matching for instanceof]
```
anton-haubner-sonarsource, line 113:
```suggestion
* https://openjdk.org/jeps/394[JEP 394: Pattern Matching for instanceof]

=== Standards

* https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.16
* https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.20.2[The JLS on instanceof]
```

anton-haubner-sonarsource (changes requested):
I really like your in-depth discussion on how class casts can be guarded and/or avoided!
My comments are mostly about making the rule description work with the SonarQube UI, and some nitpicks that do not need to be necessarily addressed.

Swalkyn:
Hmm you made lots of valid points in your review. I could have caught many of those by reading the LaYC spec more thoroughly. Sorry about that.

I'll make the requested changes!

rules/S6320/java/rule.adoc
Swalkyn: I added this here because I don't think that giving a simple example and the easiest solution helps the user. I think it's good if they think about where the logic should change (i.e. which part of the code is responsible for it)

I will skip the "How to fix it" and "Pitfalls" subsections for now, because I think it will just repeat information. We can make feedback on this for sure.

anton-haubner-sonarsource (approved):

PR #3160, opened 2023-09-26
Modify rule S3519: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone


rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+The application may crash or, even worse, the application may appear to execute correctly while losing data or producing incorrect results.
+
+Out-of-bounds memory accesses cause some of the most damaging and hard-to-track defects.
+Unbounded memory accesses may also introduce severe security vulnerabilities.
+In case a user of the program can control the contents of a buffer, for instance, a malicious attacker can the buffer overflow to overwrite critical data, execute arbitrary code, or gain unauthorized access to the system.
+
+
+== How to fix it
+
+Ensure that memory access are within the bounds of the allocated array or buffer.
```
balazs-benics-sonarsource, line 25: Shouldn't thins be plural?: memory accesses
Have you checked the text with Grammarly?

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
+#include <stdio.h>
+
+void foo() {
+  char txt[16] = "Hello, World!";
+  for (size_t i = 0; i <= sizeof(txt); ++i) { // Noncompliant: index should be between 0 and 15
+    printf("%c\n", txt[i]);
+  }
+}
```
balazs-benics-sonarsource, line 127: Unfortunately, due to loop unrolling (4), we don't detect this: https://godbolt.org/z/ezMs5jhcc
I'd recommend reworking this example.

It's also not a good practice to hardcode the length of the string at the array declaration.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+----
+void foo() {
+  char txt[16] = "Hello, World!";
+  for (size_t i = 0; i < sizeof(txt); ++i) { // Compliant: index is between 0 and 15
+    printf("%c\n", txt[i]);
+  }
+}
+----
+
+=== Going the extra mile
```
balazs-benics-sonarsource, line 142: Before the extra mile, I think it worths mentioning that one needs to be careful for zero terminating buffers whenever needed as this can be a problem (pitfall) in this context.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+
+Buffer overflows occur when a program writes data beyond the boundaries of a buffer and can lead to memory corruption and potential security vulnerabilities.
+To mitigate this risk, developers must carefully manage array and buffer sizes (, use secure coding practices, and employ techniques like input validation and bounds checking).
```
balazs-benics-sonarsource, line 145: I think we should improve the chunking of this sentence by splitting it. This way we can express the same without using parens and colons.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+
+In {cpp}, manual array or string, i.e., buffer manipulations are considered a code smell.
+
+Instead, the `std::array` type should be used to manage stack-based arrays, and the `std::vector` type should be used if a heap-based array is desired.
```
balazs-benics-sonarsource, line 149: Do we use the term "stack-based" anywhere else?
I think the accurate term is "automatic storage duration", but I guess yours is still more accessible for the reader.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+Besides always carrying their respective sizes, i.e., number of elements, `std::array` and `std::vector` implement many useful member functions such as `begin()` and `end()`, allowing one to safely and conveniently process them using algorithms from the {cpp}'s `<algorithm>` header, for instance.
```
balazs-benics-sonarsource, line 150:
```suggestion
Besides always carrying their respective sizes, i.e., the number of elements, `std::array` and `std::vector` implement many useful member functions such as `begin()` and `end()`, allowing one to safely and conveniently process them using algorithms from the {cpp}'s `<algorithm>` header, for instance.
```

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+Instead of manually concatenating two buffers using `strncat`, for instance, `std::string` allows this operation to be performed in a much more convenient manner as shown in the following code:
 
 [source,cpp]
 ----
-int array[10];
-array[9] = 0;
+#include <iostream>
+#include <string>
+
+void buz(std::string const &s) {
```
balazs-benics-sonarsource, line 188: global: I think the example names should reflect their content.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+
+=== Standards
+
+* CERT - https://wiki.sei.cmu.edu/confluence/x/wtYxBQ[ARR30-C. Do not form or use out-of-bounds pointers or array subscripts]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/i3w-BQ[STR50-CPP. Guarantee that storage for strings has sufficient space for character data and the null terminator]
+* CWE - https://cwe.mitre.org/data/definitions/119[119 Improper Restriction of Operations within the Bounds of a Memory Buffer]
+* CWE - https://cwe.mitre.org/data/definitions/121.html[121 Stack-based Buffer Overflow]
+* CWE - https://cwe.mitre.org/data/definitions/122.html[122 Heap-based Buffer Overflow]
+* CWE - https://cwe.mitre.org/data/definitions/131[131 Incorrect Calculation of Buffer Size]
+* CWE - https://cwe.mitre.org/data/definitions/788[788 Access of Memory Location After End of Buffer]
```
balazs-benics-sonarsource, line 223: In some entries you have the `.html` suffix, for others don't.
Please harmonize them.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
 == Resources
 
-* https://cwe.mitre.org/data/definitions/119[MITRE, CWE-119] - Improper Restriction of Operations within the Bounds of a Memory Buffer
-* https://cwe.mitre.org/data/definitions/131[MITRE, CWE-131] - Incorrect Calculation of Buffer Size
-* https://cwe.mitre.org/data/definitions/788[MITRE, CWE-788] - Access of Memory Location After End of Buffer
-* https://wiki.sei.cmu.edu/confluence/x/wtYxBQ[CERT, ARR30-C.] - Do not form or use out-of-bounds pointers or array subscripts
-* https://wiki.sei.cmu.edu/confluence/x/i3w-BQ[CERT, STR50-CPP.] - Guarantee that storage for strings has sufficient space for character data and the null terminator
+=== Conference presentations
+
+* CppCon 2018 - https://www.youtube.com/watch?v=0S0QgQd75Sw&ab_channel=CppCon[Software Vulnerabilities in C and C++]
```
balazs-benics-sonarsource, line 213: This link appears to be dead to me.

balazs-benics-sonarsource (changes requested):
Pretty good, I'd say.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,226 @@
...
+----
+#include <stdio.h>
+
+int main() {
+  char buffer[64] = {0};
+  if (fgets(buffer, 64, stdin) != NULL) { // Compliant: reads at most 64 - 1 characters
+    printf("buffer contents: %s", buffer);
+  }
+  return 0;
+}
```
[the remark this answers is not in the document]
pdschbrt: Yes, I've changed the code snippets in question to the C programming language.

rules/S3519/cfamily/rule.adoc
pdschbrt: I'll give it a shot and rework it a bit.

rules/S3519/cfamily/rule.adoc
pdschbrt: Let me add a `Pitfalls` section here.

rules/S3519/cfamily/rule.adoc
pdschbrt: Yes, I think "stack-based" is easier to swallow here. Especially, since we talk about "heap-based" arrays in the next half of the sentence.

rules/S3519/cfamily/rule.adoc
pdschbrt: I agree, but I guess that is off the table as this is not consistently handled in the rules that have been rewritten so far, unfortunately. From what I've seen I believe it's quite the contrary :-(

rules/S3519/cfamily/rule.adoc
pdschbrt: Oh, thanks for pointing that out. It was alive a few days ago. YouTube is strange sometimes 😄

pdschbrt: That's crazy. The link is alive. It's only dead if you try to access the video using the Island browser. When using ordinary Chrome, it's all fine.

rules/S3519/cfamily/rule.adoc
pdschbrt: This should be fine and more of a flow without the "the" as "i.e." refers to "sizes".

rules/S3519/cfamily/rule.adoc
necto: Perhaps something to report to IT?

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
+Memory accesses should be bounded to prevent buffer overflows.
+
 == Why is this an issue?
 
-Array overruns and buffer overflows happen when memory access accidentally goes beyond the boundary of the allocated array or buffer. These overreaching accesses cause some of the most damaging, and hard to track defects.
+Array overruns and buffer overflows occur when a memory access accidentally goes beyond the boundary of the allocated array or buffer.
 
+These overreaching accesses cause some of the most damaging and hard to track defects.
+Not only do these faulty accesses constitute undefined behavior, but they frequently introduce security vulnerabilities and open up the program to buffer overflow attacks, too.
```
necto, line 8: This sounds like tautology
```suggestion
Not only do these faulty accesses constitute undefined behavior, but they frequently introduce security vulnerabilities, too.
```

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+== What is the potential impact?
+
+Accessing memory that goes beyond the boundary of the allocated array or buffer results in undefined behavior.
+
+For programs that exercise undefined behavior, the compiler is no longer bound by the language specification.
+The application may crash or, even worse, the application may appear to execute correctly while losing data or producing incorrect results.
+
+Out-of-bounds memory accesses cause some of the most damaging and hard-to-track defects.
+Unbounded memory accesses may also introduce severe security vulnerabilities.
+In case a user of the program can control the contents of a buffer, for instance, a malicious attacker can the buffer overflow to overwrite critical data, execute arbitrary code, or gain unauthorized access to the system.
```
necto, line 20:
```suggestion
In case a user of the program can control the contents of a buffer, for instance, a malicious attacker can use the buffer overflow to overwrite critical data, leak secret or private information, execute arbitrary code, or gain unauthorized access to the system.
```

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+void foo(void) {
+  char *buf1 = (char *)malloc(17);
+  char *buf2 = (char *)malloc(8);
+  memset(buf1, 0, 17);
+  memset(buf2, 0, 8);
+  const char *const str = "AAAA"
+                          "BBBB"
+                          "CCCC"
+                          "DDDD";
+  strncpy(buf1, str, strlen(str));
```
necto, line 107: This is also far from best practice: if I accidentally add some characters to `str`, I'll overrun `buf1`. I think you should change both non-compliant and compliant examples to allocate `buf1` based on `strlen(str)`

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+
+[source,c,diff-id=3,diff-type=noncompliant]
+----
+#include <stdio.h>
+
+void bar(void) {
+  int array[8] = {1, 2, 3, 4, 5, 6, 7, 8};
+  array[8] = 42; // Noncompliant: index should be between 0 and 7
+  for (size_t i = 0; i < 8; ++i) { // Compliant: index is between 0 and 7
+    printf("%d\n", array[i]);
```
necto, line 128: Since this is a "**non**compliant" example, why not take the opportunity?
```suggestion
  for (size_t i = 0; i <= 8; ++i) { // Noncompliant: index is between 0 and 7
    printf("%d\n", array[i]); // Noncompliant: index reaches 8 - beyond the last element of the array
```

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+
+[source,c,diff-id=3,diff-type=compliant]
+----
+#include <stdio.h>
+
+void bar(void) {
+  int array[8] = {1, 2, 3, 4, 5, 6, 7, 8};
+  array[7] = 42; // Compliant: index is between 0 and 7
+  for (size_t i = 0; i < 8; ++i) { // Compliant: index is between 0 and 7
+    printf("%d\n", array[i]);
```
necto, line 143:
```suggestion
  for (size_t i = 0; i < 8; ++i) {
    printf("%d\n", array[i]); // Compliant: index is between 0 and 7
```

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+----
 
-=== Compliant solution
+=== Pitfalls
+
+When managing and manipulating string buffers one needs to take great care that the buffers are correctly terminated with a null byte (`'\0'`).
+Failing to correctly null terminate string buffers almost always causes buffer overruns and introduces undefined behavior.
+
+The following faulty program aims at creating a string copy.
+However, the heap-allocated memory buffer for holding the copy is too small since `strlen` calculates the length of its input string but exclude the "invisible" terminating null byte ('\0').
```
necto, line 154:
```suggestion
However, the heap-allocated memory buffer for holding the copy is too small since `strlen` calculates the length of its input string but exclude the "invisible" terminating null byte (`'\0'`).
```

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+The call to `strncpy` also uses the string length computed by `strlen` and the string copy hence lacks a null terminator.
+This causes out-of-bounds reads and introduces undefined behavior in the subsequent processing steps.
+In this example, the call to `printf` will eventually trigger such erroneous behavior.
+
+[source,c]
```
necto, line 159: I would use the diff-view here too, even though it is outside of "how to fix it", I think it can still be displayed in some products, and in the worst case, diff won't be highlighted

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+char *make_str_copy(const char *const src) {
+  size_t src_len = strlen(src);
+  char *dst = (char *)malloc(src_len); // buffer too small to hold null terminator
+  if (dst == NULL) {
+    perror("malloc failed");
+    exit(1);
+  }
```
necto, line 171:
```suggestion
```
It is not always a good idea to check the `malloc` return value for null...

In any case, it is not the focus of this rule, and you already omitted it from other examples, so I would omit it here too, for brevity

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+  // heap-based array
+  std::vector<int> heap_buf = {1, 2, 3, 4};
+  heap_buf.resize(10);
+  std::iota(heap_buf.begin(), heap_buf.end(), 1);
+  std::cout << "sum of heap_buf's values: "
+            << std::accumulate(heap_buf.begin(), heap_buf.end(), 0) << '\n';
+}
+----
+
+The `std::string` type should be used to manage buffers, since it guarantees safe buffer manipulations.
```
necto, line 257:
```suggestion
The `std::string` type should be used to manage buffers since it facilitates safe buffer manipulations.
```
you can still access string out-of-bounds

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,303 @@
...
+#include <string>
+
+void tar(std::string const &s) {
+  std::string t = std::format("Hello, World! Greetings {}\n", s);
+  std::cout << t << '\n';
+}
 ----
 
 
 == Resources
```
necto, line 286: S5945 - for the discussion about replacing arrays with C++ wrappers
Maybe also S1851 that deprecates pointers

necto: Also S5782 - overflow in POSIX functions. BTW, I think some of your examples actually belong to S5782 (e.g., overrun in `printf`)

necto (changes requested):
Note @balazs-benics-sonarsource comments in the bottom of "rules/S3519/cfamily/rule.adoc". I think there are a few things that you can draw from them

rules/S3519/cfamily/rule.adoc
balazs-benics-sonarsource: Because we cannot detect that case. It would require us to loop more than 4 times to reach the OOB access.
I asked for this refactoring [here](https://github.com/SonarSource/rspec/pull/3160#discussion_r1338471315).

rules/S3519/cfamily/rule.adoc
balazs-benics-sonarsource: Excellent addition! I love it.

rules/S3519/cfamily/rule.adoc
pdschbrt: I've added them to the "Related rules" section.

rules/S3519/cfamily/rule.adoc
pdschbrt: I addressed that issue.

rules/S3519/cfamily/rule.adoc
pdschbrt: I'm usually in favor of brevity, too. However, I feel the need to correctly handle a "failing" `malloc` in this example. Since both, this rule and our examples target low-level C code, we should highlight that many things can go sideways pretty quickly due to the very basic nature of APIs and functionalities provided by the C standard library---making them rather inconvenient to use at times. Unhandled allocation failures are often overlooked, but when manually handling buffers in C you need to get everything exactly right. The `make_str_copy` function should depict a function that is more on the realistic side.

There is only one other example that uses `malloc` which I would adjust to also have it handled correctly, if you don't mind.

necto: I think it might not be a good idea because it is rare that you can do anything useful when your system is out of memory, but these frequent error checks still clutter the code. My reasoning was similar to [this SQ question](https://stackoverflow.com/questions/2248995/is-there-a-need-to-check-for-null-after-allocating-memory-when-kernel-uses-over), and [this answer](https://stackoverflow.com/a/49993528). I do not insist here, as the only downside is the code readability, and this rule deals with the safety so probably should err on the safe side

pdschbrt:
> Note @balazs-benics-sonarsource comments in the bottom of "rules/S3519/cfamily/rule.adoc". I think there are a few things that you can draw from them

Yes, I've extended the "Going the extra mile" section and incorporated some code examples, too 👍🏻

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,399 @@
...
-* https://cwe.mitre.org/data/definitions/131[MITRE, CWE-131] - Incorrect Calculation of Buffer Size
-* https://cwe.mitre.org/data/definitions/788[MITRE, CWE-788] - Access of Memory Location After End of Buffer
-* https://wiki.sei.cmu.edu/confluence/x/wtYxBQ[CERT, ARR30-C.] - Do not form or use out-of-bounds pointers or array subscripts
-* https://wiki.sei.cmu.edu/confluence/x/i3w-BQ[CERT, STR50-CPP.] - Guarantee that storage for strings has sufficient space for character data and the null terminator
+=== Articles & blog posts
+
+* Clang Hardening - https://blog.quarkslab.com/clang-hardening-cheat-sheet.html[Clang Hardening Cheat Sheet]
+* FORTIFY_SOURCE - https://www.redhat.com/en/blog/enhance-application-security-fortifysource[Enhance application security with FORTIFY_SOURCE]
+* Stack Protection - https://developers.redhat.com/articles/2022/06/02/use-compiler-flags-stack-protection-gcc-and-clang#:~:text=Stack%20canary&text=A%20modification%20indicates%20the%20detection,%2Dfstack%2Dprotector%2Dstrong[Use compiler flags for stack protection in GCC and Clang]
```
balazs-benics-sonarsource, line 373:
```suggestion
* Stack Protection - https://developers.redhat.com/articles/2022/06/02/use-compiler-flags-stack-protection-gcc-and-clang[Use compiler flags for stack protection in GCC and Clang]
```

Stuff after the `?` or `#` in URLs are usually optional.

rules/S3519/cfamily/rule.adoc
```diff
@@ -1,41 +1,399 @@
...
+----
+
+
+=== Going the extra mile
+
+Buffer overflows occur when a program writes data beyond the boundaries of a buffer and can lead to memory corruption and potential security vulnerabilities.
+To mitigate this risk, developers must carefully manage array and buffer sizes.
+This includes using secure coding practices, and employing techniques like input validation and bounds checking.
+
+Various modern C and C++ compilers allow one to automatically instrument a program during compilation using so-called _sanitizers_.
```
balazs-benics-sonarsource, line 295:
```suggestion
Various modern C and {cpp} compilers allow one to automatically instrument a program during compilation using so-called _sanitizers_.
```

balazs-benics-sonarsource (commented):
I've looked at the diff since the last review, and I think it's good.
I only had one remark.

balazs-benics-sonarsource (approved):

PR #3163, opened 2023-09-27
Modify rule S946: Expand and adjust for LaYC

pdschbrt:
## Review

- [ ] text style and tone


rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,175 @@
+Assigning the address of an automatic object to another object that outlives it is dangerous since the address is invalidated after the first object has ceased to exist.
```
frederic-tingaud-sonarsource, line 1: An explanation of what an "automatic object" is could be useful here.

pdschbrt: That's a good point. I'll try to craft a brief description.

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,178 @@
...
+----
+
+==== Compliant solution
+
+[source,cpp,diff-id=4,diff-type=compliant]
+----
+char buffer[64];
+
+void fun(char **out) {
+  for (size_t i = 0; i < sizeof(buffer) / sizeof(buffer[0]); ++i) {
```
frederic-tingaud-sonarsource, line 149:
```suggestion
  for (char &c: buffer) {
```
or
```suggestion
  std::fill(std::begin(buffer), std::end(buffer), 42);
```

frederic-tingaud-sonarsource (commented):
So far, the code samples are in C.
In a sense, I guess many users who will trigger these problems are C developers. On the other hand, we usually do C++ samples and this kind of samples can be an occasion to nudge old-school C++ developers toward more modern code.
So I would vote for C++ and more modern samples, but you can choose to follow this vote or not.

pdschbrt:
I agree with you: The rule is quite on the `C` side of life and I didn't want to overdo it in terms of making it `C++`.

rules/S946/cfamily/metadata.json
frederic-tingaud-sonarsource, line 1: I find the title a bit wordy and it also doesn't apply well to returning the address of a local variable.
What about
"The address of an automatic object should not be persisted after its lifetime"
or
"The address of an automatic object should not be persisted after the object ceases to exist"?

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
...
+Assigning the address of an automatic object to another object that outlives it is dangerous since the address is invalidated after the first object has ceased to exist.
 
-If the address of an automatic object is assigned to another automatic object of larger scope, or to a static object, or returned from a function then the object containing the address may exist beyond the time when the original object ceases to exist (and its address becomes invalid).
+== Why is this an issue?
 
+An automatic object is an object whose lifetime is automatically managed.
+The storage for an automatic object, e.g. a local variable, is allocated at the beginning of the enclosing code block and is deallocated at the end.
 
-=== Noncompliant code example
+If the address of an automatic object is assigned to another automatic object of larger scope, a static or extern object, or if it is returned from a function (using `return` or an output parameter), then the object containing the address may exist beyond the time at which the original object ceases to exist.
```
frederic-tingaud-sonarsource, line 8: Just a suggestion:
```suggestion
If the address of an automatic object is assigned to another automatic object of larger scope, a static or extern object, or if it is returned from a function (using `return` or an output parameter), then there will be a point where the address will point to an object that ceased to exist.
```

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
-== Why is this an issue?
+Assigning the address of an automatic object to another object that outlives it is dangerous since the address is invalidated after the first object has ceased to exist.
```
frederic-tingaud-sonarsource, line 1:
```suggestion
Persisting the address of an automatic object after it ceases to exist is dangerous because any attempt to dereference it would result in undefined behavior.
```
I think what makes this description harder to read is the fact that many sentences have two different "object" in them that need to be fully qualified each time to make sense.

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
...
+  }
+  *out = &local; // Noncompliant: returning the address of an object allocated on the stack (via output parameter)
+  return &local; // Noncompliant: returning the address of an object allocated on the stack
+}
+----
+
+
+== What is the potential impact?
+
+Persisting addresses of objects with automatic storage using objects that outlive them is dangerous as it may create invalid addresses.
```
frederic-tingaud-sonarsource, line 31:
```suggestion
Persisting addresses of objects with automatic storage past their lifetime is dangerous as it may create invalid addresses.
```

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
...
+Attempts to access such invalidated addresses result in undefined behavior.
+
+For programs that exercise undefined behavior, the compiler is no longer bound by the language specification.
+The application may crash or, even worse, the application may appear to execute correctly while losing data or producing incorrect results.
+
+
+== How to fix it
+
+Do not persist addresses of objects with automatic storage using objects that outlive them.
```
frederic-tingaud-sonarsource, line 40: This sentence only repeats what the title says. Summing up the below solutions might be more useful:

```suggestion
There are multiple approaches to avoid persisting addresses of object with automatic storage after their lifetime ended:

* Make a full copy of the object instead of using an address.
* Allocate the address on the heap and deallocate it once done.
* Create the automatic object in a larger scope where it outlives all storages of its address.
* Reduce the scope of objects that store the address of the original automatic object.
```

frederic-tingaud-sonarsource (commented):
Most of these are mere suggestions. Feel free to ignore them.

rules/S946/cfamily/rule.adoc
balazs-benics-sonarsource: +1, but I would recommend avoiding the term "dereference" as we can also have dangling references.
Have you considered the term "access"? See https://godbolt.org/z/43c3esze1

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
-== Why is this an issue?
+Assigning the address of an automatic object to another object that outlives it is dangerous since the address is invalidated after the first object has ceased to exist.
 
-If the address of an automatic object is assigned to another automatic object of larger scope, or to a static object, or returned from a function then the object containing the address may exist beyond the time when the original object ceases to exist (and its address becomes invalid).
+== Why is this an issue?
 
+An automatic object is an object whose lifetime is automatically managed.
+The storage for an automatic object, e.g. a local variable, is allocated at the beginning of the enclosing code block and is deallocated at the end.
```
balazs-benics-sonarsource, line 6:
```suggestion
The storage for an automatic object, e.g. a local variable, is allocated at the beginning of the enclosing code block and is deallocated at the end.
This is commonly referred to as "allocated on the stack".
```
I think it's useful to make a connection to the common name of this so that the user could connect this to other resources.

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
...
 
-=== Noncompliant code example
+If the address of an automatic object is assigned to another automatic object of larger scope, a static or extern object, or if it is returned from a function (using `return` or an output parameter), then the object containing the address may exist beyond the time at which the original object ceases to exist.
+In that case, the address becomes invalid and attempts to dereference the invalid address result in undefined behavior.
```
balazs-benics-sonarsource, line 9:
```suggestion
In that case, the address becomes invalid, and attempts to access it is undefined behavior.
```

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
...
+  int local = 42;
+  int *ptr;
+  global = &local; // Noncompliant: assigning the address of an object allocated on the stack to a global variable
+  {
+    int i = 9001;
+    ptr = &i; // Noncompliant: assigning the address of a stack-allocated object to an object that outlives it
+  }
+  *out = &local; // Noncompliant: returning the address of an object allocated on the stack (via output parameter)
+  return &local; // Noncompliant: returning the address of an object allocated on the stack
+}
```
balazs-benics-sonarsource, line 25: It's unfortunate, but we don't detect escapes through out parameters.
And I was surprised that we don't raise S6655 for `ptr`. Then I realized that we would need to read or write (in short: access) the dangling pointer `ptr` to make use raise the issue.

I'd recommend raising a ticket for the out parameter and reworking the example.

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,176 @@
...
+
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
+int *buz() {
+  int array[256];
+  for (size_t i = 0; i < std::size(array); ++i) {
+    array[i] = i;
+  }
+  return array; // Noncompliant: returns address of stack-allocated object
+}
```
balazs-benics-sonarsource, line 105: It's unfortunate, but due to loop unrolling of known bounds limitations, see CPP-3607, we won't analyze anything after the loop statement, thus we never reach the return where we could raise this issue.

rules/S946/cfamily/metadata.json
pdschbrt: Thanks for these suggestions. The title was indeed quite long.

I slightly modified your first suggestion and ended up with "The address of an automatic object should not be persisted beyond the object's lifetime" which sounds good to me.

rules/S946/cfamily/rule.adoc
pdschbrt: Interesting. I'll create a ticket for that and introduce lit tests in the corresponding `sonar-cpp` pull request that links to the ticket -- that's how Tomek and I handled such situations previously.

rules/S946/cfamily/rule.adoc
pdschbrt: At least we already have a ticket for that. I'll rework that code a bit, since we are not really interested in analyzing loops here.

rules/S946/cfamily/rule.adoc
pdschbrt: cf. https://sonarsource.atlassian.net/browse/CPP-4734

pdschbrt:
I wonder if we should add a "Going the extra mile" section and suggest using C++ alternatives (if possible) to avoid these situations in the first place.

Then, on the other hand, that would be also quite a stretch since this rule is about persisting addresses of automatic variables.

Let me know if you are happy with the current state or if we should add something, @frederic-tingaud-sonarsource, @balazs-benics-sonarsource.

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,242 @@
...
-* https://wiki.sei.cmu.edu/confluence/x/UtcxBQ[CERT, DCL30-C.] - Declare objects with appropriate storage durations
+=== Conference presentations
+
+* CppCon 2018 - https://www.youtube.com/watch?v=uQyT-5iWUow&ab_channel=CppCon[Surprises in Object Lifetime]
+
+=== Standards
+
+* CERT - https://wiki.sei.cmu.edu/confluence/x/UtcxBQ[DCL30-C. Declare objects with appropriate storage durations]
+* CERT - https://wiki.sei.cmu.edu/confluence/display/cplusplus/EXP54-CPP.+Do+not+access+an+object+outside+of+its+lifetime[EXP54-CPP. Do not access an object outside of its lifetime]
+* CERT - https://wiki.sei.cmu.edu/confluence/display/c/MSC00-C.+Compile+cleanly+at+high+warning+levels[MSC00-C. Compile cleanly at high warning levels]
```
balazs-benics-sonarsource, line 231: For these CERT links, sometimes its's a short link, sometimes a long link.
Please consolidate this and use the format that is used across different rules.

rules/S946/cfamily/rule.adoc
```diff
@@ -1,25 +1,242 @@
...
+For programs that exercise undefined behavior, the compiler is no longer bound by the language specification.
+The application may crash or, even worse, the application may appear to execute correctly while losing data or producing incorrect results.
+
+
+== How to fix it
+
+There are multiple approaches to avoid persisting addresses of objects with automatic storage after their lifetime ended:
+
+* Make a full copy of the object instead of using its address.
+* Allocate the object on the heap and deallocate it once done.
```
balazs-benics-sonarsource, line 44: Depending on the book, it's sometimes called "heap", and othertimes called "free-store".
I believe the Scott Meyers and Bjarne books differ on this taxonomy. Maybe it's better to xref these.
```suggestion
* Allocate the object on the heap (free-store) and deallocate it once done.
```

I found a paper actually using the "free-store" terminology: https://wg21.link/P1950.

balazs-benics-sonarsource (changes requested):
2 minor remarks left.

rules/S946/cfamily/rule.adoc
pdschbrt: Thanks for pointing this out. By majority vote (using `grep`) we now use short links 😁

balazs-benics-sonarsource (approved):
I'm good. Check Fred's comments and resolve them and merge this if they don't contradict.

PR #3165, opened 2023-09-27
Modify S6437: Change text to LaYC format (APPSEC-1111)

gaetan-ferry-sonarsource:
## Review

- [ ] text style and tone


rules/S6437/docker/rule.adoc
```diff
@@ -1,36 +1,47 @@
...
+accessed at container runtime. This will ensure the secret's security and
+prevent any further unexpected disclosure. Depending on the development platform
+and the leaked secret type, multiple solutions are currently available.
 
-Note that files exposing the secrets should be securely stored and not exposed
-to a large sphere. If possible, use a secret vault or another similar
-component. For example, *Docker Swarm* provides a *secrets* service that can be
-used to handle most confidential data.
+For all secrets that must be accessed at image build time, it is recommended to
+rely on Docker Buildkit's secret mount options. This will prevent secrets from
```
sebastien-andrivet-sonarsource, line 28: I think that now, the best practice is to use Docker secrets (https://docs.docker.com/engine/swarm/secrets/). But what you said remains true. Maybe:

it is recommended to rely on Docker Buildkit's secret mount options or on Docker secrets (with a link).

sebastien-andrivet-sonarsource (approved):
LGTM with a small suggestion.

rules/S6437/docker/rule.adoc
gaetan-ferry-sonarsource: I think there is a confusion here. Docker swarm secrets is the right solution for runtime secrets. Buildkit is recommended here for build time secrets, those that are used to access private services when building the image. I don't think docker swarm can handle that situation since it is a runtime service. WDYT?

gaetan-ferry-sonarsource: I'll change the runtime secrets' recommendation accordingly tho.

sebastien-andrivet-sonarsource: Yes, you are right.

PR #3173, opened 2023-09-28
Modify rule S2589: Extend LaYC content for Python

anton-haubner-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2589/python/rule.adoc
+++ b/rules/S2589/python/rule.adoc
@@ -1,6 +1,33 @@
+
+Gratuitous boolean expressions are conditions that do not change the evaluation
+of a program.
+This issue can indicate logical errors and affect the correctness of an
+application, as well as its maintainability.
+
 == Why is this an issue?
 
-include::../description.adoc[]
+Control flow constructs like `if`-statements allow the programmer to direct the
+flow of a program depending on a boolean expression.
+However, suppose that condition is always true or always false,
+independent of any input or other circumstances.
+In that case, the program will effectively always execute only one of the
+available execution branches.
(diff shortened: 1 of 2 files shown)


rules/S2589/python/rule.adoc
```diff
@@ -1,32 +1,113 @@
...
 
-include::../description.adoc[]
+Control flow constructs like `if`-statements allow the programmer to direct the
+flow of a program depending on a boolean expression.
+However, suppose that condition is always true or always false,
+independent of any input or other circumstances.
+In that case, the program will effectively always execute only one of the
+available execution branches.
+Thus, the control flow construct and the condition no longer serve a purpose,
+they become _gratuitous_.
```
Swalkyn, line 16: I think the wording could be a bit lighter here.

```suggestion
However, if the condition is always true or always false, only one of the branches
will be executed.
In that case, the control flow construct and the condition no longer serve a purpose;
they become _gratuitous_.
```

rules/S2589/python/rule.adoc
```diff
@@ -1,32 +1,113 @@
...
+def f(a, b):
+    if a is None and b is None:
+        do_something()
+    else:
+        do_something_else()
+----
+
+=== How does this work?
+
+In the first example, the gratuitous condition `a` is a `True` constant.
```
Swalkyn, line 94: Nitpick:

```suggestion
In the first example, the gratuitous condition `a` is simply`True`.
```

Swalkyn (approved):
Looks good! I left two suggestions.

PR #3178, opened 2023-09-28
Modify rule S2259: Extend LaYC content for Python

anton-haubner-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2259/python/rule.adoc
+++ b/rules/S2259/python/rule.adoc
@@ -1,10 +1,50 @@
+
+Accessing attributes on `None` is almost always a logical error and will raise
+an `AttributeError`.
+To fix this issue, make sure not to access an attribute or method on a value
+that can be `None`.
+
 == Why is this an issue?
 
-Attributes of ``++None++`` values should never be accessed. Doing so will cause an ``++AttributeError++`` to be thrown. At best, such an exception will cause abrupt program termination. At worse, it could expose debugging information that would be useful to an attacker, or it could allow an attacker to bypass security measures.
+`None` is a built-in object that represents the absence of a value.
+It is often used as a placeholder value for variables that only sometimes hold a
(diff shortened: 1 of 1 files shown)


rules/S2259/python/rule.adoc
```diff
@@ -1,31 +1,101 @@
...
+`None` does support a fixed set of special attributes like `++__class__++` or
+`++__bool__++`, and this issue is not raised when accessing these attributes.
+
+=== What is the potential impact?
+
+include::../../../shared_content/layc/exception-impact.adoc[]
+
+If a `None` value can be induced by user input, this issue may even be
+exploited by attackers to disrupt your application or gain information from
+stack traces.
```
Swalkyn, line 28: I think the special attributes access on `None` counts as an exception. I would move that to the dedicated section for it in the LaYC format.

```suggestion
=== What is the potential impact?

include::../../../shared_content/layc/exception-impact.adoc[]

If a `None` value can be induced by user input, this issue may even be
exploited by attackers to disrupt your application or gain information from
stack traces.

=== Exceptions

`None` does support a fixed set of special attributes like `++__class__++` or
`++__bool__++`, and this issue is not raised when accessing these attributes.
```

Swalkyn (approved):
Very nice! Just one small issue on the content organization.

rules/S2259/python/rule.adoc
anton-haubner-sonarsource: Good finding! I also saw this "Exceptions" section in your recent PR, and makes a lot of sense to organize it like that.
An explanation / definition of this section seems to be missing from the [LaYC format Xtranet page](https://xtranet-sonarsource.atlassian.net/wiki/spaces/PM/pages/2715320326/LaYC+Rule+Content+Specification).
Maybe this is also something we should bring up in the feedback session on Monday?

PR #3183, opened 2023-10-02
Modify S6471(docker): Improve fix section

loris-s-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6471/docker/rule.adoc
+++ b/rules/S6471/docker/rule.adoc
@@ -80,7 +80,8 @@ ENTRYPOINT ["/app/production"]
 
 == Compliant Solution
 
-For Linux-based images:
+For Linux-based images and scratch-based images that untar a Linux
+distribution:
 
 [source,docker]
 ----
@@ -105,7 +106,8 @@ RUN net user /add nonroot
 USER nonroot
 ----
 
-For multi-stage builds, the non-root user should be on the last layer:
+For multi-stage builds, the non-root user should be on the last stage:
+
 [source,docker]
 ----
 FROM alpine as builder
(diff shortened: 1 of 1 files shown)


rules/S6471/docker/rule.adoc
```diff
@@ -105,10 +105,28 @@ RUN net user /add nonroot
 USER nonroot
 ----
 
+For multi-stage builds, the non-root user should be on the last layer:
```
egon-okerman-sonarsource, line 108: "For multi-stage builds, a non-root user should be created and set in the last stage." 
Is this more correct, or did you really mean the last layer?

rules/S6471/docker/rule.adoc
```diff
@@ -105,10 +105,28 @@ RUN net user /add nonroot
...
+COPY --from=builder bin/production /app
+USER nonroot
+ENTRYPOINT ["/app/production"]
+----
+
 If the `scratch` Dockerfile untars a Linux distribution, the "Linux image"
-solution should be applied. Else, you have a choice between using a pre-written
-`/etc/passwd` file (regardless of the host operating system) or using a
-multi-stage build.
+solution should be applied. +
```
egon-okerman-sonarsource, line 124: IMO, I would simplify this as a whole:
* 'If the `scratch` Dockerfile untars a Linux distribution, the "Linux image" solution should be applied'
  * Integrate this in the "Linux image" solution instead.
* 'Here is a solution where a `passwd` file is stored right next to the source code and copied into the resulting image:'
  * This doesn't seem very clean, I would drop this.

What remains is an explanation that when you have `scratch` as a base image, you cannot add any users. Therefore, you introduce an extra build stage to add the group and user, and then later copy over `/etc/passwd` (with one example below it).

egon-okerman-sonarsource (changes requested):
Needs some changes, I don't think it is super clear yet.

rules/S6471/docker/rule.adoc
loris-s-sonarsource: correct ! thanks

rules/S6471/docker/rule.adoc
loris-s-sonarsource: totally aligned

rules/S6471/docker/rule.adoc
```diff
@@ -105,26 +106,28 @@ RUN net user /add nonroot
...
-COPY production_binary /app
-
-ENTRYPOINT ["/app/production_binary"]
+ENTRYPOINT ["/app/production"]
 ----
 
-or you can use a multi-stage build:
+With `FROM scratch` as the base image, it is not possible to add non-privileged
+users by default. To do this, add an additionnal build stage to add the group
+and user, and later copy `/etc/passswd`.
```
loris-s-sonarsource, line 127:
```suggestion
and user, and later copy `/etc/passwd`.
```

rules/S6471/docker/rule.adoc
```diff
@@ -105,26 +106,28 @@ RUN net user /add nonroot
...
+ENTRYPOINT ["/app/production"]
 ----
 
-or you can use a multi-stage build:
+With `FROM scratch` as the base image, it is not possible to add non-privileged
+users by default. To do this, add an additionnal build stage to add the group
+and user, and later copy `/etc/passwd`.
+
+Here is an example of copying a `passwd` file that defines a user that is
+generated using `adduser` in an earlier build stage:
```
egon-okerman-sonarsource, line 130:
```suggestion
Here is an example that generates a user using `adduser`, which adds the user to the `passwd` file, in the first stage. In the next stage, this user is added by copying that file:
```
This also reads better IMO, feel free to adjust.

loris-s-sonarsource: I think I want to keep what you suggested, even if it repeats it reads clearer. Also I follow a music youtuber who says "repetition legitimizes" so I feel like it's ok... Even if it's not music hahaha

PR #3190, opened 2023-10-02
Update C++ rules

amelie-renard-sonarsource:


rules/S3608/cfamily/rule.adoc
```diff
@@ -1,20 +1,20 @@
...
+Lambdas can use variables from their enclosing scope (called "capture") either by reference or by value. Since lambdas may run asynchronously, reference capture should be used with caution because, by the time the lambda runs, the referenced variable may be out of scope, resulting in an access violation at run time.
 
 
-You can specify default capture by reference (``++[&]++``), or by value (``++[=]++``). Clearly default reference capture can cause scope issues, but default value capture can also lead to problems. That's because both forms of default capture implicitly also capture ``++*this++``, which would automatically be used if for example you referenced a method from the enclosing scope.
+You can specify default capture by reference (``++[&]++``), or by value (``++[=]++``). Clearly, default reference capture can cause scope issues, but default value capture can also lead to problems. That's because both forms of default capture implicitly also capture ``++*this++``, which would automatically be used if, for example, you referenced a method from the enclosing scope.
```
frederic-tingaud-sonarsource, line 6:
```suggestion
You can specify default capture by reference (``++[&]++``), or by value (``++[=]++``). Default reference capture can cause scope issues, but so can default value capture. Both forms of default capture implicitly also capture ``++*this++``, which would automatically be used if, for example, you referenced a method from the enclosing scope.
```

rules/S6022/cfamily/rule.adoc
```diff
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-{cpp}17 introduced std::byte. It allows you to have byte-oriented access to a memory in a type-safe unambiguous manner. Before, you had to use either ``++char++``, ``++signed char++``, or ``++unsigned char++`` to access memory as bytes. The previous approach is error-prone as ``++char++`` type allows you to accidentally perform arithmetic operations. Also, it is confusing since ``++char++``, ``++signed char++``, and ``++unsigned char++`` are also used to represent actual characters and arithmetic values.
+{cpp}17 introduced `std::byte`. It allows you to have byte-oriented access to memory in a type-safe, unambiguous manner. Before, you had to use either ``++char++``, ``++signed char++``, or ``++unsigned char++`` to access memory as bytes. The previous approach is error-prone as the ``++char++`` type allows you to accidentally perform arithmetic operations. Also, it is confusing since ``++char++``, ``++signed char++``, and ``++unsigned char++`` are also used to represent actual characters and arithmetic values.
```
frederic-tingaud-sonarsource, line 3:
```suggestion
{cpp}17 introduced `std::byte`. It allows you to have byte-oriented access to memory in a type-safe, unambiguous manner. Before, you had to use either ``++char++``, ``++signed char++``, or ``++unsigned char++`` to access memory as bytes. The previous approach was error-prone as the ``++char++`` type allows you to accidentally perform arithmetic operations. Also, it was confusing since ``++char++``, ``++signed char++``, and ``++unsigned char++`` are also used to represent actual characters and arithmetic values.
```

rules/S6029/cfamily/rule.adoc
```diff
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-Starting {cpp}17, exception specifications became a part of a function type. This implies that these two functions, for example, have different types:
+From {cpp}17, exception specifications became a part of a function type. This implies that these two functions, for example, have different types:
```
frederic-tingaud-sonarsource, line 3:
```suggestion
Since {cpp}17, exception specifications have become a part of a function type. This implies that these two functions, for example, have different types:
```

frederic-tingaud-sonarsource (approved):
A few suggestions if you like them.

rules/S3608/cfamily/rule.adoc
amelie-renard-sonarsource: Better, thanks!

PR #3196, opened 2023-10-03
Add CI task to validate file extension

cristian-ambrosini-sonarsource:
The dotnet squad would like to improve the rule specification sprint process requiring some UTs to be temporarily added under the rule folder on RSPEC repository. These test case files (`.cs` and `.vb`) will be copied to the sonar-dotnet repository during the initial phases of implementation and will serve as an initial test bed. 
However, before merging the PR on the RSPEC side we need to make sure that these test case files are deleted and they don't end up on master. 

The goal of this PR would be to add a check on the pipeline that will fail if `.cs` or `.vb` files are detected (it's acceptable if it's red until the implementation is done).

### DIFF
--- a/ci/validate_file_extensions.sh
+++ b/ci/validate_file_extensions.sh
@@ -3,7 +3,7 @@ set -exo pipefail
 
 TOPLEVEL="$(realpath .)"
 RULES_DIR="${TOPLEVEL}/rules"
-CSVB_FILES=$(find "${RULES_DIR}" -type f -name "*.cs" -o -name "*.vb")
+CSVB_FILES=($(find "${RULES_DIR}" -type f -name "*.cs" -o -name "*.vb"))
 
 echo ${#CSVB_FILES[@]}
 
--- a/ci/validate_file_extensions.sh
+++ b/ci/validate_file_extensions.sh
@@ -1,9 +1,8 @@
 #!/bin/bash
-set -euxo pipefail
+set -exo pipefail
 
 TOPLEVEL="$(realpath .)"
 RULES_DIR="${TOPLEVEL}/rules"
-
 CSVB_FILES=$(find "${RULES_DIR}" -type f -name "*.cs" -o -name "*.vb")
 
 if [ "${#CSVB_FILES[@]}" -gt 0 ]; then
--- a/ci/validate_file_extensions.sh
+++ b/ci/validate_file_extensions.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 #
-# Validates that there are no files with .cs and .vb extensions present inside rules folder.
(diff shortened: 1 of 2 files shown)


ci/validate_file_extensions.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euxo pipefail
+
+TOPLEVEL="$(realpath .)"
+RULES_DIR="${TOPLEVEL}/rules"
+CSVB_FILES=($(find "${RULES_DIR}" -type f -name "*.cs" -o -name "*.vb"))
+
+if [ ${#CSVB_FILES[@]} -gt 0 ]; then
+    echo "ERROR: '.cs' and/or '.vb' files are detected."
+    echo $CSVB_FILES
```
gregory-paidis-sonarsource, line 10: This will print only the first element of the array. You need to either:
- Loop over them and print them with `for thingy in `${CSVB_FILES[@]}"`
- Use an oneliner like `printf '%s\n' "${CSVB_FILES[@]}"`

gregory-paidis-sonarsource (approved):
LGTM, please take care of the one comment I left.

ci/validate_file_extensions.sh
necto, line 1: I think this file deserves a comment similar to your PR comment.
It looks weird out of context: What is special about .cs and .vb files?

necto (approved):

ci/validate_file_extensions.sh
```diff
@@ -0,0 +1,25 @@
...
+# Those files can be of any number for both C# (*.cs) and VB.NET (.*vb).
+# The test case files will be copied to the sonar-dotnet repository during the initial phases of implementation 
+# and will serve as an initial test bed.
+# Before merging the PR on the RSPEC side, it is important to ensure that these test case files are deleted.
+# The script make sure to fail the CI if any of those previously mentioned files are present inside the rules folder.
+set -euxo pipefail
+
+TOPLEVEL="$(realpath .)"
+RULES_DIR="${TOPLEVEL}/rules"
+CSVB_FILES=($(find "${RULES_DIR}" -type f -name "*.cs" -o -name "*.vb"))
```
sebastien-marichal, line 16: Can you add `.razor` & `.cshtml` as well?

sebastien-marichal (changes requested):
We almost forgot about Blazor / Razor files 😱

ci/validate_file_extensions.sh
cristian-ambrosini-sonarsource: Good catch! I'll add them right away

sebastien-marichal (approved):
LGTM!

PR #3199, opened 2023-10-04
Modify rule S115: Extend LaYC content

anton-haubner-sonarsource:
## Review

- [ ] text style and tone


rules/S115/common/why-is-this-an-issue.adoc
```diff
@@ -0,0 +1,11 @@
...
+program after initialization.
+Oftentimes, constants are used in multiple locations across different
+subroutines.
+
+It is important that the names of constants follow a consistent and easily
+recognizable pattern.
+This way, the unique properties of constants are immediately communicated to
+the reader when observing any usage of a constant.
+For instance, they can be assured that the constant will not change, which
+simplifies debugging.
```
Swalkyn, line 11: I would also use a simpler phrasing here, what do you think of:

```suggestion
This way, readers immediately understand that the referenced value does not change.
```

Swalkyn (changes requested):
Looks good overall, but I think some of the explanations are not using the correct tone. I left some suggestions.

rules/S115/common/why-is-this-an-issue.adoc
anton-haubner-sonarsource: Well you are right, I was overcomplicating things here.
I like your suggestion, but I think we should also explain why this is useful (debugging), so I will try to keep that aspect, while also simplifying it.

rules/S115/common/what-is-the-potential-impact.adoc
```diff
@@ -0,0 +1,5 @@
+Applying a consistent naming scheme to constants improves the readability of
+code and makes it easier to understand, which will also boost team
+collaboration.
+Conversely, to forego a naming convention will negatively impact the overall
+maintainability of an application.
```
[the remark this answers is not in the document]
anton-haubner-sonarsource: Agreed, the sentences are too complex, and I like your simplification.
I would still like to retain the aspect that there is an impact on teamwork, so I'll try to fit that in while keeping it brief.

Swalkyn (approved):

PR #3220, opened 2023-10-09
Update C++ rules

amelie-renard-sonarsource:


rules/S1836/cfamily/rule.adoc
marco-antognini-sonarsource, line 1: The comment in the spreadsheet says "I think the rule itself should be reworked.." I don't know who wrote it and I partially agree -- and at the same time, I don't think it's worth the effort.

So LGTM. The linked CERT article provides enough details.

rules/S5018/cfamily/rule.adoc
```diff
@@ -1,29 +1,29 @@
 == Why is this an issue?
 
-Move operations (move constructor, move assignment operator) are all about efficient resource stealing. When stealing resources from the source, you don't have to allocate any memory or perform any other operation that might fail. This is why most people will expect move operation to be non-throwing.
+Move operations (move constructor, move assignment operator) are about efficient resource stealing. When stealing resources from the source, you don't have to allocate any memory or perform any other operation that might fail. This is why most people will expect move operations to be non-throwing.
```
marco-antognini-sonarsource, line 3: I'm uneasy with "stealing" as may convey it's a bad thing to do. Maybe "transfer ownership" is better? WDYT?

```suggestion
Move operations (move constructor, move assignment operator) are about efficiently transferring resource ownership. When transfering resources from the source, you don't have to allocate any memory or perform any other operation that might fail. This is why most people will expect move operations to be non-throwing.
```

Feel free to adapt as you like.

rules/S5018/cfamily/rule.adoc
```diff
@@ -1,29 +1,29 @@
...
+Additionally, if a move operation fails, the source object can have been partially altered by the move, making recovery very tricky or just impossible. Therefore, to ensure robustness, some functions (for instance, ``++std::move_if_noexcept++``, used by ``++std::vector++``) will decide to copy your object if its move operations are not decorated with ``++noexcept++``. This can significantly slow down your program.
 
 
-If you can not implement your move operations so that they never throw, you may as well only provide copy operations that will be safer to use.
+If you can not implement your move operations so that they never throw, you may only provide copy operations that will be safer to use.
 
 
-Swap operations are very similar to move operations, in that they should be equivalent to moving two objects into each other. So if you are adding a swap function to your type, it should be _noexcept_ too.
+Swap operations are similar to move operations in that they should be equivalent to moving two objects into each other. So if you add a swap function to your type, it should be _noexcept_ too.
```
marco-antognini-sonarsource, line 12: For consistency:
```suggestion
Swap operations are similar to move operations in that they should be equivalent to moving two objects into each other. So if you add a swap function to your type, it should be `noexcept` too.
```

rules/S5018/cfamily/rule.adoc
```diff
@@ -1,29 +1,29 @@
...
 
 
-Note that for most classes, you should not write your own move operations, but rely on the "Rule-of-Zero" (S4963).
+Note that you should not write your move operations for most classes but rely on the "Rule-of-Zero" (S4963).
 
 
 This rule raises an issue when a move or swap operation is not _noexcept_, which can happen in two cases:
```
marco-antognini-sonarsource, line 18: For consistency
```suggestion
This rule raises an issue when a move or swap operation is not `noexcept`, which can happen in two cases:
```

rules/S5274/cfamily/rule.adoc
```diff
@@ -1,31 +1,31 @@
 == Why is this an issue?
 
-Usually, when copying an object, the source object is unchanged, which means that all resources owned by the source objects have to be duplicated during the copy operation. In the case that the source object will no longer be used, this duplication is not efficient. Since {cpp}11, a mechanism named move semantic has been added to detect such cases and replace the expensive copy by a much cheaper move operation that will steal resources.
+Usually, when copying an object, the source object is unchanged, meaning all resources owned by the source objects must be duplicated during the copy operation. If the source object is no longer used, this duplication is inefficient. Since {cpp}11, a move semantic mechanism has been added to detect such cases and replace the expensive copy with a much cheaper move operation that will steal resources.
```
marco-antognini-sonarsource, line 3: re steal v transfer.
```suggestion
Usually, when copying an object, the source object is unchanged, meaning all resources owned by the source objects must be duplicated during the copy operation. If the source object is no longer used, this duplication is inefficient. Since {cpp}11, a move semantic mechanism has been added to detect such cases and replace the expensive copy with a much cheaper move operation that will transfer resources.
```

rules/S5274/cfamily/rule.adoc
```diff
@@ -1,31 +1,31 @@
...
 
 
-The cornerstone of move semantic is the ability to detect during a "copy" if the source object will be reused or not. There are three situations:
+The cornerstone of move semantics is detecting during a "copy" whether the source object will be reused or not. There are three situations:
 
-* The object is a temporary object, with no name, and if it can't be named, it can't be used
+* The object is a temporary object with no name, and if it can't be named, it can't be used
 * The object is used in some specific places, such as a return statement
-* The user explicitly promises to the compiler that he won't care for the current value of the object any longer. He does so by using the specific cast operation named ``++std::move++``. 
+* The user explicitly promises to the compiler that he won't care for the object's current value any longer. He does so by using the specific cast operation named ``++std::move++``.
```
marco-antognini-sonarsource, line 10: Make it gender neutral with "they":
```suggestion
* The user explicitly promises to the compiler that they won't care for the object's current value any longer. They do so by using the specific cast operation named ``++std::move++``.
```

[See the second definition in the Oxford dictionary](https://www.oxfordlearnersdictionaries.com/definition/english/they)

rules/S5274/cfamily/rule.adoc
```diff
@@ -1,31 +1,31 @@
...
+* in a return statement, if the returned object is a prvalue of the same class type as the function return type
+* in the initialization of a variable, if the initializer expression is a prvalue of the same class type as the variable type
 
 This rule reports an issue when the use of ``++std::move++`` prevents the copy elision from happening.
 
 
 === Noncompliant code example
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=noncompliant]
```
marco-antognini-sonarsource, line 28: In the example below, we have two comments "Noncompliant, prevents copy elision", and one comment "Noncompliant". Given the sentence above, the comments could all be simply "Noncompliant". WDYT?

rules/S5274/cfamily/rule.adoc
```diff
@@ -1,31 +1,31 @@
...
 ----
```
marco-antognini-sonarsource, line 29: `vector` is the only thing that is not fully qualified. I suggest writing `std::vector` instead.

marco-antognini-sonarsource (commented):
I'll review the rest later.

rules/S5018/cfamily/rule.adoc
amelie-renard-sonarsource: Indeed, I agree that "stealing" is not the best word. I like "transferring ownership" better, thanks!

rules/S5274/cfamily/rule.adoc
amelie-renard-sonarsource: Indeed, you're right, thanks!

rules/S5274/cfamily/rule.adoc
amelie-renard-sonarsource: Good catch, fixed in https://github.com/SonarSource/rspec/pull/3220/commits/a35d25859fc2556c45fdf8d05af34d7662897315

rules/S5274/cfamily/rule.adoc
amelie-renard-sonarsource: Agreed, fixed in https://github.com/SonarSource/rspec/pull/3220/commits/a35d25859fc2556c45fdf8d05af34d7662897315

marco-antognini-sonarsource: You kept the first one. I'm not sure it's intentional. I don't mind either way. Feel free to mark this discussion as resolved.

rules/S6016/cfamily/rule.adoc
```diff
@@ -9,31 +9,37 @@ This will become a problem:
 * When you want to capture the current state of the object.
 * When you want to pass a copy of the object to avoid any concurrency issue.
 
-``{cpp}14`` provides a workaround to solve this problem. Where you can take the underlying object by copy using the following pattern:
+``{cpp}14`` provides a workaround to solve this problem. You can copy the underlying object by using the following pattern:
```
marco-antognini-sonarsource, line 12: For consistency with typography in other rules, make C++14 regular text:
```suggestion
{cpp}14 provides a workaround to solve this problem. You can copy the underlying object by using the following pattern:
```

Another suggestion (feel free to dismiss): it's more than a workaround.
```suggestion
{cpp}14 provides a solution to this problem. You can copy the underlying object by using the following pattern:
```

rules/S6016/cfamily/rule.adoc
```diff
@@ -9,31 +9,37 @@ This will become a problem:
...
 ----
-auto lam = [& , copyOfThis = *this] { 
-std::cout << field; // implicitly calling “this” captured by reference
+auto lam = [& , copyOfThis = *this] {
+  std::cout << field; // implicitly calling “this” captured by reference
 };
 ----
-``{cpp}17`` solves this problem by introducing an explicit consistent way to capture ``++this++`` by copy:
 
+``{cpp}17`` solves this problem by introducing an explicit, consistent way to capture ``++this++`` by copy:
```
marco-antognini-sonarsource, line 28:
```suggestion
{cpp}17 solves this problem by introducing an explicit, consistent way to capture ``++this++`` by copy:
```

rules/S6016/cfamily/rule.adoc
```diff
@@ -9,31 +9,37 @@ This will become a problem:
...
 ----
-auto lam = [&, *this] { 
-std::cout << field // implicitly calling “this” captured by copy
+auto lam = [&, *this] {
+  std::cout << field // implicitly calling “this” captured by copy
 };
 ----
-This rule will flag the ``{cpp}14`` way of capturing the current object by copy and suggest to replace it with the ``{cpp}17`` way.
+
+This rule will flag the ``{cpp}14`` way of capturing the current object by copy and suggest replacing it with the ``{cpp}17`` way.
```
marco-antognini-sonarsource, line 37:
```suggestion
This rule will flag the {cpp}14 way of capturing the current object by copy and suggest replacing it with the {cpp}17 way.
```

rules/S6168/cfamily/rule.adoc
```diff
@@ -1,16 +1,16 @@
 == Why is this an issue?
 
-``++std::jthread++``, introduced in {cpp}20, is a wrapper around ``++std::thread++``. This way, it has the same functionalities as ``++std::thread++``, making the substitution really straightforward, while adding two interesting behaviors:
+``++std::jthread++``, introduced in {cpp}20, is a wrapper around ``++std::thread++``. This way, it has the same functionalities as ``++std::thread++``, making the substitution straightforward while adding two interesting behaviors:
 
-* It joins by default in its destructor. If a ``++std::thread++`` was not joined or detached before being destroyed, a call to ``++std::terminate++`` was made. This can’t happen with ``++std::jthread++``.
+* It joins by default in its destructor. If a ``++std::thread++`` was not joined or detached before being destroyed, a call to ``++std::terminate++`` was made. This behavior can’t happen with ``++std::jthread++``.
 * It can be canceled or stopped in some situations by calling ``++request_stop()++``
```
marco-antognini-sonarsource, line 6: For consistency with the previous bullet point:
```suggestion
* It can be canceled or stopped in some situations by calling ``++request_stop()++``.
```

rules/S6168/cfamily/rule.adoc
```diff
@@ -1,16 +1,16 @@
...
 
 This rule raises an issue as soon as ``++std::thread++`` is used.
```
marco-antognini-sonarsource, line 8: I personally find "as soon as" quite heavy. Feel free to dismiss if you disagree, it's subjective.
```suggestion
This rule raises an issue when ``++std::thread++`` is used.
```

rules/S5274/cfamily/rule.adoc
amelie-renard-sonarsource: Yes, it was intentional.
For example, for the user who starts by reading the code before the description :)

rules/S6168/cfamily/rule.adoc
```diff
@@ -33,5 +33,6 @@ int main() {
 
 == Resources
 
-* https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#cp25-prefer-gsljoining_thread-over-stdthread[{cpp} Core Guidelines CP.25] - Prefer "gsl::joining_thread" over "std::thread"
+* {cpp} reference - https://en.cppreference.com/w/cpp/thread/jthread[std::jthread]
+* {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#cp25-prefer-gsljoining_thread-over-stdthread[CP.25 - Prefer "gsl::joining_thread" over "std::thread"]
```
marco-antognini-sonarsource, line 37: Should these be in different categories? (Documentation and External coding guidelines) I personally don't mind.. feel free to dismiss.

rules/S6187/cfamily/rule.adoc
```diff
@@ -1,23 +1,26 @@
...
 
-* Less code to write (and therefore fewer bugs too),
+* Less code to write (and therefore fewer bugs, too),
 * Guaranteed consistency between all the comparison operators (for instance, in this situation, ``++a < b++`` and ``++!(a >= b)++`` will always return the same value).
-* Guaranteed symmetry for comparisons: If you can write ``++a<b++``, and that operation is resolved through ``++operator<=>++``, you can also write ``++b<a++``, and get a consistent result. Achieving the same result with classical comparison operators require to double the number overloads if ``++a++`` and ``++b++`` are of different types. 
+* Guaranteed symmetry for comparisons: If you can write ``++a<b++``, and that operation is resolved through ``++operator<=>++``, you can also write ``++b<a++``, and get a consistent result. Achieving the same result with classical comparison operators requires double the number overloads if ``++a++`` and ``++b++`` have different types.
```
marco-antognini-sonarsource, line 7: Small nit on spaces around operator:
```suggestion
* Guaranteed symmetry for comparisons: If you can write ``++a < b++``, and that operation is resolved through ``++operator<=>++``, you can also write ``++b < a++``, and get a consistent result. Achieving the same result with classical comparison operators requires double the number overloads if ``++a++`` and ``++b++`` have different types.
```

Another suggestion (Grammarly complained a bit about the current wording):
```suggestion
* Guaranteed symmetry for comparisons: If you can write ``++a < b++``, and that operation is resolved through ``++operator<=>++``, you can also write ``++b < a++``, and get a consistent result. Achieving the same result with classical comparison operators requires twice as many overloads if ``++a++`` and ``++b++`` have different types.
```

rules/S6187/cfamily/rule.adoc
```diff
@@ -1,23 +1,26 @@
...
 
-Additionally, if the ``++operator<=>++`` has the defaulted implementation, the compiler can also implicitly generate a defaulted implementation of operator==, simplifying the class definition one step further.
+Additionally, if the ``++operator<=>++`` has the defaulted implementation, the compiler can implicitly generate a defaulted implementation of operator==, simplifying the class definition one step further.
```
marco-antognini-sonarsource, line 9:
```suggestion
Additionally, if the ``++operator<=>++`` has the defaulted implementation, the compiler can implicitly generate a defaulted implementation of `operator==`, simplifying the class definition one step further.
```

rules/S6187/cfamily/rule.adoc
```diff
@@ -1,23 +1,26 @@
...
 
 
-Before {cpp}20, if was common to provide only ``++operator<++`` for a class, and ask the user of this class to write all his code only using this operator (this is what ``++std::map++`` requires of its key type, for instance). It is still advised in this case to replace the operator with ``++<=>++``: The quantity of required work is similar, and the user of the class will benefit from a much greater expressivity.
+Before {cpp}20, it was common to provide only ``++operator<++`` for a class and ask the user of this class to write all his code only using this operator (this is what ``++std::map++`` requires of its key type, for instance). In this case, it is still advised to replace the operator with ``++<=>++``: the quantity of required work is similar, and the class user will benefit from a much greater expressivity.
```
marco-antognini-sonarsource, line 12:
```suggestion
Before {cpp}20, it was common to provide only ``++operator<++`` for a class and ask the user of this class to write all their code only using this operator (this is what ``++std::map++`` requires of its key type, for instance). In this case, it is still advised to replace the operator with ``++<=>++``: the quantity of required work is similar, and users of the class will benefit from a much greater expressivity.
```

rules/S6187/cfamily/rule.adoc
```diff
@@ -1,23 +1,26 @@
...
 
 
-This rule reports user-provided comparison operators (member functions or free functions) ``++<++``, ``++<=++``, ``++>++`` and ``++>=++``.
+== How to fix it
 
+This rule reports user-provided comparison operators (member functions or free functions) ``++<++``, ``++<=++``, ``++>++`` and ``++>=++`` that could be replaced by defining the ``++operator<=>++`.
```
marco-antognini-sonarsource, line 17: A ` is missing:
```suggestion
This rule reports user-provided comparison operators (member functions or free functions) ``++<++``, ``++<=++``, ``++>++`` and ``++>=++`` that could be replaced by defining the ``++operator<=>++``.
```

rules/S6187/cfamily/rule.adoc
```diff
@@ -77,6 +106,12 @@ auto operator<=>(MyEnum lhs, MyEnum rhs) {
 
 == Resources
 
+=== Documentation
+
+* {cpp} reference - https://en.cppreference.com/w/cpp/language/operator_comparison#Three-way_comparison[Three-way comparison]
+
+=== Related rules
+
 * S6186 - Redundant comparison operators should not be defined.
```
marco-antognini-sonarsource, line 115: Should we also link to S6230?

rules/S6190/cfamily/rule.adoc
```diff
@@ -17,8 +17,7 @@ void log(std::string_view message, std::string_view func, std::uint_least32_t li
 
 #define TRACE(msg) std::cout << __FILE__ <<':' <<__LINE__ <<' ' << msg // Noncompliant
```
marco-antognini-sonarsource, line 18:
```suggestion
#define TRACE(msg) std::cout << __FILE__ << ':' <<__LINE__ << ' ' << msg // Noncompliant
```

rules/S6190/cfamily/rule.adoc
```diff
@@ -36,11 +35,14 @@ std::ostream trace(std::string_view msg,
   return std::cout <<location.file_name() <<':' <<location.line() <<' '<< msg;
```
marco-antognini-sonarsource, line 35:
```suggestion
  return std::cout << location.file_name() << ':' << location.line() << ' ' << msg;
```

rules/S6195/cfamily/rule.adoc
```diff
@@ -1,16 +1,16 @@
 == Why is this an issue?
 
-`std::enable_if` is a very important part of template meta-programming in {cpp} up to {cpp}17. Based on SFINAE, it can be used to subtly tune the behavior of overload resolution based on properties of types.
+`std::enable_if` is a very important part of template meta-programming in {cpp} up to {cpp}17. Based on SFINAE ("Substitution Failure Is Not An Error"), it can be used to subtly tune the behavior of overload resolution based on properties of types.
 
-However, using `std::enable_if` correctly is not easy, and requires skills and experience, for a resulting code that is not straightforward. Since {cpp}20, new features offer first-class support for what used to require `enable_if` trickery:
+However, using `std::enable_if` correctly is not easy and requires skills and experience for a resulting code that is not straightforward. Since {cpp}20, new features offer first-class support for what used to require `enable_if` trickery:
```
marco-antognini-sonarsource, line 5: I'd strengthen this point by mentioning issues with maintenance and using `std::` consistently.
```suggestion
However, using `std::enable_if` correctly is not easy and requires skills and experience for a resulting code that is not straightforward and costly to maintain. Since {cpp}20, new features offer first-class support for what used to require `std::enable_if` trickery:
```

The second sentence can also be slightly simplified, maybe?
```suggestion
However, using `std::enable_if` correctly is not easy and requires skills and experience for a resulting code that is not straightforward and costly to maintain. Since {cpp}20, new features replace complex uses of `std::enable_if`:
```

rules/S6195/cfamily/rule.adoc
```diff
@@ -1,16 +1,16 @@
...
 
-Additionally, since those features provide a higher level of abstraction, compilers understand them better and can provide clearer diagnostics when a constraint is violated.
+Additionally, since those features provide a higher level of abstraction, compilers understand them better and can provide more straightforward diagnostics when a constraint is violated.
 
-As a consequence, `std::enable_if` is no longer the right tool and should be replaced with those facilities. Note that the replacement is not always mechanical: The expression controlling a `std::enable_if` would probably be acceptable as a `requires` condition, but better alternatives usually exist, for instance reusing an existing concept defined in the standard.
+Consequently, `std::enable_if` is no longer the right tool and should be replaced with those facilities. Note that the replacement is not always mechanical: The expression controlling a `std::enable_if` would probably be acceptable as a `requires` condition, but better alternatives usually exist, for instance, reusing an existing concept defined in the standard.
```
marco-antognini-sonarsource, line 13:
```suggestion
Consequently, `std::enable_if` is no longer the right tool and should be replaced with those facilities. Note that the replacement is not always mechanical: The expression controlling a `std::enable_if` would probably be acceptable as a `requires` condition. However, better alternatives usually exist, for instance, by reusing an existing concept defined in the standard.
```

marco-antognini-sonarsource (approved):
I raised a couple of nits. Feel free to drop suggestions you disagree with. I'd be happy to reread it if you want to, but it may not be necessary.

rules/S6168/cfamily/rule.adoc
amelie-renard-sonarsource: Indeed, that is what I usually do. Fixed in https://github.com/SonarSource/rspec/pull/3220/commits/7394fa6466cdf8fa779a7577e261a6c10cb38632

rules/S6187/cfamily/rule.adoc
amelie-renard-sonarsource: Yes, added in https://github.com/SonarSource/rspec/pull/3220/commits/cf9d919c18a7db9da20a9d8d1b688aca80e25aa9

rules/S6195/cfamily/rule.adoc
amelie-renard-sonarsource: I have rewritten the paragraph in https://github.com/SonarSource/rspec/pull/3220/commits/e9661619708b5fe4fe64bf1d115fa94f416bd6b6. WDYT?

marco-antognini-sonarsource (approved):

rules/S6195/cfamily/rule.adoc
marco-antognini-sonarsource: Looks great!

amelie-renard-sonarsource:
Thanks for your thorough review!

PR #3227, opened 2023-10-09
Modify S1128: Migrate To LayC

zsolt-kolbay-sonarsource:
## Review

- [ ] text style and tone


rules/S1128/csharp/rule.adoc
frederic-tingaud-sonarsource, line 1: Don't we want to update vbnet at the same time? The current description even has a missing word:
`Although unnecessary won't change anything to the produced application`

rules/S1128/csharp/rule.adoc
```diff
@@ -1,49 +1,83 @@
...
+Starting with C# 10, it's possible to define https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#global-modifier[global usings] for an entire project. They reduce the need for repetitive namespace inclusions, but can also mask which namespaces are truly necessary for the code at hand. Over-relying on them can lead to less transparent code dependencies, especially for newcomers to the project.
+
+=== Exceptions
+
+The rule will not raise a warning for `global using` directives, even if none of the types of that namespace are used in the project:
 
 [source,csharp]
 ----
-using System.Collections.Generic; // Noncompliant - unnecessary using
+global using System.Net.Sockets; // Compliant
```
frederic-tingaud-sonarsource, line 19:
```suggestion
global using System.Net.Sockets; // Compliant by exception
```

rules/S1128/java/rule.adoc
```diff
@@ -1,20 +1,77 @@
 == Why is this an issue?
 
-include::../description.adoc[]
+Unnecessary imports refer to importing types that are not used or referenced anywhere in the code.
```
frederic-tingaud-sonarsource, line 3: Would this description work for Kotlin too? If it does, we want to keep the description.adoc and update it with this description. Otherwise, we probably want to update Kotlin or ask the JVM team to handle it afterwards.

frederic-tingaud-sonarsource (changes requested):
The content is great, but I would also change vbnet and kotlin and remove unused files.

rules/S1128/csharp/rule.adoc
zsolt-kolbay-sonarsource: The rule for VB.NET is not implemented. Microsoft is not actively developing VB.NET anymore, so it's highly unlikely that we'll put effort into implementing this rule in VB.NET in the future. I'm removing the vbnet folder.

PR #3228, opened 2023-10-09
Modify S1131: Migrate to LayC - trailing spaces

frederic-tingaud-sonarsource:
## Review

- [ ] text style and tone


rules/S1131/swift/rule.adoc
```diff
@@ -1,27 +1 @@
...
-
-include::../description.adoc[]
-
-=== Noncompliant code example
-
-[source,swift]
-----
-// Noncompliant; there are extra spaces after 'String'
-var str:String     
-----
```
rudy-regazzoni-sonarsource, line 11: Why do we remove this example?
As it may not be the most used language, I can understand that we don't want to spend time writing a dedicated example, but since it was already here, to me it makes sense to keep it. I may be missing some context/logic regarding this, let me know in that case.
This brings to me another question: I see Python doesn't have any examples, while it's one of our flagship languages (or I believe so, couldn't find the flagship language list). Maybe it could benefit to have one?

rudy-regazzoni-sonarsource (approved):
LGTM! I like the rewording of the description.
Please see my comment.

rules/S1131/swift/rule.adoc
frederic-tingaud-sonarsource: For me, this is not a code example, but a description inside a code tag: the trailing spaces are not visible, so we end up describing the problem in a comment, not showing it in the code. "Imagine there is a space after this random line of code:". IMO this is not more informative than the full-text description.

I kept the code sample in JS because it brings value by illustrating the line continuation case, even though there too, the comment does all the heavy lifting. And the PL-SQL I kept because it makes no sense to me!

We could add a line continuation example in other languages if you think it's valuable, but they usually don't compile, so I'm not super convinced of their value.

PR #3233, opened 2023-10-10
Modify S1199: Migrate to LayC - Nested code blocks should not be used

ADarko22:


rules/S1199/javascript/rule.adoc
```diff
@@ -1,36 +1,13 @@
 == Why is this an issue?
 
-Nested code blocks can be used to create a new scope: variables declared within that block cannot be accessed from the outside, and their lifetime end at the end of the block. However this only happens when you use ES6 `let` or `const` keywords, a class declaration or a function declaration (in strict mode). Otherwise the nested block is redundant and should be removed.
+include::../description.adoc[]
```
marco-antognini-sonarsource, line 3: I prefer the previous description (modulo the grammar typos) because this version of the rule for JS intentionally accept some form of nested blocks despite what the SRP says. There is also a notion of uselessness since some blocks have no effect (variable can still be accessed outside of them) that is not present in other languages. The first sentence in the included document says the opposite. It is also the case in the previous description but the second sentence was "correcting" this. Here, the reader has to read a lot more to understand the block does not always introduce a new scope.

rules/S1199/javascript/rule.adoc
```diff
@@ -1,36 +1,13 @@
...
-
-if (condition) {
-    doSomething();
-    doOtherStuff();
-}
-----
+== How to fix it
 
-=== Exceptions
+The superfluous code blocks should be extracted into separate methods. The rule does not apply to the following cases:
```
marco-antognini-sonarsource: This list of exception should be in the first section. A "Exceptions" subsection would be nice I think.

rules/S1199/javascript/rule.adoc
```diff
@@ -60,10 +37,43 @@ if (condition) {
...
+}
+----
+
 
 == Resources
-=== Documentation
 
+=== Documentation
+* https://en.wikipedia.org/wiki/Single-responsibility_principle[Wikipedia - Single Responsibility Principle]
 * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block[MDN - block statement]
```
marco-antognini-sonarsource, line 80: These should be updated to use this format:

```
* MDN web docs - <link with title>
```

https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc#short-form-names-to-use

rules/S1199/description.adoc
```diff
@@ -1 +1,6 @@
-Nested code blocks can be used to create a new scope and restrict the visibility of the variables defined inside it. Using this feature in a method typically indicates that the method has too many responsibilities, and should be refactored into smaller methods.
+Nested code blocks create a new scope where variables declared within are inaccessible from the outside and their lifespan ends with the block.
```
marco-antognini-sonarsource, line 1:
```suggestion
Nested code blocks create new scopes where variables declared within are inaccessible from the outside, and their lifespan ends with the block.
```

rules/S1199/description.adoc
```diff
@@ -1 +1,6 @@
...
+
+Although this may appear beneficial, its usage within a function often suggests that the function is overloaded, thus it may violate the Single Responsibility Principle,
```
marco-antognini-sonarsource, line 3: I suggest breaking up the sentence to make it more accessible to non-native speakers.
```suggestion
Although this may appear beneficial, their usage within a function often suggests that the function is overloaded. Thus, it may violate the Single Responsibility Principle,
```

rules/S1199/description.adoc
```diff
@@ -1 +1,6 @@
...
+and needs to be broken down into smaller functions.
+
+The presence of superfluous blocks that don't affect the control flow or create a new scope can be misleading and might suggest possible mistakes in the code.
```
marco-antognini-sonarsource, line 6: I'd remove "superfluous" here too.

"or create a new scope" is inaccurate in C/C++/Java/C#. I would drop it from this common description and not include this file in JS.

marco-antognini-sonarsource (changes requested):
Please, never force-push while someone is doing a review. It breaks tracking of comments and I can't easily know if some of my comments were lost.

Pitfalls and Exceptions are not the same thing. For this rule, Exceptions seem more appropriate.

rules/S1199/javascript/rule.adoc
ADarko22: It seems that the Exception subsection is not allowed anymore: https://docs.google.com/presentation/d/1GJSujmu8hUhtFdWPm4X9X7Eh2FiZMAC1ueFtaZZPNUk/edit#slide=id.g21a51453c43_0_79

So I was considering having it under Pitfaulls. Does it make sense?

marco-antognini-sonarsource: The slides are most likely outdated. See https://github.com/SonarSource/rspec/blob/master/docs/description.adoc#2-learn-as-you-code-rule-format

rules/S1199/cfamily/rule.adoc
```diff
@@ -1,16 +1,22 @@
...
 
-[source,cpp]
-----
+=== Code examples
+
+==== Noncompliant code example
 
+To fix your code remove redundant blocks.
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
```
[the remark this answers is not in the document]
ADarko22: I think it's still useful to have the highlighting because not all the lines are reported. I can generate the preview fro java and it looks like this:
Screenshot 2023-10-10 at 15 50 17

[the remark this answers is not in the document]
ADarko22: So I would like to keep it

rules/S1199/java/rule.adoc
```diff
@@ -2,50 +2,72 @@
...
-include::../message.adoc[]
+[source,java,diff-id=1,diff-type=compliant]
+----
+class Example {
+
+    private final Deque<Integer> stack = new LinkedList<>();
+
+    public void evaluate(int operator) {
+      switch (operator) {
+        case ADD: {
```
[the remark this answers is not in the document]
ADarko22: What do you mean? I checked the code snippets and both should be correct

rules/S1199/cfamily/rule.adoc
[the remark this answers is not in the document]
marco-antognini-sonarsource: Indeed, it looks better than I would have expected! Thanks for the screenshot.

I wasn't able to test the CFamily version (unrelated problems) but a quick diff in the terminal seems to provide similar results. 👍

rules/S1199/cfamily/rule.adoc
```diff
@@ -68,6 +72,11 @@ void f(Cache &c, int data) {
   }
```
marco-antognini-sonarsource, line 72: A brace was missing for a long time...
```suggestion
  }
}
```

rules/S1199/java/rule.adoc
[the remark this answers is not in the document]
marco-antognini-sonarsource: The C# (for example) has this exception:

> The usage of a code block after a case is allowed.

But the Java version has no such text. I agree the example seems valid. Only the sentence seems missing.

rules/S1199/javascript/rule.adoc
```diff
@@ -41,46 +70,14 @@ if (condition) {
...
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block[MDN - block statement]
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var[MDN - var]
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let[MDN - let]
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const[MDN - const]
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/class[MDN - class declaration]
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function[MDN - function declaration]
-* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode[MDN - strict mode]
\ No newline at end of file
+* Wikipedia - https://en.wikipedia.org/wiki/Single-responsibility_principle[Single Responsibility Principle]
+* MDN -  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block[block statement]
```
marco-antognini-sonarsource, line 77: The doc seems to prefer this version:
```suggestion
* MDN web docs -  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block[block statement]
```
I did not check what was used in practice and I kind of agree "MDN" is enough. I let you decide which version is more appropriate.

marco-antognini-sonarsource (approved):
Thanks for the update. LGTM with some small nits. No need for another round of review unless you want to.

PR #3238, opened 2023-10-10
Modify S1104: Migrate To LayC - avoid public field

renaud-tognelli-sonarsource:
## Review

- [ ] text style and tone


rules/S1104/description.adoc
```diff
@@ -1,4 +1,4 @@
-Public class variable fields do not respect the encapsulation principle and has three main disadvantages:
+Public fields in public classes do not respect the encapsulation principle and have three main disadvantages:
```
ADarko22, line 1: The last sentence can be improved to not start with "By using". Instead of:

> By using private attributes and accessor methods (set and get), unauthorized modifications are prevented.

We could have:

> To prevent unauthorized modifications private attributes and accessor methods (set and get) should be used.

rules/S1104/java/rule.adoc
```diff
@@ -2,13 +2,65 @@
...
 
-include::../noncompliant.adoc[]
+include::../impacts.adoc[]
 
-include::../compliant.adoc[]
 
 === Exceptions
 
-Because they are not modifiable, this rule ignores ``++public final++`` fields. Also, annotated fields, whatever the annotation(s) will be ignored, as annotations are often used by injection frameworks, which in exchange require having public fields.
+Because they are not modifiable, this rule ignores `public final` fields. Also, annotated fields, whatever the annotation(s) will be ignored, as annotations are often used by injection frameworks, which in exchange require having public fields.
```
ADarko22, line 10: I would consider not starting the sentence with "because". What do you think?

```suggestion
This rule ignores `public final` fields because they are not modifiable. Also, annotated fields, whatever the annotation(s) will be ignored, as annotations are often used by injection frameworks, which in exchange require having public fields.
```

rules/S1104/java/rule.adoc
```diff
@@ -2,13 +2,65 @@
...
+  }
+
+}
+----
+
+=== How does this work?
+
+By having a setter and a getter the code can control how the field is accessed and modified. 
+
+For example, adding validation in the setter method will ensure that only valid values are set.
```
ADarko22, line 61: We can avoid the empty line since the example relates to the previous sentence

```suggestion
By having a setter and a getter the code can control how the field is accessed and modified.  For example, adding validation in the setter method will ensure that only valid values are set.
```

rules/S1104/impacts.adoc
```diff
@@ -0,0 +1,7 @@
+=== What is the potential impact?
+
+When any part of the code can modify a public field, this can lead to unexpected changes and hard-to-trace bugs.
```
ADarko22, line 3: What about rephrasing it?

```suggestion
Public fields can be modified by any part of the code and this can lead to unexpected changes and hard-to-trace bugs.
```

rules/S1104/impacts.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+When a field is public, it is no longer possible to change how the data is stored internally without impacting the client code of the class.
```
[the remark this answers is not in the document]
renaud-tognelli-sonarsource: It sounds much better, thanks 🤩

renaud-tognelli-sonarsource:
Thanks a lot for the suggestions, the Rule is easier to read now.

rules/S1104/csharp/rule.adoc
```diff
@@ -1,48 +1,69 @@
...
-* Member values are subject to change from anywhere in the code and may not meet the programmer's assumptions.
+Note that due to optimizations on simple properties, public fields provide only very little performance gain.
 
-By using private fields and public properties (set and get), unauthorized modifications are prevented. Properties also benefit from additional protection (security) features such as Link Demands.
+include::../impacts.adoc[]
 
+=== Exceptions
 
-Note that due to optimizations on simple properties, public fields provide only very little performance gain.
+Fields marked as ``++readonly++`` or ``++const++`` are ignored by this rule.
```
zsolt-kolbay-sonarsource, line 11:
```suggestion
Fields marked as `readonly` or `const` are ignored by this rule.
```

rules/S1104/csharp/rule.adoc
```diff
@@ -1,48 +1,69 @@
...
+
+Fields inside classes or structs annotated with the ``++StructLayoutAttribute++`` are ignored by this rule.
+
+== How to fix it
+
+Depending on your needs:
+
+* Use auto-implemented properties. For common cases, where no validation is required, auto-implemented properties are a good alternative that does not have the drawbacks of public fields.
```
zsolt-kolbay-sonarsource, line 19: Which drawbacks of public fields are mitigated with auto-properties? 
- the field can still be modified by anyone outside the class without any restrictions because it's public
- it also doesn't hide the implementation details any better than a public field, or make it easier to maintain

The only reason I can think when public auto-implemented properties have a clear advantage is the usage of certain libraries - like Windows Presentation Foundation or ASP.NET Core Razor page - where certain functionalities only work on properties (like validation or setting the initial value in XAML).

zsolt-kolbay-sonarsource (approved):
LGTM

PR #3245, opened 2023-10-10
Modify S1481: Migrate to LaYC

nicolas-gauthier-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1481/csharp/rule.adoc
+++ b/rules/S1481/csharp/rule.adoc
@@ -2,7 +2,7 @@ include::../rationale.adoc[]
 
 === Exceptions
 
-Unused locally created resources in a ``++using++`` statement are not reported.
+Unused locally created resources in a ``using`` statement are not reported.
 
 [source,csharp]
 ----
--- a/rules/S1481/csharp/rule.adoc
+++ b/rules/S1481/csharp/rule.adoc
@@ -2,7 +2,7 @@ include::../rationale.adoc[]
 
 === Exceptions
 
-Unused locally created resources in a ``using`` statement are not reported.
+Unused locally created resources in a `using` statement are not reported.
 
 [source,csharp]
 ----
--- a/rules/S1481/csharp/rule.adoc
+++ b/rules/S1481/csharp/rule.adoc
@@ -1,27 +1,4 @@
-== Why is this an issue?
-
-include::../description.adoc[]
-
-=== Noncompliant code example
-
-[source,csharp]
-----
-public int NumberOfMinutes(int hours)
-{
-  int seconds = 0;   // seconds is never used 
-  return hours * 60;
-}
-----
-
-=== Compliant solution
-
-[source,csharp]
-----
-public int NumberOfMinutes(int hours)
-{
-  return hours * 60;
-}
-----
+include::../rationale.adoc[]
 
 === Exceptions
 
@@ -32,6 +9,31 @@ Unused locally created resources in a ``++using++`` statement are not reported.
 using(var t = new TestTimer()) // t never used, but compliant.
 {
   //...
+}
+----
+
+include::../how-to-fix-it.adoc[]
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
+public int NumberOfMinutes(int hours)
+{
+  int seconds = 0;   // Noncompliant - seconds is unused 
+  return hours * 60;
+}
+----
+
+==== Compliant solution
+
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+public int NumberOfMinutes(int hours)
+{
+  return hours * 60;
 }
 ----
 
--- a/rules/S1481/how-to-fix-it.adoc
+++ b/rules/S1481/how-to-fix-it.adoc
@@ -1,3 +1,3 @@
 == How to fix it
 
-The fix for this issue is straightforward, you just need to remove the unused variable declaration.
+The fix for this issue is straightforward. Once you ensure the unused variable is not part of an incomplete implementation leading to bugs, you just need to remove it.
--- a/rules/S1481/how-to-fix-it.adoc
+++ b/rules/S1481/how-to-fix-it.adoc
@@ -1,0 +1,3 @@
+== How to fix it
+
+The fix for this issue is straightforward, you just need to remove the unused variable declaration.
--- a/rules/S1481/plsql/rule.adoc
+++ b/rules/S1481/plsql/rule.adoc
@@ -8,13 +8,13 @@ include::../how-to-fix-it.adoc[]
 
 [source,sql,diff-id=1,diff-type=noncompliant]
 ----
-declare
-  seconds integer := 0; -- Noncompliant - seconds is unused
-  hours integer := 2;
-begin
+DECLARE
+  seconds INTEGER := 0; -- Noncompliant - seconds is unused
+  hours INTEGER := 2;
+BEGIN
   minutes := hours * 60;
-  dbms_output.put_line('Number of minutes: ' || minutes);
-end;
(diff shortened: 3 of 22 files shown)


rules/S1481/plsql/rule.adoc
```diff
@@ -1,4 +1,34 @@
...
+
+[source,sql,diff-id=1,diff-type=noncompliant]
+----
+declare
+  seconds integer := 0; -- Noncompliant - seconds is unused
+  hours integer := 2;
+begin
+  minutes := hours * 60;
+  dbms_output.put_line('Number of minutes: ' || minutes);
+end;
```
rudy-regazzoni-sonarsource, line 17: Code conventions and other examples in RSPEC for plsql seems to show that keywords and reference to standard library should be in uppercase.
```suggestion
DECLARE
  seconds INTEGER := 0; -- Noncompliant - seconds is unused
  hours INTEGER := 2;
BEGIN
  minutes := hours * 60;
  DBMS_OUTPUT.PUT_LINE('Number of minutes: ' || minutes);
END;
```
Also here I find that it improves the readability by differentiating your example with the sensitive code (the important part) VS the rest of the code/blocks (non-important part)

rules/S1481/python/rule.adoc
```diff
@@ -1,21 +1,35 @@
-== Why is this an issue?
+include::../rationale.adoc[]
 
-include::../description.adoc[]
+=== Exceptions
 
-=== Noncompliant code example
+``++_++`` will not raise an issue for this rule. The following examples are compliant:
```
rudy-regazzoni-sonarsource, line 5: Nitpick: I believe `` `_` `` could be used here instead, from what I was told the ``` ``++_++`` ``` is the old format. It is still valid and can be used when the first form is ambiguous/not working with specific escaped characters, but otherwise, the simpler form should be used.

rules/S1481/python/rule.adoc
```diff
@@ -1,21 +1,35 @@
...
+==== Noncompliant code example
+
+[source,python,diff-id=1,diff-type=noncompliant]
+----
 def hello(name):
-    message = "Hello " + name # Noncompliant
+    message = "Hello " + name # Noncompliant - message is unused
     print(name)
 for i in range(10):
     foo()
```
rudy-regazzoni-sonarsource, line 27: I would add a `# Noncompliant - ...` comment here also, It was not clear to me at first that this was a sensitive example that was fixed in your compliant solution.

rules/S1481/swift/rule.adoc
```diff
@@ -1,38 +1,40 @@
-== Why is this an issue?
+include::../rationale.adoc[]
 
-include::../description.adoc[]
+=== Exceptions
 
-=== Noncompliant code example
+Simple ``++for-in++`` loop counters are ignored by this rule because while they are often legitimately unused, their declaration is required by the syntax.
```
rudy-regazzoni-sonarsource, line 5: Same nitpick as precedent: `` `for-in` ``

rules/S1481/swift/rule.adoc
```diff
@@ -1,38 +1,40 @@
...
 
-=== Compliant solution
-
-[source,swift]
-----
-public func numberOfMinutes(hours:Int) -> Int{
-  return hours * 60
+for (a, b) in someElements {  // Noncompliant - b is unused
+  print(a)
 }
```
rudy-regazzoni-sonarsource, line 15: I don't get why we have a legit Noncompliant example in an exception code block example?
In other languages exception example, you don't have such thing, and indeed I would not expect this to be here.

rules/S1481/tsql/rule.adoc
```diff
@@ -1,4 +1,30 @@
...
+
+
+==== Compliant solution
+
+[source,sql,diff-id=1,diff-type=compliant]
+----
+declare 
+@DeliveryId int = 0;
+select 'Date', 'Weight' from Claims where Id = @DeliveryId;
+go
```
rudy-regazzoni-sonarsource, line 26: Same here:
```suggestion
DECLARE
@DeliveryId INT = 0;
SELECT 'Date', 'Weight' FROM Claims WHERE Id = @DeliveryId;
GO
```

rudy-regazzoni-sonarsource (changes requested):
Nice work! Good that you added all the missing examples for the languages.
I have a few suggestions, mostly nitpicking and minor modification requests, please have a look.

rules/S1481/python/rule.adoc
nicolas-gauthier-sonarsource: Good point!

rules/S1481/swift/rule.adoc
nicolas-gauthier-sonarsource: Good point, I will remove that, it's bring confusion, I don't know why it was there

rules/S1481/how-to-fix-it.adoc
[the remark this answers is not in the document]
nicolas-gauthier-sonarsource: I would prefer to keep it, as the code examples are a sub-section of "how to fix it".
The text can be modified to be more useful, maybe something like: 
`The fix for this issue is straightforward. Once you make sure the unused variable is not part of an incomplete implementation leading to bugs you just need to remove it. `

rules/S1481/csharp/rule.adoc
```diff
@@ -2,7 +2,7 @@ include::../rationale.adoc[]
 
 === Exceptions
 
-Unused locally created resources in a ``++using++`` statement are not reported.
+Unused locally created resources in a ``using`` statement are not reported.
```
rudy-regazzoni-sonarsource, line 5: There should be only one backtick:
```suggestion
Unused locally created resources in a `using` statement are not reported.
```

rules/S1481/how-to-fix-it.adoc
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Good point, I missed that the examples are in this section, probably mixed with my own PR :sweat_smile:
I like the new text 👍

rules/S1481/python/rule.adoc
```diff
@@ -2,7 +2,7 @@ include::../rationale.adoc[]
 
 === Exceptions
 
-``++_++`` will not raise an issue for this rule. The following examples are compliant:
+``_`` will not raise an issue for this rule. The following examples are compliant:
```
rudy-regazzoni-sonarsource, line 5:
```suggestion
`_` will not raise an issue for this rule. The following examples are compliant:
```

rules/S1481/swift/rule.adoc
```diff
@@ -2,17 +2,13 @@ include::../rationale.adoc[]
 
 === Exceptions
 
-Simple ``++for-in++`` loop counters are ignored by this rule because while they are often legitimately unused, their declaration is required by the syntax.
+Simple ``for-in`` loop counters are ignored by this rule because while they are often legitimately unused, their declaration is required by the syntax.
```
rudy-regazzoni-sonarsource, line 5:
```suggestion
Simple `for-in` loop counters are ignored by this rule because while they are often legitimately unused, their declaration is required by the syntax.
```

rules/S1481/vbnet/rule.adoc
```diff
@@ -2,7 +2,7 @@ include::../rationale.adoc[]
 
 === Exceptions
 
-Unused locally created resources in a ``++Using++`` statement are not reported.
+Unused locally created resources in a ``Using`` statement are not reported.
```
rudy-regazzoni-sonarsource, line 5:
```suggestion
Unused locally created resources in a `Using` statement are not reported.
```

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my last nitpicks :wink:

PR #3258, opened 2023-10-11
Modify S125: Migrate to LayC - commented code

rudy-regazzoni-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S125/cfamily/rule.adoc
+++ b/rules/S125/cfamily/rule.adoc
@@ -8,6 +8,8 @@ This rule does not apply to code documentation using Doxygen, QDoc, markdown, or
 
 == Resources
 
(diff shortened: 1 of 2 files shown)


frederic-tingaud-sonarsource (changes requested):
I agree with the idea of not changing something that is already valid for LayC.
In that case, I would still change the description slightly to avoid harsh words like "bloat". What do you think about
```
Commented-out code distracts the focus from the actual executed code. It creates a noise that increases maintainance code. And because it is never executed, it quickly becomes out of date and invalid.

Commented-out code should be deleted and can be retrieved from source control history if required.
```

frederic-tingaud-sonarsource (approved):

PR #3260, opened 2023-10-11
Modify S1066: Migrate to LayC - merge if

frederic-tingaud-sonarsource:
## Review

- [ ] text style and tone


petertrr (changes requested):
Please see my few comments. I wonder what's the motivation to keep [non]compliant examples in the `Why is this an issue` section and not in the `How to fix it` section?

rules/S1066/description.adoc
```diff
@@ -1 +1,3 @@
-Merging collapsible ``++if++`` statements increases the code's readability.
+Nested code - blocks of code inside blocks of code - is eventually necessary, but increases complexity. https://peps.python.org/pep-0020/[The Zen of Python], a text listing the Python philosophies, states "Flat is better than nested". That applies to other programming languages too and keeping the code as flat as possible, by avoiding unnecessary nesting, is considered a good practice.
```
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: Changed. What do you think of that one?

rules/S1066/noncompliant.adoc
```diff
@@ -1,9 +1,7 @@
-=== Noncompliant code example
-
-[source,text]
+[source,java]
```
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: It is shared between Java and Kotlin, which have very similar syntax.

zsolt-kolbay-sonarsource:
Please change the title of the rule to `Mergeable "if" statements should be combined`
The change comes from the review and discussion on the _Clean Code for Dummies_ book.
I'll also add @ganncamp as a reviewer.

rules/S1066/cfamily/rule.adoc
```diff
@@ -22,6 +22,37 @@ if (condition1 && condition2) { // Compliant
...
+== How to fix it
+
+If merging the conditions seems to result in a more complex code, extracting the condition or part of it in a named function or variable is a better approach to fix readability.
+
+=== Noncompliant code example
+
+[source,cpp]
+----
+if (file.isValid()) {
+  if (file.isFile() || file.isDirectory()) {   // Noncompliant
```
petertrr, line 34: I suspect that this looks more like Java code than like C++ handling of files. For example, `file.isDirectory()` could be `std::filesystem::is_directory(path)`. I would suggest to adapt these examples.

rules/S1066/csharp/rule.adoc
```diff
@@ -2,29 +2,63 @@
...
 {        
     // ...
 }
 ----
 
+== How to fix it
+
+If merging the conditions seems to result in a more complex code, extracting the condition or part of it in a named function or variable is a better approach to fix readability.
+
+=== Noncompliant code example
```
petertrr, line 32: I believe, there should be a level 3 heading `=== Code examples` and then Noncompliant and Compliant should be level 4

ganncamp (approved):

rules/S1066/cfamily/rule.adoc
frederic-tingaud-sonarsource: That's still a kind of API you can find regularly, but I switched to a condensed STL code.

rules/S1066/csharp/rule.adoc
frederic-tingaud-sonarsource: Thanks, fixed

rules/S1066/python/rule.adoc
```diff
@@ -2,23 +2,50 @@
...
+If merging the conditions seems to result in a more complex code, extracting the condition or part of it in a named function or variable is a better approach to fix readability.
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,python]
+----
+if file.isValid():
+  if file.isFile() or file.isDirectory():     # Noncompliant
```
petertrr, line 33: More pythonic example, AFAIU, would be with [os.path](https://docs.python.org/3/library/os.path.html#os.path.isdir) methods:
```suggestion
  if file.isfile() or file.isdir():     # Noncompliant
```

petertrr (approved):
LGTM, see my final suggestion about Python

PR #3266, opened 2023-10-12
Modify rule S116: Migrate to LayC - Field names should comply with a naming convention

ADarko22:


rules/S116/csharp/rule.adoc
```diff
@@ -2,26 +2,37 @@
...
 class MyClass {
    private int myField;
 }
 ----
 
+== Resources
+
+=== Documentation
+
+* Microsoft - https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines[.NET Naming Guidelines]
```
ADarko22, line 33: is this reference relevant?

rules/S116/flex/rule.adoc
```diff
@@ -2,26 +2,36 @@
...
 public class MyClass {
    public var myField:int;
 }
 ----
 
+== Resources
+
+=== Documentation
+
+* Wikipedia - https://en.wikipedia.org/wiki/Naming_convention_(programming)[Naming Convention (programming)]
```
ADarko22, line 33: could not find any specific documentation for FLEX

rules/S116/php/rule.adoc
```diff
@@ -2,26 +2,37 @@
...
 class MyClass {
   $myField;
 }
 ----
 
+== Resources
+
+=== Documentation
+
+* php.net - https://www.php.net/manual/en/userlandnaming.rules.php[Userland Naming Guide]
```
ADarko22, line 33: is this reference valid?

rules/S116/swift/rule.adoc
```diff
@@ -2,26 +2,44 @@
...
 class MyClass {
   var myField = 1
 }
 ----
 
+== Resources
+
+=== Documentation
+
+* Swift - https://www.swift.org/documentation/api-design-guidelines/#naming[API Design Guidelines - Naming]
```
ADarko22, line 33: are the references valid?

rules/S116/cfamily/rule.adoc
```diff
@@ -2,26 +2,37 @@
 
 include::../description.adoc[]
 
-=== Noncompliant code example
+== How to fix it
```
frederic-tingaud-sonarsource, line 5: The description is short and the code examples are super straightforward. I think just having the code samples in "Why is this an issue?" would be simpler.

rules/S116/cfamily/rule.adoc
```diff
@@ -2,26 +2,37 @@
...
 
-[source,cpp]
+The examples below enforce the following regular expression ``++^[a-z][a-zA-Z0-9]*$++``:
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
 class MyClass {
   int my_field;
```
frederic-tingaud-sonarsource, line 16:
```suggestion
  int my_field; // Noncompliant
```

rules/S116/cfamily/rule.adoc
```diff
@@ -2,26 +2,37 @@
...
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=compliant]
 ----
 class MyClass {
   int myField;
```
frederic-tingaud-sonarsource, line 25:
```suggestion
  int myField;  // Compliant
```

frederic-tingaud-sonarsource (changes requested):
What would you think about having the code samples in the first tab? Given how straightforward this all is, I think that would work better.

rules/S116/csharp/rule.adoc
renaud-tognelli-sonarsource: It gives weight to the rule.
Maybe point it directly to the field name convention, wdyt?
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields

rules/S116/php/rule.adoc
renaud-tognelli-sonarsource: There is not much about fields here, it is a bit misleading I would say.

rules/S116/swift/rule.adoc
renaud-tognelli-sonarsource: same here

rules/S116/php/rule.adoc
ADarko22: True, but it still brings some value to an user who wants to dig into the topic

rules/S116/cfamily/rule.adoc
```diff
@@ -2,26 +2,35 @@
 
 include::../description.adoc[]
 
-=== Noncompliant code example
+=== Code examples
```
frederic-tingaud-sonarsource, line 5: If we put the code in "Why is this an issue", we don't need the subtitles.

```

For example with the following regular expression ``++^[a-z][a-zA-Z0-9]*$++``:

----
class MyClass {
  int my_field; // Noncompliant
};
----

Should be replaced with:

----
class MyClass {
  int myField;  // Compliant
};
----
```

frederic-tingaud-sonarsource (changes requested):

frederic-tingaud-sonarsource (approved):

PR #3269, opened 2023-10-12
Modify S1121: Migrate to LayC - Assignments should not be made from within sub-expressions

petertrr:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1121/cfamily/rule.adoc
+++ b/rules/S1121/cfamily/rule.adoc
@@ -3,43 +3,22 @@
 Assigning a value inside a condition (of an `if` statement, a `for` statement, a `while`, or a `switch`) can be confusing. It assigns the value and checks it at the same time, but it is easily confused with a simple equality check with `==` and the original intention can be unclear.
 
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-void action(int parm) {
-  if (parm = getValue()) { // Noncompliant: assigning and checking. Is it on purpose?
+  if (x = getValue()) { // Noncompliant: assigning and checking. Is it on purpose?
     doSomething();
   }
-  // ...
-}
 ----
 
+It is better to assign before the statement and use the condition for the check only:
 
-{cpp}17 introduced `if` and `switch` statements with initializers, which allows to clearly separate the assignment from the check while keeping the code compact:
-
-
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=compliant]
 ----
-void action(int parm) {
-  if (parm = getValue(); parm) { // Compliant: Assignment and check are separate
+  x = getValue();
+  if (x) {
     doSomething();
   }
-  // ...
-}
 ----
-
-For older versions of the standard, it is better to assign before the statement:
-
-[source,cpp]
-----
-void action(int parm) {
-  parm = getValue();
-  if (parm) { // Compliant
-    doSomething();
-  }
-  // ...
-}
-----
-
 
 === Exceptions
 
@@ -57,8 +36,6 @@ while ((run = keepRunning())) {
 === Documentation
 
 * CWE - https://cwe.mitre.org/data/definitions/481[481: Assigning instead of Comparing]
-
-* {cpp} Reference - https://en.cppreference.com/w/cpp/language/if#If_statements_with_initializer[If statements with initializer]
 
 === Standards
 
--- a/rules/S1121/cfamily/rule.adoc
+++ b/rules/S1121/cfamily/rule.adoc
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-Assigning a value inside a condition (of an `if` statement, a `for` statement, a `while` or a `switch`) can be confusing. It assigns the value and checks it at the same time, but it is easily confused with a simple equality check with `==` and the original intention can be unclear.
+Assigning a value inside a condition (of an `if` statement, a `for` statement, a `while`, or a `switch`) can be confusing. It assigns the value and checks it at the same time, but it is easily confused with a simple equality check with `==` and the original intention can be unclear.
 
 
 [source,cpp]
@@ -14,7 +14,7 @@ void action(int parm) {
 ----
 
 
-{cpp}17 introduced `if` and `switch` statements with initializers, which allows to clearly separate the assignment and the check while keeping a compact code:
+{cpp}17 introduced `if` and `switch` statements with initializers, which allows to clearly separate the assignment from the check while keeping the code compact:
 
 
 [source,cpp]
@@ -27,7 +27,7 @@ void action(int parm) {
 }
 ----
 
-For older versions of the standard, it's better to assign before the statement:
+For older versions of the standard, it is better to assign before the statement:
 
 [source,cpp]
 ----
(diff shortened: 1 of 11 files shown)


rules/S1121/cfamily/rule.adoc
```diff
@@ -2,20 +2,44 @@
...
+include::../how-to-fix-it.adoc[]
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+int arr[] = {0, 1, 2, 3, 4};
+if (int a; a = arr[1]) {
```
petertrr, line 27: Apparently, the logic for CFamily is quite different from the other languages, and the generic example that used to reside in `[non]compliant.adoc` is in fact allowed by sonar-cpp. I tried to come up with a reasonable yet simple example, but it may be not the best one.

rules/S1121/csharp/rule.adoc
```diff
@@ -2,33 +2,13 @@
...
-var result = str.Substring(index, length);
-if (string.IsNullOrEmpty(result))
-{
-  //...
-}
-----
-
 === Exceptions
 
-Assignments inside lambda and delegate expressions are allowed.
```
petertrr, line 28: I'm not sure if this should be listed as exception, because for me it seems natural to treat assignments in function bodies, anonymous or not, as parts of the larger code and not as sub-expressions.

rules/S1121/java/rule.adoc
```diff
@@ -2,29 +2,61 @@
...
 include::../description.adoc[]
 
-include::../noncompliant.adoc[]
-
-include::../compliant.adoc[]
-
 === Exceptions
 
-Assignments in ``++while++`` statement conditions, and assignments enclosed in relational expressions are ignored.
+This rule ignores assignments in conditions of ``++while++`` statements and assignments enclosed in relational expressions.
```
zsolt-kolbay-sonarsource, line 7:
```suggestion
This rule ignores assignments in conditions of `while` statements and assignments enclosed in relational expressions.
```

If you want to highlight a single word (without escaping certain characters) as code you don't need to use ``++, a single tick character is enough.

rules/S1121/java/rule.adoc
```diff
@@ -2,29 +2,61 @@
...
+==== Compliant solution
+
+[source,java,diff-id=1,diff-type=compliant]
+----
+String str = cont.substring(pos1, pos2);
+if (!str.isEmpty()) {
+  // do something with "str"
+}
+----
+
```
zsolt-kolbay-sonarsource, line 59: You can include `../see.adoc` here, then delete the Resources section from this file.

rules/S1121/csharp/rule.adoc
petertrr: Should it be illustrated also with a code sample? I found this in tests of sonar-dotnet:
```csharp
 foo(
                (int xx) =>
                {
                    int a;
                    a = 42;
                    return a;
                });
```

zsolt-kolbay-sonarsource: There's no dead store in the code sample. The value 42 is assigned and returned. It can be written in a shorter way, but it's not an unnecessary assignment.
How about this?
```csharp
var result = Foo(() =>
{
   int x = 100; // dead store, but ignored
   x = 200;
   return x;
}
```

zsolt-kolbay-sonarsource (approved):
LGTM

rules/S1121/cfamily/rule.adoc
```diff
@@ -1,15 +1,51 @@
...
+----
+void action(int parm) {
+  if (parm = getValue(); parm) { // Compliant: Assignment and check are separate
+    doSomething();
+  }
+  // ...
+}
+----
+
+For older versions of the standard, it's better to assign before the statement:
```
marco-antognini-sonarsource, line 30: We should avoid contractions.
```suggestion
For older versions of the standard, it is better to assign before the statement:
```

rules/S1121/cfamily/rule.adoc
```diff
@@ -18,12 +54,22 @@ while ((run = keepRunning())) {
...
+=== Documentation
+
+* CWE - https://cwe.mitre.org/data/definitions/481[481: Assigning instead of Comparing]
+
+* {cpp} Reference - https://en.cppreference.com/w/cpp/language/if#If_statements_with_initializer[If statements with initializer]
+
+=== Standards
+
+* CERT https://wiki.sei.cmu.edu/confluence/x/ZNYxBQ[EXP45-C.: Do not perform assignments in selection statements]
+* CERT https://wiki.sei.cmu.edu/confluence/x/ITZGBQ[EXP51-J.: Do not perform assignments in conditional expressions]
```
marco-antognini-sonarsource, line 66: This is the Java version. It's not relevant in this language.
```suggestion
```

rules/S1121/cfamily/rule.adoc
```diff
@@ -1,15 +1,51 @@
...
+  }
+  // ...
+}
+----
+
 
 === Exceptions
 
-Assignments explicitly enclosed in parentheses are ignored.
+This rule ignores assignments explicitly enclosed in parentheses.
```
marco-antognini-sonarsource, line 46: @frederic-tingaud-sonarsource Should we mention S1110 "Redundant pairs of parentheses should be removed" here? Or do you think it's a detail and would just add noise?


(If relevant, this suggestion is probably not the best...)
```suggestion
This rule ignores assignments explicitly enclosed in parentheses and is compatible with S1110.
```

rules/S1121/cfamily/rule.adoc
```diff
@@ -18,12 +54,22 @@ while ((run = keepRunning())) {
...
 
+=== Documentation
+
+* CWE - https://cwe.mitre.org/data/definitions/481[481: Assigning instead of Comparing]
+
+* {cpp} Reference - https://en.cppreference.com/w/cpp/language/if#If_statements_with_initializer[If statements with initializer]
+
+=== Standards
+
+* CERT https://wiki.sei.cmu.edu/confluence/x/ZNYxBQ[EXP45-C.: Do not perform assignments in selection statements]
```
marco-antognini-sonarsource, line 65:
```suggestion
* CERT - https://wiki.sei.cmu.edu/confluence/x/ZNYxBQ[EXP45-C. Do not perform assignments in selection statements]
```

rules/S1121/java/rule.adoc
```diff
@@ -2,32 +2,63 @@
...
+include::../see.adoc[]
 
-* https://cwe.mitre.org/data/definitions/481[MITRE, CWE-481] - Assigning instead of Comparing
 * https://wiki.sei.cmu.edu/confluence/x/ZNYxBQ[CERT, EXP45-C.] - Do not perform assignments in selection statements
```
marco-antognini-sonarsource, line 62: The C version is not relevant in Java.
```suggestion
```

rules/S1121/java/rule.adoc
```diff
@@ -2,32 +2,63 @@
...
 * https://wiki.sei.cmu.edu/confluence/x/ITZGBQ[CERT, EXP51-J.] - Do not perform assignments in conditional expressions
```
marco-antognini-sonarsource, line 63:
```suggestion
* CERT - https://wiki.sei.cmu.edu/confluence/x/ITZGBQ[EXP51-J. Do not perform assignments in conditional expressions]
```

rules/S1121/cfamily/rule.adoc
marco-antognini-sonarsource, line 30: @frederic-tingaud-sonarsource Isn't this version the best even in C++17+? My thinking is that the purpose of the init statement in if/switch is to introduce a variable with a limited scope, which is different from the proposed fix. WDYT?

So, I would start with the C++98 solution. If you think is has value, I would mention it's possible to use C++17 init statement in if/switch without giving an example, but I'm not 100% sure it's interesting to know that in this context.

marco-antognini-sonarsource (commented):
FYI I had a look only at the CFamily description.
Ping me if I should review the rest as well.

rules/S1121/cfamily/rule.adoc
```diff
@@ -18,12 +54,22 @@ while ((run = keepRunning())) {
...
+* CERT https://wiki.sei.cmu.edu/confluence/x/ITZGBQ[EXP51-J.: Do not perform assignments in conditional expressions]
+
+=== External guidelines
```
marco-antognini-sonarsource, line 68:
```suggestion
=== External coding guidelines
```

frederic-tingaud-sonarsource (approved):

PR #3272, opened 2023-10-12
Modify S4649: Migrate to LayC

victor-diez-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4649/css/rule.adoc
+++ b/rules/S4649/css/rule.adoc
@@ -3,7 +3,7 @@
 The `font-family` (and the shorthand `font`) CSS property specifies a prioritized list of one or more font family names and/or generic family names for the selected element.
 
 
-If none of the font names defined in a ``++font++`` or ``++font-family++`` declaration are available on the browser of the user, the browser will display the text using its default font. It's recommended to always define a generic font family for each declaration of ``++font++`` or ``++font-family++`` to get a less degraded situation than relying on the default browser font. This lets the browser select an acceptable fallback font when necessary.
+If none of the font names defined in a `font` or `font-family` declaration are available on the browser of the user, the browser will display the text using its default font. It's recommended to always define a generic font family for each declaration of `font` or `font-family` to get a less degraded situation than relying on the default browser font. This lets the browser select an acceptable fallback font when necessary.
 
 The list of generic font families is as follows:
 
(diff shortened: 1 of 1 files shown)


rules/S4649/css/rule.adoc
```diff
@@ -1,21 +1,34 @@
...
+The `font-family` (and the shorthand `font`) CSS property specifies a prioritized list of one or more font family names and/or generic family names for the selected element.
 
 
-=== Noncompliant code example
+If none of the font names defined in a ``++font++`` or ``++font-family++`` declaration are available on the browser of the user, the browser will display the text using its default font. It's recommended to always define a generic font family for each declaration of ``++font++`` or ``++font-family++`` to get a less degraded situation than relying on the default browser font. This lets the browser select an acceptable fallback font when necessary.
```
petertrr, line 6: I think here and below using double backticks with plus signs is not essential and single backticks will suffice

rules/S4649/css/rule.adoc
```diff
@@ -1,21 +1,34 @@
...
+
+
+[source,css,diff-id=1,diff-type=noncompliant]
 ----
 a { 
   font-family: Helvetica, Arial, Verdana, Tahoma; /* Noncompliant; there is no generic font family in the list */
 }
 ----
 
+You should always include at least one generic family name in a font-family list, since there's no guarantee that any given font is available.
```
petertrr, line 29: I suggest to move this text to the start of `How to fix it` section before `Code examples`

rules/S4649/css/rule.adoc
```diff
@@ -25,17 +38,6 @@ a {
...
-
-=== Message
-
-Add a generic font name.
-
-
-endif::env-github,rspecator-view[]
+* CSS Specification - https://www.w3.org/TR/CSS2/fonts.html#generic-font-families[Generic font families]
+* MDN - https://developer.mozilla.org/en-US/docs/Web/CSS/font-family[``++font-family++``]
+* MDN - https://developer.mozilla.org/en-US/docs/Web/CSS/font[``++font++``]
```
petertrr, line 43: Why have you removed the conditional section `ifdef::env-github,rspecator-view[]`?

petertrr (changes requested):
Nice enhancement of the text! Please see a couple of my suggestions

rules/S4649/css/rule.adoc
victor-diez-sonarsource: true, no special chars, I always use the full non-escaping by default. I'll change that

rules/S4649/css/rule.adoc
victor-diez-sonarsource: that section is legacy containing comments from changes in the past. it was mentioned that if it did not contain valuable information it could be safely removed.

petertrr: What about the `Message` section? I think it's still valuable, as the rspec is source of truth for the analyzer

victor-diez-sonarsource: let me ask that in the slack channel

petertrr (approved):
LGTM
I think it can be merged now, even if the discussion about message may continue

PR #3276, opened 2023-10-13
Modify S5738: Migrate to LayC - `@Deprecated` code marked for removal should never be used

petertrr:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5738/java/rule.adoc
+++ b/rules/S5738/java/rule.adoc
@@ -1,19 +1,22 @@
 == Why is this an issue?
 
-Java 9 introduced a flag for the ``++@Deprecated++`` annotation, which allows to explicitly say if the deprecated code is planned to be removed at some point or not. This is done using ``++forRemoval=true++`` as annotation parameter. The javadoc of the annotation explicitly mention the following: 
-
-
-____
-If true, it means that this API element is earmarked for removal in a future release.
-
-If false, the API element is deprecated, but there is currently no intention to remove it in a future release.
+With the introduction of Java 9, the standard annotation class `java.lang.Deprecated` has been updated with new parameters. Notably, a boolean parameter `forRemoval` has been added to clearly signify whether the deprecated code is intended to be removed in the future. This is indicated with `forRemoval=true`. The javadoc of the annotation explicitly mentions the following: 
 
 ____
+This annotation type has a boolean-valued element `forRemoval`. A value of `true` indicates intent to remove the annotated program element in a future version. A value of `false` indicates that use of the annotated program element is discouraged, but at the time the program element was annotated, there was no specific intent to remove it.
+____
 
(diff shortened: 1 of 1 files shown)


rules/S5738/java/rule.adoc
```diff
@@ -1,19 +1,20 @@
...
 
 ____
-If true, it means that this API element is earmarked for removal in a future release.
+This annotation type has a boolean-valued element `forRemoval`. A value of `true` indicates intent to remove the annotated program element in a future version. A value of `false` indicates that use of the annotated program element is discouraged, but at the time the program element was annotated, there was no specific intent to remove it.
+____
 
-If false, the API element is deprecated, but there is currently no intention to remove it in a future release.
+While it is generally recommended for developers to steer clear of using deprecated classes, interfaces, and their deprecated members, those already marked for removal pose a higher risk of causing issues in your code in the near future. Therefore, any use of such deprecated code should be avoided or eliminated.
```
renaud-tognelli-sonarsource, line 9: Small (optional) suggestion, why not use a list to present the issue of keeping deprecated code?
 
```suggestion
* While it is generally recommended for developers to steer clear of using deprecated classes, interfaces, and their deprecated members, those already marked for removal will surely block you from upgrading your dependency.
* Deprecated code should be avoided or eliminated as soon as possible to avoid accumulation and allow a smooth upgrade of dependencies.
* The deprecated code is usually no longer maintained, can contain some bugs or vulnerabilities and usually indicates that there is a better way to perform. Removing it can even lead to significant improvement of your software.
```

rules/S5738/java/rule.adoc
```diff
@@ -1,19 +1,20 @@
...
 
-____
+== How to fix it
 
-While usually deprecated classes, interfaces, and their deprecated members should be avoided rather than used, inherited or extended, those already marked for removal are much more sensitive to causing trouble in your code soon. Consequently, any usage of such deprecated code should be avoided or removed.
+Usage of deprecated classes, interfaces, and their methods explicitly marked for removal in future versions is discouraged. A developer should either migrate to alternative methods or refactor the code to avoid deprecated ones.
```
renaud-tognelli-sonarsource, line 13: I think the "in the future" makes the sentence hard to read and does not add so much value, WDYT?
```suggestion
Usage of deprecated classes, interfaces, and their methods explicitly marked for removal is discouraged. A developer should either migrate to alternative methods or refactor the code to avoid the deprecated ones.
```

renaud-tognelli-sonarsource (approved):
LGTM, made a few suggestions

rules/S5738/java/rule.adoc
petertrr: Thanks for the suggestion! Personally, I find a list not completely suitable here, so I will rephrase your suggestions just as sentences.

PR #3278, opened 2023-10-13
Modify S6660: Fix typo in code snippet

guillaume-dequenne-sonarsource:


guillaume-dequenne-sonarsource:
> LGTM! For the impact do we need to add it manually here as well? Maybe it will be done during a LaYC spring :)

Good catch on the missing impact. I think impacts have already been migrated for all rules (it's not within the scope of the LaYC sprint), I suspect this one was specified around the same time with the old format, so it wasn't migrated.
I'll update the PR to fix this.

joke1196 (approved):
LGTM!

PR #3279, opened 2023-10-13
Modify S2208: Migrate To LayC - wildcard imports

renaud-tognelli-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S2208/javascript/rule.adoc
+++ b/rules/S2208/javascript/rule.adoc
@@ -2,11 +2,20 @@
 
 On the principle that clearer code is better code, you should explicitly ``++import++`` the things you want to use in a module. Using ``++import *++`` imports everything in the module, and runs the risk of confusing maintainers. Similarly, ``++export * from "module";++`` imports and then re-exports everything in the module, and runs the risk of confusing not just maintainers but also users of the module.
 
-=== Noncompliant code example
+=== Code examples
 
-[source,javascript]
+==== Noncompliant code example
+
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
 import * as Imported from "aModule";  // Noncompliant
+----
+
+==== Compliant solution
+
+[source,javascript,diff-id=1,diff-type=compliant]
+----
+import {aType, aFunction} from "aModule";
 ----
 
 ifdef::env-github,rspecator-view[]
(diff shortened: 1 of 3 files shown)


rules/S2208/javascript/rule.adoc
```diff
@@ -2,13 +2,22 @@
...
+[source,javascript,diff-id=1,diff-type=noncompliant]
 ----
 import * as Imported from "aModule";  // Noncompliant
 ----
 
+==== Compliant solution
+
+[source,javascript,diff-id=1,diff-type=compliant]
+----
+import {aType, aFunction} from "aModule";
```
renaud-tognelli-sonarsource, line 18: I am not sure of the relevance of my proposition, as importing with an "as" makes it clear from which package the import is coming and avoids any name collision.
What can be the right compliant solution to propose here?
Is the rule even relevant for JavaScript?

yassin-kammoun-sonarsource: It's true that importing with the `as` keyword solves the problem of confusing maintainers. However, I think the rule is still relevant for JavaScript if we just discourage users from importing everything instead of being explicit about which symbols are actually used. Therefore, I would say that the compliant solution you propose is fine here.

yassin-kammoun-sonarsource (approved):
Looks good!

rules/S2208/python/rule.adoc
```diff
@@ -6,15 +6,35 @@ Importing every public name from a module using a wildcard (``++from mymodule im
...
+    from matplotlib.pyplot import *  # Ok
+----
+
+Just keep in mind that wildcard imports might still create issues in these cases. It's always better to import only what you need.
+
+== How to fix it
 
 There are two ways to avoid a wildcard import:
 
 * Replace it with ``++import mymodule++`` and access module members as ``++mymodule.myfunction++``. If the module name is too long, alias it to a shorter name. Example: ``++import pandas as pd++``
```
nicolas-gauthier-sonarsource, line 32: The ''++++'' can be removed as there is no special symbol
```suggestion
* Replace it with `import mymodule` and access module members as `mymodule.myfunction`. If the module name is too long, alias it to a shorter name. Example: `import pandas as pd`
```

rules/S2208/python/rule.adoc
nicolas-gauthier-sonarsource, line 1: The lines 79 and 84 can be removed as double empty lines are not interpreted and just replaced by one empty line

nicolas-gauthier-sonarsource (approved):
LGTM!
You did a great job adapting the rule to the LaYC format and adding language-specific elements.
I just have minor formatting comments, most of them not even on your modifications ^^

PR #3280, opened 2023-10-13
Modify rule S106: Migrate to LayC - Standard outputs should not be used directly to log anything

ADarko22:


rules/S106/description.adoc
```diff
@@ -1,8 +1,10 @@
-When logging a message there are several important requirements which must be fulfilled:
+In software development, logs serve as a record of events within an application, providing crucial insights for debugging.
+When logging, it's essential to ensure that the logs are:
```
marco-antognini-sonarsource, line 2: We should avoid contractions
```suggestion
When logging, it is essential to ensure that the logs are:
```

rules/S106/description.adoc
```diff
@@ -1,8 +1,10 @@
...
-* The format of all logged message must be uniform to allow the user to easily read the log
-* Logged data must actually be recorded
-* Sensitive data must only be logged securely
+* easily accessible
+* uniformly formatted for readability
+* properly recorded
+* securely logged when dealing with sensitive data
 
-If a program directly writes to the standard outputs, there is absolutely no way to comply with those requirements. That's why defining and using a dedicated logger is highly recommended.
+If a program directly writes to the standard outputs (i.e. {language_std_outputs}) those requirements are not met.
```
marco-antognini-sonarsource, line 9: Grammarly suggests swapping the two components of the sentence. I think it helps.

Another small nit: the list of std outputs for C++ is not exaustive. It should be "e.g.," (with a comma) instead of "i.e.". I suppose it's also not exaustive for other languages.

```suggestion
Those requirements are not met if a program directly writes to the standard outputs (e.g., {language_std_outputs}).
```

rules/S106/description.adoc
```diff
@@ -1,8 +1,10 @@
...
+That's why defining and using a dedicated logger is highly recommended.
```
marco-antognini-sonarsource, line 10:
```suggestion
That is why defining and using a dedicated logger is highly recommended.
```

marco-antognini-sonarsource (approved):
The CFamily description looks good, nice update.

rules/S106/csharp/rule.adoc
```diff
@@ -1,26 +1,53 @@
...
 include::../description.adoc[]
 
-=== Noncompliant code example
+=== Exceptions
+
+The rule doesn't raise an issue for:
 
-[source,csharp]
+* Console Applications
+* Calls in methods decorated with ``++[Conditional ("DEBUG")]++``
```
zsolt-kolbay-sonarsource, line 12:
```suggestion
* Calls in methods decorated with `[Conditional ("DEBUG")]`
```

The same for other short code samples.

rules/S106/csharp/rule.adoc
```diff
@@ -1,26 +1,53 @@
...
+* Calls included in DEBUG preprocessor branches (``++#if DEBUG++``)
+
+=== Code examples
+
+The following noncompliant code:
```
zsolt-kolbay-sonarsource, line 17:
```suggestion
==== Noncompliant code example
```

In every _rule.adoc_ file.

rules/S106/csharp/rule.adoc
```diff
@@ -1,26 +1,53 @@
...
+    {
+        // ...
+        Console.WriteLine("My Message"); // Noncompliant
+        // ...
+    }
 }
 ----
 
-=== Exceptions
+Could be replaced by:
```
zsolt-kolbay-sonarsource, line 32:
```suggestion
==== Compliant solution
```

rules/S106/csharp/rule.adoc
```diff
@@ -1,26 +1,53 @@
...
+    {
+        // ...
+        _logger.LogInformation("My Message");
+        // ...
+    }
+}
+----
 
 
 == Resources
```
zsolt-kolbay-sonarsource, line 53: Collect the common resources (like OWASP references), move them to a common file (resources.adoc) and then include them in the language-specific rule.adoc.

zsolt-kolbay-sonarsource (approved):
LGTM. Nice use of templates! Left a few organizational/formatting comments.

rules/S106/csharp/rule.adoc
marco-antognini-sonarsource: I would recommend the opposite actually.

In the doc, we have:

> This first tab could use a blog post style with a free format explaining what the rule is detecting and why.

I feel the sentences instead of titles give a better reading "blog-post-like" flow, especially with small examples like these.

ADarko22: I tend to agree with Marco

rules/S106/csharp/rule.adoc
ADarko22: There are not many common resources, it doesn't seem to be worth the effort

PR #3281, opened 2023-10-13
Modify S1110: Migrate To LayC - redundant parentheses

marco-antognini-sonarsource:
Provide consistent examples for CFamily, CSharp and Java.

Make JS, PHP, Apex, Go, Kotlin, and Scala consistent.

Python has its own syntax so inline relevant bits.

Other languages are not updated: their description is considered good enough and it would require a significant investment to not mess up the syntax in their examples.

## Review

- [x] text style and tone


rules/S1110/python/rule.adoc
```diff
@@ -1,6 +1,17 @@
...
 
-include::../description.adoc[]
+Parentheses can disambiguate the order of operations in complex expressions and make the code easier to understand.
+
+[source,python]
+----
+a = (b * c) + (d * e) # Compliant: the intent is clear.
+----
+
+Redundant parentheses are parenthesis that
```
zsolt-kolbay-sonarsource, line 10: Why put them in separate lines? (it's not going to be an issue when it's rendered, but I'm curious if there's any reason)

zsolt-kolbay-sonarsource (approved):
LGTM

rules/S1110/python/rule.adoc
marco-antognini-sonarsource: Initially, I had longer sentences and was experimenting with bullet points. I forgot to put the whole sentence back on the same line. No big conspiracy there :D

PR #3290, opened 2023-10-16
Fix a couple of small things in several CFamily rules

marco-antognini-sonarsource:
⚠️  To be rebased and merged, not squashed.

## Review

- [x] text style and tone

### DIFF
--- a/rules/S112/cfamily/rule.adoc
+++ b/rules/S112/cfamily/rule.adoc
@@ -4,7 +4,7 @@ This rule raises an issue when a generic exception (such as ``++std::exception++
 
 Throwing generic exceptions such as ``++std::exception++``, ``++std::logic_error++`` and ``++std::runtime_error++`` will have a negative impact on any code trying to catch these exceptions.
 
-From a consumer perspective, it is generally a best practice to only catch exceptions you intend to handle. Other exceptions should ideally not be caught and let to propagate up the stack trace so that they can be dealt with appropriately. When a generic exception is thrown, it forces consumers to catch exceptions they do not intend to handle, which they then have to re-throw.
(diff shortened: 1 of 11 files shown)


rules/S112/cfamily/rule.adoc
```diff
@@ -40,14 +40,14 @@ void test(File file) {
...
-When throwing an exception, it is therefore recommended to throw the most specific exception possible so that it can be handled intentionally by consumers.
+Therefore, throwing the most specific exception possible is recommended so consumers can handle it intentionally.
 
 == How to fix it
 
 To fix this issue, make sure to throw specific exceptions that are relevant to the context in which they arise. It is recommended to either:
 
-* Throw a subtype of `++std::exception++`` when one matches. For instance ``++std::invalid_argument++`` could be raised when an unexpected argument is provided to a function.
-* Define a custom exception type that derives from `++std::exception++`` or one of its subclasses.
+* Throw a subtype of `std::exception` when one matches. For instance, `std::invalid_argument` could be raised when an unexpected argument is provided to a function.
```
frederic-tingaud-sonarsource, line 49: I would prefer not to remove the passthrough tags when dealing with characters like `::`. We might get lucky that it doesn't turn into a description list here, but a seemingly small change could break it.
```suggestion
* Throw a subtype of `++std::exception++`` when one matches. For instance ``++std::invalid_argument++`` could be raised when an unexpected argument is provided to a function.
```

I notice we don't have the ``++ everywhere already. I would still recommend a simple rule like "anything else than `[A-Za-z0-9]` should have the passthrough tag, rather than requiring people to know the full extent of asciidoctor tags.

Actually, that could be a cool check to add to the validation...

frederic-tingaud-sonarsource (changes requested):
I agree with everything except simplifying the inlined code.

rules/S112/cfamily/rule.adoc
marco-antognini-sonarsource: > Actually, that could be a cool check to add to the validation...

🤔 I think I agree! But we should think about how to gradually roll out this one...

Anyway, I've fixed them. I'll squash the commits for S112 once the PR is approved.

frederic-tingaud-sonarsource (approved):

PR #3292, opened 2023-10-16
Clean rule at root

frederic-tingaud-sonarsource:
In some cases, the `rule.adoc` at root of a rule is never included anywhere and thus is dead code.
It's a maintenance cost by itself, but also it misses opportunities to inline code that seems used by two documents when in fact only one document is actually rendered. And this missed opportunity, in turn, stops us from applying the correct language tag on the code samples.


zsolt-kolbay-sonarsource (approved):
LGTM. Nice cleanup!

loris-s-sonarsource (changes requested):
I reviewed a dozens first PRs and then took random samples further along the list. I just see the issue with deprecated rules, else I like the clean-up 👍

PR #3293, opened 2023-10-16
Update C++ rules

amelie-renard-sonarsource:


rules/S1185/cfamily/rule.adoc
```diff
@@ -87,7 +87,7 @@ public:
 
 class Derived : public Base {
 public:
-  void f() final { // final
+  void f() final { // Compliant: final
```
marco-antognini-sonarsource, line 90: In the description, we don't explain what "final" means. This is probably okay as it would diverge the attention to some orthogonal point. Here, I would suggest not writing "final" twice on the same line though. Maybe something like this?

```suggestion
  void f() final { // Compliant: cannot be overridden by derived classes
```

rules/S2190/cfamily/rule.adoc
```diff
@@ -1,4 +1,36 @@
-include::../rule.adoc[]
+== Why is this an issue?
+
+Recursion happens when control enters a loop that has no exit. It can occur when a method invokes itself, when two methods invoke each other, or when `goto` statements are used to move between two code segments. Recursion can be a useful tool, but unless the method includes a provision to break out the recursion and `return`, the recursion will continue until the stack overflows and the program crashes.
+
+=== Noncompliant code example
+
+[source,cpp]
+----
+int pow(int num, int exponent) {  // Noncompliant: no condition under which pow isn't re-called
```
marco-antognini-sonarsource, line 9: I'm not fond of the double negative. An idea: feel free to improve it.
```suggestion
int pow(int num, int exponent) {  // Noncompliant: pow is unconditionally re-called
```

rules/S2479/cfamily/rule.adoc
```diff
@@ -1,24 +1,28 @@
...
 [source,cpp]
 ----
-const char* foo = "A\tB";  // Compliant, use escaped value
+const char* foo = "A\tB";  // Compliant: use escaped value
 ----
 
 === Exceptions
 
-Raw string literals has no escape character mechanism.
+Raw string literals have no escape character mechanism.
```
marco-antognini-sonarsource, line 21: The current wording isn't explicit about the consequence of this fact. WDYT of this?
```suggestion
This rule does not raise issues in raw string literals since they have no escape character mechanism.
```

marco-antognini-sonarsource (approved):
LGTM. A couple of small nits but good to go otherwise.

rules/S2190/cfamily/rule.adoc
amelie-renard-sonarsource: Good idea, I have just replaced "unconditionally" with a simpler "always". Thanks

rules/S2479/cfamily/rule.adoc
amelie-renard-sonarsource: You're right, thanks!

PR #3301, opened 2023-10-16
Update ts-node to fix the build

necto:


necto:
@frederic-tingaud-sonarsource sorry for bombarding you with unsolicited review requests. This one seems to fix the build I broke with the previous PR.
I tried to upgrade other stuff (i.e., react from 17 to 18, but that breaks too many things. perhaps for another time).

frederic-tingaud-sonarsource (approved):

PR #3306, opened 2023-10-17
Modify S1488: Migrate to LayC

yassin-kammoun-sonarsource:
## Review

- [ ] text style and tone


rules/S1488/description.adoc
```diff
@@ -1,3 +1 @@
-Declaring a variable only to immediately return or throw it is a bad practice.
-
-Some developers argue that the practice improves code readability, because it enables them to explicitly name what is being returned. However, this variable is an internal implementation detail that is not exposed to the callers of the method. The method name should be sufficient for callers to know exactly what will be returned.
+Declaring a variable only to immediately return or throw it is considered a bad practice because it adds unnecessary complexity to the code. This practice can make the code harder to read and understand, as it introduces an extra step that doesn't add any value. Instead of declaring a variable and then immediately returning or throwing it, it's generally better to return or throw the value directly. This makes the code cleaner, simpler, and easier to understand.
```
petertrr, line 1:
```suggestion
Declaring a variable only to immediately return or throw it is considered a bad practice because it adds unnecessary complexity to the code. This practice can make the code harder to read and understand, as it introduces an extra step that doesn't add any value. Instead of declaring a variable and then immediately returning or throwing it, it is generally better to return or throw the value directly. This makes the code cleaner, simpler, and easier to understand.
```
I believe we shouldn't use contractions in rules

rules/S1488/how-to-fix.adoc
```diff
@@ -0,0 +1 @@
+To fix the issue, return or throw the value directly.
```
petertrr, line 1:
```suggestion
Declaring a variable only to immediately return or throw it is considered a bad practice because it adds unnecessary complexity to the code. To fix the issue, return or throw the value directly. 
```
As the `How to fix it` sections should be sufficient even without `Why this is an issue`, I suggest to duplicate some ideas here as well

petertrr (approved):
LGTM, please see my suggestions

rules/S1488/how-to-fix.adoc
yassin-kammoun-sonarsource: 👍

rules/S1488/description.adoc
yassin-kammoun-sonarsource: 👍

PR #3308, opened 2023-10-17
Modify S1118: migrate to LayC

leonardo-pilastri-sonarsource:


rules/S1118/csharp/rule.adoc
```diff
@@ -55,7 +56,7 @@ or
 
 [source,csharp,diff-id=2,diff-type=compliant]
 ----
-public class StringUtils
+public class StringUtils // Compliant: the constructor is not public
```
zsolt-kolbay-sonarsource, line 59: Why is the constructor `protected` instead of `private`? The `protected` keyword is used when inheritance is involved, and I don't see any reason why a class with only static methods will be part of any class hierarchy.

rules/S1118/csharp/rule.adoc
```diff
@@ -1,10 +1,11 @@
+:lang: C#
```
zsolt-kolbay-sonarsource, line 1: Where is the `lang` template parameter used?

rules/S1118/csharp/rule.adoc
```diff
@@ -1,10 +1,11 @@
...
 == Why is this an issue?
 
-Utility classes, which are collections of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static[`static`] members, are not meant to be instantiated.
+include::../description.adoc[]
 
-C# adds an implicit https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/public[`public`] https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constructors[`constructor`] to every class which does not explicitly define at least one constructor. Hence, at least one https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected[`protected`] constructor should be defined if you wish to inherit this utility class. Or the `static` keyword should be added to the class declaration to prevent inheriting it.
+== Hot to fix it
 
-== How to fix it
+include::../howtofix.adoc[]
```
zsolt-kolbay-sonarsource, line 8: C# and Java behaves somewhat differently regarding static classes:
- In Java defining an inner class as static means it will not be bound to any instance of the containing class (it cannot access its instance variables), but it can still be instantiated and works like a regular class. So whether a class in Java is static or not doesn't make a difference for this rule.
- In C# a class marked with the static keyword cannot be instantiated and can only have static members. This is relevant to the C# implementation of the rule and should be added to the `How to fix it` section. (there's already a code sample in the `Compliant solution` section, so you only need to mention it without any additional code).

zsolt-kolbay-sonarsource (changes requested):
Almost there. Left a few remarks to clean up.

rules/S1118/csharp/rule.adoc
leonardo-pilastri-sonarsource: In the `howtofix.adoc` file, which is shared, I mention the specific language adds implicitly a public constructor to classes, and I wanted to mention the language name there. Let me know if it makes sense to you

rules/S1118/howtofix.adoc
```diff
@@ -0,0 +1,4 @@
+== How to fix it
+
+To prevent the class from being instantiated, you should define at least one non-public constructor
```
zsolt-kolbay-sonarsource, line 3:
```suggestion
To prevent the class from being instantiated, you should define a non-public constructor. This will prevent the compiler from implicitly generating a public parameterless constructor.
```

rules/S1118/csharp/rule.adoc
zsolt-kolbay-sonarsource: It sounds good to me. Personally I wouldn't use it, I would just rewrite the _How to fix it_ [like this](https://github.com/SonarSource/rspec/pull/3308#discussion_r1362199040) just to make the adoc files a little shorter. 
But I don't have a strong opinion on it. I'll leave it up to you.

rules/S1118/csharp/rule.adoc
```diff
@@ -1,10 +1,10 @@
 == Why is this an issue?
 
-Utility classes, which are collections of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static[`static`] members, are not meant to be instantiated.
+include::../description.adoc[]
 
-C# adds an implicit https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/public[`public`] https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constructors[`constructor`] to every class which does not explicitly define at least one constructor. Hence, at least one https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected[`protected`] constructor should be defined if you wish to inherit this utility class. Or the `static` keyword should be added to the class declaration to prevent inheriting it.
+include::../howtofix.adoc[]
 
-== How to fix it
+Or the `static` keyword should be added to the class declaration to prevent inheriting it.
```
zsolt-kolbay-sonarsource, line 7:
```suggestion
Alternatively adding the `static` keyword as class modifier will also prevent it from being instantiated.
```

Yes, a static class is sealed by definition, but the focus isn't inheritance here, it's instantiation.

zsolt-kolbay-sonarsource (approved):
LGTM

PR #3311, opened 2023-10-17
Validate asciidoc ifdef/endif

marco-antognini-sonarsource:
This check is slow. Therefore, it is only done on modified rules.

CI on the first commit is red to show the expected outputs when an invalid use of ifdef/endif is introduced. It ran on all rules since the `ci` directory was modified.

Fix kotlin:S6511


ci/asciidoc_validation/validate_environment.py
```diff
@@ -0,0 +1,87 @@
...
+        self.is_env_open = False
+        self.has_env = False
+
+    def process(self):
+        content = self.file.read_text(encoding="utf-8")
+        for line_index, line in enumerate(content.splitlines(keepends=False)):
+            if line.startswith("ifdef::"):
+                self._process_open(line_index + 1, line)
+            if line.startswith("endif::"):
+                self._process_close(line_index + 1, line)
```
frederic-tingaud-sonarsource, line 39: I don't see a check that the env is closed at the end. Am I missing something?

ci/validate_asciidoc.sh
```diff
@@ -85,6 +86,17 @@ do
     fi
     rm -f stuck
 
+    # Validate modified files' ifdef/endif commands.
+    find "${dir}" -name '*.adoc' \
+      -exec python3 "./ci/asciidoc_validation/validate_environment.py" '{}' ';' \
+      >validate_env_commands 2>&1
```
frederic-tingaud-sonarsource, line 92: This might be quite slow because Python is expensive to launch. I would recommend to output find in a file and then feed this file to the Python script. That way we only launch the Python script once even for a tooling change.

frederic-tingaud-sonarsource (changes requested):

ci/asciidoc_validation/validate_environment.py
marco-antognini-sonarsource: 🤦 thanks for pointing this out!

ci/validate_asciidoc.sh
marco-antognini-sonarsource: Indeed, from 12 minutes down to 8. Nicely spotted.

frederic-tingaud-sonarsource (approved):

PR #3312, opened 2023-10-17
Modify S1075: migrate to LayC

leonardo-pilastri-sonarsource:
## Review

- [x] text style and tone


rules/S1075/php/rule.adoc
nicolas-gauthier-sonarsource, line 1: It could be nice to have an example for PHP and migrate the rule to the new format, but I understand it's not a flagship language, and it's hard to create an example for this rule. It has only [126 issues](https://peach.sonarsource.com/issues?resolved=false&rules=php%3AS1075) in Peach, and most of them are on test code, so they are probably irrelevant.

rules/S1075/resources.adoc
```diff
@@ -0,0 +1 @@
+* https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[CERT, MSC03-J.] - Never hard code sensitive information
```
nicolas-gauthier-sonarsource, line 1: The link seems not to follow the [Guide on link formatting](https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc)
```suggestion
* Carnegie Mellon University - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[CERT, MSC03-J. Never hard code sensitive information]
```
The site is currently under maintenance; let's hope it's temporary as I could not find the content elsewhere.

rules/S1075/swift/rule.adoc
nicolas-gauthier-sonarsource, line 1: Like Go Swift has a nice example (and is implemented!), it can be migrated easily:
```
== How to fix it

=== Code examples

[source,swift,diff-id=1,diff-type=noncompliant]
----
public class Foo {
    public func listUsers() -> [User] {
        var users:[User]
        let location = "/home/mylogin/Dev/users.txt"     // Non-Compliant
        let fileContent = NSString(contentsOfFile: location, encoding: NSUTF8StringEncoding, error: nil)
        users = parse(fileContent!)
        return users
    }
}
----

==== Compliant solution

[source,swift,diff-id=1,diff-type=compliant]
----
public class Foo {
    // Configuration is a class that returns customizable properties: it can be mocked to be injected during tests. 
    private var config:Configuration
    public init(myConfig:Configuration) {
        config = myConfig
    }
    public func listUsers() -> [User] {
        var users:[User]
        // Find here the way to get the correct folder, in this case using the Configuration object
        let location = config.getProperty("myApplication.listingFile")
        // and use this parameter instead of the hard coded path
        let fileContent = NSString(contentsOfFile: location, encoding: NSUTF8StringEncoding, error: nil)
        users = parse(fileContent!)
        return users
    }
}
----
```

rules/S1075/vbnet/rule.adoc
nicolas-gauthier-sonarsource, line 1: Like PHP, It could be nice to have an example and migrate the rule to the new format, but I understand it's not a flagship language, and it's hard to create an example for this rule. It has only [9 issues](https://peach.sonarsource.com/coding_rules?q=S1075&open=vbnet%3AS1075) in Peach.

nicolas-gauthier-sonarsource (commented):
Nice! Good update on the rule description!
I have some minor formatting comments and quick-win proposals I let you decide on.

rules/S1075/rule.adoc
nicolas-gauthier-sonarsource, line 1: WDYT of removing this file as it's just referencing `description.adoc`?

nicolas-gauthier-sonarsource (approved):
LGTM!
Just a last comment to remove `rule.adoc` at the root

rules/S1075/resources.adoc
```diff
@@ -0,0 +1,3 @@
+== Resources
+
+* Carnegie Mellon University - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[CERT, MSC03-J. Never hard code sensitive information]
```
marco-antognini-sonarsource, line 3: There are two things:

The format of the link is incorrect. See https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc

This is a Java-specific rule. Linking to it from other languages may result in confusion. We decided to remove it for CFamily. https://github.com/SonarSource/rspec/pull/3296 The whole rule is not implemented, so the fix was easy for us.

I don't know what other languages prefer.

PR #3323, opened 2023-10-18
Modify S1075: Fix resources link format

leonardo-pilastri-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1075/resources.adoc
+++ b/rules/S1075/resources.adoc
@@ -1,3 +1,3 @@
 == Resources
 
-* Carnegie Mellon University - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[CERT, MSC03-J. Never hard code sensitive information]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[Never hard code sensitive information]
--- a/rules/S1075/resources.adoc
+++ b/rules/S1075/resources.adoc
@@ -1,3 +1,0 @@
-== Resources
-
-* CERT - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[Never hard code sensitive information]
--- a/rules/S1075/csharp/rule.adoc
+++ b/rules/S1075/csharp/rule.adoc
@@ -44,8 +44,6 @@ public class Foo {
 }
 ----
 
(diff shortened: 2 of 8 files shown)


leonardo-pilastri-sonarsource:
The website that the URL points to is still down, so I cannot evaluate at this time which languages would actually benefit from including the resource in their rule page.

rules/S1075/resources.adoc
```diff
@@ -1,3 +1,3 @@
 == Resources
 
-* Carnegie Mellon University - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[CERT, MSC03-J. Never hard code sensitive information]
+* CERT - https://wiki.sei.cmu.edu/confluence/x/OjdGBQ[Never hard code sensitive information]
```
nicolas-gauthier-sonarsource, line 3: Hi! I'm so sorry for the incorrect link I gave you!
I found a new working link for Java: 
https://wiki.sei.cmu.edu/confluence/display/java/MSC03-J.+Never+hard+code+sensitive+information
It seems Java-specific to me, and it shows a security concern we don't have in the rule description as we focus on testing difficulties.
The metadata of the rule has a reference to this MSC03-J:
```
"securityStandards": {
    "CERT": [
      "MSC03-J."
    ]
  }
```
IMO, you could remove all references to that or add some information in the description to justify the resource, like: `In addition, hard-coded URI can be considered sensitive information. It should not be stored in the code`.
WDYT?

nicolas-gauthier-sonarsource (commented):
Nice! I found a working link, but it created more questions; I let you give me your opinion.

rules/S1075/resources.adoc
leonardo-pilastri-sonarsource: Indeed, now that I can see the content of the page it looks like it's not really related. Also in the Automated Detection table at the bottom of the page, the row for SonarQube does not mention this rule, but only S1313 and S2068.
So it makes sense to remove this resource and instead elaborate a bit on why URI can be considered sensitive info
👍

nicolas-gauthier-sonarsource (approved):
LGTM! The CI error on links validation doesn't seem related to this PR

PR #3328, opened 2023-10-18
Modify rule S4211: Change text to education framework format (APPSEC-1206)

sebastien-andrivet-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4211/csharp/rule.adoc
+++ b/rules/S4211/csharp/rule.adoc
@@ -1,25 +1,56 @@
+Transparency attributes in .NET Framework are a set of security features designed to protect your code from unauthorized access and to protect security-critical operations. However, they can become a vulnerability if not correctly implemented.
+
 == Why is this an issue?
 
-Transparency attributes, ``++SecurityCriticalAttribute++`` and ``++SecuritySafeCriticalAttribute++`` are used to identify code that performs security-critical operations. The second one indicates that it is safe to call this code from transparent, while the first one does not.  Since the transparency attributes of code elements with larger scope take precedence over transparency attributes of code elements that are contained in the first element a class, for instance, with a ``++SecurityCriticalAttribute++`` can not contain a method with a ``++SecuritySafeCriticalAttribute++``.
-
-
-This rule raises an issue when a member is marked with a ``++System.Security++`` security attribute that has a different transparency than the security attribute of a container of the member.
-
-
-=== Noncompliant code example
+Transparency attributes can be declared at several levels. If two different attributes are declared at two different levels, the result can be surprising and not what you may expect. For example, you can declare that a class is SecuritySafeCritical and that a method of this class is SecurityCritical:
 
 [source,csharp]
+----
+    [SecuritySafeCritical]
+    public class Foo
+    {
+        [SecurityCritical]
+        public void Bar() // Is the method SecurityCritical or SecuritySafeCritical?
+        {
+        }
+    }
+----
+
+You may think the method ``Bar`` is protected by the ``SecurityCritical`` attribute and can't be called from tranparent code. But in fact, the attribute that prevails is the ``SecuritySafeCritical`` attribute at the level of the class and the ``Bar`` method can be called from transparent code.
+
+
+=== What is the potential impact?
+
+Conflicting transparency attributes in .NET Framework can lead to serious security issues. 
+
+==== Elevation of Privileges
+
+An attacker could potentially exploit these conflicting transparency attributes to perform actions with higher privileges than intended. For instance, if a member is marked as SecurityCritical but the class is declared as SecuritySafeCritical, it might be called by regular code to perform security-sensitive operations that it shouldn't be able to.
+
+==== Data Exposure
+
+If a member with conflicting attributes is involved in handling sensitive data, an attacker could exploit the vulnerability to gain unauthorized access to this data. This could lead to breaches of confidentiality and potential data loss.
+
+
+== How to fix it
+
+To fix this vulnerability in the .NET Framework, you need to ensure that all members of a class or structure have consistent transparency attributes.
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 using System;
 using System.Security;
 
 namespace MyLibrary
 {
-
-    [SecurityCritical]
+    [SecuritySafeCritical]
     public class Foo
     {
-        [SecuritySafeCritical] // Noncompliant
+        [SecurityCritical] // Noncompliant
         public void Bar()
         {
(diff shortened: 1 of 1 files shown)


rules/S4211/csharp/rule.adoc
```diff
@@ -1,25 +1,56 @@
...
 
-This rule raises an issue when a member is marked with a ``++System.Security++`` security attribute that has a different transparency than the security attribute of a container of the member.
+You may think the method ``Bar`` is protected by the ``SecurityCritical`` attribute and can't be called from tranparent code. But in fact, the attribute that prevails is the ``SecuritySafeCritical`` attribute at the level of the class and the ``Bar`` method can be called from transparent code.
 
 
-=== Noncompliant code example
+=== What is the potential impact?
 
-[source,csharp]
+Conflicting transparency attributes in .NET Framework can lead to serious security issues.
```
loris-s-sonarsource, line 24: Here, that suggestion adds the "Below.." sentence, which is normally our norm while writing. I also provide the small introduction to avoid repeating it in the section below:

```suggestion
If code that is expected to be ``++[SecurityCritical]++`` is actually made ``++[SecuritySafeCritical]++`` by the .Net runtime environment, an attacker can use it and perform unexpected malicious operations with it.

Below are some real-world scenarios that illustrate some impacts of an attacker exploiting the vulnerability.
```

rules/S4211/csharp/rule.adoc
```diff
@@ -1,25 +1,56 @@
...
+
+==== Elevation of Privileges
+
+An attacker could potentially exploit these conflicting transparency attributes to perform actions with higher privileges than intended. For instance, if a member is marked as SecurityCritical but the class is declared as SecuritySafeCritical, it might be called by regular code to perform security-sensitive operations that it shouldn't be able to.
+
+==== Data Exposure
+
+If a member with conflicting attributes is involved in handling sensitive data, an attacker could exploit the vulnerability to gain unauthorized access to this data. This could lead to breaches of confidentiality and potential data loss.
+
+
```
loris-s-sonarsource, line 34: Since the most sensitive scenario in general is " a member marked as SecurityCritical but considered SecuritySafeCritical" because of its class, I suggest removing it, since it should be in the introduction.

Here I suggest borrowing some text from existing rules, to avoid unnecessarily creating new text and to maintain consistency across rules. For example, those mentioning "privilege escalation" can provide some text that help ideating, such as the deserialization rule (S5135), or the reflection injection rule (S6173)

Here is what it could look like:

```suggestion
==== Application-specific attacks

In this scenario, the attackers succeed in calling a legitimate function expecting a
trusted caller. Depending on the application business logic,  attackers can modify
the data structure or content to escalate privileges or perform unwanted actions. +
In the context of an e-commerce application, this could be changing the number
of products or prices.

==== Full application compromise

In the worst-case scenario, the attackers succeed in calling functions
that suppose the caller is trusted, and trigger code execution.

Depending on the attacker, code execution can be used with different
intentions:

* Download the internal server's data, most likely to sell it.
* Modify data, install malware, for instance, malware that mines cryptocurrencies.
* Stop services or exhaust resources, for instance, with fork bombs.

This threat is particularly insidious if the attacked organization does not
maintain a Disaster Recovery Plan (DRP).

==== Root privilege escalation and pivot

In this scenario, the attacker can do everything described in the previous
section. The difference is that the attacker additionally manages to elevate
his privileges as an administrator and attack other servers.

Here, the impact depends on how much the target company focuses on its Defense
In Depth. For example, the entire infrastructure can be compromised through a
combination of unsafe deserialization and misconfiguration:

* Docker or Kubernetes clusters
* cloud services
* network firewalls and routing
* OS access control
```

rules/S4211/csharp/rule.adoc
```diff
@@ -1,25 +1,56 @@
...
+== How to fix it
+
+To fix this vulnerability in the .NET Framework, you need to ensure that all members of a class or structure have consistent transparency attributes.
```
loris-s-sonarsource, line 37:
```suggestion
```

This section should always start by the code, or if the code has some specifics in its logic, a small code rationale explaining what it does. But the fix should be after the code

rules/S4211/csharp/rule.adoc
```diff
@@ -51,8 +81,14 @@ namespace MyLibrary
```
loris-s-sonarsource, line 81: The fix belongs here:

```suggestion

=== How does this work?

==== Never set class-level annotations

A class should never have class-level annotations if some functions have different permission levels. Instead, make sure every function has its own correct annotation. 

If no function needs a particularly distinct security annotation in a class, just set a class-level ``++[SecurityCritical]++``.

```

loris-s-sonarsource (changes requested):
I have impactful comments, and as always I am available to discuss everything

rules/S4211/csharp/rule.adoc
```diff
@@ -1,25 +1,56 @@
+Transparency attributes in .NET Framework are a set of security features designed to protect your code from unauthorized access and to protect security-critical operations. However, they can become a vulnerability if not correctly implemented.
```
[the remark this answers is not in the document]
sebastien-andrivet-sonarsource: The last sentence is only about a specific case and I do not think that "tends" is appropriate here. I will suggest something else.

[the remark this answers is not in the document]
loris-s-sonarsource: Yes, something to note is that in general when I do a code suggestion, I do not necessarily mean "replace the text by the one I'm suggesting", It's just as suggestion of how I see it. Do as you please with it 👍

rules/S4211/csharp/rule.adoc
```diff
@@ -1,25 +1,56 @@
...
+
 == Why is this an issue?
```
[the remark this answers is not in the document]
sebastien-andrivet-sonarsource: The rule is more for the developer writing the (original) code the code will not do what he or she expects. It is highly counterintuitive. It is not really for future maintainers (especially since all that is mentioned in this rule is no longer supported by .NET). I will rephrase.

I agree for the code, I will remove it.

rules/S4211/csharp/rule.adoc
sebastien-andrivet-sonarsource: "Here I suggest borrowing some text from existing rules, to avoid unnecessarily creating new text". This is not much compatible with our usage of OpenAI prompts. We get different texts each time.

For me, the examples you suggest do not apply to this specific rule.

What is detected by this rule is old, out of date and no longer used by .NET. The whole original mechanism is weak by nature. So the impacts are in fact very theoretical.

loris-s-sonarsource (approved):
LGTM

PR #3331, opened 2023-10-18
Modify rule S1079: Change text to education framework format (APPSEC-1209)

jamie-anderson-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1079/cfamily/rule.adoc
+++ b/rules/S1079/cfamily/rule.adoc
@@ -1,42 +1,58 @@
 == Why is this an issue?
 
-The ``++%s++`` placeholder is used to read a word into a string.
+The ``++%s++`` format specifier is used to read a string into a buffer.
 
-By default, there is no restriction on the length of that word, and the developer is required to pass a sufficiently large buffer for storing it.
+By default, there is no restriction on the length of that string. The ``scanf`` family of functions will continue to read characters into the buffer until they find a whitespace character.
 
-No matter how large the buffer is, there will always be a longer word.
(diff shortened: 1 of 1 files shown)


rules/S1079/cfamily/rule.adoc
```diff
@@ -1,42 +1,58 @@
...
-No matter how large the buffer is, there will always be a longer word.
+If the input contains a sufficiently long string with no whitespace characters, it can cause memory beyond the end of the buffer to be overwritten. This is a buffer overflow vulnerability.
 
-Therefore, programs relying on ``++%s++`` are vulnerable to buffer overflows.
+=== What is the potential impact?
 
+If the buffer is stored on the stack, the attacker can overwrite pieces of memory that control program flow. The attacker can exploit a technique called return-oriented programming to execute arbitrary code.
 
-A field width specifier can be used together with the ``++%s++`` placeholder to limit the number of bytes which will by written to the buffer.
+If the buffer is stored on the heap, the attacker can corrupt other memory on the heap. This allows the attacker to control the application's state, cause unexpected behavior, or cause the program to crash.
```
gaetan-ferry-sonarsource, line 13: Writing into a buffer on the heap is more or less the same as writing on the stack from a consequences point of view. In both cases, you can reach code execution, only the exploitation technique changes.
I would merge the two paragraph and go less into the details. To me, it is enough to say that we can smash sensitive parts of the memory that control the program execution to execute arbitrary code.
Likewise, I would not mention ROP which is a technical term that most people won't get, and anyway not the only way to execute shellcode.

gaetan-ferry-sonarsource (changes requested):
LGTM with a suggestion to make the impact section more highlevel.

gaetan-ferry-sonarsource (approved):
LGTM

PR #3332, opened 2023-10-19
Modify S1144: Migrate to LaYC

zsolt-kolbay-sonarsource:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1144/apex/rule.adoc
+++ b/rules/S1144/apex/rule.adoc
@@ -1,4 +1,33 @@
-include::../rule.adoc[]
+include::../why.adoc[]
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,apex]
+----
+public class Server {
+
+    // Although marked as noncompliant (unused), including the function for completeness
+
+
+    public void start() {  // Compliant, publicly available
+        log('start');
+    }
+
+    private void clear() { // Noncompliant, not used anywhere
+        
+    }
+
+    private void log(String msg) {  // Compliant, called from 'start()'
+        System.debug(msg);
+    }
+
+    private static void unusedStaticFunction() { // Noncompliant
+
+    }
+}
+----
 
 ifdef::env-github,rspecator-view[]
 
--- a/rules/S1144/apex/rule.adoc
+++ b/rules/S1144/apex/rule.adoc
@@ -1,3 +1,6 @@
+:operationName: method
+:visibility: private
+
 include::../why.adoc[]
 
 === Code examples
--- a/rules/S1144/apex/rule.adoc
+++ b/rules/S1144/apex/rule.adoc
@@ -7,10 +7,6 @@ include::../why.adoc[]
 [source,apex]
 ----
 public class Server {
-
-    // Although marked as noncompliant (unused), including the function for completeness
-
-
     public void start() {  // Compliant, publicly available
         log('start');
     }
@@ -21,10 +17,6 @@ public class Server {
 
     private void log(String msg) {  // Compliant, called from 'start()'
         System.debug(msg);
-    }
-
-    private static void unusedStaticFunction() { // Noncompliant
-
     }
 }
 ----
--- a/rules/S1144/cfamily/rule.adoc
+++ b/rules/S1144/cfamily/rule.adoc
@@ -4,10 +4,6 @@ A function that is never called is dead code, and should be removed. Cleaning ou
 
 This rule detects functions that are never called from inside a translation unit, and cannot be called from the outside.
 
-=== Code examples
-
-==== Noncompliant code example
-
 [source,cpp]
 ----
 static void unusedStaticFunction() { // Noncompliant
@@ -28,6 +24,8 @@ private:
 ----
 
 == Resources
+
+=== External coding guidelines
 
 * MISRA {cpp}:2008, 0-1-10 - Every defined function shall be called at least once.
 
--- a/rules/S1144/cfamily/rule.adoc
+++ b/rules/S1144/cfamily/rule.adoc
@@ -2,10 +2,11 @@
 
 A function that is never called is dead code, and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
 
-
 This rule detects functions that are never called from inside a translation unit, and cannot be called from the outside.
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 [source,cpp]
 ----
--- a/rules/S1144/cfamily/rule.adoc
+++ b/rules/S1144/cfamily/rule.adoc
@@ -1,8 +1,7 @@
-== Why is this an issue?
+:operationName: function
+:visibility: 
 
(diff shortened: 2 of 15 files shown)


rules/S1144/cfamily/rule.adoc
```diff
@@ -2,10 +2,11 @@
 
 A function that is never called is dead code, and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
 
-
 This rule detects functions that are never called from inside a translation unit, and cannot be called from the outside.
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
```
frederic-tingaud-sonarsource, line 9: This is not really a Noncompliant code example. You can just remove the two titles and have the text and the code together.

rules/S1144/csharp/rule.adoc
```diff
@@ -12,13 +12,11 @@ This rule doesn't raise issues on:
```
frederic-tingaud-sonarsource, line 12: "===Exceptions" makes more sense after the code examples, when the general case has been clarified

rules/S1144/java/rule.adoc
```diff
@@ -1,52 +1,49 @@
-== Why is this an issue?
+include::../why.adoc[]
 
-``++private++`` methods that are never executed are dead code: unnecessary, inoperative code that should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
+Note that this rule does not take reflection into account, which means that issues will be raised on `private` methods that are only accessed using the reflection API.
```
frederic-tingaud-sonarsource, line 3: I would put the note and the Exceptions after the code snippets

rules/S1144/cfamily/rule.adoc
```diff
@@ -2,10 +2,11 @@
```
frederic-tingaud-sonarsource, line 2: Perhaps this could be an include to ../why.adoc

rules/S1144/python/rule.adoc
```diff
@@ -1,16 +1,15 @@
...
 
-* methods with a name starting with a single underscore (ex: ``++_mymethod++``) should be seen as non-public and might change without prior notice. They should not be used by third-party libraries or software. It is ok to use those methods inside the library defining them but it should be done with caution.
-* "class-private" methods have a name which starts with at least two underscores and ends with at most one underscore. These methods' names will be automatically mangled to avoid collision with subclasses' methods. For example ``++__mymethod++`` will be renamed as ``++_classname__mymethod++``, where ``++classname++`` is the method's class name without its leading underscore(s). These methods shouldn't be used outside of their enclosing class.
+* methods with a name starting with a single underscore (ex: `_mymethod`) should be seen as non-public and might change without prior notice. They should not be used by third-party libraries or software. It is ok to use those methods inside the library defining them but it should be done with caution.
```
frederic-tingaud-sonarsource, line 5: It's dangerous to remove the ``++ with names that start with `_`. They might be converted to italic.

rules/S1144/cfamily/rule.adoc
```diff
@@ -2,10 +2,11 @@
...
-
 This rule detects functions that are never called from inside a translation unit, and cannot be called from the outside.
 
-=== Noncompliant code example
+=== Code examples
+
+==== Noncompliant code example
 
 [source,cpp]
 ----
```
frederic-tingaud-sonarsource, line 12: Below, in `== Resources`, could you also add `=== External coding guidelines` before the MISRA reference?

frederic-tingaud-sonarsource (changes requested):
The content is great, just a few notes on the form.

rules/S1144/apex/rule.adoc
```diff
@@ -1,4 +1,33 @@
...
+        
+    }
+
+    private void log(String msg) {  // Compliant, called from 'start()'
+        System.debug(msg);
+    }
+
+    private static void unusedStaticFunction() { // Noncompliant
+
+    }
```
guillaume-dequenne-sonarsource, line 28: I think non compliant examples are best kept minimal, I'm not sure it makes sense to show in the example that issues are also raised for static methods.

guillaume-dequenne-sonarsource (changes requested):
Left a few suggestions.

rules/S1144/why.adoc
```diff
@@ -0,0 +1,5 @@
+== Why is this an issue?
+
+A function that is never called is dead code, and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
+
+This rule detects functions that are never called from inside a translation unit, and cannot be called from the outside.
```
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: In case you want to split languages that do or don't refer to "Translation unit", it is the correct term for CFamily.

rules/S1144/cfamily/rule.adoc
zsolt-kolbay-sonarsource: Which lines exactly?

frederic-tingaud-sonarsource: Line 3 to 5 are identical to the content of why.adoc so far.
But Guillaume suggested to change `why.adoc`, so that might change. If you end up changing `why.adoc`, you can ignore my comment and keep the current text as is.

rules/S1144/why.adoc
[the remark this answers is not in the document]
zsolt-kolbay-sonarsource: I used the term _translation unit_ for every language. PHP has a separate paragraph that explains how enums are treated. So I think it's fine to leave _translation unit_ in the general description.

frederic-tingaud-sonarsource (approved):

rules/S1144/why.adoc
```diff
@@ -0,0 +1,7 @@
+include::description.adoc[]
+
+== Why is this an issue?
+
+A {operationName} that is never called is dead code, and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.
+
+This rule detects {operationName}s that are never referenced from inside a translation unit, and cannot be referenced from the outside.
```
guillaume-dequenne-sonarsource, line 7: _Nitpick:_ I still feel that while this sentence is great to describe the rule in a generic way, it may be a bit confusing when reading the Python version of the rule in isolation, as the next sentence directly contradicts this one (`Python has no real private methods. Every method is accessible. `: so this rule doesn't actually raise on methods that can't be referenced from the outside).

For Python I'd either remove the sentence or reuse the one in the `description.adoc` which is more accurate for the context.

guillaume-dequenne-sonarsource (approved):
Good improvement overall!
Left a final comment for the Python description. I don't think it's too big a problem in the grand scheme of things, but would be nice to be addressed.

rules/S1144/python/rule.adoc
```diff
@@ -1,7 +1,8 @@
-:operationName: method
-:visibility: "class-private"
+This rule raises an issue when a "class-private" method is never called inside the class in where it was declared.
```
guillaume-dequenne-sonarsource, line 1: Grammar: I think the "in" is redundant:
 `inside the class in where it was declared. ` -> `inside the class where it was declared.`

PR #3339, opened 2023-10-19
Modify rule S2819: Change text to education framework format (APPSEC-1208)

sebastien-andrivet-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2819/javascript/how-to-fix-it/origin.adoc
+++ b/rules/S2819/javascript/how-to-fix-it/origin.adoc
@@ -1,6 +1,8 @@
 
-
 === Code examples
+
+When sending a message, avoid using `*` for the target origin (it means no preference). Instead define it explicitly so the message will only be dispatched to this URI.
+When receiving the message, verify the orgin to be sure that it is sent by an authorized sender.
 
 ==== Noncompliant code example
 
@@ -43,9 +45,3 @@ window.addEventListener("message", function(event) {
   console.log(event.data)
 }); 
 ----
-
-=== How does this work?
-
-When sending a message, the target origin is explicitly defined so the message will only be dispatched to this URI.
-
-When receiving the message, the orgin is verified to be sure that it is sent by an authorized sender.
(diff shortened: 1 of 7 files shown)


rules/S2819/javascript/how-to-fix-it/origin.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+  console.log(event.data)
+}); 
+----
+
+=== How does this work?
+
+When sending a message, the target origin is explicitly defined so the message will only be dispatched to this URI.
+
+When receiving the message, the orgin is verified to be sure that it is sent by an authorized sender.
```
gaetan-ferry-sonarsource, line 51: In general, I like putting the general explanation of how to fix at the top of the section just below "How to fix it". I then keep the "How does this work" section to describe the actual fixes of the code example especially if they use a specific API function or trick.

I don't think it is a strong requirement tho.

gaetan-ferry-sonarsource (approved):
LGTM in general. I added a small reorganisation suggestion. It does not need a re-review tho. Except if you think it is necessary.

rules/S2819/javascript/how-to-fix-it/origin.adoc
sebastien-andrivet-sonarsource: I agree and move the text with some changes. Can you have a look? I moved the text just after "Code examples", not before because the validation was complaining.

gaetan-ferry-sonarsource: That's weird, I do this everytime without any issue. For example here: https://github.com/SonarSource/rspec/blob/b5bb91c133a64a35dd8fd0283b105823bd4e00dd/rules/S3333/php/rule.adoc?plain=1#L22

It does not make a lot of sense to put the generic description under the code samples section.

sebastien-andrivet-sonarsource: You're right. it was just a complaint from Visual Studio Code.

PR #3341, opened 2023-10-19
Modify rule S1081: Change text to education framework format (APPSEC-1210)

sebastien-andrivet-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
+++ b/rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
@@ -1,0 +1,41 @@
+== How to fix it
+
+Fixing vulnerabilities related to insecure functions involves replacing these functions with safer alternatives and implementing proper checks.
+
+=== Code examples
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+char str[20];
+gets(str); // Noncompliant; `str` buffer size is not checked and it is vulnerable to overflows
+----
+
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
+char buffer[20];
+strcpy(buffer, input); // Noncompliant; `input` length is not checked and it may overflow `buffer`
+----
+
+
+==== Compliant solution
+
+[source,cpp,diff-id=1,diff-type=compliant]
+----
+char str[20];
+gets_s(str, sizeof(str));
+----
+
+[source,cpp,diff-id=2,diff-type=compliant]
+----
+char buffer[20];
+if(strlen(input) < sizeof(buffer))
+  strcpy(buffer, input);
+----
+
+=== How does this work?
+
+C and C++ have a number of functions that are considered insecure because they do not perform bounds checking. Functions like ``gets`` are known to be risky. Replace these with safer alternatives that include bounds checking. For example, instead of ``gets``, use ``gets_s``.
+
+The code can also checks if the size of input data is less than or equal to the size of a buffer (including the terminating null byte when dealing with strings).
--- a/rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
+++ b/rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
@@ -1,6 +1,6 @@
 == How to fix it
 
-Fixing vulnerabilities related to insecure functions involves replacing these functions with safer alternatives and implementing proper checks.
+Insecure functions should be replaced with safer alternatives that limit how much data can be written to the buffer.
 
 === Code examples
 
@@ -8,6 +8,8 @@ Fixing vulnerabilities related to insecure functions involves replacing these fu
 
 [source,cpp,diff-id=1,diff-type=noncompliant]
 ----
+#include <stdio.h>
+
 char str[20];
 gets(str); // Noncompliant; `str` buffer size is not checked and it is vulnerable to overflows
 ----
@@ -23,6 +25,9 @@ strcpy(buffer, input); // Noncompliant; `input` length is not checked and it may
 
 [source,cpp,diff-id=1,diff-type=compliant]
 ----
+#define __STDC_WANT_LIB_EXT1__
+#include <stdio.h>
+
 char str[20];
 gets_s(str, sizeof(str));
 ----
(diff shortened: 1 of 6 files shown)


rules/S1081/rationale.adoc
```diff
@@ -0,0 +1,3 @@
+An attacker typically provides input that exceeds the expected size. This could be through a text field in a user interface, a file that the program reads, or data sent over a network. The insecure function then processes this oversized input, leading to unexpected behavior.
```
jamie-anderson-sonarsource, line 1: The start of this is good, but the second sentence lacks clarity. It just says that the input is processed and bad things happen, without saying why those bad things happen.

The following paragraph talks about buffers, so I would like to see something in this paragraph that explains where the buffers fit in.

I would also like to see some mention of this being a buffer overflow vulnerability, because buffer overflows are something that people have heard about and are known to be bad.

```suggestion
An attacker typically provides input that exceeds the expected size. This could be through a text field in a user interface, a file that the program reads, or data sent over a network. The insecure function processes this input and places the result into a provided buffer.

If the input is larger than the buffer can handle, the insecure function will overwrite the memory following the buffer. This situation is known as a buffer overflow vulnerability.
```

rules/S1081/impact.adoc
```diff
@@ -0,0 +1,17 @@
+=== What is the potential impact?
+
+Using insecure functions can lead to serious security issues.
+
+==== Information disclosure
+
+In some scenarios, insecure functions can lead to information disclosure. For instance, if an attacker can cause a buffer overflow, they might be able to read memory that they're not supposed to have access to. This could potentially allow them to access sensitive information, such as passwords or encryption keys.
```
jamie-anderson-sonarsource, line 7: I probably wouldn't list this as a notable impact. The vulnerable functions that this rule targets result in a buffer overrun, i.e. overwriting data. There are ways to have this turn into an information disclosure vulnerability, but code execution or DoS are far more likely targets for an attacker.

rules/S1081/impact.adoc
```diff
@@ -0,0 +1,17 @@
+=== What is the potential impact?
+
+Using insecure functions can lead to serious security issues.
```
jamie-anderson-sonarsource, line 3: I don't know if this sentence is needed. Going straight into the subsections is probably sufficient.

rules/S1081/cfamily/rule.adoc
```diff
@@ -1,40 +1,18 @@
...
 
-[source,cpp]
-----
-gets_s(str, sizeof(str)); // Prevent overflows by enforcing a maximum size for `str` buffer
-----
+// How to fix it section
 
+include::how-to-fix-it/buffer-overflow.adoc[]
 
 == Resources
```
jamie-anderson-sonarsource, line 13: It would be worthwhile to add a link here to the Buffer Overflow page on Wikipedia.

```suggestion
== Resources

=== Articles & blog posts

* Wikipedia - https://en.wikipedia.org/wiki/Buffer_overflow[Buffer overflow]
```

jamie-anderson-sonarsource (changes requested):

rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
```diff
@@ -0,0 +1,41 @@
...
+----
+char str[20];
+gets_s(str, sizeof(str));
+----
+
+[source,cpp,diff-id=2,diff-type=compliant]
+----
+char buffer[20];
+if(strlen(input) < sizeof(buffer))
+  strcpy(buffer, input);
```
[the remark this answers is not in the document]
sebastien-andrivet-sonarsource: I have not used strncpy in the example because of its weird behavior with the terminal null character. strcpy_s is probably the proper way and it has been there since C11. I will modify the example to use strcpy_s

rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
```diff
@@ -0,0 +1,41 @@
...
+strcpy(buffer, input); // Noncompliant; `input` length is not checked and it may overflow `buffer`
+----
+
+
+==== Compliant solution
+
+[source,cpp,diff-id=1,diff-type=compliant]
+----
+char str[20];
+gets_s(str, sizeof(str));
```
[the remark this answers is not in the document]
sebastien-andrivet-sonarsource: You're right. Forget about that.

rules/S1081/impact.adoc
sebastien-andrivet-sonarsource: A side-effect of using our OpenAI template.

rules/S1081/impact.adoc
sebastien-andrivet-sonarsource: Yes, you're right

rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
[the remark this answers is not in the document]
jamie-anderson-sonarsource: The odd behaviour with the null termination character is why the final byte in the buffer is always set to 0. If the string was already null-terminated, it does nothing. If the string wasn't null-terminated, it will be now.

However, I agree that `strcpy_s` is the better option in the long term (for the compilers that support it).

rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
```diff
@@ -34,9 +36,11 @@ gets_s(str, sizeof(str));
 
 [source,cpp,diff-id=2,diff-type=compliant]
 ----
+#define __STDC_WANT_LIB_EXT1__
+#include <string.h>
+
 char buffer[20];
-if(strlen(input) < sizeof(buffer))
-  strcpy(buffer, input);
+strcpy_s(buffer, sizeof buffer, input);
```
jamie-anderson-sonarsource, line 43: There's a syntax error here with the `sizeof` operator.
```suggestion
strcpy_s(buffer, sizeof(buffer), input);
```

jamie-anderson-sonarsource (changes requested):

rules/S1081/cfamily/how-to-fix-it/buffer-overflow.adoc
sebastien-andrivet-sonarsource: This is not a syntax error. This is the proper way to use `sizeof` in this case:
* `sizeof expression` without parenthesis when using with an expression
* `sizeof(type)` with parenthesis when using with a type

jamie-anderson-sonarsource: Interesting! I wrote C/C++ for several years and don't recall ever seeing this version of `sizeof` being used. I probably did see third-party code that used it as a unary operator, but I've only ever written code that uses it as a function.

It is valid to use the operator as a function all the time. When used with an expression, the compiler simply places braces around the expression rather than treating them as part of the operator.

It's worth noting that Linux Torvalds [prefers the use of braces](https://lkml.org/lkml/2012/7/11/103):

> But "sizeof()" really *is* a function. It acts exactly like a function of it's argument. There is no reason to not treat it that way. Sure, the C standard *allows* you to not have parenthesis around an expression argument, but you should treat that as the parsing oddity it is, nothing more. There is zero reason not to have the parenthesis there.

Regardless, the code is valid as-is so I'm not going to advocate for it to be changed.

jamie-anderson-sonarsource (approved):
LGTM

PR #3346, opened 2023-10-20
Modify rule S5782: Change text to education framework format (APPSEC-1211)

sebastien-andrivet-sonarsource:
## Review

- [ ] text style and tone


rules/S5782/cfamily/how-to-fix-it/buffer-overflow.adoc
```diff
@@ -0,0 +1,25 @@
+== How to fix it
+
+=== Code examples
+
+In the following code example, ``memchr`` is used to find the first occurrence of the ``@`` character in a buffer, ``array``. The second argument does not match the size of this buffer, so it may find an occurrence of the character after the end of the buffer and read data it is not supposed to. If the returned ``pos`` pointer is later used to write characters, this could unexpectedly overwrite data.
```
gaetan-ferry-sonarsource, line 5: I would have chosen a simpler example that does not require that many explanations. Let's say a dumb buffer overflow or a buffer overread where the consequence is direct.

rules/S5782/common/resources/standards.adoc
```diff
@@ -0,0 +1,8 @@
+=== Standards
+
+* OWASP - https://owasp.org/Top10/A06_2021-Vulnerable_and_Outdated_Components/[Top 10 2021 - A06 - Vulnerable and Outdated Components]
+* OWASP - https://owasp.org/www-project-top-ten/2017/A9_2017-Using_Components_with_Known_Vulnerabilities[OWASP Top 10 2017 Category A9 - Using Components with Known Vulnerabilities]
```
gaetan-ferry-sonarsource, line 4: I don't quite get this. The components are not intrinsically vulnerable. This would have worked if the issue was about `sprintf` for example. In the case of length controlled function, they are in fact the right thing to use. TBH I don't think any OWASP entry fits here except maybe simply "Injection" and I'm not too sure.

gaetan-ferry-sonarsource (changes requested):
LGTM in general with a few improvements suggestions.

rules/S5782/cfamily/how-to-fix-it/buffer-overflow.adoc
sebastien-andrivet-sonarsource: I took the existing example. It is true that `memchr` is not so often used. I can find another example (maybe `read`) but I have to check that we effectively trigger for it.

rules/S5782/impact.adoc
```diff
@@ -0,0 +1,22 @@
...
+==== Denial of service
+
+If an attacker can trigger a buffer overflow by providing oversized input, it can cause the program to crash. If the attacker repeats this process, it can continually disrupt the service, denying access to other users. This can be particularly damaging for services that require high availability, such as online platforms or databases.
+
+In some cases, the input might cause the program to enter an infinite loop or consume excessive memory, slowing down the system or even causing it to become unresponsive. This type of attack is known as a resource exhaustion DoS attack.
+
+
+=== Exceptions
+
+Functions related to sockets using the type ``++socklen_t++`` are not checked. This is because these functions are using a C-style polymorphic pattern using ``++union++``. It relies on a mismatch between allocated memory and sizes of structures and it creates false positives.
```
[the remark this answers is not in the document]
sebastien-andrivet-sonarsource: This was in the existing RSPEC. From what I understand, since socklen_t is a union, it is difficult to properly analyze and follow which fields of the union are actually used. I can perhaps make it clearer, after some experimentation. Or perhaps simply remove the last sentence.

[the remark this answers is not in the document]
gaetan-ferry-sonarsource: I would suggest that you either:
* rewrite this paragraph to make it easier to understand
* remove everything
My opinion is that this section only makes sense if we provide information about why we exclude some functions. I don't think it adds value to just say "We exclude this because of false positive".

rules/S5782/cfamily/how-to-fix-it/buffer-overflow.adoc
gaetan-ferry-sonarsource: If you think changing this example is not worth it, fine by me.

PR #3353, opened 2023-10-23
Modify S6385: Change text to LayC (APPSEC-1234)

gaetan-ferry-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6385/description.adoc
+++ b/rules/S6385/description.adoc
@@ -1,3 +1,0 @@
-Defining a custom role for a Subscription or a Management group that allows all actions will give them the same capabilities as the built-in Owner role.
-
-This rule raises an issue when a custom role has an assignable scope set to a Subscription or a Management Group and allows all actions (``++*++``).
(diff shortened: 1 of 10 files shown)


rules/S6385/terraform/rule.adoc
```diff
@@ -8,7 +8,7 @@ include::../description.adoc[]
```
loris-s-sonarsource, line 8: The code rationale you made is not included there

rules/S6385/description.adoc
```diff
@@ -1,5 +1,3 @@
 Defining a custom role for a Subscription or a Management group that allows all actions will give them the same capabilities as the built-in Owner role.
 
 This rule raises an issue when a custom role has an assignable scope set to a Subscription or a Management Group and allows all actions (``++*++``).
```
loris-s-sonarsource, line 3: I would suggest adding a small impact section that says what happens if the Owner role is compromised. We have some stuff we wrote for cloud secrets talking about cloud intrusion impact that might fit like malware distribution or cryptomining

loris-s-sonarsource (approved):
I generated the output of the rule on my local SQ and I do not feel like the content of this rule is very "educational". It's very short and I feel like we could talk about privilege escalation or other impacts for example. It could make users understand that solving this issue matters and teach newbies about what are the possible risks of it. I'm not sure everyone understands the extent of "gives the same rights as owners" for example.

But apparently, it was already flagged as LayC, so I am a little confused but maybe it was deemed good enough in the past, so: approved!

loris-s-sonarsource (approved):
LGTM !

PR #3364, opened 2023-10-25
Modify S1860: migrate to LayC

leonardo-pilastri-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S1860/java/rule.adoc
+++ b/rules/S1860/java/rule.adoc
@@ -1,6 +1,39 @@
 == Why is this an issue?
 
-Instances of value-based classes, which are pooled and potentially reused, should not be used for synchronization. If they are, it can cause unrelated threads to deadlock with unhelpful stacktraces.
+In Java, value-based classes are those for which instances are final and immutable, like `String`, `Integer` and so on.
+When a variable of one of these types is instantiated, the JVM caches its value, and the variable is just a reference to that value.
+For example, multiple `String` variables with the same value "Hello world!" will actually refer to the same cached string literal in memory.
+
+The `synchronized` keyword tells the JVM to only allow the execution of the code contained in the following block to one `Thread` at a time.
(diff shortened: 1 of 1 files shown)


rules/S1860/java/rule.adoc
```diff
@@ -1,6 +1,39 @@
 == Why is this an issue?
 
-Instances of value-based classes, which are pooled and potentially reused, should not be used for synchronization. If they are, it can cause unrelated threads to deadlock with unhelpful stacktraces.
+In Java, value-based classes are those for which instances are final and immutable, like `String`, `Integer` and so on.
+When a variable of one of these types is instantiated, the JVM caches its value, and the variable is just a reference to that value.
+For example, multiple `String` variables with the same value "Hello world!" will actually refer to the same cached string literal in memory.
+
+The `synchronized` keyword tells the JVM to only allow the execution of the code contained in the following block to one `Thread` at a time.
+This mechanism relies on the "identity" of the object that is being synchronized between threads,
+to prevent that if object X is locked, it will still be possible to lock another object Y.
```
erwan-serandour-sonarsource, line 9: I think we always try to keep sentences on the same line, to have cleaner diff when changing the description

rules/S1860/java/rule.adoc
```diff
@@ -1,6 +1,39 @@
 == Why is this an issue?
 
-Instances of value-based classes, which are pooled and potentially reused, should not be used for synchronization. If they are, it can cause unrelated threads to deadlock with unhelpful stacktraces.
+In Java, value-based classes are those for which instances are final and immutable, like `String`, `Integer` and so on.
```
erwan-serandour-sonarsource, line 3: I would try something like that : In Java, the identity of value-based classes relies on their value and not their reference. Examples include, `String`, `Integer`, and other boxed values. 
Even if I am not sure it is 100% true

erwan-serandour-sonarsource (approved):

PR #3368, opened 2023-10-25
Modify S6469(docker): Migrate to LayC

loris-s-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6469/docker/rule.adoc
+++ b/rules/S6469/docker/rule.adoc
@@ -6,11 +6,12 @@ secrets to commands that are executed during the build without baking them
 into the image. Additionally, it can be used to access SSH agents during the
 build.
 
-The `mode` option allows you to change the permissions of the secrets or
-agents. By default, it is set to `0400` for `secret` and `0600` for `ssh`.
-
+The `mode` option is an octal value that allows you to change the permissions
+of the secrets or agents. 
 The mode option is an octal value that specifies the permissions for a
-particular file or directory. +
+particular file or directory.
+
+By default, it is set to `0400` for `secret` and `0600` for `ssh`: +
 The first digit 0 stands for special permissions (like setuid, setgid and
 sticky bit) and in this case means that no special permissions are set. +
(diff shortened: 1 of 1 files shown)


rules/S6469/docker/rule.adoc
```diff
@@ -2,44 +2,73 @@
...
 
-[source,docker]
+[source,docker,diff-id=1,diff-type=compliant]
 ----
 RUN --mount=type=secret,id=build_secret,uid=1000 ./installer.sh
 ----
 
+=== How does this work?
+
+Always set the `others` bit to `0`.
```
sebastien-andrivet-sonarsource, line 55: From the explanations, it is not clear if the default permissions (i.e. those when you do not specify `mode`, 'uid', ...) are safe or not.

rules/S6469/docker/rule.adoc
```diff
@@ -2,44 +2,73 @@
...
 RUN --mount=type=secret,id=build_secret,mode=0777 ./installer.sh # Noncompliant
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,docker]
+[source,docker,diff-id=1,diff-type=compliant]
 ----
 RUN --mount=type=secret,id=build_secret,uid=1000 ./installer.sh
```
sebastien-andrivet-sonarsource, line 50: For me, the compliant solution does not match the noncompliant one. We talk about permissions (i.e. `mode`) and the solution uses `uid`. It gives the impression that `uid` is also a permission with an octal value of `1000`.

sebastien-andrivet-sonarsource (changes requested):
I find the explanation not very clear and somewhat misleading (it was also the case of the original rule). In particular, the example.

rules/S6469/docker/rule.adoc
loris-s-sonarsource: True, I didn't notice that

sebastien-andrivet-sonarsource (approved):
LGTM except for a link.

PR #3384, opened 2023-10-30
Modify S6218: add toString case to Kotlin description

erwan-serandour-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6218/kotlin/rule.adoc
+++ b/rules/S6218/kotlin/rule.adoc
@@ -1,12 +1,12 @@
 == Why is this an issue?
 
-In data classes, the default behavior of the `equals()` method is to check the equality by field values. This works well for primitive fields or fields, whose type overrides `equals()`, but this behavior doesn't work as expected for array fields.
+In data classes, the default behavior of the ``++equals()++`` method is to check the equality by field values. This works well for primitive fields or fields, whose type overrides ``++equals()++``, but this behavior doesn't work as expected for array fields.
 
 
-By default, array fields are compared by their reference, so overriding `equals()` is highly recommended to ensure a deep equality check. The same applies to the `hashcode()` method.
+By default, array fields are compared by their reference, so overriding``++equals()++`` is highly recommended to ensure a deep equality check. The same applies to the ``++hashcode()++`` and ``++toString()++`` methods.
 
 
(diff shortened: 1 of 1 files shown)


rules/S6218/kotlin/rule.adoc
```diff
@@ -1,12 +1,12 @@
...
-In data classes, the default behavior of the `equals()` method is to check the equality by field values. This works well for primitive fields or fields, whose type overrides `equals()`, but this behavior doesn't work as expected for array fields.
+In data classes, the default behavior of the ``++equals()++`` method is to check the equality by field values. This works well for primitive fields or fields, whose type overrides ``++equals()++``, but this behavior doesn't work as expected for array fields.
 
 
-By default, array fields are compared by their reference, so overriding `equals()` is highly recommended to ensure a deep equality check. The same applies to the `hashcode()` method.
+By default, array fields are compared by their reference, so overriding``++equals()++`` is highly recommended to ensure a deep equality check. The same applies to the ``++hashcode()++`` and ``++toString()++`` methods.
 
 
-This rule reports an issue if a record class has an array field and is not overriding `equals()` or `hashcode()` methods.
+This rule reports an issue if a record class has an array field and is not overriding ``++equals()++``, ``++hashcode()++`` or ``++toString()++`` methods.
```
leonardo-pilastri-sonarsource, line 9: Let's rename `record` into `data` classes
```suggestion
This rule reports an issue if a data class has an array field and is not overriding ``++equals()++``, ``++hashcode()++`` or ``++toString()++`` methods.
```

leonardo-pilastri-sonarsource (changes requested):
Just a renaming but it looks good to me!
Also, I noticed that the title of the rule only mentions the equals method, while it is reporting on 3 different methods if they are missing. Do we want to change the rule's title as well? I feel like even if more verbose, it would be more clear to mention all 3 of them.

rules/S6218/kotlin/rule.adoc
erwan-serandour-sonarsource: nice catch

leonardo-pilastri-sonarsource (approved):
It's ok for me, I let you decide if we want to change the title of the rule as well, to reflect the complete behavior of it

PR #3394, opened 2023-11-01
Modify rule S6798: remove example razor file, breaking master CI

antonioaversa:
Fixes https://cirrus-ci.com/task/6685663169347584?logs=file_extension_tests#L9
Issue introduced by https://github.com/SonarSource/rspec/pull/3210, which has green CI at the time it was validated, but would have had red CI with the new validation checks in place now.


sebastien-marichal (approved):
LGTM!

PR #3395, opened 2023-11-01
Modify rule S6806: Update rule title to match guidelines

dorian-burihabwa-sonarsource:


rules/S6806/java/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Model attribute naming convention for Spring Expression Language (SpEL)",
+  "title": "Model attributes should follow the Java naming convention",
```
johann-beleites-sonarsource, line 2:
```suggestion
  "title": "Model attributes should follow the Java identifier naming convention",
```

johann-beleites-sonarsource (approved):
Small suggestion, otherwise LGTM

PR #3398, opened 2023-11-02
Modify rule S6444: Convert from hotspot to vulnerability

hendrik-buchwald-sonarsource:
## Review

- [ ] text style and tone


rules/S6444/impact.adoc
```diff
@@ -0,0 +1,19 @@
...
+==== System Slowdown
+
+When a regular expression without a timeout is given a complex input that takes
+a long time to process, it can consume a substantial amount of system resources.
+This can slow down the entire system, affecting not only the specific process
+involved but also other processes running on the same system.
+
+==== Service Unavailability
+
+In more severe cases, a ReDoS attack can cause the system to become
```
jamie-anderson-sonarsource, line 15: It would be worth adding something here to explain what a ReDoS is, e.g.
* `ReDoS (regular expression denial of service)`
* `regular expression denial of service (ReDoS)`

rules/S6444/common/resources/standards.adoc
```diff
@@ -0,0 +1,5 @@
+=== Standards
+
+* OWASP - https://owasp.org/www-project-top-ten/2017/A1_2017-Injection[Top 10 2017 - Category A1 - Injection]
```
jamie-anderson-sonarsource, line 3: This is not an injection issue and we should probably remove A1 as a vulnerability.

Injections require unexpected commands to be processed by some sort of command interpreter. A regex is not a command interpreter, just a fancy string processor.

The 2021 Top 10 list provides a mapping of CWEs to each of their categories, and the A03 Injection category doesn't contain any sign of CWE-400 or CWE-1333. If ReDoS was an injection issue, I would expect to see it listed.

jamie-anderson-sonarsource (changes requested):
Just a couple of small issues.

rules/S6444/impact.adoc
hendrik-buchwald-sonarsource: Good point, updated.

rules/S6444/common/resources/standards.adoc
hendrik-buchwald-sonarsource: It has the injection resource because not having a timeout usually becomes a problem when there is some kind of injection involved. For example, the regular expression is user-controlled or the regular expression is just bad and the value that is checked is user-controlled.

Considering that the `rationale.adoc` talks about user input I think it is fine to keep the reference personally. What do you think?

jamie-anderson-sonarsource (approved):
LGTM

rules/S6444/common/resources/standards.adoc
jamie-anderson-sonarsource: If the regex is user-controlled, there is a potential for an injection vulnerability here. The regex parser is the command interpreter and the regex is the command.

If the regex is under developer control, the compiled regex is just a parser/evaluator. If it's inefficient, it's no different than any other inefficient algorithm (e.g. a `O(n^2)` bubble sort on a really large data set).

I'm still not 100% convinced that this is injection, but if you see that there's an argument for keeping it then I'm not going to push for its removal.

Tim-Pohlmann:
Hey @hendrik-buchwald-sonarsource what is the state of this PR? Is it still needed?

hendrik-buchwald-sonarsource:
> Hey @hendrik-buchwald-sonarsource what is the state of this PR? Is it still needed?

Hello @Tim-Pohlmann ,

it is not clear. To my knowledge, we will keep the hotspots now. But I wouldn't bet on it either. I would just keep this open for a bit.

hendrik-buchwald-sonarsource:
Closed as requested by @nils-werner-sonarsource .

PR #3400, opened 2023-11-05
Modify rule S5852: Convert security hotspot to vulnerability (APPSEC-1271)

jamie-anderson-sonarsource:
## Review

- [ ] text style and tone


rules/S5852/metadata.json
```diff
@@ -33,9 +33,6 @@
       400,
       1333
     ],
-    "OWASP": [
-      "A1"
-    ],
```
jamie-anderson-sonarsource, line 38: I decided to remove A1 because this is not an injection issue. A command is not being injected into a command interpreter.

I considered making this a 2021 A4 (Insecure Design) issue, but it's an implementation issue rather than a design issue so it doesn't fit that either.

rules/S5852/common/fix/how-does-it-work.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+If the engine needs to backtrack over an atomic group, it must backtrack over the entire group. It cannot reevaluate the contents of the group to find alternative paths to match. This reduces the number of alternative paths.
+
+==== Nested quantifiers
+
+Nested quantifiers appear when a quantified term is inside a group that is also quantified. A simple example is the expression `(a+)+`.
+
+Backtracking will change the length of input matched by the inner quantified term. If both the current and next iterations of the group continue to match, it means that the same input can be matched in multiple ways. This creates alternate evaluation paths that can lead to catastrophic backtracking.
+
+Avoid nested quantifiers to prevent this problem from occurring. If they cannot be avoided, ensure that the group has a starting or ending sequence that does not have a quantifier. This reduces the number of iterations that are needed before finding that no match can be made.
```
jamie-anderson-sonarsource, line 27: This part took forever to write succinctly. It was originally about 5 paragraphs of text that's slowly been worked down.

I'm happy to take suggestions on how this can be improved.

rules/S5852/python/rule.adoc
```diff
@@ -1,52 +1,81 @@
...
 
+[source,python,diff-id=1,diff-type=compliant]
+----
+import re;
 
-* If applicable, define a maximum number of expected repetitions using the bounded quantifiers, like ``++{1,5}++`` instead of ``+`` for instance.
-* Refactor ``++nested quantifiers++`` to limit the number of way the inner group can be matched by the outer quantifier, for instance this nested quantifier situation ``++(ba+)+++`` doesn't cause performance issues, indeed, the inner group can be matched only if there exists exactly one ``++b++`` char per repetition of the group.
-* Optimize regular expressions with ``++possessive quantifiers++`` and ``++atomic grouping++`` (available since Python 3.11).
-* Use negated character classes instead of ``++.++`` to exclude separators where applicable. For example the quadratic regex ``++.*_.*++`` can be made linear by changing it to ``++[^_]*_.*++``
+pattern = re.compile("^(a+)++$");
```
jamie-anderson-sonarsource, line 44: I tried to use the same lookaround that works for JavaScript but it doesn't work on Python 3.10. The only solution I could find was to use the newer Python version and possessive grouping.

rules/S5852/common/resources/external-guidelines.adoc
```diff
@@ -0,0 +1 @@
+* Microsoft - https://docs.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions#backtracking-with-nested-optional-quantifiers[Backtracking with Nested Optional Quantifiers]
```
jamie-anderson-sonarsource, line 1: I'm not sure if this counts as an article or an external coding guideline. I treated it as a coding guideline given that it's in the .NET documentation.

jamie-anderson-sonarsource: Somehow I completely missed that there was a `Documentation` section in the resources. This has now been moved to that section.

rules/S5852/common/fix/how-does-it-work.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+==== Quantifiers with match limits
+
+The quantifiers `*`, `+`, and `{min,}` (and their lazy equivalents) have no limit on the number of characters they can match. During backtracking, every one of these characters creates an additional path that must be evaluated. 
+
+Using the `{min,max}` quantifier (or its lazy equivalent) reduces the number of alternative paths that can be created during backtracking.
+
+
+
+* Solve the problem without regular expressions
```
hendrik-buchwald-sonarsource, line 44:
```suggestion
* Solve the problem without regular expressions.
```

Since it is a full sentence and the other points also end with a dot, I would add a dot here as well.

rules/S5852/common/fix/how-does-it-work.adoc
```diff
@@ -0,0 +1,52 @@
...
+* Use an alternative non-backtracking regex implementations such as Google's https://github.com/google/re2[RE2] or https://github.com/uhop/node-re2/[node-re2].
+* Use multiple passes. This could mean pre- and/or post-processing the string manually before/after applying the regular expression to it or using multiple regular expressions. One example of this would be to replace `+str.split(/\s*,\s*/)+` with `+str.split(",")+` and then trimming the spaces from the strings as a second step.
+* It is often possible to make the regex infallible by making all the parts that could fail optional, which will prevent backtracking. Of course this means that you'll accept more strings than intended, but this can be handled by using capturing groups to check whether the optional parts were matched or not and then ignoring the match if they weren't. For example the regex `+x*y+` could be replaced with `+x*(y)?+` and then the call to `+str.match(regex)+` could be replaced with `+matched = str.match(regex)+` and  `+matched[1] !== undefined+`.
```
hendrik-buchwald-sonarsource, line 47:
```suggestion
* It is often possible to make the regex infallible by making all the parts that could fail optional, which will prevent backtracking. Of course this means that you will accept more strings than intended, but this can be handled by using capturing groups to check whether the optional parts were matched or not and then ignoring the match if they were not. For example the regex `+x*y+` could be replaced with `+x*(y)?+` and then the call to `+str.match(regex)+` could be replaced with `+matched = str.match(regex)+` and  `+matched[1] !== undefined+`.
```

rules/S5852/common/fix/how-does-it-work.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+
+* Solve the problem without regular expressions
```
hendrik-buchwald-sonarsource, line 50: The 3 next points seem to be duplicates from the stuff above.

rules/S5852/common/resources/external-guidelines.adoc
hendrik-buchwald-sonarsource, line 1:
```suggestion
* Microsoft Documentation - https://docs.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions#backtracking-with-nested-optional-quantifiers[Backtracking with Nested Optional Quantifiers]
```

I would add this part, so it matches more whats in the [link-formatting doc](https://github.com/SonarSource/rspec/blob/master/docs/link_formatting.adoc).

hendrik-buchwald-sonarsource (changes requested):
The text in general looks very nice. There are some small things to change though.

hendrik-buchwald-sonarsource:
I forgot to mention, validate_asciidoc and validate_metadata are also failing.

rules/S5852/common/fix/how-does-it-work.adoc
jamie-anderson-sonarsource: These were from the old rules file and should have been deleted. They're gone now.

hendrik-buchwald-sonarsource (approved):
It's a very nice explanation of a complex problem :+1:

rules/S5852/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
-  "title": "Using slow regular expressions is security-sensitive",
-  "type": "SECURITY_HOTSPOT",
+  "title": "Regular expressions should performance issues due to backtracking",
```
hendrik-buchwald-sonarsource, line 2: I just noticed the title, is this correct? It sounds a bit odd to me.

hendrik-buchwald-sonarsource (changes requested):
Missed something, see comment.

rules/S5852/metadata.json
jamie-anderson-sonarsource: Yeah, it looks like I got half-way through rewriting that then got distracted. I've just pushed a new title in the `X should Y` format.

hendrik-buchwald-sonarsource (approved):
LGTM!

nils-werner-sonarsource:
Not WIP anymore. The branch is outdated, and the context might be changed. Link to branch and closed PR is given by implementation ticket.

PR #3401, opened 2023-11-06
Fix Sonar warning.

frederic-tingaud-sonarsource:


LoghinVladDev (approved):
Nothing to add. Hard review indeed

PR #3407, opened 2023-11-06
Don't crash when the asciidoc logger receives a string

frederic-tingaud-sonarsource:
The asciidoc documentation states that the logger will receive objects with a given interface, but sometimes we receive simple strings that crash the documented code.


necto (approved):
SQ has some comments, otherwise LGTM
Let's see if the GH action really works now

PR #3443, opened 2023-11-16
CPP-4815 Make rule applicable to CFamily

tomasz-kaminski-sonarsource:


rules/S6620/cfamily/metadata.json
```diff
@@ -0,0 +1,17 @@
...
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [
+  ],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6620",
+  "sqKey": "S6620",
+  "scope": "All",
+  "defaultQualityProfiles": ["Sonar way"],
```
marco-antognini-sonarsource, line 15: We may not want it in SonarWay...

rules/S6620/cfamily/metadata.json
marco-antognini-sonarsource, line 1: Taxonomy elements are missing.

marco-antognini-sonarsource (approved):
Some comments but no need for another review.

rules/S6620/cfamily/metadata.json
tomasz-kaminski-sonarsource: This may impact other plugin working. So I would prefer to keep it that way.

rules/S6620/cfamily/metadata.json
tomasz-kaminski-sonarsource: Added.

PR #3452, opened 2023-11-21
Modify rule S6802: Add a code example

sebastien-marichal:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6802/csharp/rule.adoc
+++ b/rules/S6802/csharp/rule.adoc
@@ -6,7 +6,10 @@ The reason behind this is that Blazor rebuilds all lambda expressions within the
 
 == How to fix it
 
-Ensure to not use a delegate in elements rendered in loops by using a collection of objects containing the delegate as an https://learn.microsoft.com/en-us/dotnet/api/system.action[Action].
+Ensure to not use a delegate in elements rendered in loops by either:
+
+* using a collection of objects containing the delegate as an https://learn.microsoft.com/en-us/dotnet/api/system.action[Action],
+* extract the elements into a dedicated component and use an https://learn.microsoft.com/en-us/aspnet/core/blazor/components/event-handling#eventcallback[EventCallback] to call the delegate
 
 === Code examples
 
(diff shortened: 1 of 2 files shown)


rules/S6802/csharp/rule.adoc
```diff
@@ -70,12 +73,80 @@ Ensure to not use a delegate in elements rendered in loops by using a collection
...
+    private void DoAction(MouseEventArgs e, int button)
+    {
+        // Do something here
+    }
+}
+----
+
+==== Compliant solution
+
+[source,cs]
```
antonioaversa, line 99: I am not sure that `cs` is a valid language.
Moreover, we use `csharp` in all other examples and in all other rules.
In principle, however, it should be `razor`... that probably isn't supported (I don't see it in any razor-related rule).

rules/S6802/csharp/rule.adoc
```diff
@@ -70,12 +73,80 @@ Ensure to not use a delegate in elements rendered in loops by using a collection
...
+----
+@* MyButton.razor *@
+
+<button @onclick="OnClickCallback">
+    @ChildContent
+</button>
+
+@code {
+
```
antonioaversa, line 108: This empty line seems not necessary.

rules/S6802/csharp/rule.adoc
antonioaversa: I would probably merge the two source blocks (the following one, and the one below) in the `==== Compliant solution` section, even if that means that the `@* MyButton.razor *@` code would become entirely green.

And actually, I think it would be more correct that way since the specific proposed solution requires introducing this `MyButton.razor` component.

That requires giving a name to the component that defines the foreach, and where `MyButton` is being used. That should also be done in the non-compliant section. For instance `index.razor`.

antonioaversa (changes requested):
One blocking comment. The rest are primarily suggestions and optional improvements.

rules/S6802/csharp/rule.adoc
sebastien-marichal: My mistake!

antonioaversa (approved):
LGTM!

PR #3488, opened 2023-12-07
Fix missing glibc for sonar-scanner

necto:


necto:
The problem is caused by a bug in the pre-released version of Sonar scanner, so no need for a fix here.

PR #3492, opened 2023-12-12
Modify rule S6504: Clarify requirements for Docker (APPSEC-1350)

jamie-anderson-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6504/docker/rule.adoc
+++ b/rules/S6504/docker/rule.adoc
@@ -2,10 +2,10 @@ Ownership of an executable has been assigned to a user other than root. More
 often than not, resource owners have write permissions and thus can edit the
 resource.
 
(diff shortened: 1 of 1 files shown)


pierre-loup-tristant-sonarsource (changes requested):
I think the rule is still ambiguous as its title and message only refer to executable although in practice it applies to any file or directory with an “execute” flag.

I suggest changing the rule title to :
“Resources copied to a container should be immutable by non-root users”

The message should be something like
“Make sure this copied resource is immutable by non-root users.“

Also, the rule description should be adjusted to replace “executable” by “file/directory”

daniel-teuchert-sonarsource:
I am closing this PR because it is a duplicate of https://github.com/SonarSource/rspec/pull/3528

PR #3503, opened 2023-12-14
Fix compliant example

leveretka:


rules/S6624/kotlin/rule.adoc
```diff
@@ -29,16 +29,16 @@ dependencies {
...
-}
+const val mockitoVersion = "4.5.1"
 
 dependencies {
     testImplementation("org.mockito:mockito-core:$mockitoVersion")
     testImplementation("org.mockito:mockito-inline:$mockitoVersion")
 }
 ----
 
+Alternatively, you can put `const val mockitoVersion = "4.5.1"` in any `.kt` file in `buildSrc/src/main/kotlin` or use a more robust dependency management mechanism like Spring dependency management plugin or Versions Catalog.
```
johann-beleites-sonarsource, line 40: Maybe add a link to versions catalogs & spring dep mgmt plugin?

johann-beleites-sonarsource (approved):
Small suggestion, otherwise LGTM!

rules/S6624/kotlin/rule.adoc
leveretka: We have links a few lines above, but I can add them again

PR #3506, opened 2023-12-14
Modify rule S1001 Literals namespaces should be allowlisted from the rule (CPP-4873)

loic-joly-sonarsource:


LoghinVladDev (approved):
LGTM

PR #3508, opened 2023-12-14
Add documentation about benchmark results

gregory-paidis-sonarsource:
## Review

- [ ] text style and tone


docs/benchmarks.adoc
```diff
@@ -0,0 +1,46 @@
...
+== Benchmark table
+
+- Do not use abbreviations for statistical terms, not everyone is familiar with them.
+- Add links to Wikipedia to explain statistical terms at the header of each column.
+- Prefer `Mean` to `Median`.
+
+Ideally, the columns should look like the following example:
+
+|===
+| <What is being measured> | https://en.wikipedia.org/wiki/Mean[Mean] | https://en.wikipedia.org/wiki/Standard_deviation[Standard Deviation] | https://en.wikipedia.org/wiki/Memory_management[Allocated]
```
John-Clifton-SonarSource, line 20: I think the mean link should be to arithmetic mean since I think most people will use that won't they? (https://en.wikipedia.org/wiki/Arithmetic_mean)

docs/benchmarks.adoc
```diff
@@ -0,0 +1,46 @@
...
+| <This> | 5.042 ms | 0.1049 ms | 125 KB
+| <That> | 2.691 ms | 0.0334 ms | 85.94 KB
+|===
+
+== Benchmarking code
+
+The code that was used to generate the benchmarks should be included to provide transparency and allow others to reproduce and verify the results. Preferably, the code snippet should include the sample size, the number of iterations, and the framework/library used to run the benchmarks.
```
John-Clifton-SonarSource, line 27: Is it worth also mentioning why the sample chosen is representative?

John-Clifton-SonarSource (commented):
This is looking really good. I've put a couple of comments in that might be worth addressing.

docs/benchmarks.adoc
gregory-paidis-sonarsource: That's a good point, yes.
I will change it, thanks!

antonioaversa (approved):
LGTM!
As discussed offline, I would remove the `prefer mean over median` sentence, since it would be highly controversial in many scenarios.

docs/benchmarks.adoc
gregory-paidis-sonarsource: Hey John, I discussed this with a colleague and I cannot think how exactly to convey this.
For most of the rules we implemented during a .NET performance rule sprint, the idea was to just choose a big sample so that the runs are reliable, but I wouldn't be able to find the `N` to say _"at sample size > `N` the results are reliable"_.

For now, I merged it like this, but if you want to elaborate on this, I'd be happy to.

John-Clifton-SonarSource: Its no problem leaving it as it is.

PR #3517, opened 2023-12-19
Update description.adoc to allow for freeform text in 'how' tab

John-Clifton-SonarSource:
We allow for freeform text without fixed titles in the 'why' tab. We want to allow for this in the 'how' tab too.

I have also slightly changed the terminology to talk about a freeform 'story-telling' format because I believe that is a bit closer to how we intend people to use this.

Since I was in here, I have updated the list of fields to call out the mandatory field rather than the optional ones. The list of optional fields was not correct as it stood.

## Review

- [ ] text style and tone


loic-joly-sonarsource (approved):
❤️

PR #3519, opened 2023-12-20
Fix asciidoc w.r.t. C++

marco-antognini-sonarsource:
"C++" spelled as-is can result in unexpected rendering, such as

    C++20 was released before C++17

renders as

    C20 was released before C17

Make consistent use of `[source,cpp]`.

### DIFF
--- a/rules/S1079/cfamily/rule.adoc
+++ b/rules/S1079/cfamily/rule.adoc
@@ -20,7 +20,7 @@ Note that the ``++%s++`` format specifier always null-terminates the string in t
 
 ==== Noncompliant code example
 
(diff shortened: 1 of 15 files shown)


marco-antognini-sonarsource:
Since this is a common pitfall, I had a look at commits in the past three months that I fixed in this PR. My intent is not to blame but to raise awareness since we don't have CI checks for this (it would be quite hard to write them).

CC @jamie-anderson-sonarsource @sebastien-andrivet-sonarsource @gaetan-ferry-sonarsource

frederic-tingaud-sonarsource (approved):

PR #3523, opened 2023-12-21
Update java:S1068 exceptions to mention classes with Lombok annotations

ADarko22:
From community: https://community.sonarsource.com/t/sonarqube-doesnt-report-unused-fields-with-lombok/106265

### DIFF
--- a/rules/S1068/java/rule.adoc
+++ b/rules/S1068/java/rule.adoc
@@ -34,7 +34,7 @@ public class MyClass implements java.io.Serializable {
 }
 ----
 
-* Annotated fields (also classes annotated with Lombok annotations)
+* Annotated fields and classes annotated with Lombok annotations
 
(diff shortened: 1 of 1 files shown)


rules/S1068/java/rule.adoc
```diff
@@ -34,7 +34,7 @@ public class MyClass implements java.io.Serializable {
 }
 ----
 
-* Annotated fields
+* Annotated fields (also classes annotated with Lombok annotations)
```
johann-beleites-sonarsource, line 37: I think it would be fine to mention it on the same level as the fields, so without parentheses. But I think it's fin either way.
```suggestion
* Annotated fields and classes annotated with Lombok annotations
```

johann-beleites-sonarsource (approved):

PR #3524, opened 2023-12-21
Clean warnings

frederic-tingaud-sonarsource:


frontend/src/RulePage.tsx
```diff
@@ -330,14 +330,14 @@ function useDescription(metadata: PageMetadata, ruleid: string, language?: strin
     `https://github.com/SonarSource/rspec/blob/${metadata.branch}/rules/${ruleid}${language ? '/' + language : ''}`;
 
   function htmlReplacement(domNode: Element) {
-    if (domNode.name === 'a' && domNode.attribs && domNode.attribs['data-rspec-id']) {
+    if (domNode.name === 'a' && domNode?.attribs['data-rspec-id']) {
```
LoghinVladDev, line 333: Since domNode cannot be undefined (even previously, domNode.name is accessed without checking for undefined), is this `?.` needed?

`domNode?.attribs[...]` will only check whether `domNode` is undefined. 
If `attribs` is undefined (previous check), it will fail (undefined[...]) results in error.

Although `domNode.attribs` is typed as not undefined, so checking for any undefined/null here doesn't make sense to me.

On a first look, I would just change this to:
```suggestion
    if (domNode.name === 'a' && domNode.attribs['data-rspec-id']) {
```

LoghinVladDev (changes requested):
This change is a bit unclear to me, is there some exceptional case to this?

frontend/src/RulePage.tsx
frederic-tingaud-sonarsource: You're right that I didn't do what I thought I was doing. Fixed.

LoghinVladDev (approved):

PR #3531, opened 2024-01-05
Benchmarks Guide: Update statistical term links

gregory-paidis-sonarsource:
I am not sure links on table headers will work across all variants of rspec (SL, SQ, SC, rules.sonarsource...), so instead I suggest we use a glossary (which was also suggested as a solution in discussions with the documentation team).

### DIFF
--- a/docs/benchmarks.adoc
+++ b/docs/benchmarks.adoc
@@ -11,7 +11,7 @@ The format of the benchmarks section is as follows:
 == Benchmark table
 
 - Do not use abbreviations for statistical terms, not everyone is familiar with them.
-- At the `Benchmarks` section, add a subsection called `Glossary` right after the benchmark table and add links to Wikipedia to explain statistical terms used at the header of each column:
+- In the `Benchmarks` section, add a subsection called `Glossary` right after the benchmark table and add links to Wikipedia to explain statistical terms used in the header of each column:
 
 ==== Glossary
 
@@ -19,7 +19,7 @@ The format of the benchmarks section is as follows:
 * https://en.wikipedia.org/wiki/Standard_deviation[Standard Deviation]
 * https://en.wikipedia.org/wiki/Memory_management[Allocated]
 
-Ideally, the columns should look like the following example:
+Ideally, the columns should look like in the following example:
 
 |===
 | <What is being measured> | Mean | Standard Deviation | Allocated
--- a/docs/benchmarks.adoc
+++ b/docs/benchmarks.adoc
@@ -11,12 +11,18 @@ The format of the benchmarks section is as follows:
 == Benchmark table
 
 - Do not use abbreviations for statistical terms, not everyone is familiar with them.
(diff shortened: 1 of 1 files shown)


docs/benchmarks.adoc
```diff
@@ -11,12 +11,18 @@ The format of the benchmarks section is as follows:
 == Benchmark table
 
 - Do not use abbreviations for statistical terms, not everyone is familiar with them.
-- Add links to Wikipedia to explain statistical terms at the header of each column.
+- At the `Benchmarks` section, add a subsection called `Glossary` right after the benchmark table and add links to Wikipedia to explain statistical terms used at the header of each column:
+
+==== Glossary
```
antonioaversa, line 16: Shouldn't this header be at the third level?

docs/benchmarks.adoc
```diff
@@ -11,12 +11,18 @@ The format of the benchmarks section is as follows:
 == Benchmark table
 
 - Do not use abbreviations for statistical terms, not everyone is familiar with them.
-- Add links to Wikipedia to explain statistical terms at the header of each column.
+- At the `Benchmarks` section, add a subsection called `Glossary` right after the benchmark table and add links to Wikipedia to explain statistical terms used at the header of each column:
```
antonioaversa, line 14: At the `Benchmarks` section => In the `Benchmarks` section

Used at the header of each column => Used in the header of each column

docs/benchmarks.adoc
```diff
@@ -11,12 +11,18 @@ The format of the benchmarks section is as follows:
...
+
+==== Glossary
+
+* https://en.wikipedia.org/wiki/Arithmetic_mean[Mean]
+* https://en.wikipedia.org/wiki/Standard_deviation[Standard Deviation]
+* https://en.wikipedia.org/wiki/Memory_management[Allocated]
 
 Ideally, the columns should look like the following example:
```
antonioaversa, line 22: like **in** the following example:

antonioaversa (approved):
LGTM, minor non-blocking remarks.

docs/benchmarks.adoc
gregory-paidis-sonarsource: I don't think so, take a look at the first rule I fixed:
https://github.com/SonarSource/rspec/blob/955c813529d01242147db913c5eb3e56ff02399c/rules/S6602/resources-dotnet.adoc
(click the `raw` to see it more easily)

gregory-paidis-sonarsource: Oh I think I know what you mean.
If this was part of THIS document, yes.
I am trying to show here how it is supposed to look on a real RSPEC.
I don't know if I can make it clear here.

gregory-paidis-sonarsource:
> @gregory-paidis-sonarsource Links validation is failing: https://github.com/SonarSource/rspec/pull/3531/checks?check_run_id=20195697909 However, I don't think it has anything to do with this PR.

Yeah, there are some failures from other rules, I also think it is irrelevant.

PR #3539, opened 2024-01-17
Modify rule S1104: Add exception for [Serializable]

Tim-Pohlmann:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S1104/csharp/rule.adoc
+++ b/rules/S1104/csharp/rule.adoc
@@ -8,9 +8,11 @@ include::../impacts.adoc[]
 
 === Exceptions
 
-- Fields marked as `readonly` or `const` are ignored by this rule.
-- Fields inside classes or structs annotated with `StructLayoutAttribute` are ignored by this rule.
-- Fields inside classes or structs annotated with `Serializable` are ignored by this rule unless they are annotated with `NonSerialized`.
+Fields marked as `readonly` or `const` are ignored by this rule.
+
+Fields inside classes or structs annotated with `[StructLayout]` are ignored by this rule.
+
+Fields inside classes or structs annotated with `[Serializable]` are ignored by this rule unless they are annotated with `[NonSerialized]`.
 
 == How to fix it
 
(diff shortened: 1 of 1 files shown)


rules/S1104/csharp/rule.adoc
```diff
@@ -8,9 +8,9 @@ include::../impacts.adoc[]
 
 === Exceptions
 
-Fields marked as `readonly` or `const` are ignored by this rule.
-
-Fields inside classes or structs annotated with the `StructLayoutAttribute` are ignored by this rule.
+- Fields marked as `readonly` or `const` are ignored by this rule.
+- Fields inside classes or structs annotated with `StructLayoutAttribute` are ignored by this rule.
```
antonioaversa, line 12: Nitpicking: `StructLayoutAttribute` has the `Attribute` suffix, whereas `Serializable`, on the next bullet point, hasn't. Personally, I would:
- either add square brackets and omit the `Attribute` suffix: e.g. `[StructLayout]`
- or always include the `Attribute` suffix, without square brackets: `SerializableAttribute`, `NonSerializedAttribute`

rules/S1104/csharp/rule.adoc
```diff
@@ -8,9 +8,9 @@ include::../impacts.adoc[]
...
+- Fields inside classes or structs annotated with `Serializable` are ignored by this rule unless they are annotated with `NonSerialized`.
```
antonioaversa, line 13: Nitpicking: [this section](https://github.com/SonarSource/rspec/blob/master/docs/styling_guide.adoc#lists) of the styling guide suggests not to use punctuation on lists.

However, it mainly concern "list of entries", and I think you have enough freedom here to decide whether the list is a list of entries, or a list of full sentences. 

Probably the latter I'd say, given that you repeat the verb at every bullet point ("are ignored by this rule"). 

However, if you DRY and factorize the "are ignored by this rule", the styling guide rule would apply IMO.

antonioaversa (approved):
LGTM, I have left a couple of non-blocking nitpicks.

PR #3544, opened 2024-01-19
Docs review patch 3

kevin-hinz:
mostly minor improvements to words and simple sentence structure

### DIFF
--- a/rules/S3776/why.adoc
+++ b/rules/S3776/why.adoc
@@ -9,8 +9,8 @@ As a rule of thumb, high cognitive complexity is a sign that the code should be 
 Here are the core concepts:
 
 * **Cognitive complexity is incremented each time the code breaks the normal linear reading flow.** +
-This concerns, for example: Loop structures, Conditionals, Catches, Switches, Jumps to label and mixed operators in condition.
-* **Each nesting level adds a malus to the breaking call.** +
+This concerns, for example, loop structures, conditionals, catches, switches, jumps to labels, and binary logical operators.
+* **Each nesting level interferes with the breaking call.** +
 During code reading, the deeper you go through nested layers, the harder it becomes to keep the context in mind.
 * **Method calls are free** +
  A well-picked method name is a summary of multiple lines of code. 
--- a/rules/S3776/why.adoc
+++ b/rules/S3776/why.adoc
@@ -9,8 +9,8 @@ As a rule of thumb, high cognitive complexity is a sign that the code should be 
 Here are the core concepts:
 
 * **Cognitive complexity is incremented each time the code breaks the normal linear reading flow.** +
-This concerns, for example, loop structures, conditionals, catches, switches, jumps to labels, and binary logical operators.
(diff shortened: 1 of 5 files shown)


rules/S3776/why.adoc
```diff
@@ -9,8 +9,8 @@ As a rule of thumb, high cognitive complexity is a sign that the code should be
 Here are the core concepts:
 
 * **Cognitive complexity is incremented each time the code breaks the normal linear reading flow.** +
-This concerns, for example: Loop structures, Conditionals, Catches, Switches, Jumps to label and mixed operators in condition.
-* **Each nesting level adds a malus to the breaking call.** +
+This concerns, for example, loop structures, conditionals, catches, switches, jumps to labels, and binary logical operators.
```
kevin-hinz, line 12: "binary logical operators" was a recommended improvement (for that term)

antonioaversa: "Mixed operators" is a mix of different operators. In C# those can be:
- logical (`&&`, `||`, `^`, ...), but also non-logical (e.g. bitwise: `&`, `|`, shift: `>>`, `<<`, ...)
- binary, but also non-binary (cast, `await`, `&`, ...)

So, "binary logical operators" are a subset of all the operators.

In order to keep the sentence general enough, and valid for operators-rich languages like C#, I would propose to either keep "mixed operators in condition", or go for one of the following alternatives:
- "a mix of different operators in a condition"
- "a mix of operators in a condition"
- "conditions mixing multiple operators"
- "conditions containing multiple operators in them"

rules/S3776/why.adoc
```diff
@@ -9,8 +9,8 @@ As a rule of thumb, high cognitive complexity is a sign that the code should be
...
+* **Each nesting level interferes with the breaking call.** +
```
antonioaversa, line 13: I think that "add a malus to the breaking call" means "increases the cognitive complexity of the breaking call". However, I don't understand what "breaking" means in this context.

Because this change would affect 14 languages (the description of the rule is shared by all languages), I would either propose to keep it as is or go for a more generic alternative that can work with all languages, such as "Each nesting level increases the complexity".

WDYT?

antonioaversa (changes requested):

rules/S3776/why.adoc
kevin-hinz: “A malus” is the uncommon phrase we want to avoid; it could also be “a genus” of apple trees, ha!
Your suggestion LGTM!

rules/S3776/why.adoc
kevin-hinz: You’re right, I don’t think we want to limit the description to a subset. 
I issued an improvement with one of your suggestions.

antonioaversa (approved):
LGTM

necto:
@antonioaversa and @kevin-hinz, for future PRs, please, consider a more uniform and informative summary, both for the PR title and the squashed-commit summary

PR #3547, opened 2024-01-23
Modify rule S6864: Update examples and removal of LimitRange for now

navendu-barua-sonarsource:
updates in S6864, S6869 and S6870
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6864/kubernetes/rule.adoc
+++ b/rules/S6864/kubernetes/rule.adoc
@@ -55,21 +55,6 @@ spec:
       image: nginx
 ----
 
-[source,yaml,diff-type=noncompliant]
-----
-apiVersion: v1
-kind: Pod
-metadata:
-  name: example
-spec:
-  containers:
-    - name: web
-      image: nginx
-      resources:
-        limits:
-          memory: "[]" # Noncompliant
-----
-
 ==== Compliant solution
 
 [source,yaml,diff-id=1,diff-type=compliant]
@@ -90,7 +75,8 @@ spec:
 === How does this work?
 
 A limit can be set through the property `resources.limits.memory` of a
-container.
+container. Alternatively, a default limit for a namespace can be set with
+`LimitRange`.
 
 == Resources
 
--- a/rules/S6864/kubernetes/rule.adoc
+++ b/rules/S6864/kubernetes/rule.adoc
@@ -50,9 +50,24 @@ kind: Pod
 metadata:
   name: example
 spec:
-  containers: # Noncompliant
+  containers:
+    - name: web # Noncompliant
+      image: nginx
+----
+
+[source,yaml,diff-id=1,diff-type=noncompliant]
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example
+spec:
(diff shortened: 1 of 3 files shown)


rules/S6864/kubernetes/rule.adoc
```diff
@@ -50,9 +50,24 @@ kind: Pod
 metadata:
   name: example
 spec:
-  containers: # Noncompliant
+  containers:
+    - name: web # Noncompliant
+      image: nginx
+----
+or
```
hendrik-buchwald-sonarsource: I don't think it makes sense to have this additional example, even if we raise an issue for it, as the code just does not make sense.

rules/S6864/kubernetes/rule.adoc
```diff
@@ -75,8 +90,7 @@ spec:
 === How does this work?
 
 A limit can be set through the property `resources.limits.memory` of a
-container. Alternatively, a default limit for a namespace can be set with
```
hendrik-buchwald-sonarsource, line 78: Even if we don't detect it yet, we should still keep this sentence as it is a valid way to not be vulnerable.

hendrik-buchwald-sonarsource (changes requested):
Hey, I am not really in favor of most of these changes. Moving the noncompliant comments seems reasonable but some of the other changes I can't comprehend.

hendrik-buchwald-sonarsource:
`validate_asciidoc` is also failing.

rules/S6869/kubernetes/rule.adoc
```diff
@@ -58,28 +55,30 @@ spec:
...
-  limits:
-  - default:
-      cpu: 0.5
-    type: Container
+  containers:
+    - name: web
+      image: nginx
+      resources:
+        limits:
+          cpu: "0.5"
```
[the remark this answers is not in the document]
navendu-barua-sonarsource: While trying to find some more examples [here](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/),  I did not find one that was with floating `resources/limits/cpu`. This is why I changed it to string. Please let me know if the float values are also acceptable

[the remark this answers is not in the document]
jonas-wielage-sonarsource: I would change it back to `0.5` as this is the usual way people define this property.

rules/S6870/kubernetes/rule.adoc
```diff
@@ -39,9 +39,30 @@ kind: Pod
...
-    - name: web
+  containers:
+    - name: web # Noncompliant
       image: nginx
+      volumeMounts:
+        - name: ephemeral
+          mountPath: "/tmp"
+----
+
+[source,yaml,diff-id=2,diff-type=noncompliant]
```
jonas-wielage-sonarsource, line 50: The examples in this file with `diff-id=2` do not provide any additional value for me. Please remove them

jonas-wielage-sonarsource (approved):
LGTM, please see my suggestions before merging.

PR #3551, opened 2024-01-25
Modify rule S6865: Updated example, message and removed configuration part for now

navendu-barua-sonarsource:
Removed the example for **service account's configuration** part, as this is not yet supported
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6865/kubernetes/rule.adoc
+++ b/rules/S6865/kubernetes/rule.adoc
@@ -34,8 +34,8 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: example-pod
-spec:
-  containers: # Noncompliant
+spec: # Noncompliant
+  containers:
   - name: example-pod
     image: nginx:1.25.3
 ----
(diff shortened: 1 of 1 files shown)


jonas-wielage-sonarsource (approved):
LGTM

rules/S6865/kubernetes/rule.adoc
```diff
@@ -56,28 +56,6 @@ spec:
 
 ----
 
-The same can be achieved by setting `automountServiceAccountToken: false` in the service account's configuration:
```
daniel-teuchert-sonarsource, line 59: Why is this removed?

daniel-teuchert-sonarsource (changes requested):
Overall LGTM. Just a minor question (see below)

rules/S6865/kubernetes/rule.adoc
jonas-wielage-sonarsource: Because we are not able to detect if there is a ServiceAccount next to the Pod at the moment.
We will create a follow up ticket to include this, when we have kubernetes cross-file analysis.

rules/S6865/kubernetes/rule.adoc
```diff
@@ -56,28 +56,6 @@ spec:
...
-spec:
-  containers:
-  - name: example-pod
-    image: nginx:1.25.3
-  serviceAccountName: example-service-account
-----
-
 === How does this work?
 
 The automounting of service account tokens can be disabled by setting `automountServiceAccountToken: false` in the pod's specification or in the service account's configuration.
```
jonas-wielage-sonarsource, line 61:
```suggestion
The automounting of service account tokens can be disabled by setting `automountServiceAccountToken: false` in the pod's specification. Additionally, it can be disabled in the configuration of an accompanied service account.
```

rules/S6865/kubernetes/rule.adoc
daniel-teuchert-sonarsource: I see, thanks for clarifying!

daniel-teuchert-sonarsource (approved):
LGTM! I suggest applying the suggested change from Jonas.

PR #3582, opened 2024-01-31
Modify rule S4123: remove hint about JSdoc and add it to exception

ilia-kebets-sonarsource:
@yassin-kammoun-sonarsource 
The diff might be confusing. In that case, have a look at the result: https://github.com/SonarSource/rspec/blob/modify-S4123-jsdoc-exception/rules/S4123/javascript/rule.adoc

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S4123/javascript/rule.adoc
+++ b/rules/S4123/javascript/rule.adoc
@@ -40,21 +40,25 @@ function foo() {
   return Promise.resolve(42);
 }
 async function bar() {
-  await foo();
+  await foo(); // Compliant
 }
 ----
 
 === Exceptions
 
-The rule does not raise issues if you are awaiting a function whose definition contains JSDoc. This is due to JSDoc often mistakenly declaring a returning type without mentioning that it is resolved by a promise. For example:
+The rule does not raise issues if you are awaiting a function whose definition contains JSdoc with `@returns` or `@return` tags. This is due to JSdoc often mistakenly declaring a returning type without mentioning that it is resolved by a promise. For example:
 
 [source,js]
 ----
+async function foo () {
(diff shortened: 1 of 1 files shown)


rules/S4123/javascript/rule.adoc
```diff
@@ -20,57 +20,41 @@ const x = Promise.resolve(42);
...
-function foo() {
-  Promise.resolve(42);
-}
-async function bar() {
-  await foo(); // Noncompliant
-}
-----
-
-Make sure that you return the promise.
+The rule does not raise issues if you are awaiting a function whose definition contains JSdoc. This is due to JSdoc often mistakenly declaring a returning type without mentioning that it is resolved by a promise. For example:
```
yassin-kammoun-sonarsource, line 49:
```suggestion
The rule does not raise issues if you are awaiting a function whose definition contains JSDoc. This is due to JSDoc often mistakenly declaring a returning type without mentioning that it is resolved by a promise. For example:
```

rules/S4123/javascript/rule.adoc
```diff
@@ -20,57 +20,41 @@ const x = Promise.resolve(42);
...
-  return Promise.resolve(42);
-}
-async function bar() {
-  await foo();
+/**
+ * @return {number} // this should be {Promise<number>}
+ */
+async function foo() {
+  await Promise.resolve(42)
 }
```
yassin-kammoun-sonarsource, line 58: I think the code snippet you used as a unit test is more interesting to highlight the exception.

```js
async function foo () {
  await bar();
}

/**
 * @return {number}
 */
async function bar () {
  return 42;
}
```

PR #3552, opened 2024-01-25
Modify rule S1126: Fix code example

antonioaversa:
## Review

A dedicated reviewer checked the rule description successfully for:

- [ ] logical errors and incorrect information
- [ ] information gaps and missing content
- [ ] text style and tone
- [ ] PR summary and labels follow [the guidelines](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule)

PR #3594, opened 2024-02-01
Modify rule S2386 and S3887: Add frozen collection types

martin-strecker-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2386/csharp/rule.adoc
+++ b/rules/S2386/csharp/rule.adoc
@@ -7,7 +7,7 @@ This rule raises issues for `public static` fields with a type inheriting/implem
 
 === Noncompliant code example
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 public class A 
 {
@@ -18,7 +18,7 @@ public class A
 
 === Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
 public class A 
 {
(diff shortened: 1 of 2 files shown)


rules/S2386/csharp/rule.adoc
```diff
@@ -1,31 +1,29 @@
 == Why is this an issue?
 
-``++public static++`` mutable fields of classes which are accessed directly should be protected to the degree possible. This can be done by reducing the accessibility of the field or by changing the return type to an immutable type.
+`public static` mutable fields of classes which are accessed directly should be protected to the degree possible. This can be done by reducing the accessibility of the field or by changing the return type to an immutable type.
 
 
-This rule raises issues for ``++public static++`` fields with a type inheriting/implementing ``++System.Array++`` or ``++System.Collections.Generic.ICollection<T>++``.
+This rule raises issues for `public static` fields with a type inheriting/implementing `System.Array` or `System.Collections.Generic.ICollection<T>`.
```
antonioaversa, line 6: `<` and `>` are considered [special characters](https://docs.asciidoctor.org/asciidoc/latest/subs/special-characters/) in asciidoc.
Special characters should (at least in principle - maybe in practice it works all the same) be escaped with double plus (ref [here](https://asciidoctor.org/docs/asciidoc-writers-guide/#preventing-substitution) - `single and double plus` section).

antonioaversa (approved):
LGTM. I have left some non-blocking remarks (except maybe escaping).

antonioaversa:
@martin-strecker-sonarsource If `sonar-dotnet` is not updated in https://github.com/SonarSource/sonar-dotnet/pull/8652 to include the changes in this PR, this PR should be mentioned in [the RSPEC update ticket for 9.20](https://github.com/SonarSource/rspec/pull/3594).

rules/S2386/csharp/rule.adoc
martin-strecker-sonarsource: We use single quotes for `<>` all the time in the rspec. It did not cause any problems, yet. Is there any reason why we should use double plus escaping?

rules/S2386/csharp/rule.adoc
```diff
@@ -1,31 +1,29 @@
...
 
 === Noncompliant code example
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 public class A 
 {
   public static string[] strings1 = {"first","second"};  // Noncompliant
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: (see also https://github.com/SonarSource/rspec/pull/3552#pullrequestreview-1844319962 😁 )

martin-strecker-sonarsource:
I added a todo to the implementation PR:
https://github.com/SonarSource/sonar-dotnet/pull/8652#issue-2110087142

rules/S2386/csharp/rule.adoc
antonioaversa: I am afraid that it may just work because of the implementation of asciidoc, and that may break as soon as the implementation gets to conform to the documentation. Anyway, since we already have that many cases, let's keep it as it is.

antonioaversa (approved):

PR #3640, opened 2024-02-12
Modify rule S6876: Refine wording in title and description for clearer distinction from S6877

kaufco:


leonardo-pilastri-sonarsource (approved):
Looks good!

PR #3641, opened 2024-02-12
Modify rule S6877: Refine wording in title and description for clearer distinction from S6876

kaufco:


leonardo-pilastri-sonarsource (changes requested):
I think we can slightly improve a definition

leonardo-pilastri-sonarsource (approved):

PR #3655, opened 2024-02-16
Modify rule s2178: Update rule description

kevin-hinz:
Suggested changes by the Docs Squad:
- formalize word selection to clarify the description

### DIFF
--- a/rules/S2551/why-dotnet.adoc
+++ b/rules/S2551/why-dotnet.adoc
@@ -4,6 +4,6 @@ A shared resource refers to a resource or data that can be accessed or modified 
 
 Shared resources should not be used for https://en.wikipedia.org/wiki/Lock_(computer_science)[locking] as it increases the chance of https://en.wikipedia.org/wiki/Deadlock[deadlocks]. Any other thread could acquire (or attempt to acquire) the same lock while doing some operation, without knowing that the resource is meant to be used for locking purposes.
 
-One case of this is strings, which are https://en.wikipedia.org/wiki/Interning_(computer_science)[interned] by the runtime. This means that each string instance is immutable and stored, and then is reused everywhere it is referenced.
+One example of this is strings which are https://en.wikipedia.org/wiki/Interning_(computer_science)[interned] by the runtime. This means that each string instance is immutable and stored, and then reused everywhere it is referenced.
 
 Instead, a dedicated private `object` instance should be used for each shared resource. Making the lock-specific object `private` guarantees that the access to it is as minimal as possible, in order to avoid deadlocks or lock contention.


rules/S2551/why-dotnet.adoc
```diff
@@ -4,6 +4,6 @@ A shared resource refers to a resource or data that can be accessed or modified
 
 Shared resources should not be used for https://en.wikipedia.org/wiki/Lock_(computer_science)[locking] as it increases the chance of https://en.wikipedia.org/wiki/Deadlock[deadlocks]. Any other thread could acquire (or attempt to acquire) the same lock while doing some operation, without knowing that the resource is meant to be used for locking purposes.
 
-One case of this is strings, which are https://en.wikipedia.org/wiki/Interning_(computer_science)[interned] by the runtime. This means that each string instance is immutable and stored, and then is reused everywhere it is referenced.
+One example of this is strings which are https://en.wikipedia.org/wiki/Interning_(computer_science)[interned] by the runtime. This means that each string instance is immutable and stored, and then reused everywhere it is referenced.
```
Tim-Pohlmann, line 7: Would it make sense to capitalize `string` here, since we are talking about the VB data type `String` or can `string` be considered a regular English word for a list of characters?

Tim-Pohlmann, line 7: This might be out of scope for this PR, but the second sentence is horrible to read.

Tim-Pohlmann (approved):
LGTM. I left two questions with optional suggestions. If you make bigger changes based on them feel free to re-requested review.

rules/S2551/why-dotnet.adoc
kevin-hinz: Good observation. You are correct on the rule of capitalization:
- A "string" can be a list of characters and should only be capitalized when at the start of a sentence. 
- A `String` is a proper noun referring to the VB data type, should be formatted as code, and should be capitalized as a proper noun. 

Because it was written, "One case of this is strings...", and not "One case of this is `strings`...", and in the next paragraph it says "...a dedicated private `object` instance should be used...", I assumed the example was illustrating that _string characters should be replaced with an `object`_.

- If you can confirm that the example is indeed a `String`, I agree it should be capitalized in both instances.

kevin-hinz: Agreed. 
I will issue another, separate PR with only this line, to suggest an improvement.

Tim-Pohlmann: I don't think the distinction exists in VB. If you are talking about a "string", you are talking about a value of the data type `String`. I personally would always use `String`. Especially in this paragraph, since it discusses a technical concept used explicitly for the data type.

kevin-hinz: Great. I've created a new commit to fix the capitalization.
And will create a new PR to improve the paragraph as a whole.

PR #3680, opened 2024-02-23
Modify rule S1764: Add CFamily exception for left shift operator

enzo-pellegrini:
Related to CPP-4234.

## Review

- [x] text style and tone


rules/S1764/cfamily/rule.adoc
```diff
@@ -6,9 +6,8 @@ Using the same value on both sides of a binary operator is a code defect. In the
 
 The following are ignored:
 
-* The expression `1 << 1`
 * When an increment or decrement operator is used, ex: ``+++*p++ == *p+++++``
-* Bitwise operators `|, &, ^`
+* Bitwise operators `|, &, ^, <<`
```
michael-jabbour-sonarsource, line 10: Since we are modifying this, I was thinking about adding a sentence to mention why the rationale behind the exceptions. Instead of just saying that they are ignored. I think that a common property about these exceptions is that they don't indicate an unnecessarily complex expression that has a predictable result. Maybe mentioning this rationale here can help make the intention of the rule clear. WDYT?

Also, if it is easier to just list the operators we check instead of a large list of exceptions, we can consider doing this. WDYT?

rules/S1764/cfamily/rule.adoc
```diff
@@ -6,9 +6,8 @@ Using the same value on both sides of a binary operator is a code defect. In the
 
 The following are ignored:
```
michael-jabbour-sonarsource, line 7:
```suggestion
This rule only reports cases where operators with identical operands produce a definite result and can be rewritten for clarity. Consequently, the following are ignored in this rule:
```
@enzo-pellegrini Do you think that this makes the general goal clearer?

rules/S1764/cfamily/rule.adoc
michael-jabbour-sonarsource: See https://github.com/SonarSource/rspec/pull/3680/files#r1517818169.

PR #3682, opened 2024-02-23
Modify rule S2178, S2551: Update rule description

kevin-hinz:
Changes suggested by the Docs Squad:

Modify rule S3927: Update rule description
- use full word in place of short-hand text

Modify rule S2551: Clarify the example given
- improve word choice to bring clarity to the illustration


Tim-Pohlmann (changes requested):
These changes change the meaning a bit. I left comments on the google doc.

PR #3690, opened 2024-02-27
Modify Rule S6877: Fix tags and update description

alban-auzeill:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6877/java/rule.adoc
+++ b/rules/S6877/java/rule.adoc
@@ -2,8 +2,9 @@
 
 Java 21 introduces the new Sequenced Collections API, which applies to all collections with a defined sequence on their elements, such as `LinkedList`, `TreeSet`, and others (see https://openjdk.org/jeps/431[JEP 431]).
 For projects using Java 21 and onwards, use this API instead of workaround implementations that were necessary before Java 21.
+One of the features of the new Sequenced Collections API is `SequencedCollection.reversed()` which returns a lightweight view of the original collection, in the reverse order.
 
-This rule reports when a reverse copy of a list is created using a list constructor plus `Collections.reverse(list);`, but a reverse view `list.reversed()` would have been sufficient.
+This rule reports when reverse view would have been sufficient instead of a reverse copy of a sequenced collection created using a list constructor plus a `Collections.reverse(collection);` call.
 
 If feasible, a view should be preferred over a copy because a view is a lightweight iterator without modification of the list itself.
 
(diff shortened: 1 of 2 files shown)


rules/S6877/java/rule.adoc
```diff
@@ -3,38 +3,66 @@
 Java 21 introduces the new Sequenced Collections API, which applies to all collections with a defined sequence on their elements, such as `LinkedList`, `TreeSet`, and others (see https://openjdk.org/jeps/431[JEP 431]).
 For projects using Java 21 and onwards, use this API instead of workaround implementations that were necessary before Java 21.
 
-This rule reports when a reverse copy of a list is created, but a reverse view would have been sufficient because the resulting list - as well as the original list in the scope of the resulting list - is not modified (`add`, `remove`).
-`SequencedCollection.reversed()` returns a view of the original list that reflects changes in the original list, while `Collection.reverse(List)` creates a new list that is not connected with the original list.
+This rule reports when a reverse copy of a list is created using a list constructor plus `Collections.reverse(list);`, but a reverse view `list.reversed()` would have been sufficient.
```
leonardo-pilastri-sonarsource, line 6: I would still keep a brief introduction of the api we are talking about
```suggestion
One of the features of the new Sequenced Collections API is `SequencedCollection.reversed()` which returns a lightweight view of the original collection, in the reverse order.

This rule reports when a reverse copy of a sequenced collection is created using a constructor plus `Collections.reverse(collection);`, but a reverse view would have been sufficient.
```

rules/S6877/java/rule.adoc
```diff
@@ -3,38 +3,66 @@
...
-If feasible, a view should be preferred over a copy because:
-
-1. It is a lightweight collection that only functions as a view and it does not allocate memory for the data contained in the underlying collection.
-2. It abstracts from the fact that the collection instance is a list. This results in more generic code that is easier to port, for instance, when you want to use another collection data structure.
+If feasible, a view should be preferred over a copy because a view is a lightweight iterator without modification of the list itself.
 
 == How to fix it
 
-Replace `Collection.reverse(list)` with `list.reversed()`.
+Remove `Collections.reverse(list);` and replace `list` with `list.reversed()` after.
```
leonardo-pilastri-sonarsource, line 12: I would avoid mentioning a list, as the scope of the rule is broader
```suggestion
Remove `Collections.reverse(sequencedCollection);` and replace `sequencedCollection` with `sequencedCollection.reversed()`.
```

leonardo-pilastri-sonarsource (changes requested):
Just a couple of things

rules/S6877/java/rule.adoc
alban-auzeill: There was no `Collections.reverse(sequencedCollection);` because the method has this signature:
```java
java.util.Collections#reverse(List<?> list)
```
Even if it's possible to have `sequencedCollection.reversed()`, I would prefer to speak only about lists.

PR #3693, opened 2024-02-27
Modify rule S1696: promote C# to SonarWay

cristian-ambrosini-sonarsource:
S1696 CSharp:
- RSPEC cleanup
- Promote rule to `SonarWay` profile

### DIFF
--- a/rules/S1696/csharp/rule.adoc
+++ b/rules/S1696/csharp/rule.adoc
@@ -36,11 +36,29 @@ public int GetLengthPlusTwo(string str)
 {
     int length = 2;
 
-    if (str != null) 
+    if (str is not null) 
     {
         length += str.Length;
     }
     else 
+    {
+        log.info("argument was null");
+    }
+    return length;
+}
+----
+
+or:
+
+[source,csharp]
+----
+public int GetLengthPlusTwo(string str) 
+{
+    int length = 2;
+
+    length += str?.Length ?? 0;
+
+    if (str is null)
     {
         log.info("argument was null");
     }
--- a/rules/S1696/csharp/rule.adoc
+++ b/rules/S1696/csharp/rule.adoc
@@ -15,16 +15,14 @@ Instead of catching NullReferenceException, it's better to prevent it from happe
 ----
 public int GetLengthPlusTwo(string str) 
 {
-    int length = 2;
-    try 
+    try
     {
-       length += str.Length;
+        return str.Length + 2;
     }
-    catch (NullReferenceException e) 
+    catch (NullReferenceException e)
     {
-        log.info("argument was null");
+        return 2;
     }
-    return length;
 }
 ----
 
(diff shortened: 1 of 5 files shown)


rules/S1696/csharp/rule.adoc
```diff
@@ -21,9 +28,9 @@ public int GetLengthPlusTwo(string str)
 }
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
```
sebastien-marichal, line 33: I would use the `is not null` in the example. It is more accurate and it was mentioned in the old version of the description.

rules/S1696/csharp/rule.adoc
```diff
@@ -1,10 +1,17 @@
...
 
-``++NullReferenceException++`` should be avoided, not caught. Any situation in which ``++NullReferenceException++`` is explicitly caught can easily be converted to a ``++null++`` test, and any behavior being carried out in the catch block can easily be moved to the "is null" branch of the conditional.
+Catching `NullReferenceException` is generally considered a bad practice because it can hide bugs in your code. Instead of catching this exception, you should aim to prevent it. This makes your code more robust and easier to understand.
+In addition, constantly catching and handling `NullReferenceException` can lead to performance issues. Exceptions are expensive in terms of system resources, so they should be used cautiously and only for exceptional conditions, not for regular control flow.
 
-=== Noncompliant code example
+== How to fix it
 
-[source,csharp]
+Instead of catching NullReferenceException, it's better to prevent it from happening in the first place. You can do this by using null checks or null conditional operators (`?.`) before accessing members of an object.
```
sebastien-marichal, line 8: Since you mention the condition operators, you could add a new pair of Noncompliant/Compliant code example.

rules/S1696/metadata.json
```diff
@@ -34,7 +34,7 @@
     ]
   },
   "defaultQualityProfiles": [
-
+    "Sonar way"
```
sebastien-marichal, line 37: I speak I, I am not sure this the way to do it, but I won't block the PR for this.

sebastien-marichal (changes requested):
I left some suggestions

rules/S1696/metadata.json
sebastien-marichal: I am fine with both, no need to change it.

sebastien-marichal (changes requested):
Sorry, I am not so sure anymore about the conditional access example.
I will let you decide whether you want to keep it

rules/S1696/csharp/rule.adoc
```diff
@@ -1,70 +1,51 @@
...
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
-public int GetLengthPlusTwo(string str) 
+public int GetLengthPlusTwo(string str)
```
cristian-ambrosini-sonarsource, line 33: I'll keep just 1 compliant example, I think it's enough.
I preferred the extended version compared to the more cryptic `public int GetLengthPlusTwo(string str) => str is null ? 2 : str.Length + 2;`, I think it's easier to understand for less experienced devs

rules/S1696/csharp/rule.adoc
```diff
@@ -41,30 +48,30 @@ public int GetLengthPlusTwo(string str)
...
-(visible only on this page)
-
-=== Message
-
-Do not catch NullReferenceException; test for null instead.
+or:
 
+[source,csharp]
+----
+public int GetLengthPlusTwo(string str)
```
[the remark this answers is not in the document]
cristian-ambrosini-sonarsource: Kept only one noncompliant/compliant pair

rules/S1696/csharp/rule.adoc
sebastien-marichal: Fine by me!

sebastien-marichal (approved):
LGTM!

PR #3696, opened 2024-02-27
CPP-4962 S6169 Update rspec to cover if consteval and its effects on is_constant_evaluated

tomasz-kaminski-sonarsource:
## Review

- [ ] text style and tone


tomasz-kaminski-sonarsource:
The main section are pretty done, so you can review them. I left two additional sections I would like to cover.

rules/S6169/cfamily/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Call to \"std::is_constant_evaluated\" should not be gratuitous",
+  "title": "Uses of \"std::is_constant_evaluated\" and \"if consteval\" should not be redundant",
```
loic-joly-sonarsource, line 2: When I read it, I understand that `std::is_constant_evaluated` should not be redundant with `if consteval`.
```suggestion
  "title": "\"std::is_constant_evaluated\" and \"if consteval\" should only be used when necessary",
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
+This rule raises an issue when `std::is_constant_evaluated()` or `if consteval` are used in a context where they always produce the same result, i.e. are always `true` or always `false`.
+
 == Why is this an issue?
 
-``++std::is_constant_evaluated++`` is used to determine whether or not a context is constant-evaluated. This can be useful when, for example, two different implementations are provided for an algorithm: one, usually slow, for compile-time and the other one, faster, for runtime.
+The `std::is_constant_evaluated` function (introduced in {cpp}20) and the `if consteval` language construct (introduced in {cpp}23) are used to determine whether the evaluation is performed at compile-time or runtime.
```
loic-joly-sonarsource, line 5: Should we add parentheses?
```suggestion
The `std::is_constant_evaluated()` function (introduced in {cpp}20) and the `if consteval` language construct (introduced in {cpp}23) are used to determine whether the evaluation is performed at compile-time or runtime.
```

loic-joly-sonarsource, line 5:
```suggestion
The `std::is_constant_evaluated` function (introduced in {cpp}20) and the `if consteval` language construct (introduced in {cpp}23) are used to determine whether the evaluation is performed at compile-time or at runtime.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+This can be useful when, for example, two different implementations are provided for an algorithm: 
+one, that does not perform any IO operations and is compatible with compile-time evaluation, and the other one, at runtime that emits log entries as a side-effect.
```
loic-joly-sonarsource, line 7:
```suggestion
one that does not perform any IO operations and is compatible with compile-time evaluation, and the other one that also emits log entries at runtime.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+
+These constructs should be used inside functions that are `constexpr`, and thus can be evaluated both at compile-time and runtime.
```
loic-joly-sonarsource, line 9:
```suggestion
These constructs should be used inside functions that are `constexpr`, and thus can be evaluated both at compile-time and at runtime.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+a call to `std::is_constant_evaluated` will always return the same result, similarly `if (not) consteval` will always evaluate the same branch,
+making their use redundant.
+
+This rule will raise issues for contexts where expressions are always evaluated at compile-time or always evaluated at runtime.
+
+In contexts that are always evaluated at compile-time:
+
+* `+std::is_constant_evaluated()+` will always return `true`.
+* `+if consteval { /* then-branch */ }+` will always be `true` and evaluate the `then-branch`.
+* `+if !consteval { /* then-branch */ } else { /* else-branch */}+` will always be `false` and evaluate the `else-branch`.
```
loic-joly-sonarsource, line 21: I'm not sure this part is adding lots of value.

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+
+* The initialization of a variable declared `constexpr` or `constinit`.
++
+[source,cpp]
+----
+constexpr int size = std::is_constant_evaluated() ? 10 : 20; // Noncompliant: will always return true
+constinit int val = std::is_constant_evaluated() ? 20 : 30; // Noncompliant: will always return true
+---- 
+
+* All expressions inside the immediate context. This includes bodies of `consteval` function, `then` branches of `if consteval`, and `else` branches of `if not consteval`.
```
loic-joly-sonarsource, line 48:
```suggestion
* All expressions inside an _immediate context_. This includes bodies of `consteval` function, `then` branches of `if consteval`, and `else` branches of `if not consteval`.
```
Or even:
```suggestion
* All expressions inside what is called an _immediate context_: This includes bodies of `consteval` function, `then` branches of `if consteval`, and `else` branches of `if not consteval`.
```

loic-joly-sonarsource, line 48:
```suggestion
* All expressions inside the immediate context. For instance, bodies of `consteval` function, `then` branches of `if consteval`, and `else` branches of `if not consteval`.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    /* Branch is never taken */
+  }
+  
+  if not consteval { // Noncompliant: always false
+    /* Branch is never taken */
+  } else {
+    /* Branch is always taken */
+  }
+
+  bool ce = std::is_constant_evaluated(); // Noncompliant: will always returns true
```
loic-joly-sonarsource, line 65: Maybe redundant with the next line?

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+  return std::is_constant_evaluated(); // Noncompliant: will always returns true
+}
+
+constexpr bool possiblyCompileTimeFunc() {
+  if consteval { // Compliant: depends on the call side
```
loic-joly-sonarsource, line 70:
```suggestion
  if consteval { // Compliant: depends on the call site
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    if consteval { // Noncompliant: will always be true
+      /* .... */ 
+    }
+    if not consteval { // Noncompliant: will always be false
+      /* .... */
+    }
+    return std::is_constant_evaluated(); // Noncompliant: will always returns true
+  }
+
+  if ! consteval { // Compliant: depends on the call side
```
loic-joly-sonarsource, line 80:
```suggestion
  if ! consteval { // Compliant: depends on the call site
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    /* Runtime branch */
+  } else {
+    return std::is_constant_evaluated(); // Noncompliant: will always returns true
+  }
+
+  return std::is_constant_evaluated(); // Compliant: depends on call side
```
loic-joly-sonarsource, line 86:
```suggestion
  return std::is_constant_evaluated(); // Compliant: depends on call site
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+}
+----
+
+In contexts that are always evaluated at runtime:
+
+* `+std::is_constant_evaluated()+` will always return `false`.
+* `+if consteval { /* then-branch */ } else { /* else-branch */}+` will always be `false` and evaluate `else-branch`.
+* `+if !consteval { /* then-branch */ }+` will always be `true` and evaluate `then-branch`.
+
```
loic-joly-sonarsource, line 95: Same remark

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+They include:
+
+* The body of functions that are not `constexpr` or `consteval`.
```
loic-joly-sonarsource, line 98:
```suggestion
* The body of functions that are neither `constexpr` nor `consteval`.
```

loic-joly-sonarsource, line 98: What about:
```cpp
int main() {
    static constinit int i = std::is_constant_evaluated() ? 42 : 0;
    return i;
}
```

(I think this is corner-casy enough that it can just be ignored)

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    /* Always taken branch */
+  }
+
+  if not consteval {  // Noncompliant: always true
+    /* Always taken branch */
+  } else {
+    /* Never taken branch */
+  }
+
+  bool ce = std::is_constant_evaluated(); // Noncompliant: will always returns false
```
loic-joly-sonarsource, line 119: Maybe redundant with next line

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+  return std::is_constant_evaluated(); // Noncompliant: will always returns false
+}
+
+constexpr bool possiblyCompileTimeFunc() {
+  if not consteval { // Compliant: depends on the call side
```
loic-joly-sonarsource, line 124:
```suggestion
  if not consteval { // Compliant: depends on the call site
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    if consteval { // Noncompliant: will always be false
+      /* ... */ 
+    }
+    if not consteval { // Noncompliant: will always be true
+      /* .... */
+    }
+    return std::is_constant_evaluated(); // Noncompliant: will always returns false
+  }
+
+  if consteval { // Compliant: depends on the call side
```
loic-joly-sonarsource, line 134:
```suggestion
  if consteval { // Compliant: depends on the call site
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    /* Compile-time branch */
+  } else {
+    return std::is_constant_evaluated(); // Noncompliant: will always returns false
+  }
+
+  return std::is_constant_evaluated(); // Compliant: depends on call side
```
loic-joly-sonarsource, line 140:
```suggestion
  return std::is_constant_evaluated(); // Compliant: depends on call site
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+  bool ce = std::is_constant_evaluated(); // Noncompliant: will always returns false
+  return std::is_constant_evaluated(); // Noncompliant: will always returns false
+}
+
+constexpr bool possiblyCompileTimeFunc() {
+  if not consteval { // Compliant: depends on the call side
+    if consteval { // Noncompliant: will always be false
+      /* ... */ 
+    }
+    if not consteval { // Noncompliant: will always be true
```
loic-joly-sonarsource, line 128: I globally like those examples, but they are a little bit verbose. Maybe having just `if consteval` of `if not consteval` (not both) would be enough?

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+----
+
+Otherwise, the lambda call operator is implicitly considered to be `constexpr`,
+regardless if it is marked so.
+This means that the lambda can be invoked at compile-time, 
+and uses of `std::is_constant_evaluated()` and `if consteval` are not redundant.
 
+However, when a lambda is invoked locally only in compile-time or runtime context,
+checking the evaluation context is still redundant.
+In particular, this is clear, when the lambda is immediately invoked. The rule will raise issues in that case:
```
loic-joly-sonarsource, line 165: I might miss a case, but could it be simplified into:
```suggestion
However, when a lambda is immediately invoked in compile-time or runtime context, it becomes implicitly consteval, and checking the evaluation context is redundant.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
 
-``++std::is_constant_evaluated++`` will always return ``++true++`` when called in:
+=== When do `constexpr` functions become immediate (`consteval`)?
 
-* the condition of ``++if constexpr++``
-* the condition of ``++static_assert++``
-* ``++consteval++`` functions
+An immediate (`consteval`) function can be invoked only at compile-time,
+and thus requires that all arguments are known at compile-time,
+i.e. either there are constants or the function is invoked in the immediate context:
```
loic-joly-sonarsource, line 182:
```suggestion
i.e. either they are constants or the function is invoked in an _immediate context_:
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+}();
+----
 
-``++std::is_constant_evaluated++`` will always return ``++true++`` when called in:
+=== When do `constexpr` functions become immediate (`consteval`)?
 
-* the condition of ``++if constexpr++``
-* the condition of ``++static_assert++``
-* ``++consteval++`` functions
+An immediate (`consteval`) function can be invoked only at compile-time,
```
loic-joly-sonarsource, line 180:
```suggestion
An immediate (`consteval`) function can only be invoked at compile-time,
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+and thus requires that all arguments are known at compile-time,
+i.e. either there are constants or the function is invoked in the immediate context:
+[source,cpp]
+----
+consteval int process(int);
+float process(float);
```
loic-joly-sonarsource, line 186: Why this overload?

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+  return process(x);
+}
+----
 
+The same behavior is applied to lambdas, both generic and non-generic,
+if they contain an immediate invocation.
 
-=== Noncompliant code example
+As a consequence, uses of `std::is_constant_evaluated()` and `if consteval` may be redundant,
+in case of immediate escalated lambda or function template instantiations.
```
loic-joly-sonarsource, line 217:
```suggestion
As a consequence, uses of `std::is_constant_evaluated()` and `if consteval` are also redundant when used in an _immediate escalated_ lambda or function template instantiations.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+}
+----
 
-And it will always return ``++false++`` in:
+In the case of non-template functions, this can be fixed by putting the call
+to immediate function inside an `if consteval` block.
+However, in the case of templates, it is possible that depending on the template parameters,
+an immediate or runtime function will be called.
+In such case, the compiler will automatically change the enclosing function to an immediate function,
+in a process referred to as immediate escalation:
```
loic-joly-sonarsource, line 198:
```suggestion
in a process referred to as _immediate escalation_:
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+}
+----
 
+The same behavior is applied to lambdas, both generic and non-generic,
+if they contain an immediate invocation.
 
-=== Noncompliant code example
+As a consequence, uses of `std::is_constant_evaluated()` and `if consteval` may be redundant,
+in case of immediate escalated lambda or function template instantiations.
+This rule will raise issues if such uses are redundant for all possible specializations
```
loic-joly-sonarsource, line 218:
```suggestion

This rule will raise issues if such uses are redundant for all possible specializations
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+};
+
+template<typename T>
+constexpr auto = conditionalGenericLambda = [] (auto x) {
+  process(x); // Calls consteval function depending on the argument type
+  return std::is_constant_evaluated(); // Compliant: not all specializations are immediate
+};
+
+template<typename T>
+constexpr auto alwaysGenericLambda = [](T x, int t) {
```
loic-joly-sonarsource, line 250:
```suggestion
constexpr auto alwaysImmediateGenericLambda = [](T x, int t) {
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+  return std::is_constant_evaluated(); // Noncompliant: all specializations are immediate
+}
+
+constexpr auto nonGenericLambda = [](int x) {
+  process(x); // Always call consteval function
+  return std::is_constant_evaluated(); // Noncompliant: lambda is immediate
+};
+
+template<typename T>
+constexpr auto = conditionalGenericLambda = [] (auto x) {
```
loic-joly-sonarsource, line 244:
```suggestion
constexpr auto = conditionallyImmediateGenericLambda = [] (auto x) {
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+  process(t); // Always call consteval function
+  return std::is_constant_evaluated(); // Noncompliant: all specializations are immediate
+};
+----
+
+== How to fix it
+
+Depending on the context, the issue may be fixed by:
+
+* changing the context of invocation: removing `constexpr` from `if` or changing function from `consteval` to `constexpr`.
```
loic-joly-sonarsource, line 260:
```suggestion
* changing the context of invocation: replacing an `if constexpr` with a simple `if` or changing a function from `consteval` to `constexpr`.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -31,10 +277,9 @@ constexpr double power(double b, int x) {
 }
 ----
 
+==== Compliant solution
```
loic-joly-sonarsource, line 280: I'd like to see an additional compliant solution, tagged C++23, based on `if consteval` (this might require removing the diff)

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+== How to fix it
+
+Depending on the context, the issue may be fixed by:
+
+* changing the context of invocation: removing `constexpr` from `if` or changing function from `consteval` to `constexpr`.
+* removing dead code: replacing `std::is_constant_evaluated()` with the produced value, removing `if consteval` and dead branches.
+
+=== Code examples
+
+Changing `if constexpr` into `if`, so the condition is no longer always evaluated at compile-time.
```
loic-joly-sonarsource, line 265: I think this would be a more descriptive title.
```suggestion
=== Changing `if constexpr` into `if`, so the condition is no longer always evaluated at compile-time.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +290,89 @@ constexpr double power(double b, int x) {
 }
 ----
 
+Removing dead branches and inlining the result of `std::is_constant_evaluated()`.
```
loic-joly-sonarsource, line 293:
```suggestion
=== Removing dead branches and inlining the result of `std::is_constant_evaluated()`.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +290,89 @@ constexpr double power(double b, int x) {
...
+
+  if ! consteval {
+    /* Code C */
+  }
+ 
+  return std::is_constant_evaluated(); // Compliant: result depends on evaluation
+}
+----
+
+Changing the function to be declared as `constexpr` to allow runtime evaluation:
```
loic-joly-sonarsource, line 336:
```suggestion
=== Changing the function to be declared as `constexpr` to allow runtime evaluation:
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +290,89 @@ constexpr double power(double b, int x) {
...
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
+consteval bool onlyCompileTimeFunc() {
```
loic-joly-sonarsource, line 342:
```suggestion
consteval bool compileTimeOnlyFunc() {
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +290,89 @@ constexpr double power(double b, int x) {
...
+
+  return std::is_constant_evaluated(); // Noncompliant: will always returns true
+}
+----
+
+==== Compliant solution
+
+[source,cpp,diff-id=3,diff-type=compliant]
+----
+constexpr bool onlyCompileTimeFunc() {
```
loic-joly-sonarsource, line 357:
```suggestion
constexpr bool compileTimeOrRuntimeFunc() {
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +290,89 @@ constexpr double power(double b, int x) {
...
+
+== Resources
+
+=== Documentation
+* {cpp} reference - https://en.cppreference.com/w/cpp/language/consteval[consteval specifier]
+* {cpp} reference - https://en.cppreference.com/w/cpp/language/if#Consteval_if[Consteval if]
+
+=== Standards
+
+* Open Standards - https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2564r3.html[P2564R3] `consteval` needs to propagate up
```
loic-joly-sonarsource, line 377: In the past, we would have avoided referencing ISO C++ papers, because they were considered too arcane for our users. Did this policy change?

tomasz-kaminski-sonarsource: I wouldn't consider this paper to be too arcane, and instead rather good description of the problem.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: It is required for explanation.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I do not think this is allowed by rspec format.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: `std::is_constant_evaluated()` is `bool`, not a function.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This is covered by first section and initialier of `cosntexpr` variable. You could create same case for `if consteval`

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I think we should explain what this functions due, and how they actually work, so I see value in this explanation. We ask user to remove redundant `if consteval` and calls, and they need to know how to remove them.
This ties to rule implementation and messages, and explain what we mean by saying `if consteval` is true, which is not obvious.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I disagreee. Adding `const` here will trigger attempt constant evaluation, regardless if the funciton is `consteval`, `constexpr` or runtime. I think it is worth to show not only return.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: And the same answer.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: See: https://godbolt.org/z/K5f93Ex4Y

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I do not think this is allowed by rspec format.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I think here being verbose, and redundant helps to convey message. I do not have else branches, and could have.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This is not true. Lambda does not become `consteval` because it is called at compile time, and for sure is not `consteval` if it's only at runtime,

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I do not think this is allowed by rspec format.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Added

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,273 @@
...
+    /* Branch is never taken */
+  } else {
+    /* Branch is always taken */
+  }
+
+  bool ce = std::is_constant_evaluated(); // Noncompliant: will always returns true
+  return std::is_constant_evaluated(); // Noncompliant: will always returns true
+}
+
+constexpr bool possiblyCompileTimeFunc() {
```
marco-antognini-sonarsource, line 69: This `possiblyCompileTimeFunc` example and the next one are almost identical. When reading the description, it confused me -- I was looking for the nuances to make sure I was understanding the rule correctly. I expect users to be equally confused. Maybe one of them should be removed?

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,329 @@
...
+void runtime() {
+  static int s = 20; // Evaluted at compile time.
+}
+----
++
+Evaluating such variables at compile time, avoids order of initialization issues.
+It is recommended to mark such variables as `constexpr` (if they can be made `const`) or `constinit`.
+
+* For local variables that are declared `const` and have integral and enumeration type.+
++
```
marco-antognini-sonarsource, line 168: The `+` is currently displayed. Did you intend to remove the empty line?

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,329 @@
...
+  if consteval {  // Noncompliant: always true
+    return true;
+  }
+  return false;
+}();
+----
+
+=== When do `constexpr` functions become immediate (`consteval`)?
+
+An immediate (`consteval`) function can only be invoked at compile-time,
```
marco-antognini-sonarsource, line 236: Why the parenthesis around `consteval`? Do you mean `immediate function === consteval function`? Something else? Maybe introduce what "immediate" means in the first sentence of this section.

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,329 @@
...
+and thus requires that all arguments are known at compile-time,
+i.e. either they are constants or the function is invoked in an _immediate context_:
+[source,cpp]
+----
+consteval int process(int);
+float process(float);
+
+constexpr void foo(int x) {
+  return process(x); // ill-formed, the process cannot be called at compile-time
+}
```
marco-antognini-sonarsource, line 246:
```suggestion
constexpr auto foo(int x) {
  return process(x); // ill-formed, the process cannot be called at compile-time
}
```

Additionally, I would explain why it cannot be called at compile-time. Here, because of the overload, the reader could assume it's because of `process`. But it's because of `x`.

Or maybe you meant to use float?

Or maybe I did not understand something else. 🙃

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,329 @@
...
+  return std::is_constant_evaluated(); // Noncompliant: all specializations are immediate
+}
+
+constexpr auto nonGenericLambda = [](int x) {
+  process(x); // Always call consteval function
+  return std::is_constant_evaluated(); // Noncompliant: lambda is immediate
+};
+
+template<typename T>
+constexpr auto = conditionallyImmediateGenericLambda = [] (auto x) {
```
marco-antognini-sonarsource, line 300:
```suggestion
constexpr auto conditionallyImmediateGenericLambda = [] (auto x) {
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +348,103 @@ constexpr double power(double b, int x) {
...
+    if consteval { // Noncompliant: will always be true
+      /* Code A */ 
+    }
+    if not consteval { // Noncompliant: will always be false
+      /* Code B */
+    }
+    return std::is_constant_evaluated(); // Noncompliant: will always returns true
+  }
+
+  if ! consteval {
```
marco-antognini-sonarsource, line 382: I think we should use only one style in the example.
I would use `not` since it is more natural (for me) because it is between two words (no symbols) and thus "easier" on the brain.

```suggestion
  if not consteval {
```

tomasz-kaminski-sonarsource: I disagree, I want examples to cover options, so it is clear that these two options

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: It was explained paragraph before, I do not see a point of repeating it.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Removed.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: This whole section explains whta immediate is. Please suggest concrete rewrite.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Fixed

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: They focus only on parts that were already explain. First `possiblyCompileTimeFunc` only illustrates always true cases. And this shows only always false cases. I do not think removing them helps.
Please provide concrete suggestionsfor change, that will still preserve the flow of rspec.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Removed one overload and change the name of the funciton so it will not be confused with one having multiple overloads.

rules/S6169/cfamily/rule.adoc
```diff
@@ -141,30 +141,30 @@ constexpr bool possiblyCompileTimeFunc() {
 }
 ----
 
-=== Why issue is raised for non-compile time variables?
+=== When the issue is raised for variables that are neither `constexpr` nor `constinit`?
 
 For some variables, the compile will try to initialize them at compile time,
```
marco-antognini-sonarsource, line 146:
```suggestion
For some variables, the compiler will try to initialize them at compile time,
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -233,21 +233,32 @@ constexpr bool ce = [] () {
...
 to immediate function inside an `if consteval` block.
+[source,cpp]
+----
+consteval int handle(int);
+
+constexpr void foo(int x) {
+  if consteval {
+    return handle(x); // OK, the handle is evaluated only at compile-time
+  }
+}
```
marco-antognini-sonarsource, line 259: I would remove the `return` (since it's a void function and control flow is orthogonal to the question)

rules/S6169/cfamily/rule.adoc
marco-antognini-sonarsource: My confusion arises from the use of parenthesis. The new wording makes it easier to understand. One nit grammar though:

```
An immediate (including one marked `consteval`) function can only be invoked at compile-time,
and thus requires that all arguments are known at compile-time,
i.e. either they are constants or the function is invoked in an _immediate context_:
```

the `(including one marked consteval)` part should be after `function`, not before (`one` is currently a "forward" reference and it's confusing, if not incorrect).

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,340 @@
...
+    /* Compile-time branch */
+  } else {
+    return std::is_constant_evaluated(); // Noncompliant: will always returns false
+  }
+
+  return std::is_constant_evaluated(); // Compliant: depends on call site
+}
+----
+
+=== When the issue is raised for variables that are neither `constexpr` nor `constinit`?
```
marco-antognini-sonarsource, line 144:
```suggestion
=== When is the issue raised for variables that are neither `constexpr` nor `constinit`?
```

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Done

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Done.

rules/S6169/cfamily/rule.adoc
marco-antognini-sonarsource: On 2nd read, I think I understand better what you wanted to achieve.

I think you did a great job with the sub-sections below -- the titles helped me understand the flow. I wonder if it would also help to transform these two sentences into titles:
 * In contexts that are always evaluated at compile-time:
 * In contexts that are always evaluated at runtime:
Or add a title before if transforming is not great, like `# Compile-time/Runtime contexts`. It would help visually mark the examples as _definitely_ different.

WDYT? I don't want to press too much on this -- it's just an idea and it's very subjective.

rules/S6169/cfamily/rule.adoc
marco-antognini-sonarsource, line 1: What follows in a grammar nit -- feel free to dismiss if you want to move forward, as the text is still understandable.

The use of `will + inf` in this document can often, if not always, be replaced with the present tense. What I mean by that is that facts (that don't change) should be presented/explained using the present tense.

Example:

> When used inside a context that is either always evaluated at compile-time or always evaluated at runtime, a call to std::is_constant_evaluated ~will always return~ _always returns_ the same result, similarly if (not) consteval ~will always evaluate~ _always evaluates_ the same branch, making their use redundant.

rules/S6169/cfamily/rule.adoc
marco-antognini-sonarsource: In lit tests, I would definitely agree. Here, a bit less. But it's subjective so I'm fine keeping as-is.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: I try to use the additional titles to cover something extra, more detailed information, that are not important for understanding the issue. For me this generic explanation of always compile/run time is essential and part of main body of the issue, thus I would prefer not exclude it.

I think the main body could be researched to avoid such duplication, but I do not think it is worth it now. We already are spending a week on this rule. If you can thing of slight renames or commentary on the examples that would help to highlight the difference, I would be ok with having them.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Just edit them in place.

rules/S6169/cfamily/rule.adoc
marco-antognini-sonarsource: Okay, let's not overengeneer this then. 👍

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+// The lambda is provably invoked at compile-time only:
+constexpr bool ce = [] () {
+  if consteval {  // Noncompliant: always true
+    return true;
+  }
+  return false;
+}();
+----
+
+=== When do `constexpr` functions become immediate (`consteval`)?
```
tomasz-kaminski-sonarsource, line 234: Using `consteval` is misleading. 
```suggestion
=== When do `constexpr` functions become immediate (compile-time only)?
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  }
+  
+  if not consteval { // Noncompliant: always false
+    /* Branch is never taken */
+  } else {
+    /* Branch is always taken */
+  }
+
+  bool ce = std::is_constant_evaluated(); // Noncompliant: will always returns true
+  return std::is_constant_evaluated(); // Noncompliant: will always returns true
```
marco-antognini-sonarsource, line 66:
```suggestion
  bool ce = std::is_constant_evaluated(); // Noncompliant: always true
  return std::is_constant_evaluated(); // Noncompliant: always returns true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+}
+
+constexpr bool possiblyCompileTimeFunc() {
+  if consteval { // Compliant: depends on the call site
+    if consteval { // Noncompliant: will always be true
```
marco-antognini-sonarsource, line 71:
```suggestion
    if consteval { // Noncompliant: always true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+      /* .... */ 
+    }
+    if not consteval { // Noncompliant: will always be false
```
marco-antognini-sonarsource, line 74:
```suggestion
    if not consteval { // Noncompliant: always false
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+      /* .... */
+    }
+    return std::is_constant_evaluated(); // Noncompliant: will always returns true
```
marco-antognini-sonarsource, line 77:
```suggestion
    return std::is_constant_evaluated(); // Noncompliant: always returns true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  }
+
+  if ! consteval { // Compliant: depends on the call site
+    /* Runtime branch */
+  } else {
+    return std::is_constant_evaluated(); // Noncompliant: will always returns true
```
marco-antognini-sonarsource, line 83:
```suggestion
    return std::is_constant_evaluated(); // Noncompliant: always returns true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+
+  return std::is_constant_evaluated(); // Compliant: depends on call site
+}
+----
+
+In contexts that are always evaluated at runtime:
+
+* `+std::is_constant_evaluated()+` will always return `false`.
+* `+if consteval { /* then-branch */ } else { /* else-branch */}+` will always be `false` and evaluate `else-branch`.
+* `+if !consteval { /* then-branch */ }+` will always be `true` and evaluate `then-branch`.
```
marco-antognini-sonarsource, line 94:
```suggestion
* `+std::is_constant_evaluated()+` always returns `false`.
* `+if consteval { /* then-branch */ } else { /* else-branch */}+` always evaluates `else-branch`.
* `+if !consteval { /* then-branch */ }+` always evaluates `then-branch`.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  }
+
+  if not consteval {  // Noncompliant: always true
+    /* Always taken branch */
+  } else {
+    /* Never taken branch */
+  }
+
+  bool ce = std::is_constant_evaluated(); // Noncompliant: will always returns false
+  return std::is_constant_evaluated(); // Noncompliant: will always returns false
```
marco-antognini-sonarsource, line 120:
```suggestion
  bool ce = std::is_constant_evaluated(); // Noncompliant: always false
  return std::is_constant_evaluated(); // Noncompliant: always returns false
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+}
+
+constexpr bool possiblyCompileTimeFunc() {
+  if not consteval { // Compliant: depends on the call site
+    if consteval { // Noncompliant: will always be false
```
marco-antognini-sonarsource, line 125:
```suggestion
    if consteval { // Noncompliant: always false
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+      /* ... */ 
+    }
+    if not consteval { // Noncompliant: will always be true
```
marco-antognini-sonarsource, line 128:
```suggestion
    if not consteval { // Noncompliant: always true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+      /* .... */
+    }
+    return std::is_constant_evaluated(); // Noncompliant: will always returns false
```
marco-antognini-sonarsource, line 131:
```suggestion
    return std::is_constant_evaluated(); // Noncompliant: always returns false
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  }
+
+  if consteval { // Compliant: depends on the call site
+    /* Compile-time branch */
+  } else {
+    return std::is_constant_evaluated(); // Noncompliant: will always returns false
```
marco-antognini-sonarsource, line 137:
```suggestion
    return std::is_constant_evaluated(); // Noncompliant: always returns false
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  }
+
+  return std::is_constant_evaluated(); // Compliant: depends on call site
+}
+----
+
+=== When is the issue raised for variables that are neither `constexpr` nor `constinit`?
+
+For some variables, the compiler will try to initialize them at compile time,
+and only if such initialization is not possible, will evaluate them at runtime.
```
marco-antognini-sonarsource, line 147:
```suggestion
For some variables, the compiler tries to initialize them at compile time.
They are initialized at runtime only if such initialization is not possible.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+
+This happens in two scenarios:
+
+* For variables with static and thread storage duration, like global, static, and thread-local variables.
++
+[source,cpp]
+----
+int x = 10;           // Evaluated at compile time
+int const y = init(); // Evaluated at compile time, if `init()` is constant-expression
+std::mutex m;         // Evaluated at compile time, the selected constructor is constexpr
```
marco-antognini-sonarsource, line 157:
```suggestion
int const y = init(); // Evaluated at compile time if `init()` is constant-expression
std::mutex m;         // Evaluated at compile time because the selected constructor is constexpr
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+void cpp03Code() {
+  int const size = 5; // Evaluated at compile time
+  int arr[size] = {}; // OK, size is constant
+  for (int i = 0; i < size; ++i) {
+    /* .... */
+  }
+}
+----
++
+Evaluating such variables at compile time was introduced in the first {cpp} standard before `constexpr` was introduced, to allow patterns like the above.
```
marco-antognini-sonarsource, line 180: A small orthogonal suggestion to make the flow easier.
```suggestion
Evaluating such variables at compile time was already possible before `constexpr` was introduced to allow patterns like the above.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+It is recommended to mark such variables as `constexpr` since {cpp}11.
 
-``++std::is_constant_evaluated++`` will always return ``++true++`` when called in:
+Due to the above special rules, the `std::is_constant_evaluated()` and `if consteval` will always be `true` within such implicit constant initialization.
```
marco-antognini-sonarsource, line 183:
```suggestion
Due to the above special rules, `std::is_constant_evaluated()` and `if consteval` are always `true` within such implicit constant initialization.
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+The call operator of a lambda can be explicitly marked `consteval`. In such cases, it can only be invoked at compile-time, and the rule raises an issue:
 
-=== Noncompliant code example
+[source,cpp]
+----
+[] () consteval {
+  if consteval {  // Noncompliant: always true
+    /* .... */
+  }
+  return std::is_constant_evaluated(); // Noncompliant: will always returns true
```
marco-antognini-sonarsource, line 210:
```suggestion
  return std::is_constant_evaluated(); // Noncompliant: always returns true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+----
+
+Otherwise, the lambda call operator is implicitly considered to be `constexpr`,
+regardless if it is marked so.
+This means that the lambda can be invoked at compile-time, 
+and uses of `std::is_constant_evaluated()` and `if consteval` are not redundant.
+
+However, when a lambda is invoked locally only in compile-time or runtime context,
+checking the evaluation context is still redundant.
+In particular, this is clear, when the lambda is immediately invoked. The rule will raise issues in that case:
```
marco-antognini-sonarsource, line 221:
```suggestion
In particular, this is obvious when the lambda is immediately invoked. The rule raises issues in that case:
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
 
 [source,cpp]
 ----
+// The lambda is provably invoked at compile-time only:
```
marco-antognini-sonarsource, line 225:
```suggestion
// The lambda is provably invoked only at compile-time:
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+constexpr void foo(int x) {
+  if consteval {
+    handle(x); // OK, the handle is evaluated only at compile-time
+  }
+}
+----
+
+However, in the case of templates, it is possible that depending on the template parameters,
+an immediate or runtime function will be called.
+In such case, the compiler will automatically change the enclosing function to an immediate function,
```
marco-antognini-sonarsource, line 262:
```suggestion
In such case, the compiler automatically changes the enclosing function to an immediate function,
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  return process(x);
+}
+----
+
+The same behavior is applied to lambdas, both generic and non-generic,
+if they contain an immediate invocation.
+
+As a consequence, uses of `std::is_constant_evaluated()` and `if consteval` are also redundant when used in an _immediate escalated_ lambda or function template instantiations.
+
+This rule will raise issues if such uses are redundant for all possible specializations
```
marco-antognini-sonarsource, line 283:
```suggestion
This rule raises issues if such uses are redundant for all possible specializations
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+
+template<typename T>
+constexpr bool conditionallyImmediate(T x) {
+  process(x); // Calls consteval function depending on the argument type
+  return std::is_constant_evaluated(); // Compliant: not all specializations are immediate
+}
+
+template<typename T>
+constexpr bool alwaysImmediate(T x, int t) {
+  process(t); // Always call consteval function
```
marco-antognini-sonarsource, line 299:
```suggestion
  process(t); // Always calls consteval function
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+  return std::is_constant_evaluated(); // Noncompliant: all specializations are immediate
+}
+
+constexpr auto nonGenericLambda = [](int x) {
+  process(x); // Always call consteval function
```
marco-antognini-sonarsource, line 304:
```suggestion
  process(x); // Always calls consteval function
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+
+template<typename T>
+constexpr auto conditionallyImmediateGenericLambda = [] (auto x) {
+  process(x); // Calls consteval function depending on the argument type
+  return std::is_constant_evaluated(); // Compliant: not all specializations are immediate
+};
+
+template<typename T>
+constexpr auto alwaysImmediateGenericLambda = [](T x, int t) {
+  process(t); // Always call consteval function
```
marco-antognini-sonarsource, line 316:
```suggestion
  process(t); // Always calls consteval function
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,27 +1,338 @@
...
+=== Code examples
+
+Changing `if constexpr` into `if`, so the condition is no longer always evaluated at compile-time.
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
 constexpr double power(double b, int x) {
   if constexpr (std::is_constant_evaluated()) {  // Noncompliant: will always be true
```
marco-antognini-sonarsource, line 337:
```suggestion
  if constexpr (std::is_constant_evaluated()) {  // Noncompliant: always true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +357,103 @@ constexpr double power(double b, int x) {
...
+
+Removing dead branches and inlining the result of `std::is_constant_evaluated()`.
+
+==== Noncompliant code example
+
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
+constexpr bool possiblyCompileTimeFunc() {
+  if consteval {
+    if consteval { // Noncompliant: will always be true
```
marco-antognini-sonarsource, line 382:
```suggestion
    if consteval { // Noncompliant: always true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +357,103 @@ constexpr double power(double b, int x) {
...
+      /* Code A */ 
+    }
+    if not consteval { // Noncompliant: will always be false
```
marco-antognini-sonarsource, line 385:
```suggestion
    if not consteval { // Noncompliant: always false
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +357,103 @@ constexpr double power(double b, int x) {
...
+      /* Code B */
+    }
+    return std::is_constant_evaluated(); // Noncompliant: will always returns true
```
marco-antognini-sonarsource, line 388:
```suggestion
    return std::is_constant_evaluated(); // Noncompliant: always returns true
```

rules/S6169/cfamily/rule.adoc
```diff
@@ -45,4 +357,103 @@ constexpr double power(double b, int x) {
...
+[source,cpp,diff-id=3,diff-type=noncompliant]
+----
+consteval bool compileTimeOnlyFunc() {
+  if consteval {  // Noncompliant: always true
+    /* Code A */
+  } else {
+    /* Code B */
+  }
+
+  return std::is_constant_evaluated(); // Noncompliant: will always returns true
```
marco-antognini-sonarsource, line 430:
```suggestion
  return std::is_constant_evaluated(); // Noncompliant: always returns true
```

rules/S6169/cfamily/rule.adoc
marco-antognini-sonarsource: Suggestions added

rules/S6169/cfamily/rule.adoc
loic-joly-sonarsource: Yes, my suggestion is not good... My point is that I found it difficult to read your sentence...  What about:
```
However, when a lambda is immediately invoked, its evaluation context is known and does not need to be tested.
```

tomasz-kaminski-sonarsource: I wanted to make this sentence cover case, if we extend imply to cover for example passing the lambda to `std::any_of`, so say this as special example.

rules/S6169/cfamily/rule.adoc
loic-joly-sonarsource: It is not clear that the first section has priority over the second section. Right now, we say that, unconditionally, bodies of functions not declared constexpr/consteval are always evaluated at runtime, while it's only partially true.
But being more accurate would mean being much more verbose...

tomasz-kaminski-sonarsource: Also rspec, goes in tandem with rule imp, that will raise the issue in that case, and point to the constexpr variable initialization in secondary. So, I do not think this will be that much confusion for the users.

tomasz-kaminski-sonarsource: Added example with separate paragrahp

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Update the examples to give them some subsections.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Update the examples to give them some subsections.

rules/S6169/cfamily/rule.adoc
tomasz-kaminski-sonarsource: Update the examples to give them some subsections.

rules/S6169/cfamily/rule.adoc
```diff
@@ -1,48 +1,479 @@
...
+
+Depending on the context, the issue may be fixed by:
+
+* changing the context of invocation: replacing an `if constexpr` with a simple `if` or changing a function from `consteval` to `constexpr`.
+* removing dead code: replacing `std::is_constant_evaluated()` with the produced value, removing `if consteval` and dead branches.
+
+=== Inside `if constexpr` condition
+
+Changing `if constexpr` into `if` leads to the condition is no longer always evaluated at compile-time,
+and now result of `std::is_constant_evaluated()` depends onn the call site.
```
marco-antognini-sonarsource, line 347:
```suggestion
Changing `if constexpr` into `if` leads to the condition is no longer always evaluated at compile-time,
and now the result of `std::is_constant_evaluated()` depends on the call site.
```

marco-antognini-sonarsource (approved):

PR #3699, opened 2024-02-28
Move dotnet keywords to shared_content

Tim-Pohlmann:
## Review

- [ ] text style and tone


shared_content/dotnet/csharp_keywords.adoc
martin-strecker-sonarsource, line 1: Consider `csharp_dictionary.adoc` as we also have "concept" and other vocabulary in it.

martin-strecker-sonarsource (changes requested):

martin-strecker-sonarsource (approved):
LGTM. Please merge yourself.

rules/S4586/csharp/rule.adoc
```diff
@@ -1,7 +1,4 @@
-:keyword_null: null
-:keyword_async: async
-:concept_method: method
-:typeparameter_TResult: <TResult>
+include::../../../shared_content/dotnet/csharp_dictionary.adoc[]
```
antonioaversa, line 1: One possible remark on the folder structure.

In multiple rules that have both C# and VB.NET, we use the following approach:
- we have some of the content of the rule shared in a file at the root of the rule directory (e.g. `S1048/how-dotnet.adoc`)
- that file does code inclusion based on language: e.g. `include::{language}/noncompliant-code.adoc[]`
- the language-specific `rule.adoc` does `:language: csharp` or `:language: vbnet` first thing, before including any section in common between the two languages (defined in `..`)

In such a scenario, if we want to use the dictionary, and not repeat twice the language info, we would need to do an inclusion like the following:
```
include::../../../shared_content/dotnet/{language}_dictionary.adoc[]
```

Now, I know that inclusions such as `include::{language}/noncompliant-code.adoc[]` work, because we use them in multiple places (see S1048). My question is: do inclusion also work with part of the name of a folder (i.e. `{language}_dictionary`)?

antonioaversa (approved):
Approved with a [remark](https://github.com/SonarSource/rspec/pull/3699/files#r1506237881).

rules/S4586/csharp/rule.adoc
Tim-Pohlmann: I don't think that's needed. If the dictionary is included in `rule.adoc` it can be used in all included files, language-specific or not.

PR #3701, opened 2024-02-28
Modify rule S2701: promote C# to SonarWay

cristian-ambrosini-sonarsource:
S2701 CSharp:
- RSPEC cleanup
- Promote rule to `SonarWay` profile

### DIFF
--- a/rules/S2701/csharp/rule.adoc
+++ b/rules/S2701/csharp/rule.adoc
@@ -14,8 +14,6 @@ include::how-nunit.adoc[]
 
 include::how-xunit.adoc[]
 
-include::../rspecator.adoc[]
-
 == Resources
 
 === Documentation
@@ -24,3 +22,5 @@ include::../rspecator.adoc[]
 * https://xunit.net/docs/getting-started/netcore/cmdline[xUnit Documentation]
 * Microsoft Learn - https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-mstest[MSTest Documentation]
 * https://github.com/xunit/xunit/issues/2027[Xunit doesn't have an Assert.Fail() operation]
+
+include::../rspecator.adoc[]
--- a/rules/S2701/csharp/rule.adoc
+++ b/rules/S2701/csharp/rule.adoc
@@ -6,10 +6,6 @@ Most of the testing frameworks provide more explicit assertion methods that will
 
 === Exceptions
 
-`Assert.True` and `Assert.False` are not raising this issue for xUnit because `Assert.Fail` was introduced in 2020 version 2.4.2 and both `Assert.True(false, message)` and `Assert.False(true, message)` were ways to simulate it.
-
-=== Exceptions
-
 In the context of xUnit, `Assert.True` and `Assert.False` are not flagged by this rule. This is because `Assert.Fail` was only introduced in 2020 with version `2.4.2`. Prior to this, developers used `Assert.True(false, message)` and `Assert.False(true, message)` as workarounds to simulate the functionality of `Assert.Fail()`.
 
 include::how-mstest.adoc[]
--- a/rules/S2701/csharp/rule.adoc
+++ b/rules/S2701/csharp/rule.adoc
@@ -6,7 +6,11 @@ Most of the testing frameworks provide more explicit assertion methods that will
 
 === Exceptions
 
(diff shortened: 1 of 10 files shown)


rules/S2701/csharp/rule.adoc
```diff
@@ -1,30 +1,26 @@
...
-(visible only on this page)
+include::how-nunit.adoc[]
 
-include::../message.adoc[]
+include::how-xunit.adoc[]
 
-'''
-== Comments And Links
-(visible only on this page)
+include::../rspecator.adoc[]
```
sebastien-marichal, line 17: Probably not a big issue, I think this should be after the Resources section

rules/S2701/csharp/rule.adoc
```diff
@@ -1,30 +1,26 @@
...
 
-[source,csharp]
-----
-bool b = true;
-NUnit.Framework.Assert.AreEqual(true, b);
-Xunit.Assert.NotSame(true, b);
-Microsoft.VisualStudio.TestTools.UnitTesting.Assert.AreEqual(true, b);
-System.Diagnostics.Debug.Assert(true);
-----
+`Assert.True` is not raising this issue for xUnit because `Assert.Fail` was introduced in 2020 version 2.4.2 and `Assert.True(false, message)` was a way to simulate it.
```
sebastien-marichal, line 9: Just to be sure, do we only ignore `Assert.True(false)` and not `Assert.False(true)` as well?

sebastien-marichal (approved):
LGTM!
Left only nitpick comments

rules/S2701/csharp/rule.adoc
cristian-ambrosini-sonarsource: Exactly, `AssertFalse(true, ...)` will report, this is at least the current rule implementation

rules/S2701/csharp/rule.adoc
cristian-ambrosini-sonarsource: Indeed!

rules/S2701/csharp/rule.adoc
sebastien-marichal: Having both ignored could save a few FPs maybe, WDYT?
Look like to me a low hanging FP.

cristian-ambrosini-sonarsource: On Sourcegraph `Assert.True(false` reports 7.2k+ results, while `Assert.False(true` returns 2.9k. I think during the implementation [this comment](https://github.com/xunit/xunit/issues/2027#issuecomment-549434940) was the prominent source for this and we skipped only Assert.True. Maybe you are right and we should play safe and avoid reporting on Assert.False. We will create some FN as a drawback. I will update this rspec and do the change in sonar-dotnet

sebastien-marichal: In which case `Assert.False(true)` would be an FN?

cristian-ambrosini-sonarsource: Assert.False(false,..)

cristian-ambrosini-sonarsource: https://github.com/SonarSource/sonar-dotnet/pull/8823

rules/S2701/csharp/rule.adoc
```diff
@@ -1,30 +1,30 @@
...
-bool b = true;
-NUnit.Framework.Assert.AreEqual(true, b);
-Xunit.Assert.NotSame(true, b);
-Microsoft.VisualStudio.TestTools.UnitTesting.Assert.AreEqual(true, b);
-System.Diagnostics.Debug.Assert(true);
-----
+`Assert.True` and `Assert.False` are not raising this issue for xUnit because `Assert.Fail` was introduced in 2020 version 2.4.2 and both `Assert.True(false, message)` and `Assert.False(true, message)` were ways to simulate it.
 
-ifdef::env-github,rspecator-view[]
+=== Exceptions
```
sebastien-marichal, line 11: There is 2 Exceptions sections

sebastien-marichal (changes requested):

rules/S2701/csharp/rule.adoc
cristian-ambrosini-sonarsource: There you go, thanks for taking a second look 😄

sebastien-marichal (approved):
LGTM!

PR #3704, opened 2024-02-28
Modify rule S2053: Add better code examples and more how to fix sections

gaetan-ferry-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2053/python/how-to-fix-it/cryptodome.adoc
+++ b/rules/S2053/python/how-to-fix-it/cryptodome.adoc
@@ -6,7 +6,7 @@ include::../../common/fix/code-rationale.adoc[]
 
 ==== Noncompliant code example
 
-[source,python,diff-id=1,diff-type=noncompliant]
+[source,python,diff-id=3,diff-type=noncompliant]
 ----
 from Crypto.Protocol.KDF import scrypt
 
@@ -15,7 +15,7 @@ digest = scrypt(password, salt=b"F3MdWpeHeeSjlUxvKBnzzA", key_len=32, N=2**17, r
 
 ==== Compliant solution
 
-[source,python,diff-id=1,diff-type=compliant]
+[source,python,diff-id=3,diff-type=compliant]
 ----
 import secrets
 from Crypto.Protocol.KDF import scrypt
--- a/rules/S2053/python/how-to-fix-it/cryptodome.adoc
+++ b/rules/S2053/python/how-to-fix-it/cryptodome.adoc
@@ -29,5 +29,5 @@ digest = scrypt(password, salt=salt, key_len=32, N=2**17, r=8, p=1)
 include::../../common/fix/salt.adoc[]
 
 Here, the compliant code example ensures the salt is random and has a sufficient
-length by calling the `secrets.token_bytes` function. This one internally uses a
-cryptographically secure pseudo random number generator.
+length by calling the `secrets.token_bytes` function. This function internally
+uses a cryptographically secure pseudo-random number generator.
--- a/rules/S2053/python/how-to-fix-it/cryptodome.adoc
+++ b/rules/S2053/python/how-to-fix-it/cryptodome.adoc
@@ -1,0 +1,33 @@
+== How to fix it in Cryptodome
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+==== Noncompliant code example
+
+[source,python,diff-id=1,diff-type=noncompliant]
+----
+from Crypto.Protocol.KDF import scrypt
+
+digest = scrypt(password, salt=b"F3MdWpeHeeSjlUxvKBnzzA", key_len=32, N=2**17, r=8, p=1) # Noncompliant
+----
+
+==== Compliant solution
+
+[source,python,diff-id=1,diff-type=compliant]
+----
+import secrets
+from Crypto.Protocol.KDF import scrypt
+
+salt = secrets.token_bytes(32)
+digest = scrypt(password, salt=salt, key_len=32, N=2**17, r=8, p=1)
+----
+
+=== How does this work?
+
+include::../../common/fix/salt.adoc[]
+
+Here, the compliant code example ensures the salt is random and has a sufficient
+length by calling the `secrets.token_bytes` function. This one internally uses a
+cryptographically secure pseudo random number generator.
--- a/rules/S2053/python/how-to-fix-it/cryptography.adoc
+++ b/rules/S2053/python/how-to-fix-it/cryptography.adoc
@@ -6,7 +6,7 @@ include::../../common/fix/code-rationale.adoc[]
 
 ==== Noncompliant code example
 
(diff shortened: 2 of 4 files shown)


rules/S2053/python/how-to-fix-it/cryptodome.adoc
```diff
@@ -0,0 +1,33 @@
...
+----
+
+==== Compliant solution
+
+[source,python,diff-id=3,diff-type=compliant]
+----
+import secrets
+from Crypto.Protocol.KDF import scrypt
+
+salt = secrets.token_bytes(32)
```
jamie-anderson-sonarsource, line 23: I'm divided on the use of `secrets.token_bytes()` for this purpose.

According to the documentation for the `token_*` functions:
> The [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) module provides functions for generating secure tokens, suitable for applications such as password resets, hard-to-guess URLs, and similar.

These functions are not really supposed to be used for unspecified random value generation. However, because they do return a cryptographically-secure sequence of bytes (because they wrap `os.urandom()` behind the scenes) they often get used for generating things like keys, salts and nonces.

I've done quite a bit of reading on this, including the email chain behind [PEP-506](https://peps.python.org/pep-0506/), and nothing clarifies the best option here.

I was going to suggest the use of `secrets.SystemRandom` / `random.SystemRandom` here but the class does not provide a method of generating random bytes.

I wish there was a better option here, but I would suggest using `os.urandom()` for generating any random value that isn't strictly being used as a token.

rules/S2053/python/how-to-fix-it/cryptodome.adoc
```diff
@@ -0,0 +1,33 @@
...
+digest = scrypt(password, salt=salt, key_len=32, N=2**17, r=8, p=1)
+----
+
+=== How does this work?
+
+include::../../common/fix/salt.adoc[]
+
+Here, the compliant code example ensures the salt is random and has a sufficient
+length by calling the `secrets.token_bytes` function. This one internally uses a
+cryptographically secure pseudo random number generator.
```
jamie-anderson-sonarsource, line 33: If we're ok with the use of `secrets.token_bytes` as a function:
```suggestion
length by calling the `secrets.token_bytes` function. This function internally
uses a cryptographically secure pseudo-random number generator.
```

If we switch to `os.urandom`:
```suggestion
length by calling the `os.urandom` function. This function generates random
sequences of bytes that are suitable for cryptographic use.
```

rules/S2053/python/how-to-fix-it/cryptography.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+==== Compliant solution
+
+[source,python,diff-id=2,diff-type=compliant]
+----
+import secrets
+from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+from cryptography.hazmat.primitives import hashes
+
+salt = secrets.token_bytes(32)
```
jamie-anderson-sonarsource, line 25: See my previous comment on this matter.

rules/S2053/python/how-to-fix-it/cryptography.adoc
```diff
@@ -0,0 +1,35 @@
...
+digest = PBKDF2HMAC(hashes.SHA256(), length=32, salt=salt, iterations=100000).derive(password)
+----
+
+=== How does this work?
+
+include::../../common/fix/salt.adoc[]
+
+Here, the compliant code example ensures the salt is random and has a sufficient
+length by calling the `secrets.token_bytes` function. This one internally uses a
+cryptographically secure pseudo random number generator.
```
jamie-anderson-sonarsource, line 35: See my previous comment on this matter.

rules/S2053/python/how-to-fix-it/std.adoc
```diff
@@ -8,25 +8,26 @@ include::../../common/fix/code-rationale.adoc[]
...
 
 [source,python,diff-id=1,diff-type=compliant]
 ----
-import crypt
+import hashlib
+import secrets
 
-salt = crypt.mksalt(crypt.METHOD_SHA256)
-hash = crypt.crypt(password, salt)
+salt = secrets.token_bytes(32)
```
jamie-anderson-sonarsource, line 23: See my previous comment on this matter.

rules/S2053/python/how-to-fix-it/std.adoc
```diff
@@ -8,25 +8,26 @@ include::../../common/fix/code-rationale.adoc[]
...
 ----
 
 === How does this work?
 
 include::../../common/fix/salt.adoc[]
 
 Here, the compliant code example ensures the salt is random and has a sufficient
-length by calling the `crypt.mksalt` function. This one internally uses a
+length by calling the `secrets.token_bytes` function. This one internally uses a
 cryptographically secure pseudo random number generator.
```
jamie-anderson-sonarsource, line 33: See my previous comment on this matter.

jamie-anderson-sonarsource (changes requested):

rules/S2053/python/how-to-fix-it/cryptodome.adoc
egon-okerman-sonarsource: `secrets.token_bytes()` really just calls `os.urandom()` behind the scenes and does nothing more, so there is no functional difference at all. Considering `secrets` was added in Python 3.6, I would prefer `os.urandom()` only because of its better version support.

gaetan-ferry-sonarsource: `token_bytes` calls `random.SystemRandom` who is only a wrapper to `os.urandom`. The `os.urandom` provides this notice:

> See also
>
> The [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) module provides higher-level functions.
> For an easy-to-use interface to the random number generator provided by your platform, please see
> [random.SystemRandom](https://docs.python.org/3/library/random.html#random.SystemRandom).

I originally thought it would be better to point to a higher-level library, for example, to make sure people are aware that they exist and end up using them for general secure RNG use cases. For example, I would not like people to rely on `os.urandom` for generating numbers in a range or choosing a random array element or shuffling something, etc.

For this random.SystemRandom might be better suited. However, I like that `token_bytes` has a sane default key size.

WDYT?

egon-okerman-sonarsource: > The os.urandom provides this notice
I think this is mostly a reference to methods like `token_hex` or `token_urlsafe` which are higher-level than `os.urandom`. Like you said, `secrets.token_bytes`, `random.SystemRandom` and `os.urandom` are different API calls with the same result.

> However, I like that `token_bytes` has a sane default key size.
This seems fine to me. Let's go with that.

egon-okerman-sonarsource (approved):
LGTM

jamie-anderson-sonarsource (approved):
LGTM after the discussion around `secrets.token_bytes()`

PR #3713, opened 2024-02-29
Modify rule S2551: clarify example used in the description

kevin-hinz:
Changes suggested by the Docs Squad:
- the example used to illustrate the rule did not consider if the case was in C# or VB
- In addition, the description was confusing so was simplified to make the case easier to understand.

### DIFF
--- a/rules/S2551/csharp/rule.adoc
+++ b/rules/S2551/csharp/rule.adoc
@@ -2,7 +2,7 @@ include::../why-dotnet.adoc[]
 
 For example, a `string` should never be used for locking. When a `string` is https://en.wikipedia.org/wiki/Interning_(computer_science)[interned] by the runtime, it can be shared by multiple threads, breaking the locking mechanism.
 
-Instead, a dedicated private `object` instance should be used for each shared resource. This minimizes access to the lock instance, avoiding deadlocks or lock contention.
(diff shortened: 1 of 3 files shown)


rules/S2551/csharp/rule.adoc
```diff
@@ -1,5 +1,9 @@
 include::../why-dotnet.adoc[]
 
+For example, a `string` should never be used for locking. When a `string` is https://en.wikipedia.org/wiki/Interning_(computer_science)[interned] by the runtime, it can be shared by multiple threads, breaking the locking mechanism.
+
+Instead, a dedicated private `object` instance should be used for each shared resource. This minimizes access to the lock instance, avoiding deadlocks or lock contention.
```
Tim-Pohlmann, line 5: For me `and` sounds better here. But I'm neither a native speaker nor a technical writer. So consider this an optional suggestion. But if you apply it, also apply it to the VB.NET rule.

```suggestion
Instead, a dedicated private `object` instance should be used for each shared resource. This minimizes access to the lock instance, avoiding deadlocks and lock contention.
```

Tim-Pohlmann (approved):
Very nice! One small optional suggestion.

rules/S2551/csharp/rule.adoc
kevin-hinz: Thanks for the suggestion.
I made the improvement, added a commit, and all checks have passed.

PR #3716, opened 2024-03-01
Modify rule S1948: Clarify the description (SONARJAVA-4814)

johann-beleites-sonarsource:
SONARJAVA-4814 Improved S1948 RSPEC to better explain why public collections are a problem


kaufco (approved):
Lgtm!

PR #3719, opened 2024-03-01
Modify rule S5542: Change text for rule update

jamie-anderson-sonarsource:
Rule S5542 will be modified so that it is no longer raised on CBC mode. A new rule will be introduced that triggers on the use of unauthenticated encryption modes, which will cover CBC.

See APPSEC-1560.

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5542/rationale.adoc
+++ b/rules/S5542/rationale.adoc
@@ -10,16 +10,19 @@ When selecting encryption algorithms, tools, or combinations, you should also
 consider two things:
 
 1. No encryption is unbreakable.
-2. The strength of an encryption algorithm is usually measured by the effort required to crack it within a reasonable time frame.
-
+2. The strength of an encryption algorithm is usually measured by the effort
+   required to crack it within a reasonable time frame.
 
 For these reasons, as soon as cryptography is included in a project, it is
 important to choose encryption algorithms that are considered strong and secure
 by the cryptography community.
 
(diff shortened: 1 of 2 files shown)


rules/S5542/rationale.adoc
```diff
@@ -10,16 +10,19 @@ When selecting encryption algorithms, tools, or combinations, you should also
...
 
-For AES, the weakest modes are CBC (Cipher Block Chaining) and ECB (Electronic Codebook) 
-because they are either vulnerable to padding oracles or do
-not provide authentication mechanisms.
+For AES, the weakest mode is ECB (Electronic Codebook). Repeated blocks of data
+are encrypted to the same value, making them easy to identify and reducing the
+difficulty of recovering the original cleartext.
+
+Unauthenticated modes such as CBC (Cipher Block Chaining) may be used but are
+prone to attacks that manipulate the ciphertext. They must be used with caution.
```
jamie-anderson-sonarsource, line 25: I have added this note because the rule update will take a while to put in place. We will still raise S5542 on uses of CBC while this is in place.

I've tried to use wording here that can potentially be left in place after the rule change. We want to push developers towards authenticated encryption modes, even if CBC is no longer detected by S5542. (It should however be detected by an upcoming new rule around unauthenticated encryption.)

sebastien-andrivet-sonarsource (approved):
LGTM

PR #3736, opened 2024-03-06
Modify rule S1696: Fix link

marco-antognini-sonarsource:
I was looking for double dashes in cfamily rules for CPP-4910 and randomly stumbled upon S1696 for C#.

The link has two issues:
 1. `--` is transformed into em-dash `—​`.
 2. `]` in the link text was interpreted as the closing bracket instead of a regular text character.

In both cases, the fix is to escape such character.

FWIW, I did not attempt to search for other broken links in C#.

## Review

- [ ] text style and tone


gregory-paidis-sonarsource (approved):
LGTM, thanks for fixing this!

PR #3752, opened 2024-03-11
Modify rules S1871 and S3923: Update description in accordance with the new understanding

guillem-bartina-sonarsource:
Update the description of these two closely related rules in accordance with the new understanding (see ticket).

Implementation ticket: [CPP-2002](https://sonarsource.atlassian.net/browse/CPP-2002)

## Review

- [ ] text style and tone

[CPP-2002]: https://sonarsource.atlassian.net/browse/CPP-2002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rules/S1871/cfamily/exceptions.adoc
+++ b/rules/S1871/cfamily/exceptions.adoc
@@ -1,6 +1,6 @@
 === Exceptions
 
-Branches in an `if` or `switch` chain with trivial body are ignored. A body is considered trivial if it contains only a single line of code, possibly with a trailing `break` in case of `switch` chains.
+Branches in an `if` or `switch` with a _simple_ body are ignored. A body is considered simple if it contains a single line of code, possibly followed by a `break` in case of `switch` chains.
 
 [source,cpp]
 ----
--- a/rules/S1871/cfamily/exceptions.adoc
+++ b/rules/S1871/cfamily/exceptions.adoc
@@ -1,6 +1,6 @@
 === Exceptions
 
-Branches in an `if` or `switch` chain that have a _simple_ body are ignored. A body is considered simple if it contains a single line of code, possibly followed by a `break` in case of `switch` chains.
+Branches in an `if` chain or `switch` statement that have a _simple_ body are ignored. A body is considered simple if it contains a single line of code, possibly followed by a `break` in the case of `switch` statements.
 
 [source,cpp]
 ----
--- a/rules/S1871/cfamily/exceptions.adoc
+++ b/rules/S1871/cfamily/exceptions.adoc
@@ -1,0 +1,14 @@
+=== Exceptions
+
+Branches in an `if` or `switch` chain with trivial body are ignored. A body is considered trivial if it contains only a single line of code, possibly with a trailing `break` in case of `switch` chains.
+
+[source,cpp]
+----
+if (a == 1) {
+  doSomething();  // Compliant, usually this is done on purpose to increase the readability
+} else if (a == 2) {
+  doSomethingElse();
+} else {
+  doSomething();
+}
+----
--- a/rules/S1871/cfamily/exceptions.adoc
+++ b/rules/S1871/cfamily/exceptions.adoc
@@ -1,6 +1,6 @@
 === Exceptions
 
-Branches in an `if` or `switch` with a _simple_ body are ignored. A body is considered simple if it contains a single line of code, possibly followed by a `break` in case of `switch` chains.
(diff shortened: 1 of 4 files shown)


rules/S1871/cfamily/exceptions.adoc
```diff
@@ -0,0 +1,14 @@
+=== Exceptions
+
+Branches in an `if` or `switch` chain that have a _simple_ body are ignored. A body is considered simple if it contains a single line of code, possibly followed by a `break` in case of `switch` chains.
```
marco-antognini-sonarsource, line 3: Nit: "switch chain" is not a thing AFAIK.

```suggestion
Branches in `if` or `switch` statements that have a _simple_ body are ignored. A body is considered simple if it contains a single line of code, possibly followed by a `break` in the case of ``switch``es.
```

rules/S3923/cfamily/rule.adoc
```diff
@@ -1,4 +1,32 @@
...
+
+int func() {
+  if (a > 10) return 1;
+  return 2; // implicit default branch
+}
+
+int b = a > 10 ? 1 : 2;
+                  // ^ default branch by construction
+----
+
```
marco-antognini-sonarsource, line 16: Anything before "Why is this an issue?" will be displayed on all tabs on SonarQube/Cloud/... Example: [SQ](https://peach.aws-prd.sonarsource.com/coding_rules?open=cpp%3AS1763&rule_key=cpp%3AS1763) [Rules](https://rules.sonarsource.com/cpp/RSPEC-1763/)

We should have at most a sentence or two in this section. See https://github.com/SonarSource/rspec/blob/master/docs/description.adoc

I think the example is not bad but should be in the next section.

rules/S3923/cfamily/rule.adoc
```diff
@@ -1,4 +1,32 @@
...
+In the following code:
+
+include::../common/code-example.adoc[]
+
+Either there is a copy-paste error, or the `if` or `switch` chain or ternary operator is unecessary.
+
+== How to fix it
+
+Either fix the copy-paste error or remove the conditional structure.
```
marco-antognini-sonarsource, line 29: I don't think it's worth having a separate tab for one sentence given how trivial the fix is. Keep it in "why is this an issue".

marco-antognini-sonarsource (changes requested):

rules/S3923/cfamily/metadata.json
frederic-tingaud-sonarsource, line 1: I would change the impact at the root of the rule because that impact is the same for all implementations.

rules/S3923/cfamily/rule.adoc
```diff
@@ -1,4 +1,37 @@
...
+
+In the following code:
+
+include::../common/code-example.adoc[]
+
+Either there is a copy-paste error that should be fixed; or the `if` chain, `switch` statement or ternary operator is unecessary and should be removed.
+
+=== Considerations
+
+_All branches_ includes the default branch, which is the branch that is executed when none of the conditions of the structure are satisfied. This branch may present itself explicitly, implicitly or by construction.
```
frederic-tingaud-sonarsource, line 13: Custom titles are allowed in this section, but I'm not convinced it brings much value. One sentence we often use is "This rule triggers when...", which might work well here. For example:

```suggestion
This rule triggers when all branches, _including the default branch_, are identical.
The default branch is the branch that is executed when none of the conditions of the structure are satisfied. This branch may present itself explicitly, implicitly, or by construction.
```

rules/S3923/cfamily/rule.adoc
```diff
@@ -1,4 +1,37 @@
...
+    return 1;
+  default: // explicit default branch
+    return 2;
+}
+
+int b = a == 0 ? 1 : 2;
+                  // ^ default branch by construction
+----
 
 ifdef::env-github,rspecator-view[]
```
frederic-tingaud-sonarsource, line 36: We should also mention S1871 in Resources/Related rules.

frederic-tingaud-sonarsource (changes requested):

rules/S3923/cfamily/metadata.json
guillem-bartina-sonarsource: Yes, this was my initial idea. But wouldn't we have to synchronize with the rest of the lang squads so that they change the metadata in their analyzers?

marco-antognini-sonarsource (approved):
LGTM, but wait for at least @frederic-tingaud-sonarsource's approval before merging.

frederic-tingaud-sonarsource (approved):

PR #3775, opened 2024-03-18
Modify rule S5205,S5213: Cover std::move_only_function (CPP-5047)

marco-antognini-sonarsource:
Apply some general NFC improvements to S5213 and its companion rule S5205. Apply LaYC format where relevant.

Remove link to C++ Core Guideline: it is mostly empty (no example) and has an ambiguous point about std::function.

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S5205/cfamily/rule.adoc
+++ b/rules/S5205/cfamily/rule.adoc
@@ -1,23 +1,26 @@
 == Why is this an issue?
 
-When you want to define a function that can accept a function pointer as an argument, there are three ways in {cpp} to declare the parameter type:
+When you want to define a function that is parametrized by another function, there are three ways to declare its parameter type:
 
-* A function pointer: ``++void f(void (*callback)());++``
-* A ``++std::function++``: ``++void f(std::function<void()> callback);++``
-* A template argument: ``++template<class Callback> void f(Callback callback);++``
+* A function pointer: +
+  ``++void f(void (*callback)());++``
+* A typed-erased function wrapper such as ``++std::function++``: +
+  ``++void f(std::function<void()> callback);++``
+* A template parameter: +
+  ``++template <class Callback> void f(Callback callback);++``
 
-Using a function pointer is an inferior solution, for the following reasons:
+Using a function pointer is an inferior solution for the following reasons:
 
 * Only a function pointer can be passed as an argument, while the other options offer the caller more flexibility because they can take more advanced functors, such as lambdas with some captured state
 * The syntax is obscure
 * It typically has worse performance than the template parameter solution.
 
-See S5213 for a discussion of how to choose between ``++std::function++`` and a template parameter.
+See S5213 for a discussion on choosing between ``++std::function++`` and a template parameter.
 
 
 === Noncompliant code example
 
-[source,cpp]
+[source,cpp,diff-id=1,diff-type=noncompliant]
 ----
 using Criterion = bool (*)(DataPoint const&);
 void filter(DataSet* data, Criterion criterion); // Noncompliant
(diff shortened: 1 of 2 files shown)


rules/S5205/cfamily/rule.adoc
```diff
@@ -1,58 +1,77 @@
 == Why is this an issue?
 
-When you want to define a function that can accept a function pointer as an argument, there are three ways in {cpp} to declare the parameter type:
+When you want to define a function that is parametrized by another function, there are three ways to declare its parameter type:
```
LoghinVladDev, line 3: I believe the subject in the second part of the sentence is a bit unclear.

What about
```suggestion
When you want to receive a function pointer as a parameter in a function definition, there are three ways to declare its parameter type:
```

rules/S5213/cfamily/rule.adoc
```diff
@@ -1,73 +1,115 @@
...
+You can use it to store a lambda capturing a non-copyable object such as `std::unique_ptr`, which cannot be done with `std::function`.
 
+The abstraction offered by `std::function` and `std::move_only_function` comes at a cost: a compiler typically cannot inline a call to these types due to the type erasure.
 
-As a conclusion, if the functor can be known at compile-time, you should prefer using a template parameter, if it has to be dynamic, ``++std::function++`` will give you greater flexibility.
+Additionally, these wrappers can be "empty", meaning they do not currently represent any callable.
+While invoking an empty `std::function` throws a `std::bad_function_call`, invoking an empty `std::move_only_function` results in undefined behavior.
 
+// We could also explain that move_only_function is better at const-, noexpect-, and r-valuesness-correctness.
+// But it would bring little to the discussion, if not only confusion.
```
LoghinVladDev, line 27: Maybe in a potential "Going the extra mile" section?

It would be easy to compare type-erasure vs templates when enforcing certain invocable properties (const, noexcept, r-value, return correctness - signature vs the `std::is_*_invocable_*` and concepts).

It might be going too far, but I believe this is why that section exists.

Edit : I see that the concepts & `std::is_*_invocable_*` are also mentioned below in "Working with template-parameters" and examples, so this is only to be considered if worth it and does not involve too much effort

rules/S5205/cfamily/rule.adoc
```diff
@@ -1,58 +1,77 @@
...
+template <class Criterion> // Compliant, uses the more efficient template argument
+void filter(DataSet& data, Criterion criterion);
 
 using Callback = std::function<void(EventInfo const&)>;
 class Button {
 public:
-    void addOnClick(Callback c) {myOnClickHandler = c;} // Compliant, uses the more flexible std::function
+  void addOnClick(Callback c) { // Compliant, uses the more flexible std::function
+    myOnClickHandler = c;
+  }
```
LoghinVladDev, line 53: For when a std::function is passed as parameter
```suggestion
  void addOnClick(Callback const& c) { // Compliant, uses the more flexible std::function
    myOnClickHandler = c;
  }
```
or
```suggestion
  void addOnClick(Callback c) { // Compliant, uses the more flexible std::function
    myOnClickHandler = std::move(c);
  }
```

LoghinVladDev (changes requested):
Minor findings, need to have another look, but looks good!

rules/S5205/cfamily/rule.adoc
marco-antognini-sonarsource: I wanted to avoid "function pointer" in this sentence because it may not be what you want to use in the end. That's why I went for "parametrized by another function".

That said, I appreciate the sentence is not easy to get... Happy to get any other suggestions.

rules/S5213/cfamily/rule.adoc
marco-antognini-sonarsource: > Maybe in a potential "Going the extra mile" section?

Maybe indeed... I'm pondering the ROI. If it was a blog post, why not.

But since the issue is raised on `std::function` and `std::move_only_function`, readers are probably already aware of the benefit of `std::move_only_function` and thus of what you mention, at least from a high-level point of view, aren't they?

I'm also not sure how to write such a section... I fear I would either digress too much or gloss over some important aspects. Finding the right balance for the audience isn't easy here.

I think, if it's alright with you, I'll keep this in mind and maybe come back to it in another PR if time allows. WDYT?

rules/S5205/cfamily/rule.adoc
LoghinVladDev: Okay, then a rephrase putting the parameter as the subject? Or just get rid of the "pointer" all together? (keeping the function as the subject, easier to read IMO)
```suggestion
When you want to receive a function as a parameter in a function definition, there are three ways to declare its parameter type:
```

rules/S5213/cfamily/rule.adoc
LoghinVladDev: Yeah, any way you put it, going into detail here requires a blogpost, or even a book chapter, as much as I'd want to have a paragraph summarising everything. Best put it in the 'maybe blogpost' drawer

rules/S5205/cfamily/rule.adoc
LoghinVladDev: Upon re-reading, this repeats "parameter", and without it, the same reference to the subject ("its") becomes ambiguous. I'm fine with either

LoghinVladDev:
Would it make sense to mention `std::move` of `std::move_only_function` as an exception?

marco-antognini-sonarsource:
> Would it make sense to mention `std::move` of `std::move_only_function` as an explicit exception?

`std::move` is not a specific exception actually, and `std::move_only_function` is considered exactly as `std::function`.

https://godbolt.org/z/aavqvc6hY

All of this falls into

> [this rule does not consider the parameter if] it participates in other operations.

from the conclusion.

Does that make sense?

LoghinVladDev (approved):

PR #3777, opened 2024-03-18
Modify rule S5566: Cover std::ranges::contains (CPP-5058)

LoghinVladDev:
## Review

- [x] text style and tone

### DIFF
--- a/rules/S5566/cfamily/rule.adoc
+++ b/rules/S5566/cfamily/rule.adoc
@@ -48,7 +48,7 @@ bool areAllDesired(std::vector<int> values) {
 }
 
 bool containsDesired(std::vector<int> values, int desired) {
-  for (int val : values) {
+  for (int val : values) { // Noncompliant
     if (val == desired) {
       return true;
     }
@@ -89,13 +89,13 @@ bool asDesired(const int v);
 
 bool areAllDesired(std::vector<int> values) {
   return std::ranges::all_of(values, asDesired);
-  // Or until C++20:
+  // Or, before C++20:
   return std::all_of(std::begin(values), std::end(values), asDesired);
 }
 
 bool containsDesiredCpp23(std::vector<int> values, int desired) {
   return std::ranges::contains(values, desired);
-  // Or until C++23:
+  // Or, before C++23:
   return std::any_of(std::begin(values), std::end(values), [desired](int val) { return val == desired; });
 }
 
@@ -110,7 +110,7 @@ void displayNumber(int i) {
 void f(vector<int> &v) {
 
   std::ranges::for_each(v, displayNumber);
-  // Or until C++20:
+  // Or, before C++20:
   std::for_each(v.begin(), v.end(), displayNumber);
 
   auto sum = 0;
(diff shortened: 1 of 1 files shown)


LoghinVladDev:
To be merged together with https://github.com/SonarSource/sonar-cpp/pull/3413

rules/S5566/cfamily/rule.adoc
```diff
@@ -77,14 +87,22 @@ using namespace std;
 
 bool asDesired(const int v);
 
-bool areAllDesired2(std::vector<int> values) {
+bool areAllDesired(std::vector<int> values) {
```
LoghinVladDev, line 90: Any idea for why this was `areAllDesired2` previously? If so, should this naming be kept for `containsDesired` as well (below)?

rules/S5566/cfamily/rule.adoc
```diff
@@ -19,6 +19,7 @@ Currently, this rule supports:
 * ``++std::all_of++`` (since {cpp}11) and ``++std::ranges::all_of++`` (since {cpp}20): returns ``++true++`` if all elements in the given range are matching the given predicate, ``++false++`` otherwise
 * ``++std::none_of++`` (since {cpp}11) and ``++std::ranges::none_of++`` (since {cpp}20): returns ``++true++`` if no elements in the given range are matching the given predicate, ``++false++`` otherwise
 * ``++std::any_of++`` (since {cpp}11) and ``++std::ranges::any_of++`` (since {cpp}20): returns ``++true++`` if at least one element in the given range is matching the given predicate, ``++false++`` otherwise
+* ``++std::contains++`` and ``++std::ranges::contains++`` (since {cpp}23): returns ``++true++`` if at least one element in the given range is equal to the given value, ``++false++`` otherwise
```
marco-antognini-sonarsource, line 22: There is no `std::contains` function. There are container-specific functions, but phrased like this it's confusing. Some were already introduced in C++20 (map). Since this should be covered by [S6171](https://sonarsource.github.io/rspec/#/rspec/S6171), I would remove this mention (PR title too).

rules/S5566/cfamily/rule.adoc
marco-antognini-sonarsource: Probably a copy-paste from the test file?

Anyway, since there are a couple of variations for the compliant solution based on STD version, I would suggest the following:
 * Keep the name without `2`
 * Remove the `diff-id`/`diff-type` from `[source]`. 
 * In `f`, show `std::ranges::for_each` before `std::for_each(v.begin(), v.end()`.
 * Merge `areAllDesiredCpp20` into `areAllDesired` with the best version first (C++20)  (like it's done for `f`)
 * Merge `containsDesiredCpp23` into `containsDesired` in a similar fashion.

The alternative to keep the diff would be to split the compliant solution, but I don't think it would be beneficial for readability...

WDYT?

rules/S5566/cfamily/rule.adoc
marco-antognini-sonarsource, line 22: I would be tempted to add links to C++ reference in the "Documentation" section since we do it for pretty much every rule nowadays.

marco-antognini-sonarsource (changes requested):
Minor tweaks only

rules/S5566/cfamily/rule.adoc
LoghinVladDev: I (wrongly) assumed the iterator+sentinel variant was introduced outside `std::ranges`, but it wasn't. Thanks for pointing that out

rules/S5566/cfamily/rule.adoc
LoghinVladDev: Makes sense, addressed

rules/S5566/cfamily/rule.adoc
LoghinVladDev: Added

rules/S5566/cfamily/rule.adoc
```diff
@@ -114,6 +128,11 @@ void f(vector<int> &v) {
 
 * {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/e49158a/CppCoreGuidelines.md#es71-prefer-a-range-for-statement-to-a-for-statement-when-there-is-a-choice[ES.71: Prefer a range-`for`-statement to a `for`-statement when there is a choice]
 * {cpp} Core Guidelines - https://github.com/isocpp/CppCoreGuidelines/blob/e49158a/CppCoreGuidelines.md#p3-express-intent[P.3: Express intent]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/for_each[`std::for_each`]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/all_any_none_of[`std::all_of`, `std::any_of`, `std::none_of`]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/ranges/for_each[`std::ranges::for_each`]
+* {cpp} reference - https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of[`std::ranges::all_of`, `std::ranges::any_of`, `std::ranges::none_of`]
```
LoghinVladDev, line 134: Would `std::ranges::*_of` work, or keep it explicit?

marco-antognini-sonarsource (commented):
A couple of nits

marco-antognini-sonarsource (approved):
LGTM, thanks

PR #3780, opened 2024-03-18
Modify rule S6881: SONARJAVA-4832 Modify description of S6881 to reflect reduced scope

johann-beleites-sonarsource:


irina-batinic-sonarsource (commented):
Why do we have only the Noncompliant code example in the first Code example but no Compliant Solution?

PR #3786, opened 2024-03-19
S5148: Adjust description to fit improved detection

daniel-teuchert-sonarsource:
This PR should be merged once https://sonarsource.atlassian.net/browse/SONARHTML-220 is implemented.

### DIFF
--- a/rules/S5148/html/rule.adoc
+++ b/rules/S5148/html/rule.adoc
@@ -2,24 +2,22 @@ include::../description.adoc[]
 
 include::../ask-yourself.adoc[]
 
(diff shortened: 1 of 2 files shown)


loris-s-sonarsource (approved):
LGTM but why did you remove the inclusion of recommended.adoc while keeping the file?

daniel-teuchert-sonarsource:
> LGTM but why did you remove the inclusion of recommended.adoc while keeping the file?

Because that file is used by the JavaScript part of this rule and the content is not really applicable to HTML.

PR #3794, opened 2024-03-20
Modify S2955 C#: Fix compliant solution

gregory-paidis-sonarsource:
Provide a better solution, that does not box value types to `object`.
For more details, see [this thread](https://github.com/SonarSource/sonar-dotnet/issues/8933#issuecomment-2004644300).

### DIFF
--- a/rules/S2955/csharp/rule.adoc
+++ b/rules/S2955/csharp/rule.adoc
@@ -30,7 +30,7 @@ bool IsDefault<T>(T value)
 ----
 bool IsDefault<T>(T value)
 {
-  if(EqualityComparer<T>.Default.Equals(value, default(T)))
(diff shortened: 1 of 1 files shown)


rules/S2955/csharp/rule.adoc
```diff
@@ -30,7 +30,7 @@ bool IsDefault<T>(T value)
 ----
 bool IsDefault<T>(T value)
 {
-  if(object.Equals(value, default(T)))
+  if(EqualityComparer<T>.Default.Equals(value, default(T)))
```
antonioaversa, line 33:
```suggestion
  if (EqualityComparer<T>.Default.Equals(value, default(T)))
```

antonioaversa (approved):
LGTM, left a minor, non-blocking remark.

PR #3797, opened 2024-03-20
Modify rule S6045: Use LaYC format and update description for C++23 (CPP-5024)

marco-antognini-sonarsource:
This ended up being bigger than I initially expected when picking up the ticket. Hope it was worth it...

Any feedback you can give me is welcome. :-)

Note that I purposely did not update https://sonarsource.github.io/rspec/#/rspec/S6021 in this PR. That would be too much work for this "easy" ticket.

## Review

- [ ] text style and tone


rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
 == Why is this an issue?
 
-{cpp}14 has introduced transparent comparators: the function objects that support heterogeneous comparison (i.e., comparison of values of different types, such as ``++std::string++`` and ``++char const*++``). When using such comparator, the search-optimized containers, namely, ``++std::set++``, ``++std::multiset++``, ``++std::map++``, and ``++std::multimap++``, enable additional lookup-function overloads that support types different from the ``++key_type++``.
+Transparent function objects are function-like types that support heterogeneous operations. There are essentially two kinds of such types: transparent comparators and transparent hashers. In other words, a transparent comparator for strings would support comparing any two of `std::string`, ``++char const*++``, and ``++std::string_view++``.
```
loic-joly-sonarsource, line 3: I think it's broader than just these types...
```suggestion
Transparent function objects are function-like types that support heterogeneous operations. There are essentially two kinds of such types: transparent comparators and transparent hashers. For instance, a transparent comparator for strings would support comparing a `std::string` with a ``++char const*++``, a ``++std::string_view++``....
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
-  std::size_t operator()(std::string_view sv) const {
-    std::hash<std::string_view> hasher;
-    return hasher(sv);
-  }
-};
+// Given a container c with a non-transparent comparator:
+std::set<std::string> c = ...;
+
+// Calling "find" with a C-style string (char const*)
+c.find("Nemo");
```
loic-joly-sonarsource, line 15:
```suggestion
auto it = c.find("Nemo");
```
I don't like discarding the value of find. And I don't like too verbose examples. I think just storing it might be a good balance...

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
+// is equivalent to
+c.find(std::string{"Nemo"});
```
loic-joly-sonarsource, line 17:
```suggestion
auto it = c.find(std::string{"Nemo"});
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
+
+// Calling C++20 "contains" with a std::string_view sv
+// does not compile since conversion has to be explicit:
+//   c.contains(sv);
+// It has to be rewritten like this:
+c.contains(std::string(sv));
```
loic-joly-sonarsource, line 23:
```suggestion
//   if (c.contains(sv)) { ... }
// It has to be rewritten like this:
if (c.contains(std::string(sv))) { ... }
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
 ----
 
-Prefer using a transparent comparator with associative ``++std::string++`` containers to avoid creating the temporary. Note that transparent comparators are strongly discouraged if used with types that are not directly comparable as it will lead to the creation of ``++O(log(container.size())))++`` temporaries with lookup functions such as ``++find++``, ``++count++``, and ``++lower_bound++``.
+Using heterogeneous comparison and hashing directly benefits the application performance since unnecessary temporaries can be avoided. This is particularly true when the ``++key_type++`` is `std::string`.
```
loic-joly-sonarsource, line 26: I don't think the performance improvement is more true for string than for other types.

 ```suggestion
Using heterogeneous comparison and hashing directly benefits the application performance since unnecessary temporaries can be avoided. Since so many string-like classes exist, and since suitable transparent function objects work with strings, there is no good reason not to use them.```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
 
-Custom non-transparent functor (comparator, equality or hasher) may have different semantics than corresponding operators on `std:::string`. In such case, the heterogeneous lookup can still be enabled, by declaring the ``++is_transparent++`` nested type in the functor, and adjusting the implementation to accept either ``++std::string_view++`` or any type (i.e. turning it into a template). The later change is required to avoid the creation of `std::string` temporaries for each invocation and thus degradation of performance.
+For this reason, this rule detects `std::set`, `std::map`, ``++std::unordered_set++``, ``++std::unordered_map++``, `std::multiset`, `std::multimap`, ``++std::unordered_multiset++``, and ``++std::unordered_multimap++`` types that use `std::string` as key and do not enable heterogeneous operations.
```
loic-joly-sonarsource, line 28: Shouldn't this be at the end?

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
+    "Dory", "Marlin", "Nemo", "Emo", "Darla"
+  };
+  sea.find("Nemo"); // This leads to a temporary std::string{"Nemo"}.
+
+  std::string_view hero{"Nemo"};
+  sea.contains(std::string(hero)); // Extra temporary std::string.
 }
+----
+
+Unfortunately, the default `std::less<std::string>` is not transparent.
```
loic-joly-sonarsource, line 63: I don't think it's unfortunate... See S6021

```suggestion
The default comparator for `std::set<std::string>` is `std::less<std::string>`  which  is not transparent.
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
+  std::set<std::string, CaseInsensitiveCompare> m = { // Noncompliant
+    "Dory", "Marlin", "Nemo", "Emo", "Darla"
+  };
+
+  m.lower_bound("nemo"); // This leads to a temporary std::string{"Nemo"}.
 }
 ----
 
-=== Compliant solution
+However, like `std::less<std::string>`, `CaseInsensitiveCompare` is not transparent, and the code contains undesired temporary strings.
```
loic-joly-sonarsource, line 117:
```suggestion
However, like `std::less<std::string>`, `CaseInsensitiveCompare` is not transparent, and the code will trigger the construction of undesired temporary strings.
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
 
-[source,cpp]
+==== Compliant solution
+
+A comparator needs to explicitly declare itself as transparent. For this purpose, the {cpp} standard assumes that a comparator with an inner type named `is_transparent` is heterogeneous. The type itself does not matter.
```
loic-joly-sonarsource, line 121: The comparator is not heterogeneous; the container is.
```suggestion
A comparator needs to explicitly declare itself as transparent. This is achieved by having and inner type named `is_transparent` in the comparator. The type itself does not matter.
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
+
+Furthermore, the comparator needs to provide heterogeneous comparisons. There are multiple ways to achieve this:
+
+. Its `operator()` could be templated, like `std::less<void>` does.
+. It could provide the relevant overloads for the software.
+. Or it can provide one overload that handles any kind of key.
```
loic-joly-sonarsource, line 127: I don't get this one.

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
 
-void g() {
-  // UpToTenLess is now transparent
-  std::set<std::string, UpToTenLess> m = { "Dory", "Marlin", "Nemo", "Emo"};
-  m.find("Nemo"); // std::string_view is created out of raw C-string literal
-  std::string_view n{"Nemo"};
-  m.find(n); // No need to create a std::string
-}
-----
+Transparent comparators/hashers are strongly discouraged when used with types that are not directly comparable as it will lead to the creation of ``++O(log(container.size()))++`` temporaries with lookup functions such as ``++find++``, ``++count++``, ``++lower_bound++``, ...
```
loic-joly-sonarsource, line 201: Directly add reference to S6021, in which case, we don't really need to explain too much here.

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,228 @@
...
+However, like `std::less<std::string>`, `CaseInsensitiveCompare` is not transparent, and the code contains undesired temporary strings.
 
-[source,cpp]
+==== Compliant solution
+
+A comparator needs to explicitly declare itself as transparent. For this purpose, the {cpp} standard assumes that a comparator with an inner type named `is_transparent` is heterogeneous. The type itself does not matter.
+
+Furthermore, the comparator needs to provide heterogeneous comparisons. There are multiple ways to achieve this:
+
+. Its `operator()` could be templated, like `std::less<void>` does.
```
marco-antognini-sonarsource, line 125:
```suggestion
. Its `operator()` could be templated and written in a generic way, like `std::less<void>` does.
```

rules/S6045/cfamily/rule.adoc
marco-antognini-sonarsource: The last one? I intended to explain you can have one overload that supports comparing any relevant type. I guess I should mention that the keys need to be converted to the operator parameter type for the sentence to make sense...

What about this, combined with the above suggestion?

```diff
-. Its `operator()` could be templated, like `std::less<void>` does.
+. Its `operator()` could be templated and written in a generic way, like `std::less<void>` does.
 . It could provide the relevant overloads for the software.
```

```suggestion
. It can provide one overload whose lightweight parameter type can be constructed from key-like types.

With the first and third options, the implementation is reusing existing functions and types by delegating the comparison operation to them.
```

rules/S6045/cfamily/rule.adoc
marco-antognini-sonarsource: Once we migrate S6021 to LaYC format, why not. For now, I prefer linking to it in the Ressource part.

rules/S6045/cfamily/rule.adoc
loic-joly-sonarsource: would support comparing a `std::string` with string-like types (such as ``++char const*++`` or  ``++std::string_view++``).

rules/S6045/cfamily/rule.adoc
loic-joly-sonarsource: An excellent and very common example

rules/S6045/cfamily/rule.adoc
marco-antognini-sonarsource: Rewritten in https://github.com/SonarSource/rspec/pull/3797/commits/14cd49763f. Let me know what you think.

rules/S6045/cfamily/rule.adoc
marco-antognini-sonarsource: I've merged the change in 04a666beb44ee018e3acaca3143cb19dbbd7adaf. Let me know if you have more feedback.

marco-antognini-sonarsource (commented):
@loic-joly-sonarsource I've tried to address your comment.

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,234 @@
...
+
+Furthermore, the comparator needs to provide heterogeneous comparisons. There are multiple ways to achieve this:
+
+. Its `operator()` could be templated and written in a generic way, like `std::less<void>` does.
+. It could provide the relevant overloads for the software.
+. It can provide one overload whose lightweight parameter type can be constructed from key-like types.
+
+With the first and third options, the implementation is reusing existing functions and types by delegating the comparison operation to them.
+
+When working with `std::string`, the last option may often be the simplest thanks for `std::string_view`.
```
loic-joly-sonarsource, line 135: I think this sentence has become redundant with the next one.
```suggestion
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,234 @@
...
+
+For instance, to implement a transparent case-insensitive comparator for strings, we can leverage the fact that ``++string_view++`` offers lightweight conversions. Furthermore, this class provides the same functionality we need from `std::string`, which allows us to update `CaseInsensitiveCompare` by changing two lines:
```
loic-joly-sonarsource, line 137: I don't understand this part: 
> Furthermore, this class provides the same functionality we need from `std::string`

Maybe:
```suggestion
For instance, to implement a transparent case-insensitive comparator for strings, we can leverage the fact that ``++string_view++`` offers lightweight conversions. Furthermore, since ``++string_view++`` and `string` have very similar interfaces, the implementation of the comparator can remain unchanged:
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,234 @@
...
+
+A comparator needs to explicitly declare itself as transparent. This is achieved by having an inner type named `is_transparent` in the comparator. The type itself does not matter.
+
+Furthermore, the comparator needs to provide heterogeneous comparisons. There are multiple ways to achieve this:
+
+. Its `operator()` could be templated and written in a generic way, like `std::less<void>` does.
+. It could provide the relevant overloads for the software.
+. It can provide one overload whose lightweight parameter type can be constructed from key-like types.
+
+With the first and third options, the implementation is reusing existing functions and types by delegating the comparison operation to them.
```
loic-joly-sonarsource, line 133: I'm not sure I like this sentence: 
- The 2nd version may also delegate to other functions
- I don't think it really add value...

Could it just be dropped?

```suggestion
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,234 @@
...
 
-void g() {
-  // UpToTenLess is now transparent
-  std::set<std::string, UpToTenLess> m = { "Dory", "Marlin", "Nemo", "Emo"};
-  m.find("Nemo"); // std::string_view is created out of raw C-string literal
-  std::string_view n{"Nemo"};
-  m.find(n); // No need to create a std::string
-}
-----
+Transparent comparators/hashers are strongly discouraged when used with types that are not directly comparable or expensive to create. For example, in the example above, if `CaseInsensitiveCompare` had the inner type ``++is_transparent++`` but its `operator()` still had ``std::string`` arguments, each comparison inside ``++lower_bound++`` would create a temporary `std::string`. This would be significantly worse than the original version, which creates only one temporary object.
```
loic-joly-sonarsource, line 209:
```suggestion
Transparent comparators/hashers are strongly discouraged when used with types that are not directly comparable or expensive to create. For example, in the example above, if `CaseInsensitiveCompare` had the inner type ``++is_transparent++`` but its `operator()` still had ``std::string`` arguments, each internal comparison performed by ``++lower_bound++`` would create a temporary `std::string`. This would be significantly worse than the original version, which creates only one temporary object.
```

loic-joly-sonarsource (approved):
I still made a few suggestions, but you can directly merge after you either applied or disregarded them, they are pretty small now.

rules/S6045/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks, it's easier to read!

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,232 @@
...
-  m.find("Nemo"); // std::string_view is created out of raw C-string literal
-  std::string_view n{"Nemo"};
-  m.find(n); // No need to create a std::string
-}
-----
+Transparent comparators/hashers are strongly discouraged when used with types that are not directly comparable or expensive to create. For example, in the example above, if `CaseInsensitiveCompare` had the inner type ``++is_transparent++`` but its `operator()` still had ``std::string`` arguments, each internal comparison performed by ``++lower_bound++`` would create a temporary `std::string`. This would be significantly worse than the original version, which creates only one temporary object.
+
+The standard comparator types existed before {cpp}14, but their type parameter had to be provided. {cpp}14 introduced a default value for the template parameter, `void`, alongside a template specialization that is the transparent comparator. Therefore, when creating an object of such type, it is essential to write ``++std::less<>++`` and not ``++std::less<std::string>++``, for example.
 
+---
```
marco-antognini-sonarsource, line 211: Leftover
```suggestion
```

rules/S6045/cfamily/rule.adoc
```diff
@@ -1,118 +1,231 @@
...
+
+A comparator needs to explicitly declare itself as transparent. This is achieved by having an inner type named `is_transparent` in the comparator. The type itself does not matter.
+
+Furthermore, the comparator needs to provide heterogeneous comparisons. There are multiple ways to achieve this:
+
+. Its `operator()` could be templated and written in a generic way, like `std::less<void>` does.
+. It could provide the relevant overloads for the software.
+. It can provide one overload whose lightweight parameter type can be constructed from key-like types.
+
+
```
marco-antognini-sonarsource, line 133:
```suggestion
```

PR #3805, opened 2024-03-21
Modify S6432: Add conciseness and clear actionability to the message

loris-s-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6432/message.adoc
+++ b/rules/S6432/message.adoc
@@ -1,7 +1,7 @@
 === Message
 
 * Primary location
-** Use a dynamically-generated initialization vector (IV) to avoid IV-key pair reuse.
+** Use a random number generator as initialization vector, instead of hardcoded values.
 
 * Secondary location
 ** The initialization vector is configured here.


rules/S6432/message.adoc
```diff
@@ -1,8 +1,8 @@
 === Message
 
 * Primary location
-** Use a dynamically-generated initialization vector (IV) to avoid IV-key pair reuse.
+** Use a random number generator as initialization vector, instead of hardcoded values.
```
daniel-teuchert-sonarsource, line 4:
```suggestion
** Use a dynamically-generated initialization vector (IV) instead of hardcoded values.
```
I would change the first part of this sentence since you are not using the generator as the IV but rather the generated values.

daniel-teuchert-sonarsource (changes requested):
Some change necessary

rules/S6432/message.adoc
loris-s-sonarsource: I initially added this somehow incorrect sentence because I think there's a possibility of users "dynamically-generating" IVs and still hardcoding them. This might sound far-fetched but I can see a probability of users who do not really understand who would use a third party randomly generated value to generate the IV, and then hardcoding it.

I wanted people to understand that the RNG should not have any intermediary between rng() and encrypt(iv, etc.).

With the new version, we talk about hardcoded values so that's an improvement already, let's start with this small improvement and see afterwards

daniel-teuchert-sonarsource (approved):
LGTM!

PR #3816, opened 2024-03-26
Modify rules S6185,S6484,S6494,S6495: RSPEC improvements (CPP-5056)

marco-antognini-sonarsource:
NB: to make Grammarly work a bit better, I ensured sentences start and end on the same line. We are not consistent (sometimes the sentence is broken over multiple lines, sometimes a whole paragraph is on one line). I did not attempt to fix this inconsistency.

NB2: Link validation fails in a JS/TS rule.

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6185/cfamily/rule.adoc
+++ b/rules/S6185/cfamily/rule.adoc
@@ -38,6 +38,10 @@ std::string greeting(int n) {
 * {cpp} reference - https://en.cppreference.com/w/cpp/utility/format/format[`std::format`]
 * {cpp} reference - https://en.cppreference.com/w/cpp/string/basic_string/to_string[``++std::to_string++``]
 
+=== Articles & blog posts
+
+* {cpp} Stories - https://www.cppstories.com/2022/custom-stdformat-cpp20/[Formatting Custom types with std::format from C++20]
+
 === Related rules
 
 * S6484 - Concatenated "std::format" outputs should be replaced by a single invocation
--- a/rules/S6185/cfamily/rule.adoc
+++ b/rules/S6185/cfamily/rule.adoc
@@ -31,3 +31,9 @@ std::string greeting(int n) {
 }
 ----
 
+== Resources
+
+=== Documentation
+
+* {cpp} reference - https://en.cppreference.com/w/cpp/utility/format/format[`std::format`]
+* {cpp} reference - https://en.cppreference.com/w/cpp/string/basic_string/to_string[``++std::to_string++``]
--- a/rules/S6185/cfamily/rule.adoc
+++ b/rules/S6185/cfamily/rule.adoc
@@ -6,7 +6,7 @@
 Before {cpp}20, one popular way to obtain the same result was the conversion of the values with ``++std::to_string++`` and piecewise string concatenation.
 
 
-``++std::format++`` is strictly superior. It is more efficient because it constructs the string in place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
+``++std::format++`` is strictly superior. It is more efficient because it constructs the string in-place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
 
 
 This rule reports string concatenation cases that can be replaced by ``++std::format++`` to improve performance and readability.
--- a/rules/S6185/cfamily/rule.adoc
+++ b/rules/S6185/cfamily/rule.adoc
@@ -40,10 +40,10 @@ std::string greeting(int n) {
 
 === Articles & blog posts
 
-* {cpp} Stories - https://www.cppstories.com/2022/custom-stdformat-cpp20/[Formatting Custom types with std::format from C++20]
+* {cpp} Stories - https://www.cppstories.com/2022/custom-stdformat-cpp20/[Formatting Custom types with std::format from {cpp}20]
 
 === Related rules
 
 * S6484 - Concatenated "std::format" outputs should be replaced by a single invocation
-* S6494 - C++ formatting functions should be used instead of C printf-like functions
+* S6494 - {cpp} formatting functions should be used instead of C printf-like functions
 * S6495 - "std::format" should be used instead of standard output manipulators
--- a/rules/S6185/cfamily/rule.adoc
+++ b/rules/S6185/cfamily/rule.adoc
@@ -6,10 +6,10 @@
 Before {cpp}20, one popular way to obtain the same result was the conversion of the values with ``++std::to_string++`` and piecewise string concatenation.
 
 
-``++std::format++`` is strictly superior. It is more efficient because it constructs the string in-place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
+``++std::format++`` is strictly superior. It is more efficient because it constructs the string in place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
 
 
-This rule reports string concatenation cases that can be replaced by ``++std::format++`` and gain in speed and readability.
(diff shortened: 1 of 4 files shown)


rules/S6484/cfamily/rule.adoc
```diff
@@ -1,12 +1,12 @@
 == Why is this an issue?
 
-`std::format` accepts a format string composed of ordinary text and replacement fields (surrounded with `{}`) that are replaced with a textual representation of the next `std::format` arguments.
+`std::format` accepts a format string composed of ordinary text and replacement fields (surrounded with `{}`) that are replaced with a textual representation of the following `std::format` arguments.
```
loic-joly-sonarsource, line 3: For me, "the following" means you're going to describe them in the next sentence...
```suggestion
`std::format` accepts a format string composed of ordinary text and replacement fields (surrounded with `{}`) that are replaced with a textual representation of the remaining `std::format` arguments.
```

rules/S6494/cfamily/rule.adoc
```diff
@@ -2,13 +2,11 @@
...
-* The `std::print` interface family ({cpp}23) allows printing formatted text
+* The `std::format` interface family ({cpp}20) allows formatting text into a string.
+* The `std::print` interface family ({cpp}23) allows printing formatted text.
 
-Firstly, {cpp} formatting facilities perform validation of the format string against the type
-of the formatted argument. If the validation fails, it is reported as a compilation error
-for the calls of `std::print` and `std::format`.
-In addition `std::vformat`, `std::vprint_unicode`, and `std::vprint_nonunicode` report validation failures by throwing an instance of `std::format_error`.
+Firstly, {cpp} formatting facilities perform validation of the format string against the type of the formatted argument. If the validation fails, it is reported as a compilation error for the calls of `std::print` and `std::format`.
+In addition, `std::vformat`, `std::vprint_unicode`, and `std::vprint_nonunicode` report validation failures by throwing an instance of `std::format_error`.
```
loic-joly-sonarsource, line 9: Not sure of the "Firstly" "In addition" here. 
```suggestion
{cpp} formatting facilities perform validation of the format string against the type of the formatted argument. If the validation fails, it is reported as a compilation error for the calls of `std::print` and `std::format`. `std::vformat`, `std::vprint_unicode`, and `std::vprint_nonunicode`, that can't validate at compile-time, report failures by throwing an instance of `std::format_error`.
```

rules/S6494/cfamily/rule.adoc
```diff
@@ -97,3 +92,13 @@ which may not be actionable.
 * {cpp} reference - https://en.cppreference.com/w/cpp/io/vprint_nonunicode[`std::vprint_nonunicode`]
 * {cpp} reference - https://en.cppreference.com/w/cpp/io/basic_ostream/vprint_unicode[`std::vprint_unicode(std::ostream)`]
 * {cpp} reference - https://en.cppreference.com/w/cpp/io/basic_ostream/vprint_nonunicode[`std::vprint_nonunicode(std::ostream)`]
```
loic-joly-sonarsource, line 94: Why only those 3? They don't seem the most relevant...
Maybe a global link to https://en.cppreference.com/w/cpp/utility/format

loic-joly-sonarsource (changes requested):

rules/S6494/cfamily/rule.adoc
marco-antognini-sonarsource: Sounds better, indeed. One nit though: "that can't validate at at compile-time" --> it's a design choice. I would rephrase it a bit:

```suggestion
{cpp} formatting facilities perform validation of the format string against the type of the formatted argument. If the validation fails, it is reported as a compilation error for the calls of `std::print` and `std::format`. When the format string is not available at compile-time, `std::vformat`, `std::vprint_unicode`, and `std::vprint_nonunicode` can be used. They will report failures at runtime by throwing an instance of `std::format_error`.
```

WDYT?

rules/S6494/cfamily/rule.adoc
marco-antognini-sonarsource: The last two are from `<ostream>` and are not available from the page you linked.
I'm not sure how important that is, though. Maybe we should have fewer links...

What about a link to:
 * std::format
 * std::print
 * std::print(ostream)
 * <format>
 * <print>

and we "ignore" the rest here, assuming the reader can navigate to them?

rules/S6494/cfamily/rule.adoc
loic-joly-sonarsource: LGTM

rules/S6185/cfamily/rule.adoc
```diff
@@ -6,10 +6,10 @@
 Before {cpp}20, one popular way to obtain the same result was the conversion of the values with ``++std::to_string++`` and piecewise string concatenation.
 
 
-``++std::format++`` is strictly superior. It is more efficient because it constructs the string in-place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
+``++std::format++`` is strictly superior. It is more efficient because it constructs the string in place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
```
marco-antognini-sonarsource, line 9: It was correct before...
```suggestion
``++std::format++`` is strictly superior. It is more efficient because it constructs the string in-place instead of copying substrings one by one. It is also often shorter and easier to read because the format pattern is presented in a single piece and not scattered across the concatenation expression.
```

rules/S6494/cfamily/rule.adoc
loic-joly-sonarsource: I think we need `format_to`, because we use it so much in the example... (or should we change the example? For instance by returning a string?
I'm not sure of the value of having both  `std::format` and `<format>`, it's easy to navigate from one to thee other.

marco-antognini-sonarsource: - links: std::print(ostream) `<format>` `<print>`
- update example: keep same interface with format_to_n, + alternative with format

marco-antognini-sonarsource: 1. done.
2. re-reading the examples, I found more issues. I split them into smaller functions and gave one example of redesign.

rules/S6494/cfamily/rule.adoc
```diff
@@ -35,43 +31,46 @@ This rule raises issues for calls of the `printf`, `fprintf`, `sprintf` and `snp
 
 [source,cpp]
 ----
-void printFunc(char* out, size_t n) {
-  sprintf(out, "%u %s", 10u, “text”); // Noncompliant
+void printTextIntoBuffer(char* out) {
```
loic-joly-sonarsource, line 34:
```suggestion
void printTextIntoBuffer(char* out) { // Assumes the buffer pointed-to by out is large enough
```

rules/S6494/cfamily/rule.adoc
```diff
@@ -35,43 +31,46 @@ This rule raises issues for calls of the `printf`, `fprintf`, `sprintf` and `snp
...
 ----
-void printFunc(char* out) {
-  std::format_to(out, “{:d} {:s}”, 10u, text); // Compliant
-  // or
-  std::format_to(out, “{} {}”, 10u, text); // Compliant
-
-  std::format_to_n(out, ”{:d} % {:G}”, 10, 10.0L); // Compliant
-  // or
-  std::format_to_n(out, ”{} % {:G}”, 10, 10.0L); // Compliant
+void printTextIntoBuffer(char* out) {
```
loic-joly-sonarsource, line 52:
```suggestion
void printTextIntoBuffer(char* out) { // Assumes the buffer pointed-to by out is large enough
```

loic-joly-sonarsource (approved):
The examples are much better now.

rules/S6494/cfamily/rule.adoc
marco-antognini-sonarsource: Makes sense. I've added the comment on a separate line to keep them short. ad02ba96e8

PR #3824, opened 2024-03-26
Modify rules S5275,S3630: Briefly mention problems with implicit lifetime (CPP-5022)

marco-antognini-sonarsource:
I intentionally kept it very short -- I fear the alternative would be a full-size blog post. Let me know what you think.

## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3630/cfamily/rule.adoc
+++ b/rules/S3630/cfamily/rule.adoc
@@ -6,9 +6,7 @@ In some cases, ``++reinterpret_cast++`` can be simply replaced by a more focused
 
 If the goal is to access the binary representation of an object, ``++reinterpret_cast++`` leads to undefined behavior. Since {cpp}20, the best option is to use ``++std::bit_cast++``: it allows reinterpreting a value as being of a different type of the same length, preserving its binary representation. Before {cpp}20, `std::memcpy` must be used. See S6181 for more details.
 
-S5275 covers specific scenarios with ``++reinterpret_cast++`` that lead to undefined behavior.
-
-This rule raises an issue when ``++reinterpret_cast++`` is used.
+This rule raises an issue whenever ``++reinterpret_cast++`` is used. S5275 also covers ``++reinterpret_cast++`` but focuses on scenarios leading to undefined behavior.
 
 === Noncompliant code example
 
--- a/rules/S3630/cfamily/rule.adoc
+++ b/rules/S3630/cfamily/rule.adoc
@@ -6,6 +6,8 @@ In some cases, ``++reinterpret_cast++`` can be simply replaced by a more focused
 
 If the goal is to access the binary representation of an object, ``++reinterpret_cast++`` leads to undefined behavior. Since {cpp}20, the best option is to use ``++std::bit_cast++``: it allows reinterpreting a value as being of a different type of the same length, preserving its binary representation. Before {cpp}20, `std::memcpy` must be used. See S6181 for more details.
 
+S5275 covers specific scenarios with ``++reinterpret_cast++`` that lead to undefined behavior.
+
 This rule raises an issue when ``++reinterpret_cast++`` is used.
 
 === Noncompliant code example
--- a/rules/S3630/cfamily/rule.adoc
+++ b/rules/S3630/cfamily/rule.adoc
@@ -1,14 +1,12 @@
 == Why is this an issue?
 
-Because ``++reinterpret_cast++`` does not perform any type safety validations, it is capable of performing dangerous conversions between unrelated types, often leading to undefined behavior.
+Because ``++reinterpret_cast++`` does not perform any type safety validations, it can perform dangerous conversions between unrelated types, often leading to undefined behavior.
 
-In some cases, `reinterpret_cast` can be simply replaced by a more focused cast, such as `static_cast`.
+In some cases, ``++reinterpret_cast++`` can be simply replaced by a more focused cast, such as ``++static_cast++`` or ``++dynamic_cast++``.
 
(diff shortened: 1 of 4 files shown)


rules/S3630/see.adoc
loic-joly-sonarsource, line 1: ❤️

rules/S5275/cfamily/rule.adoc
```diff
@@ -1,12 +1,14 @@
 == Why is this an issue?
 
-Because ``++reinterpret_cast++`` ignores the type system, it is capable of performing dangerous conversions between unrelated types which can lead to undefined behavior.
+Because ``++reinterpret_cast++`` ignores the type system, it can perform dangerous conversions between unrelated types, leading to undefined behavior.
```
loic-joly-sonarsource, line 3: I think it is slightly unclear in the new wording what lead to UB. It could be "reinterpret_cast" (in which case the UB does not always happen) or it could be the "dangerous conversions" (in which case the wording is right)
If you don't like: "which could lead", what about:
```suggestion
Because ``++reinterpret_cast++`` ignores the type system, it can perform dangerous conversions between unrelated types that are undefined behavior.
```
Or, as in the previous rule:
```suggestion
Because ``++reinterpret_cast++`` ignores the type system, it can perform dangerous conversions between unrelated types, often leading to undefined behavior.
```

rules/S5275/cfamily/rule.adoc
```diff
@@ -1,12 +1,14 @@
...
 
-* when it is used to make the compiler believe that an object in memory is from a different type from its real type (for instance, casting a ``++long*++`` to ``++double*++``, because accessing a ``++long++`` as if it was a ``++double++`` is undefined behavior (even if ``++sizeof(long) == sizeof(double)++``),
-* when it is used to cast between different levels of a complex inheritance hierarchy (a ``++static_cast++`` would apply pointer offsets to take into account multiple inheritance, for instance, but ``++reinterpret_cast++`` does not)
+* When it is used to cast between different levels of a complex inheritance hierarchy. A ``++static_cast++`` would apply pointer offsets to account for multiple inheritances, but ``++reinterpret_cast++`` does not.
+
+* When the cast breaks the rules for implicit object lifetimes since this has undefined behavior. This happens when casting to a type, for example, with a user-provided destructor or no trivial default constructor because ``++reinterpret_cast++`` does not execute any code.
```
loic-joly-sonarsource, line 11: I think the text is hard to read because it mixes a description of the detected pattern and a justification of why this pattern is detected.
```
This rule reports an issue for these problematic uses of ``++reinterpret_cast++``, which are all undefined behavior:

* When it is used to make the compiler believe that an object in memory has a different type from its actual type, such as casting a ``++long*++`` to ``++double*++``, even if their sizes in memory are the same (``++sizeof(long) == sizeof(double)++``).
* When it is used to cast between different levels of a complex inheritance hierarchy. A ``++static_cast++`` would apply pointer offsets to account for multiple inheritances, but ``++reinterpret_cast++`` does not.
* When the cast breaks the rules for implicit object lifetimes. For example,  this happens when casting to a type with a user-provided destructor or no trivial default constructor because ``++reinterpret_cast++`` does not execute any code.
```

rules/S3630/cfamily/rule.adoc
```diff
@@ -1,26 +1,26 @@
...
 
-In some cases, `reinterpret_cast` can be simply replaced by a more focused cast, such as `static_cast`.
+In some cases, ``++reinterpret_cast++`` can be simply replaced by a more focused cast, such as ``++static_cast++`` or ``++dynamic_cast++``.
 
-If the goal is to access the binary representation of an object, `reinterpret_cast` leads to undefined behavior. Before {cpp}20, the correct way is to use `memcpy` to copy the object's bits. Since {cpp}20, a better option is available: ``++std::bit_cast++`` allows to reinterpret a value as being of a different type of the same length preserving its binary representation (see also S6181).
+If the goal is to access the binary representation of an object, ``++reinterpret_cast++`` leads to undefined behavior. Since {cpp}20, the best option is to use ``++std::bit_cast++``: it allows reinterpreting a value as being of a different type of the same length, preserving its binary representation. Before {cpp}20, `std::memcpy` must be used. See S6181 for more details.
 
+S5275 covers specific scenarios with ``++reinterpret_cast++`` that lead to undefined behavior.
```
loic-joly-sonarsource, line 9: Usually, when we have 2 rules that cover mostly the same cases, we try to provide our users with advice on which one(s) to activate.

marco-antognini-sonarsource: I wasn't sure how to formulate this here, given that both are in Sonar Way, and one is a strict subset of the other.

Maybe there is more to do than mention this in the RSPEC -- maybe they should be non-overlapping sets?

Let's have a quick chat about this.

rules/S5275/cfamily/rule.adoc
marco-antognini-sonarsource: Thanks, I like your suggestion very much!

marco-antognini-sonarsource: Done in f3076cc361

rules/S3630/cfamily/rule.adoc
marco-antognini-sonarsource: I've updated the description as we discussed. But the more I think about it, the more I feel the two rules should not interect. If you agree, I'll raise a ticket to improve this separately. Regardless, let me know if the current wording is okay in the current context.

marco-antognini-sonarsource:
Moving back to draft to avoid merging this by mistake.

rules/S3630/cfamily/rule.adoc
loic-joly-sonarsource: Agree to open another ticket. Maybe the solution is to merge the rules? This is just one more instance of a recurring issue we have with closely related rules. Maybe we should try to find a higher-level solution...

The current working is good enough I think.

marco-antognini-sonarsource: [CPP-5202](https://sonarsource.atlassian.net/browse/CPP-5202)

PR #3828, opened 2024-03-27
Modify rule S3470: Clarify that we allow std class specialization

frederic-tingaud-sonarsource:
## Review

- [ ] text style and tone


rules/S3470/cfamily/rule.adoc
```diff
@@ -12,17 +12,18 @@ ____
...
 ____
 
-You may think that it's legitimate to reopen ``++std++`` to define a version of extension points (``++std::swap++``, ``++std::hash++``...) that work with your types, but it's not necessary:  If you call these extension points according to the correct pattern, the user-defined version will be found too.
+You may think that it's legitimate to reopen `std` to define a version of extension points (``++std::swap++``, ``++std::hash++``...) that work with your types, but it's not necessary:  If you call these extension points according to the correct pattern, the user-defined version will be found too.
 
+One exception that is allowed by the standard is the specialization of standard class templates. In that case, the specialization must respect the requirement of the original template and has to be for a "program-defined type" (a type that is specific to the program, by opposition to a type from the standard). That is for example the recommended approach to customize ``++std::out_ptr++``.
```
loic-joly-sonarsource, line 24: I think the structure of the text was hard to follow.
Additionally, mention `inout_ptr` too, I think it can be specialized too.
```suggestion
However, the standard allows specializing standard class templates in namespace `std`. In that case, the specialization must respect the requirement of the original template and has to be for a "program-defined type" (a type that is specific to the program, by opposition to a type from the standard).

You may therefore think that it's legitimate to reopen `std` to define a version of extension points (``++std::swap++``, ``++std::hash++``...) that work with your types, but it's not necessary:  If you call these extension points according to the correct pattern, the user-defined version will be found too.

The only extension points for which the specialization is the recommended approach are ``++std::out_ptr++`` and ``++std::inout_ptr++``.
```

PR #3833, opened 2024-03-28
Modify S2076(python): Improve code samples

loris-s-sonarsource:
# Why
For Python particularly, users like to take our code samples and test them. Our analyzers will typically not raise an issue on the code samples we showed here and this PR aims to solve this.

### DIFF
--- a/rules/S2076/python/how-to-fix-it/paramiko.adoc
+++ b/rules/S2076/python/how-to-fix-it/paramiko.adoc
@@ -12,7 +12,7 @@ shell control characters, the expected `ping` command behavior will be changed.
 [source,python,diff-id=1,diff-type=noncompliant]
 ----
 from flask import Flask, request
-from paramiko import SSHClient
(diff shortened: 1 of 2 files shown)


rules/S2076/python/how-to-fix-it/paramiko.adoc
```diff
@@ -11,22 +11,37 @@ shell control characters, the expected `ping` command behavior will be changed.
 
 [source,python,diff-id=1,diff-type=noncompliant]
 ----
-client = SSHClient()
-client.connect("example.org", username=USER, password=PASS)
-client.exec_command(request.args.get("cmd")) # Noncompliant
+from flask import Flask, request
+from paramiko import SSHClient
```
loris-s-sonarsource, line 15:
```suggestion
from paramiko.client import SSHClient
```

daniel-teuchert-sonarsource (approved):
LGTM!

PR #3834, opened 2024-03-28
Make sure a label is added to each PR

loic-joly-sonarsource:


.github/workflows/update_coverage.yml
```diff
@@ -1,7 +1,7 @@
 name: Update rule coverage
 on:
   schedule:
-    - cron: '17 2 * * *'
+    - cron: '17 0 * * *'
```
frederic-tingaud-sonarsource, line 4: This change shouldn't be here

.github/workflows/ensure_label.yml
```diff
@@ -0,0 +1,20 @@
+name: PR should have a language label
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+
+jobs:
+  label:
+    runs-on: ubuntu-20.04
```
frederic-tingaud-sonarsource, line 8: Most of our workflows use ubuntu-latest and I think this one would benefit from it.
```suggestion
    runs-on: ubuntu-latest
```

frederic-tingaud-sonarsource (changes requested):

.github/workflows/update_coverage.yml
loic-joly-sonarsource: Agreed... Not sure how it appeared...

frederic-tingaud-sonarsource (approved):

PR #3853, opened 2024-04-08
Fix Azure Resource Manager rules coverage

mstachniuk:
## Review

- [ ] text style and tone

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -56,7 +56,7 @@ frontend_tests_task:
     <<: *CONTAINER_DEFINITION
     dockerfile: ci/frontend-tests-dockerfile
     cpu: 1
-    memory: 2G
+    memory: 3G
   node_modules_cache:
     folder: frontend/node_modules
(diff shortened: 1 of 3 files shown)


necto (approved):
Please also set an existing language for the snippet highlighting in rspec-tools/rspec_tools/utils.py.
Right now it maps this "language" to "azureresourcemanager" highlight.js does not know. You can map it to any similar language that highlight.js does recognize, or `text` if you find no appropriate language.

mstachniuk:
1. It doesn't make sense to do such mapping rspec-tools? In our docs, for each code snipped we define the language, e.g. [source,json,diff-id=11,diff-type=noncompliant]
2. ARM can have a code written in JSON and Bicep format. The second has not support in highlight.js 
3. It is already defined: https://github.com/SonarSource/rspec/blob/master/rspec-tools/rspec_tools/utils.py#L79
I merged as it is as the additional change will not change anything in highlighting.

PR #3857, opened 2024-04-09
Modify rule S6398: Add How to fix it in Spring

daniel-teuchert-sonarsource:
This PR should be merged once https://sonarsource.atlassian.net/browse/SONARSEC-5033 is implemented.


rules/S6398/java/how-to-fix-it/spring.adoc
```diff
@@ -0,0 +1,44 @@
+== How to fix it in Spring
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+In this example the json String is parsed using the `BasicJsonParser` class from the `org.springframework.boot.json` package. This class is used to parse JSON strings into a `Map` object. Since the attacker can control the json string they can control the keys and values of the `Map` object.
```
gaetan-ferry-sonarsource, line 7: I think this reads strangely.

```suggestion
In this example a json string is parsed using the `BasicJsonParser` class from the `org.springframework.boot.json` package. This class is used to parse JSON strings into a `Map` object. Since the attacker can control the json string they can control the keys and values of the `Map` object.
```

```suggestion
In this example the `input` json string is parsed using the `BasicJsonParser` class from the `org.springframework.boot.json` package. This class is used to parse JSON strings into a `Map` object. Since the attacker can control the json string they can control the keys and values of the `Map` object.
```

gaetan-ferry-sonarsource (approved):
LGTM with a small improvement suggestion

PR #3862, opened 2024-04-10
Modify rules S6830, S1452: Update dead links

johann-beleites-sonarsource:


rules/S1452/java/rule.adoc
```diff
@@ -62,7 +62,7 @@ List<Lifeform> getLifeforms() { ... }       // Compliant, using supertype instea
 === Articles & blog posts
 
 * https://medium.com/javarevisited/variance-in-java-and-scala-63af925d21dc[Sinisa Louc - A Complete Guide to Variance in Java and Scala]
-* https://kotlinexpertise.com/kotlin-generics-and-variance-vs-java[Kotlin Expertise Blog - Kotlin Generics and Variance (Compared to Java)]
+* https://web.archive.org/web/20240206045705/https://kotlinexpertise.com/kotlin-generics-and-variance-vs-java/[Kotlin Expertise Blog - Kotlin Generics and Variance (Compared to Java)]
```
irina-batinic-sonarsource, line 65: The link is valid now, but do we really want to have it here?
It is a **Kotlin Expertise Blog** - Kotlin Generics and Variance (Compared to Java) and for Java devs, I don't think it is useful.
If we implement this rule for Kotlin, I think it would be more appropriate to put it there.
WDYT?

johann-beleites-sonarsource: I see where you're coming from. The article has several sections right at the start talking only about Java variance, explaining it pretty well. I know this article is not only about Java - we're also okay about having this article for Kotlin, even though it doesn't only talk about Kotlin. IMO, this article contains some valuable explanations for Java and I don't see the fact that it doesn't focus only on Java as a reason not to include it. (Also, I don't have a better alternative ready).

irina-batinic-sonarsource (approved):

PR #3897, opened 2024-04-18
Modify rule S3649: Add Couchbase example

hendrik-buchwald-sonarsource:
Implementation ticket: https://sonarsource.atlassian.net/browse/SONARSEC-5078

## Review

- [ ] text style and tone


rules/S3649/java/how-to-fix-it/couchbase.adoc
```diff
@@ -0,0 +1,63 @@
+== How to fix it in Couchbase
+
+=== Code examples
+
+The following code is vulnerable to SQL injection because user-controlled data
```
gaetan-ferry-sonarsource, line 5: Shouldn't this be SQL++ ? I guess it's OK to keep SQL.
```suggestion
The following code is vulnerable to SQL++ injection because user-controlled data
```

rules/S3649/java/how-to-fix-it/couchbase.adoc
```diff
@@ -0,0 +1,63 @@
...
+is inserted directly into a query string: The application assumes that incoming
```
gaetan-ferry-sonarsource, line 6:
```suggestion
is inserted directly into a query string. The application assumes that incoming
```

rules/S3649/java/how-to-fix-it/couchbase.adoc
```diff
@@ -0,0 +1,63 @@
...
+        return result.rowsAs(Person.class);
+    }
+}
+----
+
+==== Compliant solution
+
+[source,java,diff-id=31,diff-type=compliant]
+----
+import org.springframework.data.couchbase.core.CouchbaseTemplate;
```
gaetan-ferry-sonarsource, line 33: It might make sense to add the import for `Query` here. It wasn't obvious to me what Query it was here. Especially considering that Spring has multiple occurrences of Query types across packages.

```suggestion
org.springframework.data.couchbase.core.query.Query;
```

On a side note, I find it weird that this section is about couchbase but the solution uses spring code. I understand it considering we are working around Spring but for a user not relying on Spring it might not be obvious to adapt the fix.

gaetan-ferry-sonarsource (approved):
LGTM in general. I still proposed some improvement that you might want to consider. Or not.

rules/S3649/java/how-to-fix-it/couchbase.adoc
hendrik-buchwald-sonarsource: Good point

PR #3899, opened 2024-04-18
Modify rule S3740: added examples and explanation about "cast", "instanceof" and wildcards

kaufco:


leonardo-pilastri-sonarsource (approved):
LGTM

PR #3901, opened 2024-04-22
S2951: Rename

pavel-mikula-sonarsource:
Rules should not be prefixed with their language name. We generally don't do that, as it doesn't make sense.


sebastien-marichal (approved):
LGTM!

PR #3910, opened 2024-04-23
Deprecate S6803

mary-georgiou-sonarsource:
After discussing the impact of this rule we decided to deprecate it.

- The rule applies < .NET8
- It has a very low impact
- It applies only to sealed components


costin-zaharia-sonarsource (approved):
LGTM!

PR #3921, opened 2024-05-03
Modify rule S2529: Fix encoding issue in PL/SQL

sebastien-marichal:


rules/S2529/plsql/rule.adoc
```diff
@@ -1,6 +1,6 @@
 == Why is this an issue?
 
-Developers should use the ``++FROM ... OUTER JOIN++`` syntax rather than the Oracle join operator (``+``). The reason is that outer join queries that use ``+`` are subject to several restrictions which do not apply to the ``++FROM ... OUTER JOIN++`` syntax. For instance, a ``++WHERE++`` condition containing the ``+`` operator cannot be combined with another condition using the ``++OR++`` logical operator.
+Developers should use the `FROM ... OUTER JOIN` syntax rather than the Oracle join operator (`+`). The reason is that outer join queries that use `+` are subject to several restrictions which do not apply to the `FROM ... OUTER JOIN` syntax. For instance, a `WHERE` condition containing the `+` operator cannot be combined with another condition using the `OR` logical operator.
```
pavel-mikula-sonarsource, line 3: Both, the original encoding and the new one seems valid, so while this is an improvement, it's practically a no-op.

What version of rule-api did you use to generate the rspec in plsql?

pavel-mikula-sonarsource (approved):
LGTM

rules/S2529/plsql/rule.adoc
sebastien-marichal: I download the latest one from [repox](https://repox.jfrog.io/ui/native/sonarsource/com/sonarsource/rule-api/rule-api/): 2.8.0.3873

PR #3923, opened 2024-05-06
Modify S2365: Fix code examples

sebastien-marichal:
See https://github.com/SonarSource/sonar-dotnet/issues/9254

### DIFF
--- a/rules/S2365/csharp/rule.adoc
+++ b/rules/S2365/csharp/rule.adoc
@@ -9,11 +9,21 @@ include::../why-dotnet.adoc[]
 [source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 private List<string> _foo = new List<string> { "a", "b", "c" };
-public IEnumerable<string> Foo  // Noncompliant: expensive ToList call
-{ 
+private string[] _bar = new string[] { "a", "b", "c" };
+
+public IEnumerable<string> Foo  // Noncompliant: collection _foo is copied
+{
     get
     {
-        return (string[])_foo.Clone(); 
+        return _foo.ToList();
+    }
+}
+
+public IEnumerable<string> Bar // Noncompliant: array _bar is copied
+{
+    get
+    {
+        return (string[])_bar.Clone();
     }
 }
 ----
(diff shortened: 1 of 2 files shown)


rules/S2365/csharp/rule.adoc
```diff
@@ -9,11 +9,21 @@ include::../why-dotnet.adoc[]
 [source,csharp,diff-id=1,diff-type=noncompliant]
 ----
 private List<string> _foo = new List<string> { "a", "b", "c" };
```
CristianAmbrosini, line 11: (optional) What about adding using statements as well? Specifically the System.Linq
```cs
using System.Collections.Generic;
using System.Linq;
```

rules/S2365/csharp/rule.adoc
```diff
@@ -23,9 +33,16 @@ public IEnumerable<string> Foo  // Noncompliant: expensive ToList call
 [source,csharp,diff-id=1,diff-type=compliant]
 ----
 private List<string> _foo = new List<string> { "a", "b", "c" };
+private string[] _bar = new string[] { "a", "b", "c" };
```
CristianAmbrosini, line 36: I would take this opportunity to change the field names as well, they should not begin with an underscore IMO

rules/S2365/vbnet/rule.adoc
```diff
@@ -8,11 +8,18 @@ include::../why-dotnet.adoc[]
 
 [source,vbnet,diff-id=1,diff-type=noncompliant]
 ----
-Dim foo = {"a", "b", "c" }
-Property Foo() As String() ' Noncompliant
+Private _foo As New List(Of String) From {"a", "b", "c"}
```
CristianAmbrosini, line 11: Field names

CristianAmbrosini (changes requested):

rules/S2365/csharp/rule.adoc
sebastien-marichal: The example context is implicitly set to a class.
Adding usings would be incorrect. I would rather not add them.

CristianAmbrosini (approved):
LGTM!

PR #3928, opened 2024-05-10
Forbid adding a link to rules.sonarsource.com

frederic-tingaud-sonarsource:


rspec-tools/rspec_tools/validation/sanitize_asciidoc.py
```diff
@@ -5,6 +5,7 @@
 * Inline code with backquotes is correctly escaped and balanced
 * Include commands are not appended to other code
```
necto, line 6: Is this list up to date?

rspec-tools/rspec_tools/validation/sanitize_asciidoc.py
```diff
@@ -52,6 +53,8 @@
 
 CPP = re.compile(r"\b[Cc]\+\+")
 
+RULES_SONARSOURCE = re.compile(r"https:\/\/rules\.sonarsource\.com\/(.*)\/RSPEC-\d+")
```
necto, line 56: there are also
```
link://rules.sonarsource.com[rules]
http://rules.sonarsource.com[rules]
```
As of now, the only mentions of `rules.sonarsource.com` in the rules are in comments and links. Also, I see no reason to link to rules.sonarsource.com from the rule descriptions themselves. other than referring to other rules, and we have a dedicated mechanism for that.

Why not search for only the ``rules.sonarsource.com`` substring?

necto (approved):

rspec-tools/rspec_tools/validation/sanitize_asciidoc.py
frederic-tingaud-sonarsource: Indeed, it's not.

rspec-tools/rspec_tools/validation/sanitize_asciidoc.py
frederic-tingaud-sonarsource: I never saw the form `link://rules.sonarsource.com` is it an asciidoc specific format? It doesn't work in my VSCode preview.

I don't see a good reason to accept `rules.sonarsource.com` when it doesn't refer to a rule; I just wanted to make sure that we don't have false positives and I wouldn't have an alternative to propose if it blocks a PR.

necto: I haven't seen `link://...` either, but it worked on https://asciidoclive.com/.
I don't insist. It seems people rarely make the mistake of linking to `rules.sonarsoruce.com` directly, so this corner case will unlikely occur anyway.

PR #3930, opened 2024-05-15
Rule S6781: Remove redundant code from net-core snippet

costin-zaharia-sonarsource:
## Review

- [ ] text style and tone


rules/S6781/csharp/how-to-fix/net-core.adoc
```diff
@@ -54,10 +54,6 @@ public class LoginExampleController : ControllerBase
     [HttpPost]
     public IActionResult Post([FromBody] LoginModel login)
     {
-        // Code to validate the login information is omitted
-
-        var key = _config["Jwt:Key"] ??
-            throw new ApplicationException("JWT key is not configured.");
         var securityKey = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(key)); // Noncompliant
```
costin-zaharia-sonarsource, line 57: In this code snippet, the key is set at the top of the file. If we override it here, we're changing the meaning and the example will duplicate the previous one.

gregory-paidis-sonarsource: Why did you choose to delete the comment as well?
It seems to appear in all other snippets.

costin-zaharia-sonarsource: oh, the comment is fine, I'll add it back. Thanks!

PR #3933, opened 2024-05-15
Modify rule S5500: mention std::ranges::move and rewrite RSPEC (CPP-5219)

marco-antognini-sonarsource:
Companion PR: SonarSource/sonar-cpp#3622


marco-antognini-sonarsource:
validate_links fails for unrelated reasons.

loic-joly-sonarsource (approved):
I'll let you decide if you want to apply the suggestion

rules/S5500/cfamily/rule.adoc
```diff
@@ -74,11 +87,11 @@ public:
 };
 ----
 
-Alternatively, `insertAllShapes` could also be rewritten like this using {cpp}23:
+Alternatively, `insertAllVisibleShapes` could also be rewritten like this using {cpp}23:
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: I'm not sure I follow, there is already one above. Did you mean something else?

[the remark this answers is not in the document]
marco-antognini-sonarsource: @loic-joly-sonarsource ping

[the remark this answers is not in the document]
marco-antognini-sonarsource: @loic-joly-sonarsource I would say, if you have small suggestions, please share them. If you are not convinced with the approach or have larger suggestions, I'd rather not spend too much time on it and revert to the previous version for now. It can be improved later.

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
+
+==== Noncompliant code example
+
+When the parameter represents a single object you want to move, it is not sufficient to use `&&` after its type in the parameter list.
+
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+void DrawingStore::insertVisibleShape(Shape&& shape) {
+  if (shape.isVisible()) {
+    store.emplace_back(shape); // Noncompliant, call to std::move is expected.
```
loic-joly-sonarsource, line 90:
```suggestion
    store.emplace_back(shape); // Noncompliant, calls Shape's copy constructor
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
   }
-};
+}
 ----
 
+With the above implementation, the `Shape` object appended in `store` is created using ``Shape``'s _copy_ constructor.
```
loic-joly-sonarsource, line 95:
```suggestion
```
With the suggested modification in the comment, I think this no longer brings value.

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
 [source,cpp,diff-id=1,diff-type=compliant]
 ----
-class Shape {
-// ...
-public:
-  bool isVisible() const;
-};
+void DrawingStore::insertVisibleShape(Shape&& shape) {
+  if (shape.isVisible()) {
+    store.emplace_back(std::move(shape)); // Compliant
```
loic-joly-sonarsource, line 105:
```suggestion
    store.emplace_back(std::move(shape)); // Compliant, calls Shape's move constructor
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
+When you want to transfer the content of multiple objects into another container, it also makes sense to define the parameter as rvalue with `&&`.
+
+==== Noncompliant code example
+
+While the following code looks fine and compiles, it does actually _copy_ the elements. In fact, `shapes` is left unchanged.
+
+[source,cpp,diff-id=2,diff-type=noncompliant]
+----
+void DrawingStore::insertAllShapes(std::vector<Shape>&& shapes) {
+  for (auto& s : shapes) {
```
loic-joly-sonarsource, line 123: Should we use `auto &&` here?

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
+    if (s.isVisible()) {
+      store.emplace_back(std::move(s)); // Compliant
     }
   }
-};
+}
 ----
 
-Alternatively, `insertAllShapes` could also be rewritten like this using {cpp}23:
+// We purposely do not go into the details of "move-from" states and the fact that `shapes` has still the same number of elements while some of them are in this "moved-from" state.
```
loic-joly-sonarsource, line 146:
```suggestion
// We purposely do not go into the details of "moved-from" states and the fact that `shapes` has still the same number of elements while some of them are in this "moved-from" state.
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
+
+Notice that in this solution, the for-loop variable `s` remains an lvalue reference with a single `&`.
```
loic-joly-sonarsource, line 148: Adjust this if the previous loop is rewritten with `auto&&`

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
-
-    // Alternatively:
-    for (auto&& s : shapes | std::views::as_rvalue) {
-      if (s.isVisible()) {
-        store.emplace_back(std::move(s)); // Compliant
-      }
+void DrawingStore::insertAllVisibleShapes(std::vector<Shape>&& shapes) {
+  for (auto&& s : shapes | std::ranges::views::as_rvalue) {
+    if (s.isVisible()) {
+      store.emplace_back(std::move(s)); // Compliant
```
loic-joly-sonarsource, line 161:
```suggestion
      store.emplace_back(s); // Compliant
```
Or did I miss something?

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
+
+[source,cpp,diff-id=3,diff-type=compliant]
 ----
-Shape&& updateShape(Shape&& shape) {
-  /* ... */
-  // C++23: Implicit move, equivalent to `std::move(shape)`
-  // C++20: No move and ill-formed as Shape&& reference cannot bound to Shape&
-  return shape;
+void DrawingStore::insertAllShapes(std::vector<Shape>&& shapes) {
+  // Compliant: use "move" instead of "copy".
```
loic-joly-sonarsource, line 191:
```suggestion
  // Compliant: uses "move" instead of "copy".
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
 }
 ----
 
-Due to all of the above, this rule does not treat `return p` as an exception in {cpp} standard prior to {cpp}23,
-and requires the explicit move `return std::move(p)`.
+==== Noncompliant code example
 
-In contrast to returning local (stack) variables, named return value optimization (NRVO)
-does not apply to function parameters, so an explicit `std::move` call has no impact on optimizations.
+However, sometimes `std::ranges::move` cannot be used, for example when the transfer of container should be conditional.
```
loic-joly-sonarsource, line 198:
```suggestion
However, sometimes `std::ranges::move` cannot be used, for example when not all elements should be moved.
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
+  // Noncompliant: the elements of shapes are not moved.
+  std::ranges::copy_if(
+    shapes,
+    std::back_inserter(store),
+    &Shape::isVisible
+  );
+}
+----
+
+Again, the elements are copies instead of moved.
```
loic-joly-sonarsource, line 213:
```suggestion
Again, the elements are copied instead of being moved.
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
 
+==== Compliant solution
+
+While a solution using {cpp}20 exists, it is fairly verbose and error-prone.
```
loic-joly-sonarsource, line 217:
```suggestion
While a solution using {cpp}20 exists, based on `std::make_move_iterator`, it is fairly verbose and error-prone.
```
Without giving too many details, help user search for the right terms.

loic-joly-sonarsource (commented):
I like the rewrite a lot.

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: The type deduction will give `Shape&` in both case. I feel `auto&` is less misleading.  `Shape&&` does not compile (rvalue reference to type 'Shape' cannot bind to lvalue of type 'Shape').

Type deduction may be confusing here, so I'll spell the type `Shape&` instead.

marco-antognini-sonarsource:
```suggestion
  for (Shape& s : shapes) {
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
-    for (auto& s : shapes) {
-      if (s.isVisible()) {
-        store.emplace_back(std::move(s)); // Compliant
-      }
+As in the previous example, a call to `std::move` is required to fix the implementation:
+
+[source,cpp,diff-id=2,diff-type=compliant]
+----
+void DrawingStore::insertAllVisibleShapes(std::vector<Shape>&& shapes) {
+  for (auto& s : shapes) {
```
marco-antognini-sonarsource, line 138:
```suggestion
  for (Shape& s : shapes) {
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
   }
-};
+}
 ----
 
-Alternatively, `insertAllShapes` could also be rewritten like this using {cpp}23:
+// We purposely do not go into the details of "move-from" states and the fact that `shapes` has still the same number of elements while some of them are in this "moved-from" state.
+
+Notice that in this solution, the for-loop variable `s` remains an lvalue reference with a single `&`.
+Furthermore, writing ``++for (auto& s : std::move(shapes))++`` would not fix the issue because this call to `std::move` has no effect here.
```
marco-antognini-sonarsource, line 149:
```suggestion
Furthermore, writing ``++for (Shape& s : std::move(shapes))++`` would not fix the issue because this call to `std::move` has no effect here.
```

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
-      &Shape::isVisible
-    );
-
-    // Alternatively:
-    for (auto&& s : shapes | std::views::as_rvalue) {
-      if (s.isVisible()) {
-        store.emplace_back(std::move(s)); // Compliant
-      }
+void DrawingStore::insertAllVisibleShapes(std::vector<Shape>&& shapes) {
+  for (auto&& s : shapes | std::ranges::views::as_rvalue) {
```
marco-antognini-sonarsource, line 159:
```suggestion
  for (Shape&& s : shapes | std::ranges::views::as_rvalue) {
```

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: The rvalue `s` would not be moved. :-)

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: Makes sense. Slight adaptation:
```suggestion
While a solution based on ``++std::make_move_iterator++`` exists before {cpp}23, it is fairly verbose and error-prone.
```
I'm still not adding the link at the end in "Resources" though.

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: Ditto, I'm now using `Shape&` everywhere.

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: See my other comment in this review.

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: See my other comment in this review.

rules/S5500/cfamily/rule.adoc
loic-joly-sonarsource: In that case, what would you think of:
```
void DrawingStore::insertVisibleShape(Shape&& shape) { // Noncompliant, Shape is never moved
  if (shape.isVisible()) {
    store.emplace_back(shape); // Calls Shape's copy constructor.
  }
}
```

Which would be corrected as:
```
void DrawingStore::insertVisibleShape(Shape&& shape) { // Compliant
  if (shape.isVisible()) {
    store.emplace_back(std::move(shape)); // Calls Shape's move constructor.
  }
}
```
To me, the benefits are:
- As you mention, even if a call to std::move is expected, it could be called in other places
- I think it more clearly describes what actually happens in the code because:
   - Inline comments avoid having to refer to a place in the code in the next sentence
   - There is a symmetry in the inline comments that highlight the difference

If you still disagree, I'm not going to fight. The RSPEC is correct as it is. I just feel it could be improved.

rules/S5500/cfamily/rule.adoc
loic-joly-sonarsource: My point was exactly that with `auto&&,` some people would erroneously believe that it would be deduced to `Shape&&,` and it was an opportunity to open their eyes.
However, I agree with your change of removing deduction.

rules/S5500/cfamily/rule.adoc
loic-joly-sonarsource: Oops... Still not used to `as_rvalue`...

rules/S5500/cfamily/rule.adoc
```diff
@@ -4,143 +4,248 @@ Rvalue reference arguments allow the efficient transfer of the ownership of obje
...
 ----
 
-Alternatively, `insertAllShapes` could also be rewritten like this using {cpp}23:
+// We purposely do not go into the details of "moved-from" states and the fact that `shapes` has still the same number of elements while some of them are in this "moved-from" state.
+
+Notice that in this solution, the for-loop variable `s` remains an lvalue reference with a single `&`.
+Furthermore, writing ``++for (Shape& s : std::move(shapes))++`` would not fix the issue because this call to `std::move` has no effect here.
+The call to `std::move` has to be on `s`, not `shapes`.
+
+Thankfully, {cpp}23 offers a new feature that makes the code more regular: with ``++std::ranges::views::as_rvalue++``, it is now possible to make `s` an rvalue too, making the intent of the code clearer.
```
loic-joly-sonarsource, line 152: I think that with the current text, the exact benefits of `as_const` are unclear, because the issue is separated from its solution.

```suggestion
Writing ``++for (Shape& s : std::move(shapes))++`` would not fix the issue because this call to `std::move` has no effect here.
The call to `std::move` has to be on `s`, not `shapes`.

Notice that in this solution, the for-loop variable `s` remains an lvalue reference with a single `&`. In {cpp}23, it is possible to make it a rvalue too, with ``++std::ranges::views::as_rvalue++``, making the intent of the code clearer.
```

loic-joly-sonarsource (approved):
I'm still suggesting changes, but feel free to discard them if you disagree with them.

rules/S5500/cfamily/rule.adoc
marco-antognini-sonarsource: Your idea is not bad, but it would take me a bit of time to adapt and make other sections consistent.

It also has a downside: the diff highlighting will include the first line of the (non)compliant code because of the comment, but only the comment changes. Since the diff is on the whole line (and not on words), this will distract the readers a bit.

So I'll keep the current wording for now. We can always revisit this in the future.

PR #2626, opened 2023-07-25
Modify rule S6330: Default Queue encryption is now SSE-SQS (APPSEC-339)

sebastien-andrivet-sonarsource:
## Review

A dedicated reviewer checked the rule description successfully for:

- [x] logical errors and incorrect information
- [x] information gaps and missing content
- [x] text style and tone
- [x] PR summary and labels follow [the guidelines](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule)

PR #3941, opened 2024-05-22
Modify rule S6330: Default Queue encryption is now SSE-SQS (iac only)

petertrr:
Extracted from #2626 as the changes in sonar-iac have been implemented already


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #3942, opened 2024-05-22
Modify rule S6964: Update message and description

mary-georgiou-sonarsource:


rules/S6964/csharp/rule.adoc
zsolt-kolbay-sonarsource, line 1: There's one more thing that should be mentioned (which I should have added before merging the rule): the analyzer will only raise for Model classes that are in the same project as the Controller class that references them. This is due to a limitation of Roslyn (see [this issue](https://github.com/SonarSource/sonar-dotnet/issues/9243)), something that we can't handle right now, but it should be mentioned in the RSPEC.

zsolt-kolbay-sonarsource (approved):
LGTM. Left one comment that should be addressed.

PR #3949, opened 2024-05-30
Modify rule S1699: Improve rule description for CPP-5114

marco-antognini-sonarsource:


tomasz-kaminski-sonarsource (approved):
Looks good from the content perspective. I also like how it is organized.

PR #3956, opened 2024-06-03
Modify rule S6973: Adding PyTorch optimizers

joke1196:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6973/python/how-to-fix-it/pytorch.adoc
+++ b/rules/S6973/python/how-to-fix-it/pytorch.adoc
@@ -9,9 +9,9 @@ Specify the hyperparameters when instantiating the optimizer
 [source,python,diff-id=2,diff-type=noncompliant]
 ----
 from my_model import model
-from torch.optim import Adadelta
+from torch.optim import AdamW
 
-optimizer = Adadelta(model.parameters(), lr = 0.001) # Noncompliant : weight_decay is not specified, different values can change the behaviour of the optimizer significantly
+optimizer = AdamW(model.parameters(), lr = 0.001) # Noncompliant : weight_decay is not specified, different values can change the behaviour of the optimizer significantly
 ----
 
 ==== Compliant solution
@@ -19,7 +19,7 @@ optimizer = Adadelta(model.parameters(), lr = 0.001) # Noncompliant : weight_dec
 [source,python,diff-id=2,diff-type=compliant]
 ----
 from my_model import model
-from torch.optim import Adadelta
+from torch.optim import AdamW
 
-optimizer = Adadelta(model.parameters(), lr = 0.001, weight_decay = 0.003) # Compliant
+optimizer = AdamW(model.parameters(), lr = 0.001, weight_decay = 0.003) # Compliant
 ----
--- a/rules/S6973/python/how-to-fix-it/pytorch.adoc
+++ b/rules/S6973/python/how-to-fix-it/pytorch.adoc
@@ -1,0 +1,25 @@
+== How to fix it in PyTorch
+
+Specify the hyperparameters when instantiating the optimizer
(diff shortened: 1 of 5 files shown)


rules/S6973/python/how-to-fix-it/pytorch.adoc
```diff
@@ -0,0 +1,25 @@
...
+----
+
+==== Compliant solution
+
+[source,python,diff-id=2,diff-type=compliant]
+----
+from my_model import model
+from torch.optim import Adadelta
+
+optimizer = Adadelta(model.parameters(), lr = 0.001, weight_decay = 0.003) # Compliant
```
ghislainpiot, line 24: nit: We might want to switch the exemple to Adam/AdamW to make it more relatable to users ?

rules/S6973/python/rule.adoc
```diff
@@ -1,18 +1,16 @@
-This rule raises an issue when a Scikit-learn model is instantiated without specifying the important hyperparameters. 
-
-
+This rule raises an issue when a machine learning model or optimizer is instantiated without specifying the important hyperparameters.
```
ghislainpiot, line 1: I'm sorry I didn't catch this last time, but I don't think we should call it a "model," especially now that we are including Pytorch in the rule.

rules/S6973/python/rule.adoc
```diff
@@ -1,18 +1,16 @@
...
 
-When instantiating a Scikit-learn estimator, it will use default values for the hyperparameters that are not specified.
-Relying on the default values can lead to non-reproducible results across diffferent versions of the library. 
+When instantiating an estimator or an optimizer, it will use default values for the hyperparameters that are not specified.
+Relying on the default values can lead to non-reproducible results across different versions of the library. 
 
 Furthermore, the default values might not be the best choice for the specific problem at hand and can lead to suboptimal performance.
 
 Here are the estimators and the parameters considered by this rule : 
 [cols="1,1"]
```
ghislainpiot, line 11: I am also sorry for not catching this last time, but I find the table hard to read. It's the same for the `numpy.Generator` rule. I don't know how we could fix it however.

ghislainpiot (changes requested):
The bulk of the rule is good, it is just small details and things I did not catch in the previous sprint.

rules/S6973/python/rule.adoc
joke1196: I am not sure we can do much as some formatting features are disabled it seems, i.e. striping even or odd rows

ghislainpiot (approved):
LGTM!

rules/S6973/python/rule.adoc
jean-jimbo-sonarsource: I'll follow up on formatting tables.

rules/S6973/python/rule.adoc
```diff
@@ -1,90 +1,66 @@
-This rule raises an issue when a Scikit-learn model is instantiated without specifying the important hyperparameters. 
-
-
+This rule raises an issue when a machine learning estimator or optimizer is instantiated without specifying the important hyperparameters. 
 
 == Why is this an issue?
 
-When instantiating a Scikit-learn estimator, it will use default values for the hyperparameters that are not specified.
-Relying on the default values can lead to non-reproducible results across diffferent versions of the library. 
+When instantiating an estimator or an optimizer, it will use default values for the hyperparameters that are not specified.
```
jean-jimbo-sonarsource, line 5: Suggestion: "When instantiating an estimator or an optimizer, default values for any hyperparameters that are not specified will be used."

jean-jimbo-sonarsource (changes requested):
Suggestion

PR #3965, opened 2024-06-04
CPP-5126 S3630 Add exceptions and updated description

tomasz-kaminski-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3630/cfamily/rule.adoc
+++ b/rules/S3630/cfamily/rule.adoc
@@ -1,12 +1,12 @@
 == Why is this an issue?
 
-Because ``++reinterpret_cast++`` does not perform any type safety validations, it is capable of performing dangerous conversions between unrelated types, often leading to undefined behavior.
+Because ``++reinterpret_cast++`` does not perform any safety validations, it is capable of dangerous conversions between unrelated types, often leading to undefined behavior.
 
 In some cases, `reinterpret_cast` can be simply replaced by a more focused cast, such as `static_cast`.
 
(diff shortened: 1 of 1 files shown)


rules/S3630/cfamily/rule.adoc
```diff
@@ -47,7 +49,13 @@ This rule raises an issue when ``++reinterpret_cast++`` is used.
   }
 ----
 
+=== Exceptions
 
+This rule does not raise an issue when `reinterpret_cast` is used to convert a pointer to:
```
loic-joly-sonarsource, line 54: It would be good to explain why these exceptions are in place...
```suggestion
Since those conversions have a well-defined behavior, this rule does not raise an issue when `reinterpret_cast` is used to convert a pointer to:
```

rules/S3630/cfamily/rule.adoc
```diff
@@ -47,7 +49,13 @@ This rule raises an issue when ``++reinterpret_cast++`` is used.
...
+
+* ``++char*++``, ``++unsigned char*++``, ``++std::byte*++`` or `const` pointers to these types,
```
loic-joly-sonarsource, line 56:
```suggestion
* ``++char*++``, ``++unsigned char*++``, ``++std::byte*++`` or `const` variants of these types, 
```
When I read pointer to these types, I understood `char ** const`

loic-joly-sonarsource (approved):

PR #3969, opened 2024-06-04
Logging rules: add "how to fix" section

costin-zaharia-sonarsource:
## Review

- [ ] text style and tone


rules/S6669/why-dotnet.adoc
```diff
@@ -3,15 +3,6 @@
...
 
-The default regex will consider compliant the followings:
-
-* `logger`
-* ``++_logger++``
-* `Logger`
-* ``++_Logger++``
-* `log`
-* ``++_log++``
-
```
costin-zaharia-sonarsource, line 14: Moved to "how to fix" section.

rules/S6669/csharp/rule.adoc
```diff
@@ -1,12 +1,27 @@
 include::../why-dotnet.adoc[]
 
+== How to fix it
+
+Update the name of the field or property to follow the configured naming convention. By default, the following names are considered compliant:
+
+* `{logger}`
+* ``++{_logger}++``
+* `{Logger}`
+* ``++{_Logger}++``
```
mary-georgiou-sonarsource, line 10: Why the use of `++`?

costin-zaharia-sonarsource: It's because of the `_` character. It has a special meaning in ASCIDOC (italic formatting) and it needs to be escaped. With `++` :)

mary-georgiou-sonarsource (approved):
LGTM - one minor comment regarding `++`. Do we need them?

PR #3970, opened 2024-06-04
Modify rule S1479: allow statements with single-line clauses only

antonioaversa:
## Review

- [ ] text style and tone


rules/S1479/csharp/rule.adoc
```diff
@@ -1,44 +1,62 @@
 == Why is this an issue?
 
-When https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-switch-statement[switch] statements have large sets of case clauses, it is usually an attempt to map two sets of data. A https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2[Dictionary] should be used instead to make the code more readable and maintainable.
+When https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-switch-statement[switch] statements have large sets of `case` clauses, it is usually an attempt to map two sets of data.
```
martin-strecker-sonarsource, line 3: This is one of many use cases of switch. switch statements are procedural and are most of the time used for control-flow. That means for side effects like in 
```cs
string message;
switch(mouse.PressedButton)
{
    case 1:
        message  = "left button";
        DoSomething();
        break;
    case 2:
        message  = "right button";
        if (btn_Ok.Enabled)
        {
            DoSomethingElse();
        }
        break;
    default:
        message  = "some button";
        break;
}
```
We can not assume the "map two sets" case is the most common use case.
Also, the rule implementation has nothing to do with the map case. It does not check for this use case. As such, the rule description is misleading.

rules/S1479/csharp/metadata.json
```diff
@@ -1,3 +1,4 @@
 {
+    "title": "\"switch\" statements with many \"case\" clauses should have only one statement",
```
martin-strecker-sonarsource, line 2: This title is misleading. See the discussion below about why we think so.

rules/S1479/csharp/rule.adoc
```diff
@@ -1,44 +1,62 @@
...
+When https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-switch-statement[switch] statements have large sets of `case` clauses, it is usually an attempt to map two sets of data. 
+
+When the `case` clauses of such `switch` statement contain multiple lines, the code becomes hard to read and maintain. 
+
+For example, the https://www.sonarsource.com/docs/CognitiveComplexity.pdf[Cognitive Complexity] is going to be particularly high. Another example is the introduction of local variables in some of the `case` clauses, which may require introducing local scopes with curly braces, to avoid name clashing.
+
+In such scenarios, it's better to rethink the approach to data retrieval, possibly making use of a collection such as a https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2[Dictionary].
```
martin-strecker-sonarsource, line 9: > data retrieval

Here again: switch statements are about control flow and do not necessarily care about data.

martin-strecker-sonarsource (changes requested):

antonioaversa:
@martin-strecker-sonarsource As discussed offline, I am expecting a counter-proposal on how you would change the first paragraph of the description (which is common to all languages) and the title of the rule, before moving forward.

rules/S1479/csharp/rule.adoc
```diff
@@ -47,37 +62,102 @@ public class TooManyCase
 
 [source,csharp,diff-id=1,diff-type=compliant]
 ----
-using System.Collections.Generic;
+public int MapChar(char ch, int value)
+{
+    switch(ch) // Noncompliant
```
martin-strecker-sonarsource, line 67:
```suggestion
    switch(ch) // Compliant: All 5 cases are single line statements
```

rules/S1479/csharp/rule.adoc
```diff
@@ -47,37 +62,102 @@ public class TooManyCase
...
+    }
+}
+----
+
+==== Compliant solution
+
+[source,csharp,diff-id=2,diff-type=compliant]
+----
+public int MapChar(char ch, int value) =>
+    ch switch
```
martin-strecker-sonarsource, line 133:
```suggestion
    ch switch // Compliant
```

martin-strecker-sonarsource (changes requested):
This is so much better than before 👍. 
The only concern I have is the duplication of the non-compliant example in C#. Is this the recommended way to do it? I probably would just have added the second compliant solution below the first one (without `diff-id`) with a small sentence like "For this example, a switch expression is more concise and clear:"

PR #3971, opened 2024-06-04
Modify Rule S3624 description and deprecate S4963 (CPP-5113)

frederic-tingaud-sonarsource:
Matching implementation PR: https://github.com/SonarSource/sonar-cpp/pull/3578


rules/S3624/cfamily/rule.adoc
```diff
@@ -1,109 +1,143 @@
+When a class cannot follow the Rule of Zero, its special members should follow one of a few logical patterns.
+
 == Why is this an issue?
 
-In {cpp}, you should not directly manipulate resources (a database transaction, a network connection, a mutex lock) but encapsulate them in RAII (_Resource Acquisition Is Initialization_) wrapper classes that will allow you to manipulate them safely. When defining one of those wrapper classes, you cannot rely on the compiler-generated special member functions to manage the class' resources for you (see the Rule-of-Zero, S4963). You must define those functions yourself to make sure the class' resources are properly copied, moved, and destroyed.
+Most classes should not directly handle resources, but instead, use members that perform resource handling for them:
```
enzo-pellegrini, line 5: Not sure of the 

> but instead

it's not too readable

enzo-pellegrini, line 5: Maybe

> use members

Is not the clearest way to say that you should have class members with of a class that handles the resource for you, it would be more clear if you said maybe "use classes that perform resource handling for them as members" although this is too long. But use members to mean that you have fields of a class that wraps resource handling was not immediate to understand.

enzo-pellegrini (commented):
Just a couple one comment, I hope it's useful

rules/S3624/cfamily/rule.adoc
```diff
@@ -79,25 +134,10 @@ All copy and move operations are deleted.
 
 === Standards
 
-* MISRA C++23 15.0.1 - "Special member functions" shall be provided appropriately
+* MISRA {cpp}23 15.0.1 - "Special member functions" shall be provided appropriately
```
loic-joly-sonarsource, line 137: I think we could add links to https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c20-if-you-can-avoid-defining-default-operations-do and https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all

loic-joly-sonarsource (approved):
Some small comments, but overall I'm ok.

necto (commented):
Good rewriting!
I would suggest two things:
- Some non-compliant examples and an explanation of how they lead to bugs would go a long way to motivate this rule. As it is now, it does not explain why it is in my interest to follow it.
- I recommend placing each sentence, or phrase on a separate line. It brings a few benefits, compared to long-running lines of text:
  - If you change a couple of words, I will be able to see that immediately in the diff, unlike in the case of long lines where I would have to read two long paragraphs and try to understand what changed
  - When searching for an author of a paragraph, git blame reports authors per-line, so if the paragraph consists of multiple lines as opposed to a single one, I can more precisely pin-point who to contact.

PR #3983, opened 2024-06-06
RULEAPI-770 Fix bug due to multiple sonarpedia.json files

johann-beleites-sonarsource:
When there are multiple sonarpedia.json files for the same language, the rules picked up for the next sonarpedia.json file will override the ones picked up previously. With this fix, the rules are not overwritten but all are collected, so one repository can have multiple sonarpedia.json files for the same language.

### DIFF
--- a/rspec-tools/rspec_tools/coverage.py
+++ b/rspec-tools/rspec_tools/coverage.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import json
+import collections
 from git import Repo
 from git import Git
 from pathlib import Path
@@ -133,7 +134,7 @@ class Coverage:
         self.rule_implemented(rule_id, language, analyzer, version)
 
 def all_implemented_rules():
-  implemented_rules = {}
+  implemented_rules = collections.defaultdict(list)
   for sp_file in Path('.').rglob('sonarpedia.json'):
     print(sp_file)
     sonarpedia_path=sp_file.parents[0]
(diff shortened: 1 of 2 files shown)


rspec-tools/rspec_tools/coverage.py
```diff
@@ -141,7 +141,12 @@ def all_implemented_rules():
       sonarpedia = load_json(sp_file)
       path = str(sonarpedia_path) + '/' + sonarpedia['rules-metadata-path'].replace('\\', '/')
       languages = sonarpedia['languages']
-      implemented_rules.update(get_implemented_rules(path, languages))
+
+      implemented_rules_in_path = get_implemented_rules(path, languages)
```
frederic-tingaud-sonarsource, line 145: I see that we have the same pattern already in this file, but it can actually be simplified a bit like this:
```python
import collections
...
implemented_rules = collections.defaultdict(list)
...
for lang, rules in implemented_rules_in_path.items():
  implemented_rules[lang].append(rules)
```

frederic-tingaud-sonarsource (changes requested):
The change is functionally good for me.
I'm no expert in Python, but I think we can make it simpler by using some nice Dictionary features.

frederic-tingaud-sonarsource (approved):

rspec-tools/rspec_tools/coverage.py
johann-beleites-sonarsource: Indeed, this is nicer, thanks for the suggestion

PR #3987, opened 2024-06-11
Modify rule S5769: Remove dead link

karim-ouerghemmi-sonarsource:
I believe it's Ok to simply remove it. I didn't find an alternative and the link is not on the wayback machine https://wayback-api.archive.org/web/20240000000000*/https://zevolving.com/use-of-field-symbols-vs-work-area


nicolas-gauthier-sonarsource (approved):
LGTM! Thanks!

PR #3996, opened 2024-06-14
Create specialized single-language rules

jamie-anderson-sonarsource:
This change allows single-language rules to use a specialized template when they're created.

The reason for this change is that the AppSec squad wants to improve the efficiency with which new `secrets` rules can be created. These rules only exist for a single language, `secrets`, and much of the rule content and metadata is similar/identical between rules.

The best way to realize this efficiency improvement for `secrets` rules is to have a specialized template that gets used when a new RSPEC is created. This template is used when a rule is created for a single language, and that language is `secrets`.

When a rule is created with a single language, the appropriate template folder is checked for a folder whose name matches the language. If such a folder is found, it will be used as the template for the language. If no such folder is found, the existing language-neutral template will be used instead.

I have chosen not to lock this feature down to just the `secrets` language. I can't think of any other languages that could make use of this language-specific template functionality right now, but it's possible that it will be useful for new languages in the future.

## Review

- [ ] text style and tone

### DIFF
--- a/rspec-tools/rspec_template/single_language/secrets/metadata.json
+++ b/rspec-tools/rspec_template/single_language/secrets/metadata.json
@@ -1,0 +1,56 @@
+{
+  "title": "SECRET_TYPE should not be disclosed",
+  "type": "VULNERABILITY",
+  "code": {
+    "impacts": {
+      "SECURITY": "HIGH"
+    },
+    "attribute": "TRUSTWORTHY"
+  },
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "30min"
+  },
+  "tags": [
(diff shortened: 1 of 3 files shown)


jamie-anderson-sonarsource:
The `validate_links` check is failing due to links in rules S1694 and S6914. They are unrelated to this change.

At the time of writing this comment, the `validate_metadata` task is failing due to an issue on rule S5334. That rule is not touched by this PR, so I am re-running the check in the hope that it's a transient error.

gaetan-ferry-sonarsource (commented):
LGTM. I'd rather have the opinion from someone from RSPEC before approving though.

frederic-tingaud-sonarsource (approved):

PR #4019, opened 2024-06-28
Modify S3247: Improve description and update example

sebastien-marichal:
RSPEC update for https://github.com/SonarSource/sonar-dotnet/issues/2120

### DIFF
--- a/rules/S3247/csharp/rule.adoc
+++ b/rules/S3247/csharp/rule.adoc
@@ -1,11 +1,21 @@
 == Why is this an issue?
 
-Because the ``++is++`` operator performs a cast if the object is not null, using ``++is++`` to check type and then casting the same argument to that type, necessarily performs two casts. The same result can be achieved more efficiently with a single cast using ``++as++``, followed by a null-check.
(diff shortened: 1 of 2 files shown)


rules/S3247/csharp/rule.adoc
```diff
@@ -1,11 +1,21 @@
 == Why is this an issue?
 
-Because the ``++is++`` operator performs a cast if the object is not null, using ``++is++`` to check type and then casting the same argument to that type, necessarily performs two casts. The same result can be achieved more efficiently with a single cast using ``++as++``, followed by a null-check.
+In C#, the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#is-operator[`is` type testing operator] can be used to check if the run-time type of an object is compatible with a given type. If the check is successful, it means the code performed a cast. Performing another cast following the check result into a duplicate cast.
```
mary-georgiou-sonarsource, line 3:
```suggestion
In C#, the https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#is-operator[`is` type testing operator] can be used to check if the run-time type of an object is compatible with a given type. If the object is not null, then the `is` operator performs a cast, and so performing another cast following the check result is redundant.
```

mary-georgiou-sonarsource (approved):
LGTM - minor polishing

PR #4027, opened 2024-07-02
Modify rule S5259: Document #pragma once (CPP-5133)

loic-joly-sonarsource:


frederic-tingaud-sonarsource (approved):

rules/S5259/cfamily/rule.adoc
```diff
@@ -12,6 +12,7 @@ The include guard pattern is made up of four parts:
 
 The rule raises an issue when the name in the second part differs from the first (usually because of a typo or a copy/paste issue).
 
+Because the include guard pattern is cumbersome, virtually every compiler provides a non-standard alternative: `+#pragma once+`. This directive prevents multiple inclusions of the file that contains it without needing to invent a unique macro name for each file. Note that it relies on the notion of _same file_, which can be tricky to determine. Additionally, it will not work with build systems that copy headers in different places.
```
loic-joly-sonarsource, line 15:
```suggestion
Because the include guard pattern is cumbersome, virtually every compiler provides a non-standard alternative: ``++#pragma once++``. This directive prevents multiple inclusions of the file that contains it without needing to invent a unique macro name for each file. Note that it relies on the notion of _same file_, which can be tricky to determine. Additionally, it will not work with build systems that copy headers in different places.
```

rules/S5259/cfamily/rule.adoc
```diff
@@ -34,6 +37,19 @@ The rule raises an issue when the name in the second part differs from the first
 #endif
 ----
 
+An alternative is to use `#pragma once` instead:
```
frederic-tingaud-sonarsource, line 40:
```suggestion
An alternative is to use ``++#pragma once++`` instead:
```

frederic-tingaud-sonarsource (approved):

PR #4028, opened 2024-07-03
Modify rule S824: Clarify the most vexing parse (CPP-5122)

loic-joly-sonarsource:
New suggested message (the rule implementation has not been updated yet):
> Move this function declaration outside of this block scope, or if the intent was to declare a variable, use a syntax that avoids the most vexing parse.

### DIFF
--- a/rules/S824/cfamily/rule.adoc
+++ b/rules/S824/cfamily/rule.adoc
@@ -8,15 +8,15 @@ Additionally, when a function is declared at block scope, the intent is often no
 ----
 void f() {
   int a;
-  int b();
+  string b();
   short c(short (a));
 }
 ----
 
 * `b` could be interpreted as:
 
-** A variable of type `int` with empty initialization or
-** A function with no argument and returning an `int`. 
+** A variable of type `string` with empty initialization or
+** A function with no argument and returning a `string`. 
 
 +
 The second interpretation is selected.
@@ -35,16 +35,18 @@ There are several ways to write the code differently so that `b` and `c` can onl
 ----
 void f() {
   int a;
-  int b {};
+  string b {};
   auto c = short (a);
 }
 ----
 
 By raising issues on local function declaration, this rule helps detect when a function is inadvertently declared.
 
+
+== How to fix it
 === Noncompliant code example
 
-[source,cpp]
+[source,cpp,diff-id=2,diff-type=noncompliant]
 ----
 class A {
 };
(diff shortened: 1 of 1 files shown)


rules/S824/cfamily/rule.adoc
```diff
@@ -1,10 +1,46 @@
...
+It is rarely useful to declare a function at block scope. Such a function will not get special access to any name in its enclosing scope, and therefore, it is equivalent but clearer to declare it instead in the enclosing namespace.
 
+Additionally, when a function is declared at block scope, the intent is often not to declare a function but instead to declare and initialize a variable. This problem is nicknamed the _most vexing parse_ and stems from the fact that some syntaxes can be ambiguous, and that in that case the language unintuitively favors function declaration:
 
-Additionally, where a declaration statement could either declare a function or an object, the compiler will choose to declare the function. To avoid potential developer confusion over the meaning of a declaration, functions should not be declared at block scope.
+[source,cpp,diff-id=1,diff-type=noncompliant]
+----
+void f() {
+  int a;
+  int b();
```
frederic-tingaud-sonarsource, line 11: I'm not sure `int` is the best case for that. As a beginner, I think I never tried to write `int b();` because a default int didn't make sense to me. I did try to write `std::string b();` to initialize an empty string, though.

rules/S824/cfamily/rule.adoc
```diff
@@ -1,10 +1,46 @@
...
+void f() {
+  int a;
+  int b {};
+  auto c = short (a);
+}
+----
+
+By raising issues on local function declaration, this rule helps detect when a function is inadvertently declared.
 
 === Noncompliant code example
```
frederic-tingaud-sonarsource, line 45: We should perhaps rewrite or remove this part? It feels weird to have a `=== Noncompliant code example` section just after two code examples.

frederic-tingaud-sonarsource (changes requested):

rules/S824/cfamily/rule.adoc
loic-joly-sonarsource: changed to string.

rules/S824/cfamily/rule.adoc
loic-joly-sonarsource: I tried to go in the opposite direction... I put it in its own _How to fix it_ section and added a compliant example.

frederic-tingaud-sonarsource (approved):

rules/S824/cfamily/rule.adoc
frederic-tingaud-sonarsource: It works better.

PR #4039, opened 2024-07-05
BUILD-4733: Use COVERAGE_GITHUB_TOKEN from the vault instead of the default GITHUB_TOKEN

tomverin:


rspec-tools/rspec_tools/cli.py
necto, line 1: @tomverin is `GITHUB_TOKEN` still available?
I believe, these uses are fine with the normal `GITHUB_TOKEN` because they are concerned with changes done to this repository (SonarSource/rspec) only.

rspec-tools/tests/test_cli.py
necto, line 1: To be kept in sync with rspec_tools/cli.py

.github/workflows/update_quickfix_status.yml
necto, line 1: Same as with cli.py, I believe the default `GITHUB_TOKEN` should suffice.

.github/workflows/main.yml
necto, line 1: Normally this workflow should not need access to any repository than this one (SonarSource/rspec), so, I believe, `GITHUB_TOKEN` is enough

.github/workflows/create_new_rspec.yml
necto, line 1: Same as with main.yml, `GITHUB_TOKEN` should be enough

.github/workflows/add_language.yml
necto, line 1: Same as with main.yml, `GITHUB_TOKEN` should be enough

necto (approved):
Am I right that the difference between `COVERAGE_GITHUB_TOKEN` and `GITHUB_TOKEN` is in the access to closed-source repositories, like, sonar-abap?
I believe it is only needed for coverage, indeed, and the regular `GITHUB_TOKEN` should be enough for all the other actions, which are done locally on this repo only.
I leave it up to you, RE, whether you want to use a single powerful token or two different ones.
Otherwise, looks good, unfortunately I can't test it locally.

rspec-tools/rspec_tools/cli.py
tomverin: yes it's still available but since the other token is giving the same permissions and more, it was easier to move everything

PR #4049, opened 2024-07-09
Fix language in rule creation script

frederic-tingaud-sonarsource:


marco-antognini-sonarsource (approved):
I glanced at it; it looks okay.

PR #4065, opened 2024-07-16
Revert Coverage script to functional state

frederic-tingaud-sonarsource:
The coverage script has been broken since Jun 26 and we've been pushing a lot of fixes since that not only didn't fix it, but reduced readability and didn't improve security (addition of a "COVERAGE" token that was used for all operations even unrelated to coverage, with the token having more authorizations than necessary for coverage calculation). I think we should revert and look into a correct solution offline without impacting the whole analysis team.


necto (approved):
Makes sense. I will be off starting tomorrow, so I suggest adding `on\workflow_dispatch` to the coverage workflow and triggering it manually today in case it doesn't work and you want me to have a look at a follow-up fix

PR #4089, opened 2024-07-23
Fix coverage script when there is no new coverage

frederic-tingaud-sonarsource:
The previous change to the task `'Push the updated coverage to master'` made it passthrough because it used `{{` as part of the condition, which seems to be forbidden.
Changing the condition to what we usually use to check default_branch in other workflow scripts throughout our repositories.


necto (approved):

necto:
Please don't forget to add some explanation in the commit message

PR #4109, opened 2024-08-05
Deprecate rule S2387

sebastien-marichal:
Superseded by [CS0108](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0108) in C#
Superseded by [BC40004](https://learn.microsoft.com/en-us/dotnet/visual-basic/misc/bc40004) in VB.NET

### DIFF
--- a/rules/S2387/csharp/metadata.json
+++ b/rules/S2387/csharp/metadata.json
@@ -1,5 +1,5 @@
 {
-  "defaultQualityProfiles": [
-    
-  ]
+  "status": "deprecated",
+  "tags": [],
+  "defaultQualityProfiles": []
 }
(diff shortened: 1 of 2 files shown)


rules/S2387/csharp/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "defaultQualityProfiles": [
-    
-  ]
+  "status": "deprecated",
+  "tags": [],
```
zsolt-kolbay-sonarsource, line 3: Why do you override the `tag` element?

zsolt-kolbay-sonarsource (approved):
LGTM!

PR #4111, opened 2024-08-05
Deprecate Rule S2325 for C#

zsolt-kolbay-sonarsource:
The rule is covered by the [CA1822](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822) analyzer from Microsoft.


sebastien-marichal (changes requested):
[CA1822](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822) is enabled by default as a suggestion.
Does is show up during the build/analysis and in SQ?

zsolt-kolbay-sonarsource:
I checked it, you're right. It CA1822 doesn't show up in SQ.
[I'm promoting the rule to Sonar-Way](https://github.com/SonarSource/rspec/pull/4123).

PR #4115, opened 2024-08-05
CPP-5564 S2335 Make more characters properly terminate an escape sequence

loic-joly-sonarsource:
See also https://github.com/SonarSource/sonar-cpp/pull/3880


rules/S2335/cfamily/rule.adoc
```diff
@@ -21,12 +27,14 @@ int c1 = '\141t'; // Noncompliant
 ----
 const char *s2 = "\x41" "g"; // Compliant - terminated by end of literal
 const char *s3 = "\x41\x67"; // Compliant - terminated by another escape
-int c2 = '\141\t'; // Compliant - terminated by another escape
+int c2 = '\141\t'; // Compliant (with a different value) - terminated by another escape
```
frederic-tingaud-sonarsource, line 30: I don't understand what this means.

frederic-tingaud-sonarsource (changes requested):

rules/S2335/cfamily/rule.adoc
loic-joly-sonarsource: This is supposed to be the compliant version of some non-compliant code, but the value of `c2` is not the same as in the non-compliant code.

Once again, the framework of having non-compliant followed by compliant makes it more complex to write the RSPEC. Without these titles, I would have move this compliant example right after the non-compliant one, with a different variable name.

```
[source,cpp]
----
const char *s1 = "\x41g";  // Noncompliant
int c1 = '\141t'; // Noncompliant
int c2 = '\141\t'; // Compliant - terminated by another escape
----
```

frederic-tingaud-sonarsource: I preferred the previous version, without `(with a different value)`. The Compliant code block shows various examples of compliant declarations, there is no need to try to create a link between `c1` and `c2`.

marco-antognini-sonarsource (approved):

PR #4120, opened 2024-08-06
Modifiy S2674: Promote it to SonarWay

sebastien-marichal:
And update to LaYC format


rules/S2674/csharp/rule.adoc
```diff
@@ -1,47 +1,63 @@
...
 ----
 
-=== Compliant solution
+==== Compliant solution
 
-[source,csharp]
+[source,csharp,diff-id=1,diff-type=compliant]
 ----
-public void DoSomething(string fileName) 
+byte[] DoSomething(string fileName)
```
zsolt-kolbay-sonarsource, line 38:
```suggestion
byte[] ReadFile(string fileName)
```

zsolt-kolbay-sonarsource (approved):
LGTM! The conversion to LAYC is a nice improvement.

PR #4123, opened 2024-08-07
Modify S2325: Promote C# rule to Sonar-Way

zsolt-kolbay-sonarsource:
Although it rule is covered by the [CA1822](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822) Roslyn analyzer, but that doesn't show up in SQ, because by default it's only a Suggestion rather than a Warning.

### DIFF
--- a/rules/S2325/csharp/rule.adoc
+++ b/rules/S2325/csharp/rule.adoc
@@ -1,65 +1,86 @@
 == Why is this an issue?
 
-Methods and properties that don't access instance data can be ``++static++`` to prevent any misunderstanding about the contract of the method.
+Methods and properties that don't access instance data should be marked as `static` for the following reasons:
 
+* Clarity and Intent: Marking a method/property as static makes it clear that the method does not depend on instance data and can be called without creating an instance of the class. This improves the readability of the code by clearly conveying the member's intended use.
+* Performance: Instance methods/properties in C# require an instance of the class to be called. This means that even if the it doesn't use any instance data, the runtime still needs to pass a reference to the instance during the call. For static methods and properties, this overhead is avoided, leading to slightly better performance.
+* Memory Usage: Since instance methods implicitly carry a reference to the instance (the caller object), they can potentially prevent the garbage collector from collecting the instance whem it is not otherwise referenced. Static members do not carry this overhead, potentially reducing memory usage.
+* Testing: Static members can be easier to test since they do not require an instance of the class. This can simplify unit testing and reduce the amount of boilerplate code needed to set up tests.
 
-=== Noncompliant code example
+=== Exceptions
 
-[source,csharp]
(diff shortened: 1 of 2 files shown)


sebastien-marichal (changes requested):
Can you update the sections to LaYC format?

rules/S2325/csharp/rule.adoc
```diff
@@ -1,65 +1,86 @@
...
 
+* Clarity and Intent: Marking a method/property as static makes it clear that the method does not depend on instance data and can be called without creating an instance of the class. This improves the readability of the code by clearly conveying the member's intended use.
+* Performance: Instance methods/properties in C# require an instance of the class to be called. This means that even if the it doesn't use any instance data, the runtime still needs to pass a reference to the instance during the call. For static methods and properties, this overhead is avoided, leading to slightly better performance.
+* Memory Usage: Since instance methods implicitly carry a reference to the instance (the caller object), they can potentially prevent the garbage collector from collecting the instance whem it is not otherwise referenced. Static members do not carry this overhead, potentially reducing memory usage.
+* Testing: Static members can be easier to test since they do not require an instance of the class. This can simplify unit testing and reduce the amount of boilerplate code needed to set up tests.
```
sebastien-marichal, line 8: I am not really convinced this is a good point to have here; feels a bit forced.
While this is a valid point, IMO, it does not have the same impact as the other.
I will let you decide if you want to keep it.

rules/S2325/csharp/rule.adoc
```diff
@@ -1,65 +1,86 @@
 == Why is this an issue?
 
-Methods and properties that don't access instance data can be ``++static++`` to prevent any misunderstanding about the contract of the method.
+Methods and properties that don't access instance data should be marked as `static` for the following reasons:
 
+* Clarity and Intent: Marking a method/property as static makes it clear that the method does not depend on instance data and can be called without creating an instance of the class. This improves the readability of the code by clearly conveying the member's intended use.
+* Performance: Instance methods/properties in C# require an instance of the class to be called. This means that even if the it doesn't use any instance data, the runtime still needs to pass a reference to the instance during the call. For static methods and properties, this overhead is avoided, leading to slightly better performance.
+* Memory Usage: Since instance methods implicitly carry a reference to the instance (the caller object), they can potentially prevent the garbage collector from collecting the instance whem it is not otherwise referenced. Static members do not carry this overhead, potentially reducing memory usage.
```
sebastien-marichal, line 7: This seems to be very specific. It took me a bit of time to think of an actual usage where it would be true, and I am not 100% sure.
The only usage I could think of is when you use the method as a delegate somewhere.

Could the wording be too complex?

sebastien-marichal (approved):
LGTM!
I have some open non-blocking questions.

PR #4134, opened 2024-08-09
Modify S4052: Promote C# rule to Sonar way

zsolt-kolbay-sonarsource:
Also converted the rule description to LayC.


sebastien-marichal (approved):
LGTM!

PR #4163, opened 2024-08-19
Modify rule S2068: Update documentation to not recommend the customization

ghislainpiot:
SONARPY-2077


joke1196 (approved):
LGTM. Thanks for the clean up as well!

PR #4168, opened 2024-08-20
Modify rule S3715: Mention empty initializer before C23

michael-jabbour-sonarsource:
Do not merge before MMF-3874 is merged


michael-jabbour-sonarsource:
@loic-joly-sonarsource I didn't add examples for it, as it may cause confusion by placing it in the non-compliant section. Let me know if you think an example can be useful...

loic-joly-sonarsource (approved):

PR #4178, opened 2024-08-22
Modify rule S3562: improve resources and add rspecator

antonioaversa:
## Review

- [ ] text style and tone


rules/S3562/dart/rule.adoc
```diff
@@ -9,7 +9,7 @@ Enum-like classes are classes that:
 * are defined as non-abstract;
 * have only one private non-factory constructor
 * have two or more static const fields whose type is the enclosing class and
```
leveretka, line 11: you probably don't need `and` here, if you create another list item from the second part of the sentence
```suggestion
* have two or more static const fields whose type is the enclosing class
```

leveretka (approved):
One tiny suggestion. Other than that looks fine!

rules/S3562/dart/rule.adoc
antonioaversa: Indeed, "and" removed.
I have also removed the semicolon, that was only on the first item of the list, as we have an explicit style rule in the rspec repo about punctuation on lists that suggests not using them

PR #4190, opened 2024-08-23
Modify rules S6320;S6416;S6555;S6707;S6976 Update examples

sallaigy:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6555/java/rule.adoc
+++ b/rules/S6555/java/rule.adoc
@@ -2,11 +2,13 @@
 
 A reference to ``++null++`` should never be dereferenced/accessed. Doing so will cause a ``++NullPointerException++`` to be thrown. At best, such an exception will cause abrupt program termination. At worst, it could expose debugging information that would be useful to an attacker, or it could allow an attacker to bypass security measures.
 
+== How to fix it
 
+=== Code examples
 
-=== Noncompliant code example
+==== Noncompliant code example
 
-[source,java]
+[source,java,diff-id=1,diff-type=noncompliant]
 ----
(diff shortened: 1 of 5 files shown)


rules/S6555/java/rule.adoc
```diff
@@ -8,14 +8,11 @@ A reference to ``++null++`` should never be dereferenced/accessed. Doing so will
...
-  return condition ? "name" : null;
-}
-
-public boolean isNameEmpty() {
-  return getName().length() == 0; // Noncompliant; the result of getName() could be null, but isn't null-checked
+public boolean isNameEmpty(String givenName, String familyName) {
+  if (givenName != null && familyName != null) {
+    return givenName.isEmpty() && familyName.isEmpty();
+  }
+  return familyName.isEmpty(); // Noncompliant; familyName may be null
```
chrislain-razafimahefa-sonarsource, line 15: There is no 'compliant' section for this rule?

chrislain-razafimahefa-sonarsource (approved):
LGTM! S6555 does not have a compliant section!?

rules/S6555/java/rule.adoc
sallaigy: The original rule example didn't have one either, so I did not include one here either. Would you prefer that we add one?

chrislain-razafimahefa-sonarsource: True, I also saw that the original one did not have one. It seems weird not to have one when all the other rules do, so I think we should, if you have no objections.

chrislain-razafimahefa-sonarsource (approved):
LGTM!

PR #4215, opened 2024-08-30
Modify rule S2755: Simplify how to fix it section

hendrik-buchwald-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/docs/header_names/allowed_framework_names.adoc
+++ b/docs/header_names/allowed_framework_names.adoc
@@ -49,6 +49,7 @@
 * Java Cryptography Extension
 * Apache HttpClient
 * Couchbase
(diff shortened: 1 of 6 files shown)


daniel-teuchert-sonarsource (changes requested):
I think the SAXBuilder compliant example (from jdom2.adoc) can also be adjusted.
E.g. the following should be sufficient if I understand correctly:
```
import org.jdom2.input.SAXBuilder;

public void decode() {
    SAXBuilder builder = new SAXBuilder();
    builder.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
} 
```

PR #4235, opened 2024-09-04
Modify rule S2637(cfamily): CPP-5602 allow assigning nullptr to params

necto:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2637/cfamily/metadata.json
+++ b/rules/S2637/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "\"nonnull\" pointers should not be set to null",
+  "title": "Null should not be passed to \"nonnull\" pointers",
   "tags": [
     "cwe",
     "symbolic-execution",
--- a/rules/S2637/cfamily/metadata.json
+++ b/rules/S2637/cfamily/metadata.json
@@ -1,5 +1,5 @@
 {
-  "title": "Null should not be passed to \"nonnull\" pointers",
+  "title": "\"nonnull\" parameters and return values of \"returns_nonnull\" functions should not be null",
   "tags": [
     "cwe",
     "symbolic-execution",
--- a/rules/S2637/cfamily/rule.adoc
+++ b/rules/S2637/cfamily/rule.adoc
@@ -13,7 +13,7 @@ A commonly used attribute is `nonnull` which can be used to mark a function's re
 #include <string.h>
 
 __attribute__((returns_nonnull)) int *
-make_array_copy(int *src, size_t len)__attribute__((nonnull(1)));
+make_array_copy(int *src, size_t len) __attribute__((nonnull(1)));
 
 int * make_array_copy(int *src, size_t len) {
   int *dst = (int *)malloc(len * sizeof(int));
@@ -44,14 +44,14 @@ int* foo(__attribute__((nonnull)) int* x) {
 Failing to adhere to the attribute may introduce serious program errors.
 In particular, the compiler does not enforce that values marked as `nonnull` are indeed non-null at runtime; it is the developers' responsibility to adhere to the attribute.
 These values are typically _not_ null-checked before use.
-Passing null (i.e., `NULL`, `0` or `nullptr`) argument to a parameter as `nonnull`
-or returning null from a function marked as `returns_nonnull` is hence likely to cause a null-pointer dereference.
+Passing null (i.e., `NULL`, `0` or `nullptr`) as an argument to a parameter that is marked as `nonnull`
+or returning null from a function marked as `returns_nonnull` is, hence, likely to cause a null-pointer dereference.
 Compilers may even apply optimizations based on this attribute and might, for instance, _remove_ an explicit null-check if the parameter is declared as `nonnull` -- even in code outside of the function with the attribute.
 
 Note that the `nonnull` attribute is a GNU extension (see https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute[nonnull] and https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-returns_005fnonnull-function-attribute[returns_nonnull]) which many compiler vendors have implemented.
 
 Note that it is allowed to _assign_ null to a parameter marked as `nonnull`.
-This attribute specifies only the function call contract
+This attribute is only concerned with the function call contract
 and does not control the evolution of the parameter variable.
 For example, a linked-list search could be implemented as follows:
 
@@ -59,7 +59,7 @@ For example, a linked-list search could be implemented as follows:
 ----
 struct List {
   int value;
-  List* next; // NULL for a tail node.
+  List* next; // nullptr for a tail node.
 };
 
 
@@ -90,20 +90,20 @@ Because compilers may apply optimizations based on the `nonnull` attribute, not 
 ----
 struct List {
   int value;
-  List *next; // NULL for a tail node.
+  List *next; // nullptr for a tail node.
 };
 
 size_t len(List *n) __attribute__((nonnull));
-size_t len(List *n) {
-  if (!n) return 0; // Impossible branch according to the attribute
-  return 1 + len(n->next);
+size_t len(List *l) {
+  if (!l) return 0; // Impossible branch according to the attribute
+  return 1 + len(l->next);
 }
 ----
 
 
 == How to fix it
 
-Ensure not to pass null values when `nonnull` arguments are expected and not to return a null value when a function marked as `returns_nonnull`.
+Ensure not to pass null values when `nonnull` arguments are expected and not to return a null value when a function is marked as `returns_nonnull`.
 This especially holds for library functions, which frequently require `nonnull` pointer parameters.
 
 On other occasions, it might be more appropriate to remove the attribute.
--- a/rules/S2637/cfamily/rule.adoc
+++ b/rules/S2637/cfamily/rule.adoc
@@ -15,7 +15,7 @@ A commonly used attribute is `nonnull` which can be used to mark a function's re
 __attribute__((returns_nonnull)) int *
 make_array_copy(int *src, size_t len) __attribute__((nonnull(1)));
 
-int * make_array_copy(int *src, size_t len) {
+int *make_array_copy(int *src, size_t len) {
   int *dst = (int *)malloc(len * sizeof(int));
   if (dst == NULL) {
     perror("malloc failed");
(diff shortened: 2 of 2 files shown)


rules/S2637/cfamily/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "\"nonnull\" pointers should not be set to null",
+  "title": "Null should not be passed to \"nonnull\" pointers",
```
pdschbrt, line 2: I wonder if we can be even more specific. How about something like:
`Null should not be passed as an argument to a parameter marked \"nonnull\"`

rules/S2637/cfamily/rule.adoc
```diff
@@ -13,7 +13,9 @@ A commonly used attribute is `nonnull` which can be used to mark a function's re
 #include <string.h>
 
 __attribute__((returns_nonnull)) int *
-make_array_copy(__attribute__((nonnull)) int *src, size_t len) {
+make_array_copy(int *src, size_t len)__attribute__((nonnull(1)));
```
pdschbrt, line 16: I'm not familiar with this syntax, but do we actually start at `1` or is it `0` to denote the very first function parameter?

rules/S2637/cfamily/rule.adoc
```diff
@@ -42,11 +44,35 @@ int* foo(__attribute__((nonnull)) int* x) {
 Failing to adhere to the attribute may introduce serious program errors.
 In particular, the compiler does not enforce that values marked as `nonnull` are indeed non-null at runtime; it is the developers' responsibility to adhere to the attribute.
 These values are typically _not_ null-checked before use.
-Setting a value marked as `nonnull` to null (i.e., `NULL`, `0` or `nullptr`) is hence likely to cause a null-pointer dereference.
+Passing null (i.e., `NULL`, `0` or `nullptr`) argument to a parameter as `nonnull`
+or returning null from a function marked as `returns_nonnull` is hence likely to cause a null-pointer dereference.
```
pdschbrt, line 48:
```suggestion
Passing null (i.e., `NULL`, `0` or `nullptr`) as an argument to a parameter that is marked as `nonnull`
or returning null from a function marked as `returns_nonnull` is, hence, likely to cause a null-pointer dereference.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -42,11 +44,35 @@ int* foo(__attribute__((nonnull)) int* x) {
...
 Compilers may even apply optimizations based on this attribute and might, for instance, _remove_ an explicit null-check if the parameter is declared as `nonnull` -- even in code outside of the function with the attribute.
 
 Note that the `nonnull` attribute is a GNU extension (see https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute[nonnull] and https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-returns_005fnonnull-function-attribute[returns_nonnull]) which many compiler vendors have implemented.
 
+Note that it is allowed to _assign_ null to a parameter marked as `nonnull`.
+This attribute specifies only the function call contract
+and does not control the evolution of the parameter variable.
```
pdschbrt, line 55:
```suggestion
This attribute is only concerned with the function call contract
and does not control the evolution of the parameter variable.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -42,11 +44,35 @@ int* foo(__attribute__((nonnull)) int* x) {
...
+For example, a linked-list search could be implemented as follows:
+
+[source,cpp]
+----
+struct List {
+  int value;
+  List* next; // NULL for a tail node.
```
pdschbrt, line 62:
```suggestion
  List* next; // nullptr for a tail node.
```
But I'm also fine with `NULL`.

rules/S2637/cfamily/rule.adoc
pdschbrt, line 16: In one of the other usages of this attribute, we do have an additional space ` ` between the function's parameter list and the `__attribute__`.
```suggestion
make_array_copy(int *src, size_t len) __attribute__((nonnull(1)));
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -58,29 +84,26 @@ In practice, dereferencing a null pointer may lead to program crashes, or the ap
...
+Because compilers may apply optimizations based on the `nonnull` attribute, not respecting `nonnull` can also introduce more complex bugs such as resource leaks or infinite recursion as indicated in the following code snippet:
 
 [source,cpp]
 ----
-struct Node {
-  int data;
-  Node *next; // NULL for a tail node.
+struct List {
+  int value;
+  List *next; // NULL for a tail node.
```
pdschbrt, line 93: ?
```suggestion
  List *next; // nullptr for a tail node.
```
But I'm fine with `NULL` here, too.

pdschbrt, line 93: Both, `List *next` and `List* next` are used. Can we settle on one style? Preferably `List *next`.
Let's not add `List * next` 😁

rules/S2637/cfamily/rule.adoc
```diff
@@ -58,29 +84,26 @@ In practice, dereferencing a null pointer may lead to program crashes, or the ap
...
 };
 
-size_t len(__attribute__((nonnull)) Node *n) {
-  size_t l = 0;
-  while (n) {
-    ++l;
-    n = n->next;
-  }
-  return l;
+size_t len(List *n) __attribute__((nonnull));
```
pdschbrt, line 96: We don't need an index here? (I really don't know, since I always use the attribute directly next to the argument.)

rules/S2637/cfamily/rule.adoc
```diff
@@ -58,29 +84,26 @@ In practice, dereferencing a null pointer may lead to program crashes, or the ap
...
+size_t len(List *n) {
+  if (!n) return 0; // Impossible branch according to the attribute
+  return 1 + len(n->next);
```
pdschbrt, line 99: Let's be consistent with naming our pointer-to-list-typed parameters.
```suggestion
size_t len(List *l) {
  if (!l) return 0; // Impossible branch according to the attribute
  return 1 + len(l->next);
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -58,29 +84,26 @@ In practice, dereferencing a null pointer may lead to program crashes, or the ap
...
 }
 ----
 
 
 == How to fix it
 
-Ensure not to pass null values when non-null arguments are expected, do not return a null value when a non-null return value is expected, and do not assign null to parameters marked as non-null.
+Ensure not to pass null values when `nonnull` arguments are expected and not to return a null value when a function marked as `returns_nonnull`.
```
pdschbrt, line 106:
```suggestion
Ensure not to pass null values when `nonnull` arguments are expected and not to return a null value when a function is marked as `returns_nonnull`.
```

rules/S2637/cfamily/rule.adoc
```diff
@@ -92,8 +115,8 @@ On other occasions, it might be more appropriate to remove the attribute.
 
 [source,cpp,diff-id=1,diff-type=noncompliant]
 ----
-__attribute__((returns_nonnull))
-int* foo(__attribute__((nonnull)) int* x) {
+int* foo(int* x) __attribute__((nonnull));
```
pdschbrt, line 118: Does this syntax mean that any pointer input parameter and return value might not be null?

rules/S2637/cfamily/rule.adoc
```diff
@@ -108,8 +131,8 @@ void bar() {
 
 [source,cpp,diff-id=1,diff-type=compliant]
 ----
-__attribute__((returns_nonnull))
-int* foo(__attribute__((nonnull)) int* x) {
+int* foo(int* x) __attribute__((nonnull));
```
pdschbrt, line 134: Same question.

rules/S2637/cfamily/rule.adoc
```diff
@@ -108,8 +131,8 @@ void bar() {
...
+int* foo(int* x) {
```
pdschbrt, line 135: And can we have the `*` always next to the variable/and function (and not next to the type)? Or, in case you like the other spelling better, we should at least be consistent.

rules/S2637/cfamily/rule.adoc
pdschbrt: I'm reading this on Github, but I currently fail to see the difference between the non-compliant and compliant code snippets. Could you point out the difference to me?

pdschbrt (changes requested):
Thanks for taking care of this and improving this information, @necto!

I've added a handful of minor comments.

rules/S2637/cfamily/metadata.json
pdschbrt: I just realized that my suggestion does not cover the return case. But perhaps you could also give it another thought and we might be able to come up with some better that is still reasonably short.

rules/S2637/cfamily/rule.adoc
necto: We start from 1, see [compiler explorer](https://compiler-explorer.com/z/EaqajEf77)

rules/S2637/cfamily/metadata.json
necto: How about this?
> "nonnull" parameters and return values of "returns_nonnull" functions should not be null

rules/S2637/cfamily/rule.adoc
necto: having no index means it applies to all parameters.
I think it is more readable for 1-argument function, but should not be used if the function takes multiple arguments.

rules/S2637/cfamily/rule.adoc
necto: Only the former. It is orthogonal with `returns_nonnull`.
I removed `returns_nonnull` because we have another example for it, and it does not change anything in this example

rules/S2637/cfamily/rule.adoc
necto: fixed in d1844af83e

rules/S2637/cfamily/rule.adoc
necto: fixed in d1844af83e

rules/S2637/cfamily/metadata.json
pdschbrt: Agreed. That's good!

rules/S2637/cfamily/rule.adoc
pdschbrt: Thanks for clarifying. It still looks weird, though 😅

rules/S2637/cfamily/rule.adoc
pdschbrt: Makes sense. Nice!

rules/S2637/cfamily/rule.adoc
pdschbrt: Thanks for clarifying.

rules/S2637/cfamily/rule.adoc
pdschbrt: Awesome 🎉

rules/S2637/cfamily/rule.adoc
pdschbrt: Yeah, it's more readable in that case. Cool.

pdschbrt (approved):
Good work.

If you adjust the title to what you've suggested, it is ready to be merged 🚀

PR #4238, opened 2024-09-04
Modify rule S6147: refactor directory structure, set quickfix and code-snippet language

necto:
I recommend review by commit


rules/S6147/cfamily/rule.adoc
```diff
@@ -1 +1,110 @@
-include::../rule.adoc[]
+== Why is this an issue?
```
pdschbrt, line 1: Aren't we missing a one-liner title here?

pdschbrt (changes requested):
It seems you only moved stuff around and fixed the encoding.

However, I think we lost the title while doing the clean-up. Please recover it.

rules/S6147/cfamily/rule.adoc
necto: AFAIK, no we are not.
The one-liner title comes from metadata.json, and it is added during rendering on top by the rspec frontend page for convenience.

necto: I simply inlined this `include::[]` directive, copying verbatim the (previously redacted) content of `../rule.adoc`

pdschbrt: I may have misremembered this. Thanks for double-checking.

pdschbrt (approved):

PR #4242, opened 2024-09-05
Modify rule S3329: Correct example code

daniel-teuchert-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3329/java/how-to-fix-it/java-cryptography-extension.adoc
+++ b/rules/S3329/java/how-to-fix-it/java-cryptography-extension.adoc
@@ -61,7 +61,7 @@ public void encrypt(String key, String plainText) {
     SecretKeySpec keySpec = new SecretKeySpec(key.getBytes(StandardCharsets.UTF_8), "AES");
 
     try {
-        Cipher cipher = Cipher.getInstance("AES/GCM/NoPadding");
+        Cipher cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
         cipher.init(Cipher.ENCRYPT_MODE, keySpec, iv);
 
     } catch(NoSuchAlgorithmException|InvalidKeyException|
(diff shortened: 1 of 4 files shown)


rules/S3329/java/how-to-fix-it/java-cryptography-extension.adoc
```diff
@@ -61,7 +61,7 @@ public void encrypt(String key, String plainText) {
     SecretKeySpec keySpec = new SecretKeySpec(key.getBytes(StandardCharsets.UTF_8), "AES");
 
     try {
-        Cipher cipher = Cipher.getInstance("AES/CBC/NoPadding");
+        Cipher cipher = Cipher.getInstance("AES/GCM/NoPadding");
```
gaetan-ferry-sonarsource, line 64: While I get that this was part of the original ticket, I don't get why we want to switch from CBC to GCM for the compliant code example. I think we should stick to using CBC with the pkcs5 padding as used in the noncompliant example.

rules/S3329/kotlin/how-to-fix-it/java-cryptography-extension.adoc
```diff
@@ -66,7 +66,7 @@ fun encrypt(key: String, plainText: String) {
     val keySpec = SecretKeySpec(key.toByteArray(StandardCharsets.UTF_8), "AES")
 
     try {
-        val cipher = Cipher.getInstance("AES/CBC/NoPadding")
+        val cipher = Cipher.getInstance("AES/GCM/NoPadding")
```
gaetan-ferry-sonarsource, line 69: While I get that this was part of the original ticket, I don't get why we want to switch from CBC to GCM for the compliant code example. I think we should stick to using CBC with the pkcs5 padding as used in the noncompliant example.

gaetan-ferry-sonarsource (changes requested):
I don't understand why we recommend using GCM in this rule. It does not seem to answer the issue we raise. Also, this is only done for java and kotlin while the Python example sticks to CBC for the compliant code sample.

gaetan-ferry-sonarsource (approved):
LGTM

PR #4243, opened 2024-09-06
Modify rule S1607: Allow to add an explanation comment after the test disabling

yassin-kammoun-sonarsource:


rules/S1607/javascript/rule.adoc
```diff
@@ -6,7 +6,7 @@ This rule raises an issue when a test construct from Jasmine, Jest, Mocha, or No
 
 == How to fix it in Jasmine
 
-A comment should be added before the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
+A comment should be added before or after the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
```
ilia-kebets-sonarsource, line 9:
```suggestion
A comment should be added before, on, or after the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
```

rules/S1607/javascript/rule.adoc
```diff
@@ -35,7 +35,7 @@ describe('foo', function() {
 
 == How to fix it in Jest
 
-A comment should be added before the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
+A comment should be added before or after the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
```
ilia-kebets-sonarsource, line 38:
```suggestion
A comment should be added before, on, or after the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
```

rules/S1607/javascript/rule.adoc
```diff
@@ -64,7 +64,7 @@ describe('foo', function() {
 
 == How to fix it in Mocha
 
-A comment should be added before the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
+A comment should be added before or after the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
```
ilia-kebets-sonarsource, line 67:
```suggestion
A comment should be added before, on, or after the line of the unit test explaining why the test was disabled. Alternatively, if the test is no longer relevant, it should be removed entirely.
```

PR #4250, opened 2024-09-09
Avoid race condition on validate_links cache

frederic-tingaud-sonarsource:


marco-antognini-sonarsource (approved):
LGTM, sounds like a smart fix to the problem.

PR #4251, opened 2024-09-09
Add rule S4328 to Sonar way quality profile

vdiez:


yassin-kammoun-sonarsource (approved):
Looks good to me. Should we merge this already or do you want to see how the improvement of the rule implementation goes first?

vdiez:
i'm working on the implementation still. I may change the rspec if I add some exceptions

PR #4303, opened 2024-09-19
Modify rule S3776: Clarify the behavior of function nesting

yassin-kammoun-sonarsource:


yassin-kammoun-sonarsource:
@SonarSource/analysis-js-squad Could someone review this please?

PR #4332, opened 2024-09-26
Add a new language identifier for Ansible

hendrik-buchwald-sonarsource:


rspec-tools/rspec_tools/utils.py
```diff
@@ -77,6 +78,7 @@
     'objectivec': 'objectivec',
     'vb': 'vb',
     # these languages are not supported by highlight.js as the moment:
+    'ansible': 'ansible',
```
pierre-loup-tristant-sonarsource, line 81:
```suggestion
    'objectivec': 'objectivec',
    'vb': 'vb',
     'ansible': 'yaml',
    # these languages are not supported by highlight.js as the moment:
```
Let's do the sane as kubernetes here.

PR #4362, opened 2024-10-02
Modify S6781(C#): Add alternative message

loris-s-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6781/csharp/rule.adoc
+++ b/rules/S6781/csharp/rule.adoc
@@ -31,7 +31,10 @@ ifdef::env-github,rspecator-view[]
 
 === Message
 
-JWT secret keys should not be disclosed.
+* When secrets are stored in configuration files:
+** Make sure that the config file containing the secret is inacessible to unauthorized users.
+* Other cases:
+** JWT secret keys should not be disclosed.
 
 === Highlight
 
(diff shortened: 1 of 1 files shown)


rules/S6781/csharp/rule.adoc
```diff
@@ -31,7 +31,10 @@ ifdef::env-github,rspecator-view[]
 
 === Message
 
-JWT secret keys should not be disclosed.
+* When secrets are stored in configuration files:
+** Make sure that the config file containing the secret is inacessible to unauthorized users.
```
jamie-anderson-sonarsource, line 35: I think that it would be better if the message mentioned JWT secret keys. There's only a concern when it comes to production secrets, so I've had a go at incorporating that into the message.

```suggestion
** Make sure that production JWT secret keys are not stored in source control.
```

jamie-anderson-sonarsource: I should also not that `inaccessible` is incorrectly spelled in the original message.

jamie-anderson-sonarsource (changes requested):
I think that this message needs improvement.

rules/S6781/csharp/rule.adoc
daniel-teuchert-sonarsource: I agree with Jamie's suggestion.

jamie-anderson-sonarsource (approved):
LGTM with the suggestion applied

PR #4380, opened 2024-10-07
SONARPY-2005: Modify rule S1721: Update rule for in (...)

Seppli11:
## Review

- [ ] text style and tone


Seppli11:
I'm not sure if it would be wise to add another example for the `for ... in (...)` example, or maybe the description could be more clearer

joke1196:
I think it is fine like this. The change you made is clear.

joke1196:
Before merging could you modify the PR name to be `Modify rule S... : ...` as it seems to be the standard. Thanks!

joke1196 (approved):
LGTM. Thanks for taking care of this.

PR #4398, opened 2024-10-09
Modify rule S5500: fix level of "How to fix it" section (CPP-5804)

marco-antognini-sonarsource:


frederic-tingaud-sonarsource (commented):
Excellent find!
There are multiple such errors. Could you see to add a rule in `rspec-tools\validation\description.py`?

PR #4407, opened 2024-10-10
Detect title as subsections

marco-antognini-sonarsource:
Existing violations were fixed:

 * #4408
 * #4409

PR #4408, opened 2024-10-10
Modify rules S2710,S5685: Fix How to fix it section level

marco-antognini-sonarsource:


marco-antognini-sonarsource:
@SonarSource/analysis-python-squad WYI, we are introducing a check in #4407 to prevent this typo from happening again in the future.

joke1196 (approved):
LGTM! Thank you for that!

PR #4409, opened 2024-10-10
Modify rule S2699: Fix How to fix it section level

marco-antognini-sonarsource:


rules/S2699/csharp/rule.adoc
```diff
@@ -32,7 +32,7 @@ By enforcing the presence of assertions, this rule aims to enhance the reliabili
 
 Test methods that include a call to a custom assertion method will not raise any issues.
 
-=== How to fix it
+== How to fix it
```
marco-antognini-sonarsource, line 35: This actually does not work [because](https://cirrus-ci.com/task/5126377391783936?logs=asciidoc_validation#L4):

```
Rule csharp:S2699 is mixing "How to fix it" with "How to fix it in FRAMEWORK NAME" sections. Either use a single "How to fix it" or one or more "How to fix it in FRAMEWORK"
```

marco-antognini-sonarsource, line 35: This actually does not work [because](https://cirrus-ci.com/task/5126377391783936?logs=asciidoc_validation#L4):

```
Rule csharp:S2699 is mixing "How to fix it" with "How to fix it in FRAMEWORK NAME" sections. Either use a single "How to fix it" or one or more "How to fix it in FRAMEWORK"
```

marco-antognini-sonarsource (commented):
@SonarSource/analysis-dotnet-squad I would need your input on this PR.

Essentially, we discovered that a few rules were using the wrong level for "How to fix it", resulting in the section appearing in the "Why is this an issue?" tab. We want to detect those errors and prevent them from happening in #4407. Hence, my poor attempt here.

Turns out that for this rule, you use frameworks, but there is also a generic How to fix it section. How do you propose updating the rspec so that we can introduce this check?

rules/S2699/csharp/rule.adoc
CristianAmbrosini: Hi Marco! So the "default" How to fix it section mentions a compliant solution using a custom assertion method, which is not part of the commonly used frameworks (MSTest, NUnit, xUnit, which, mentioned right after). 
We can call this generic section "How to Fix It in custom assertion framework" but it's very odd, without considering that the framework should be listed in the [allowed frameworks names](https://github.com/SonarSource/rspec/blob/8158b8b61a41c3cb853920e3bdf4bff3cf575d21/docs/header_names/allowed_framework_names.adoc#L18). What's your suggestion?

marco-antognini-sonarsource, line 35:
```suggestion
// This is on purpose not spelled "How to fix it" because:
//  - we cannot mix the generic section with framework-specific sections
//  - we cannot use a subsection with that name.
=== How can you fix it?
```

marco-antognini-sonarsource: What about having a subsection (not tab) named "How can you fix it?"? This way, the decision is obviously intentional to name it differently and it does not collide with CI validation rules.

WDYT?

CristianAmbrosini: I'm afraid it might be confusing, but I don't really see other solutions, so let's do that 👍

marco-antognini-sonarsource: I've applied the change but it seems I've messed up something while testing this on Friday -- my validation run failed. I'll fix it and ping you when it's _actually_ ready.

marco-antognini-sonarsource: @CristianAmbrosini  Sorry for the noise, I made a mistake and merged the wrong commit into my test branch. [All is good now](https://cirrus-ci.com/task/6038138240368640).

CristianAmbrosini (approved):
LGTM!

PR #4413, opened 2024-10-14
Add guidelines to format references to code elements

loic-joly-sonarsource:
Since the information is split into different files, I thought it better to duplicate the new guideline (I did not want to fully reorganize the doc...)

### DIFF
--- a/docs/description.adoc
+++ b/docs/description.adoc
@@ -274,12 +274,11 @@ In case no language is appropriate for a code block (for example shared examples
 === References within code blocks
 
 When referencing a name within a comment in a code example, use double quotes to make it clear it refers to an existing element in the code.
+
 [source,cpp]
 ----
 int i = 0;
-cout << noexcept(++i); // Noncompliant, "i" is not incremented -> Double quotes, ok
-cout << noexcept(++i); // Noncompliant, i is not incremented -> No quotes, bad
-cout << noexcept(++i); // Noncompliant, `i` is not incremented -> Backticks, bad
+cout << noexcept(++i); // Noncompliant, "i" is not incremented
 ----
 
 === Diff view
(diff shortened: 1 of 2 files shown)


docs/description.adoc
```diff
@@ -271,6 +271,17 @@ tsql:: use `sql`
 
 In case no language is appropriate for a code block (for example shared examples between multiple languages), you can use `text` as the language.
 
+=== References within code blocks
+
+When referencing a name within a comment in a code example, use double quotes to make it clear it refers to an existing element in the code.
+[source,cpp]
+----
+int i = 0;
+cout << noexcept(++i); // Noncompliant, "i" is not incremented -> Double quotes, ok
```
frederic-tingaud-sonarsource, line 280: I wouldn't write down bad examples here.
This documentation is already large and people skim through it. Unless you read very carefully, it is hard to see the small "ok" vs "bad" at the very end of the line.
On top of that, the rest of the doc doesn't have the same "ok / bad" pattern.

docs/styling_guide.adoc
```diff
@@ -109,3 +109,17 @@ Use it when referencing variable names, file names, tokens, and all kinds of spe
...
+
+When referencing elements from the code within a normal sentence, use the `backticks` (```) to format it. This includes variable names, function names, class names, and so on.
+
+When referencing the same elements within a comment in a code block, surrpond it with double quotes.
+[source,cpp]
+----
+int i = 0;
+// Write
+cout << noexcept(++i); // Noncompliant, "i" is not incremented -> Double quotes
+// Avoid
```
frederic-tingaud-sonarsource, line 122: Same here

frederic-tingaud-sonarsource (changes requested):

docs/styling_guide.adoc
loic-joly-sonarsource: Discussed: Not the same. We'll keep it that way.

frederic-tingaud-sonarsource (approved):

PR #4417, opened 2024-10-14
BUILD-6576 Print response text when failing to check a link

dorian-burihabwa-sonarsource:


edward-gonzales-sonarsource:
@dorian-burihabwa-sonarsource Shall we close this stale PR?

dorian-burihabwa-sonarsource:
Indeed, no need for this any more.

PR #4418, opened 2024-10-14
Modify rule S7007: Update example to trigger the rule

jamie-anderson-sonarsource:
While validating this rule, it was found that the noncompliant example in the RSPEC did not trigger. This is because the detection also looks for the secret ID, which is missing from the example.

The example has now been updated to include a secret ID and it now correctly triggers the rule.

## Review

- [ ] text style and tone


hendrik-buchwald-sonarsource (approved):
LGTM

PR #4426, opened 2024-10-18
Add Jira integration

pavel-mikula-sonarsource:
To review:

Make sure the files are the same as these
https://github.com/SonarSource/sonar-dotnet-enterprise/blob/master/.github/workflows/PullRequestClosed.yml
https://github.com/SonarSource/sonar-dotnet-enterprise/blob/master/.github/workflows/RequestReview.yml
https://github.com/SonarSource/sonar-dotnet-enterprise/blob/master/.github/workflows/SubmitReview.yml
Make sure that PullRequestCreated.yml has the correct Jira project on the last line
FYI:
Special GitHub tokens were added here https://github.com/SonarSource/re-terraform-aws-vault/pull/2437


marco-antognini-sonarsource (approved):
As discussed, we would like to prevent closing Jira tickets (at least CPP ones) if they have an open subtask.

AFAICT, this is orthogonal to this PR in terms of implementation.
If you agree with this statement, feel free to merge.

PR #4446, opened 2024-10-28
Modify rule S2245: Clarify the naming of random number generators

daniel-teuchert-sonarsource:
## Review

- [ ] text style and tone


rules/S2245/ask-yourself.adoc
```diff
@@ -1,7 +1,7 @@
 == Ask Yourself Whether
 
 * the code using the generated value requires it to be unpredictable. It is the case for all encryption mechanisms or when a secret value, such as a password, is hashed.
-* the function you use generates a value which can be predicted (pseudo-random).
+* the function you use is NOT a cryptographically secure pseudorandom number generator (CSPRNG).
```
pierre-loup-tristant-sonarsource, line 4: We usually avoid using negations in the "Ask Yourself Whether" part because it decreases readability.
This is probably why the previous sentence was chosen, even though I agree it's imperfect.

pierre-loup-tristant-sonarsource (approved):
LGTM.
I am not too sure if the update of the "Ask Yourself Whether" section is really an improvement (I let you decide).

rules/S2245/ask-yourself.adoc
daniel-teuchert-sonarsource: Good point! I tried to improve this. IMO, the original wording needed to be changed to be more aligned with the rest of the RSPEC on which PRNGs are okay to use in security-sensitive scenarios.

PR #4452, opened 2024-10-29
Modify S2053(C#): Improve code sample

loris-s-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2053/csharp/how-to-fix-it/dot-net.adoc
+++ b/rules/S2053/csharp/how-to-fix-it/dot-net.adoc
@@ -25,7 +25,9 @@ using System.Security.Cryptography;
 
 public static void hash(string password)
 {
-    var hashed = new Rfc2898DeriveBytes(password, 32, 10000, HashAlgorithmName.SHA256);
+    var saltSize = 32;
+    var iterations = 100_000;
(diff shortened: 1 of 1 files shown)


rules/S2053/csharp/how-to-fix-it/dot-net.adoc
```diff
@@ -25,7 +25,9 @@ using System.Security.Cryptography;
 
 public static void hash(string password)
 {
-    var hashed = new Rfc2898DeriveBytes(password, 32, 10000, HashAlgorithmName.SHA256);
+    var saltSize = 32;
+    var iterations = 100_000;
```
hendrik-buchwald-sonarsource, line 29:
```suggestion
    var iterations = 10_000;
```

Was it a deliberate change? Otherwise, I would say let's stay with 10k. A quick Google search also said that it is the recommended value.

loris-s-sonarsource: This is actually what we recommend in S5344: 
image

hendrik-buchwald-sonarsource: Ah, okay. Let's use it then.

PR #4453, opened 2024-10-30
Modify rule S1155: Adopt standard library method names

mostafa-mohammed-sonarsource:
Avoid confusion from mentioning `is_empty()` which is not part of the standard library, and rephrase the relevant parts of the RSPEC. Part of CPP-4776.

## Review

- [x] text style and tone


rules/S1155/cfamily/rule.adoc
```diff
@@ -2,7 +2,7 @@
 
 When you call `empty()` or `is_empty()`, it clearly communicates the code's intention, which is to check if the collection is empty. Using `size() == 0` for this purpose is less direct and makes the code slightly more complex.
 
-Moreover, depending on the implementation, the `size()`, `length()`, or `count()` methods can have a time complexity of `O(n)` where `n` is the number of elements in the collection. On the other hand, `empty()` and `is_empty()` simply check if there is at least one element in the collection, which is a constant time operation, `O(1)`.
+Moreover, depending on the implementation, the `size()`, `length()`, or `count()` methods can have a time complexity of `O(n)` where `n` is the number of elements in the collection. On the other hand, `empty()` and `is_empty()` simply check if there is at least one element in the collection, which is a constant time operation, `O(1)`. Note that only `empty()` is part of the standard library, but we report on methods with similar names.
```
michael-jabbour-sonarsource, line 5: I am not sure, but does this clarification also apply to `length()` or `count()`? If so, I would suggest restructuring the whole sentence to mention only methods that are part of the standard library, then adding a sentence in the end saying that the rule is also raised when methods with similar names are used on user-defined classes. WDYT?

michael-jabbour-sonarsource: Tagging @marco-antognini-sonarsource just to make sure my understanding of the ticket is correct :smile:

marco-antognini-sonarsource: I think this is a good idea. You can push it even a notch further: the rule is titled

> "empty()" or "is_empty()" should be used to test for emptiness

Removing the mention of `is_empty` here would also make sense and lighten the title.
The same information can be provided in the issue message. I see this on Peach and it's very reasonable:

> Use "isEmpty()" to check whether the container is empty or not.

----

BTW, the PR is titled "Clarify that only empty() is O(1)" but the current change seems to be about something else. This is not a big deal because if you apply @michael-jabbour-sonarsource's suggestion, the PR would get outdated anyway.

michael-jabbour-sonarsource: This is really helpful, thank you very much @marco-antognini-sonarsource :pray:

marco-antognini-sonarsource (commented):
This looks good. A small grammar nit. I leave the official review/approval to @michael-jabbour-sonarsource :)

rules/S1155/cfamily/rule.adoc
```diff
@@ -1,8 +1,8 @@
...
+When you call `empty()`, it clearly communicates the code's intention, which is to check if the collection is empty. Using `size() == 0` for this purpose is less direct and makes the code slightly more complex.
 
-Moreover, depending on the implementation, the `size()`, `length()`, or `count()` methods can have a time complexity of `O(n)` where `n` is the number of elements in the collection. On the other hand, `empty()` and `is_empty()` simply check if there is at least one element in the collection, which is a constant time operation, `O(1)`.
+Moreover, in the standard library, depending on the implementation, the `size()` method can have a time complexity of `O(n)` where `n` is the number of elements in the collection. On the other hand, `empty()` simply checks if there is at least one element in the collection, which is a constant time operation, `O(1)`. Note that this rule also identifies similar method names in user defined types, where the semantics and complexity may differ.
```
michael-jabbour-sonarsource, line 5:
```suggestion
Moreover, in the standard library, depending on the implementation, the `size()` method can have a time complexity of `O(n)` where `n` is the number of elements in the collection. On the other hand, `empty()` simply checks if there is at least one element in the collection, which is a constant time operation, `O(1)`. Note that this rule also identifies similar method names in user-defined types, where the semantics and complexity may differ.
```
One more grammar nit :smile:

michael-jabbour-sonarsource (approved):
LGTM! Let's merge this once SonarSource/sonar-cpp#4178 is updated with the new metadata.

PR #4462, opened 2024-10-31
Allow 'INFO' and 'BLOCKER' for rule quality severity

erwan-serandour:
With the introduction of the Multi-Quality Rule mode, there are now five different rule quality severities : ["INFO," "LOW," "MEDIUM," "HIGH," "BLOCKER"]. Allow INFO and BLOCKER for rule quality severity.


frederic-tingaud-sonarsource:
Hi Erwan, Thanks for the PR. At the same time as we introduce the 5 levels MQR, I would like to run a script that migrates all existing rules to these new levels. If we don't do both at the same time, it will be complex to know which rule is using the 5-levels scale and which one is using the 3-levels scale.
I'll try to do this early next week.

andrei-epure-sonarsource:
are the `"func": "Constant\/Issue",` changes safe? they seem unrelated.

andrei-epure-sonarsource:
out of curiosity - I skimmed through the changes and most seem HIGH->BLOCKER changes.

Were there any changes related to the new INFO level?

andrei-epure-sonarsource (approved):

marco-antognini-sonarsource:
> > are the `"func": "Constant\/Issue",` changes safe? they seem unrelated.
> 
> They are safe. We have an unnecessary escape here and when loading in memory, then dumping the JSON file, these get lost.

I vaguely remember a similar discussion... because what you say is the theory. 😅 

We have [this specific code](https://github.com/SonarSource/sonar-rule-api/blob/490c9baf6fdb78f99656ee028d9cc0ee409c79da/src/main/java/com/sonarsource/ruleapi/services/RuleFilesService.java#L120). Might be worth checking this does not break anything silently...

balazs-benics-sonarsource:
> are the `"func": "Constant\/Issue",` changes safe? they seem unrelated.

I'd propose splitting the PR, 1 adding the new categories, and 1 for changing the `Constant\/Issue` thingy. WDYT?
This is a huge PR, which deserves really narrow motivation.

frederic-tingaud-sonarsource:
I change the PR to put back the `\` in `"Constant\/Issue"`.

frederic-tingaud-sonarsource (approved):

PR #4502, opened 2024-11-14
Modify rule S6249: fix Terraform code examples

GabinL21:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6249/terraform/rule.adoc
+++ b/rules/S6249/terraform/rule.adoc
@@ -8,6 +8,7 @@ include::../recommended.adoc[]
 
 No secure policy is attached to this bucket:
 
+[source,terraform]
 ----
 resource "aws_s3_bucket" "mynoncompliantbucket" { # Sensitive 
   bucket = "mynoncompliantbucketname"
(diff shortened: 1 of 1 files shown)


rules/S6249/terraform/rule.adoc
```diff
@@ -31,9 +33,9 @@ resource "aws_s3_bucket_policy" "mynoncompliantbucketpolicy" {
       {
         Sid       = "HTTPSOnly"
         Effect    = "Deny"
-        Principal = [
-          "arn:aws:iam::123456789123:root"
-        ] # secondary location: only one principal is forced to use https
+        Principal = {
```
GabinL21, line 36: The previous code example was "right" too, but we're not currently detecting it with our rule implementation. 

This change ensures that we're documenting a case that is working with our current implementation, while the implementation itself will be updated in other tickets.

jonas-wielage-sonarsource: Pasting the noncompliant example does not raise an issue in the current implementation (if I didn't do any mistake)

jonas-wielage-sonarsource (commented):
Please change the PR title to reflect the RSPEC guidelines.
"Modify rule S6249: ..."

rules/S6249/terraform/rule.adoc
jonas-wielage-sonarsource: Nevermind, was a mistake on my setup.

PR #4528, opened 2024-11-22
Modify S7133: Update rule description

mary-georgiou-sonarsource:


rules/S7133/description-dotnet.adoc
```diff
@@ -11,6 +11,9 @@ This rule raises if you acquire a lock with one of the following methods, and do
 * https://learn.microsoft.com/en-us/dotnet/api/system.threading.spinlock.enter[SpinLock.Enter]
 * https://learn.microsoft.com/en-us/dotnet/api/system.threading.spinlock.tryenter[SpinLock.TryEnter]
 
+This rule will raise an issue when the code uses the https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose[disposable pattern]. This pattern makes locking easy to use and delegates the responsibility to the caller. Users should accept issues in sunc cases, as they should appear only once per https://learn.microsoft.com/en-us/dotnet/standard/threading/overview-of-synchronization-primitives[synchronization primitive].
```
pavel-mikula-sonarsource, line 14:
```suggestion
This rule will raise an issue when the code uses the https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose[disposable pattern]. This pattern makes locking easy to use and delegates the responsibility to the caller. Users should accept issues in such cases, as they should appear only once per https://learn.microsoft.com/en-us/dotnet/standard/threading/overview-of-synchronization-primitives[synchronization primitive].
```

pavel-mikula-sonarsource, line 14: synchronization primitive was unfortunate wording form me, it only covers the underlying primitives, not the more complex locks (like `ReaderWriterLock`). Let's use "once for each synchronization type"

pavel-mikula-sonarsource (changes requested):
Minor adjustment

PR #4532, opened 2024-11-25
Modify rule S3490 Improve RSPEC: explain that `= default` can be in TU/out of line CPP-5407

loic-joly-sonarsource:


rules/S3490/cfamily/rule.adoc
```diff
@@ -16,7 +16,10 @@ In most cases, you should just remove the code of the redundant function. In som
 For default constructors, you can often use the default version if you use in-class initialization instead of the initializer list. You must make it explicitly defaulted if your class has any other constructor.
 
 
-For destructors, you may want to use the ``++=default++`` syntax to be able to declare it as virtual (see S1235).
+For destructors, you may want to use the ``++= default++`` syntax in the following cases:
+
+* When you want to declare the destructor as _virtual_ (see S1235).
+* When your class contains smart pointers to incomplete types, and you want to delay the destructor definition to the point where the types are complete so that the smart pointer can propertly delete them. This commontly happens when using the PIMPL idiom. In that case, declare the destructor in the class and define it pout of line with `= default` when the type is complete.
```
frederic-tingaud-sonarsource, line 22:
```suggestion
* When your class contains smart pointers to incomplete types, and you want to delay the destructor definition to the point where the types are complete. This commonly happens when using the PIMPL idiom. In that case, declare the destructor in the class and define it out-of-line with `= default` when the type is complete so that the smart pointer can properly delete them.
```

PR #4534, opened 2024-11-25
Modify rule S5411: explain that @NonNull values are ignored.

tomasz-tylenda-sonarsource:
Modify S5411 explaining that @NonNull values are ignored. The rule was recently improved in SONARJAVA-5184 to cover more @NonNull variables.

### DIFF
--- a/rules/S5411/java/rule.adoc
+++ b/rules/S5411/java/rule.adoc
@@ -5,7 +5,7 @@ When boxed type `java.lang.Boolean` is used as an expression to determine the co
 
 It is safer to avoid such conversion altogether and handle the `null` value explicitly.
 
-Note, however, that no issues will be raised for Booleans that have already been null-checked.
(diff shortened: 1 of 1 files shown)


rules/S5411/java/rule.adoc
```diff
@@ -38,6 +38,35 @@ if(b != null){
 }
 ----
 
+=== Exceptions
+
+The issue is not raised if the expression is annotated `@NonNull` / `@NotNull`.
+This is useful if a boxed type is an instantiation of a generic type parameter
+and cannot be avoided.
```
dorian-burihabwa-sonarsource, line 45: This is nitpicking, but I think it is best to keep a sentence, even if it is long, on the same line so that changes are easier to spot in future changes.
```suggestion
This is useful if a boxed type is an instantiation of a generic type parameter and cannot be avoided.
```

PR #4537, opened 2024-11-26
Update S6584 and S6595 with gdebi package manager

rudy-regazzoni-sonarsource:


rules/S6595/docker/rule.adoc
```diff
@@ -7,6 +7,10 @@ Leaving unnecessary files in Docker image increases its size.
 The Docker images should be small and only contain necessary data.
 The cache index is obsolete after installation.
 
+=== Exceptions
+
+The rule does not raise when the update is followed by a `gdebi` package installation, as it can still install required dependencies when installing from file, which would require apt cache to be updated.
```
GabinL21, line 12:
```suggestion
The rule does not raise when the update is followed by a `gdebi` package installation, as it can still install required dependencies when installing from a file, which would require the `apt` cache to be updated.
```

For the "apt" one, I guess it would be clearer and more consistent to put it into backquotes

The others are coming from Grammarly

GabinL21 (approved):
LGTM! Check my small comment before merging 😉

PR #4542, opened 2024-11-26
Modify rule S2068: Focus on passwords only

yassin-kammoun-sonarsource:


vdiez (approved):
LGTM, but why did you copy/paste from root description? I don't see any changes

yassin-kammoun-sonarsource:
> LGTM, but why did you copy/paste from root description? I don't see any changes

It's using the term "credential" everywhere instead of "password".

PR #4559, opened 2024-12-05
Modify rule S3541: Deprecate the rule (CPP-5904)

loic-joly-sonarsource:


rules/S3541/cfamily/metadata.json
frederic-tingaud-sonarsource, line 1: The tags should also be removed

frederic-tingaud-sonarsource (changes requested):

frederic-tingaud-sonarsource (approved):

PR #4567, opened 2024-12-10
[DO NOT SQUASH]Improve coverage calculation

frederic-tingaud-sonarsource:


rspec-tools/rspec_tools/coverage.py
```diff
@@ -14,7 +14,6 @@
   'sonar-architecture',
   'sonar-cobol',
   'sonar-cpp',
-  'sonar-css',
```
marco-antognini-sonarsource, line 17: why this change?

rspec-tools/rspec_tools/coverage.py
```diff
@@ -191,11 +190,16 @@ def checkout_repo(repo):
   else:
     return Repo(repo)
 
+
+VERSION_RE = re.compile(r'\d[\d\.]+')
```
marco-antognini-sonarsource, line 194: Why introduce this filtering? (I'm not sure why we filter out tags with `-` in the first place...)

I assume there's a good reason that you've faced with specific repos.

marco-antognini-sonarsource (approved):

PR #4571, opened 2024-12-11
[NFC] Only fetch HEADs of tags & master for coverage computation

marco-antognini-sonarsource:
Download 375M instead of 712M.

### DIFF
--- a/rspec-tools/rspec_tools/coverage.py
+++ b/rspec-tools/rspec_tools/coverage.py
@@ -187,7 +187,9 @@ def checkout_repo(repo):
   if token:
     git_url=f"https://oauth2:{token}@github.com/SonarSource/{repo}"
   if not os.path.exists(repo):
-    return Repo.clone_from(git_url, repo)
+    repo = Repo.clone_from(git_url, repo, depth=1, single_branch=True)
+    repo.remote('origin').fetch('refs/tags/*:refs/tags/*', depth=1)
+    return repo
   else:
     return Repo(repo)
 


marco-antognini-sonarsource:
@frederic-tingaud-sonarsource 

I'm closing this PR because the effort to update the tests doesn't seem worth the speedup for a nightly run.

That said, it is interesting that there is any diff (see 59adaa1).

Other optimization ideas:

1. Do not check out any files: We spend a lot of time checking out versions that are not relevant (no sonarpedia.json) or writing files we don't care about on disk (implementations code).
It should be possible to use [`Tree`s](https://gitpython.readthedocs.io/en/stable/reference.html#module-git.objects.tree) to iterate over [`Blob`s](https://gitpython.readthedocs.io/en/stable/reference.html#module-git.objects.blob) for a given revision and find the relevant files. It's not hard but it requires modifying the implementation in many places because the assumption that files are on the disk is heavily baked in.

2. Alternatively, we could use a [sparse-checkout](https://git-scm.com/docs/git-sparse-checkout) and only checkout .json files. It is easy to do on the command line:
```
git clone $url $dir --sparse --depth 1 --single-branch master
git -C $dir fetch origin 'refs/tags/*:refs/tags/*' --depth 1
git -C $dir sparse-checkout set --no-cone '*.json'
```
This should not cause problems with the current implementation given we only care about JSON files.
But [GitPython](https://gitpython.readthedocs.io/en/stable/index.html) does not have any sparse checkout facility...

As a side note, if we need to extend this coverage tool, I'd suggest dropping `GitPython` and relying on `git` directly.

PR #4572, opened 2024-12-11
Add rule to product mapping file

frederic-tingaud-sonarsource:


marco-antognini-sonarsource (changes requested):
@frederic-tingaud-sonarsource If the goal is to merge this as-is, I would suggest explaining in the commit message how it was generated (and that we plan to make its generation automated).

If you share how you generated it, I can assess if the content is correct. Otherwise, it's just magic data. :-)

frederic-tingaud-sonarsource:
Replaced by #4576

PR #4575, opened 2024-12-12
RULEAPI-826 Compute rule / product mapping

frederic-tingaud-sonarsource:


rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +256,123 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
...
+def collect_coverage_per_product(rules_dir):
+  git_repo = checkout_repo('sonar-enterprise')
+  bundle_map = get_packaged_plugins(git_repo)
+  tags = git_repo.tags
+  tags.sort(key = lambda t: t.commit.committed_date)
+  versions = [tag.name for tag in tags if is_interesting_version(tag.name)]
+  plugin_versions = {}
+  for version in versions:
+    plugin_versions[version] = get_plugin_versions(git_repo, version)
+  build_rule_per_product(rules_dir, bundle_map, plugin_versions)
```
marco-antognini-sonarsource, line 378: I'm concern we do not sort things consistently with `collect_coverage_for_all_versions` (committed date vs tag).

Maybe the rest of the code handle this nicely, in which case I would remove this sort and test the implementation with a shuffle to make sure it's resilient.

I also think we can have a lightweight test here that cover scenarios like:

* SQS 2025.1 is released with plugin X.1 and thus S1
* (Assume 2025.1 is an LTA)
* SQS 2025.2 is released with plugin X.2 and drops S1
* bugfix SQS 2025.1.1 is released still with X.1 (or X.1.1) and with S1

or this one:

* SQS 2025.1 is released with plugin X.1 and _no_ S1
* (Assume 2025.1 is an LTA)
* SQS 2025.2 is released with plugin X.2 and introduces S1
* bugfix SQS 2025.1.1 is released still with X.1 (or X.1.1) and thus without S1

In terms of test implementation, I would suggest:

1. pass the `git_repo` as a parameter
2. create a fake `sonar-entreprise` for a couple of scenarios -- we just need tags & specific files.
3. create a fake plugin coverage JSON file for those scenarios
4. a big of pytest magic should be enough -- we may not even need to mock/patch anything

I appreciate we do not want complicated unit tests for each function (although it would be nice), but we can strike a balance between having nothing and something that cover the main corner cases from end-to-end.

rspec-tools/rspec_tools/coverage.py
marco-antognini-sonarsource, line 1: In terms of schema design, I realise now after the various changes that we have a suboptimal schema. Having `'SonarQube Server': { 'edition': 'Developer', 'version': xyz }` would be easier to consume than looking at which keys exists in the JSON object.

rspec-tools/rspec_tools/repo_plugin_mapping.json
marco-antognini-sonarsource, line 1: We should document this file and share its existance.

marco-antognini-sonarsource (changes requested):

rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +256,123 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
   collect_coverage_for_version(repo, git_repo, version, coverage)
   coverage.save_to_file(RULES_FILENAME)
 
+
+def get_plugin_versions(git_repo, version):
+  g = Git(git_repo)
+  repo_dir = git_repo.working_tree_dir
+  try:
+    with pushd(repo_dir):
```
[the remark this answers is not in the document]
frederic-tingaud-sonarsource: We need to access build.gradle from inside the sonar-enterprise repository folder otherwise git commands will not work.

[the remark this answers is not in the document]
marco-antognini-sonarsource: I don't I follow. We do `g.show(f'{version}:build.gradle')`, so we're not looking at the files on disk. And I would expect `g` to know which directory to use (otherwise, I'd git GitPython 🥇 for being crap :p).

rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +256,123 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
...
+        versions[m[2]] = m[3]
+      return versions
+  except Exception as e:
+    print(f"Sonar Enterprise {version} checkout failed: {e}")
+    raise
+
+def get_packaged_plugins(git_repo):
+  g = Git(git_repo)
+  repo_dir = git_repo.working_tree_dir
+  with pushd(repo_dir):
```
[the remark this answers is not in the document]
marco-antognini-sonarsource: discussed: `g` does not work when the PWD is not the expected one... sad about GitPython here.

rspec-tools/rspec_tools/coverage.py
```diff
@@ -49,6 +49,12 @@
 
 RULES_FILENAME = 'covered_rules.json'
 
+DEPENDENCY_RE = re.compile(r'''\bdependency\s+['"](?:com|org)\.sonarsource\.[\w.-]+:(?P<plugin_name>[\w-]+):(?P<version>\d+(\.\d+)+)['"]''')
+
+BUNDLED_SIMPLE = r'''['"](?:com|org)\.sonarsource\.(?:[\w.-]+):(?P<plugin_name>[\w-]+)['"]'''
```
marco-antognini-sonarsource, line 54:
```suggestion
BUNDLED_SIMPLE = r'''['"](?:com|org)\.sonarsource\.[\w.-]+:(?P<plugin_name>[\w-]+)['"]'''
```

rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +255,119 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
...
+  sys.exit(f'Couldnt find plugin {plugin}')
+
+
+def build_rule_per_product(bundle_map, plugin_versions):
+  rules_coverage = load_json(RULES_FILENAME)
+  rule_per_product = defaultdict(lambda: defaultdict(lambda: {}))
+  repo_plugin_mapping = load_json(Path(__file__).parent / 'repo_plugin_mapping.json')
+  for lang, rules in rules_coverage.items():
+    for rule, since in rules.items():
+      if not isinstance(since, str):
```
marco-antognini-sonarsource, line 338:
```suggestion
      if not isinstance(since, str):
        # The rule has an "until", therefore it does not exist anymore
        # and we don't need to report it in the produce mapping.
```

rspec-tools/rspec_tools/coverage.py
```diff
@@ -197,9 +202,13 @@ def is_version_tag(name):
...
-  if not is_version_tag(key):
-    return [0]
-  return list(map(int, key.split('.')))
+  v = key.removeprefix('sqcb-').removeprefix('sqs-')
+  if not is_version_tag(v):
+    if v == 'master':
+      return [0]
+    else:
+      sys.exit(f'Unexpected version {key}')
+  return list(map(int, v.split('.')))
```
marco-antognini-sonarsource, line 211: Let's make the flow more linear between condition and action
```suggestion
  if is_version_tag(v):
    return list(map(int, v.split('.')))
  if v == 'master':
    return [0]
  sys.exit(f'Unexpected version {key}')
  
```

rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +255,119 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
...
+def get_packaged_plugins(git_repo):
+  g = Git(git_repo)
+  repo_dir = git_repo.working_tree_dir
+  with pushd(repo_dir):
+    bundle_map = {}
+    for key, bundle in BUNDLES.items():
+      bundle_map[key] = []
+      content = g.show(f'master:{bundle}')
+      for m in re.finditer(BUNDLED_RE, content):
+        if m.groupdict()['plugin_name'] != None:
```
marco-antognini-sonarsource, line 285: I'll make some minor adjustments: `m.groupdict()['plugin_name']` can be simplified with `m['plugin_name']`.

rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +255,119 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
...
+EDITIONS =['Developer', 'Enterprise', 'Datacenter']
+
+
+def fill_product_mapping(plugin: str, bundle_map: dict, version: str, plugin_versions: dict, product_per_rule_per_lang: dict):
+  if plugin in bundle_map['Community Build']:
+    product_per_rule_per_lang['SonarQube Community Build'] = lowest_community_build_version(plugin_versions, plugin, version)
+    product_per_rule_per_lang['SonarQube Server'] = {
+      'minimal-edition': EDITIONS[0],
+      'since-version': lowest_server_version(plugin_versions, plugin, version)
+      }
```
marco-antognini-sonarsource, line 320: We'll get a hard error (thanks to the `sys.exit`) if a rule is released for the first time in a Community Build.

rspec-tools/rspec_tools/coverage.py
marco-antognini-sonarsource: e056c0f888

rspec-tools/rspec_tools/coverage.py
marco-antognini-sonarsource: f5d9bbf6ab

rspec-tools/rspec_tools/coverage.py
```diff
@@ -49,6 +49,12 @@
 
 RULES_FILENAME = 'covered_rules.json'
 
+DEPENDENCY_RE = re.compile(r'''\bdependency\s+['"](?:com|org)\.sonarsource\.[\w.-]+:(?P<plugin_name>[\w-]+):(?P<version>\d+(\.\d+)+)['"]''')
```
marco-antognini-sonarsource, line 52: Right now, in the code, we say `major >= 8` however this regex does not match for all plugins.

Example:
https://github.com/SonarSource/sonar-enterprise/blob/f869d840f12a4dfa69533e17d3b27558617a7f82/build.gradle#L128C35-L128C41

marco-antognini-sonarsource: 9.x looks compatible though: https://github.com/SonarSource/sonar-enterprise/blob/ce743843018827cb30c05735e7971523de34bb48/build.gradle#L171-L209

marco-antognini-sonarsource: Starting 10.0 we have an entry that does not match the regex but we should not care about this one so that's fine: https://github.com/SonarSource/sonar-enterprise/blob/75dae6a655f49ac78b0a786f6c19b15085561af9/build.gradle#L218

10.1 introduced a second one, but again we should be good. https://github.com/SonarSource/sonar-enterprise/blob/53c01c35c264c7e3d76cf5fb955de406f36b115e/build.gradle#L254-L255

marco-antognini-sonarsource: I've pushed 3b745315e4 to:

1. support `>=9` only
2. detect missing versions if variables are re-introduced for plugins in the future.

_If_ we need older coverage, we can accommodate that in a later PR. @frederic-tingaud-sonarsource can you validate this is OK for now?

rspec-tools/rspec_tools/coverage.py
frederic-tingaud-sonarsource: Oh, good point!
One cool thing we could do is handle it gracefully and return something like "Coming soon" or "Next release" in that case.

marco-antognini-sonarsource: I would suggest keeping this as-is for now and fixing this with RULEAPI-830 and a proper test (i.e., after RULEAPI-828) to focus on the essential right now.

rspec-tools/rspec_tools/coverage.py
frederic-tingaud-sonarsource: Thanks for catching these! It looks good to me.

rspec-tools/rspec_tools/coverage.py
marco-antognini-sonarsource: Tests will be covered by RULEAPI-828

rspec-tools/rspec_tools/coverage.py
frederic-tingaud-sonarsource: I made the change because it is a scenario that is supposed to happen very quickly after the release of the first LTA, as SQ Community Builds have a shorter cycle than SQ Enterprise.
I also realized that there was a bug in how we handled "master" branches, which once fixed, would trigger the same exception.

rspec-tools/rspec_tools/coverage.py
```diff
@@ -246,3 +254,125 @@ def update_coverage_for_repo_version(repo, version, rules_dir):
...
+  rules_coverage = load_json(RULES_FILENAME)
+  rule_per_product = defaultdict(lambda: defaultdict(lambda: {}))
+  repo_plugin_mapping = load_json(Path(__file__).parent / 'repo_plugin_mapping.json')
+  for lang, rules in rules_coverage.items():
+    for rule, since in rules.items():
+      if not isinstance(since, str):
+        # The rule has an "until", therefore it does not exist anymore
+        # and should not appear in the product mapping.
+        continue
+      target_repo, version = since.split(' ')
```
marco-antognini-sonarsource, line 345: We should probably skip the `master` version here since it's not available in the product anyway and we do not look at `master` versions of sonar-enterprise.

rspec-tools/rspec_tools/coverage.py
marco-antognini-sonarsource, line 1: I had a look at the data we produce and there is something fishy. I'll hold off merging this because I don't want the script to cause trouble during the holidays when we are not there -- it would not be a nice Xmas gift.

Looking at [S100](https://github.com/SonarSource/rspec/blob/e9a57f6817528b0bf4c5ad755e022082bef3026e/frontend/public/rule_product_mapping.json#L2C1-L122C5) (_very_ old rule), we say:

* for C/C++/ObjC: introduced in 9.0 Dev edition ✅ 
* for Apex: enterprise edition (probably correct), introduced in 10.7: does not match [reality](https://peach.aws-prd.sonarsource.com/coding_rules?languages=apex&q=S100&open=apex%3AS100) SQS report it being there since Nov 02, 2018.
* for C#: Dev edition (sounds correct) but introduced in SQS 10.8 / sqcb-24.12 but SQ [reports](https://peach.aws-prd.sonarsource.com/coding_rules?languages=cs&q=S100&open=csharpsquid%3AS100) the rule being introduced in 2015...

I don't know where we do things incorrectly...

marco-antognini-sonarsource (changes requested):

PR #4587, opened 2024-12-23
NET-913 Modify rule S1264: Improve description to match the implementation

sebastien-marichal:
[NET-913](https://sonarsource.atlassian.net/browse/NET-913)

Part of [NET-255](https://sonarsource.atlassian.net/browse/NET-255)

Fixes https://github.com/SonarSource/sonar-dotnet/issues/8249

## Review

- [ ] text style and tone

[NET-913]: https://sonarsource.atlassian.net/browse/NET-913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NET-255]: https://sonarsource.atlassian.net/browse/NET-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rules/S1264/csharp/rule.adoc
+++ b/rules/S1264/csharp/rule.adoc
@@ -1,4 +1,66 @@
-include::../rule.adoc[]
+== Why is this an issue?
+
+Using a `for` loop without its typical structure (initialization, condition, increment) can be confusing. In those cases, it is better to use a `while` loop as it is more readable.
+
+The initializer section should contain a variable declaration to be considered as a valid initialization.
+
+== How to fix it
+
+Replace the `for` loop with a `while` loop.
+
+=== Code example
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=1,diff-type=noncompliant]
+----
+for (;condition;) // Noncompliant; both the initializer and increment sections are missing
+{
+    // Do something
+}
+----
+
+==== Compliant solution
+
+[source,csharp,diff-id=1,diff-type=compliant]
+----
+while (condition)
+{
+    // Do something
+}
+----
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=2,diff-type=noncompliant]
+----
+int i;
+
+for (i = 0; i < 10;) // Noncompliant; the initializer section should contain a variable declaration
+{
+    // Do something
+    i++;
+}
+----
+
+==== Compliant solution
+
+[source,csharp,diff-id=2,diff-type=compliant]
+----
+int i = 0;
+
+while (i < 10)
+{
+    // Do something
+    i++;
+}
+----
+
+== Resources
+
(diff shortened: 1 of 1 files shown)


rules/S1264/csharp/rule.adoc
```diff
@@ -1,4 +1,66 @@
...
+}
+----
+
+==== Noncompliant code example
+
+[source,csharp,diff-id=2,diff-type=noncompliant]
+----
+int i;
+
+for (i = 0; i < 10;) // Noncompliant; the initializer section should contain a variable declaration
```
CristianAmbrosini, line 39: I the message correct? `the initializer section should contain a variable declaration`
Shouldn't it be `the initializer section should contain an increment statement`?

sebastien-marichal: The initializer section here contains only an initialization without a variable declaration.
The increment section is missing in this example.
The wording could be better

CristianAmbrosini: Is the rule reporting also for missing variable declarations?
ie:
```cs
var i;
for (i = 0; i <100; i++) { }
```

sebastien-marichal: Not it won't, for now it is checking if both the initializer and the increment section are valid.

CristianAmbrosini: Then `// Noncompliant; the initializer section should contain a variable declaration` is not a good comment. It shouldn't raise for that and it's not why the rule is raising in that case, it's just the missing incrementer. Do I miss something?

sebastien-marichal: It does raise because of the missing declaration.
If there were a variable declaration, the rule won't raise even though there is no incrementer.

The condition for the rule to raise is:
- missing initialization section 
- and, missing increment section

In the rule we check if there is a [variable declaration](https://github.com/SonarSource/sonar-dotnet-enterprise/blob/master/analyzers/src/SonarAnalyzer.CSharp/Rules/UseWhileLoopInstead.cs#L36). 
So if the content of the initialization section is not a declaration and the increment section is empty, we raise the rule.

sebastien-marichal: I think I will drop this example, it brings more confusion than it helps to understand the rule.
I have added in the description the following:
```
The initializer section should contain a variable declaration to be considered as a valid initialization.
```
And I think this should be enough for this case then.

CristianAmbrosini (approved):
LGTM!

PR #4595, opened 2025-01-06
Modify S3169: Add benchmarks

martin-strecker-sonarsource:
## Review

- [ ] text style and tone


rules/S3169/csharp/rule.adoc
```diff
@@ -1,9 +1,9 @@
...
 
-There's no point in chaining multiple ``++OrderBy++`` calls in a LINQ; only the last one will be reflected in the result because each subsequent call  completely reorders the list. Thus, calling ``++OrderBy++`` multiple times is a performance issue as well, because all of the sorting will be executed, but only the result of the last sort will be kept.
+There's no point in chaining multiple `OrderBy` calls in a LINQ; only the last one will be reflected in the result because each subsequent call  completely reorders the list. Thus, calling `OrderBy` multiple times is a performance issue as well, because all of the sorting will be executed, but only the result of the last sort will be kept.
 
 
-Instead, use ``++ThenBy++`` for each call after the first. 
+Instead, use `ThenBy` for each call after the first. 
 
 
 === Noncompliant code example
```
sebastien-marichal, line 9: This could be adapted to the LaYC format:

```
== How to fix it
=== Code examples
==== Compliant solution
```

With the correct `diff-id` and `diff-type` in the code snippet.

rules/S3169/csharp/rule.adoc
```diff
@@ -27,6 +27,83 @@ var x = personList
   .ToList();
 ----
 
+=== Benchmarks
+
+[options="header"]
+|===
+| Method                | Runtime              | Mean     | StdDev   | Allocated
+| OrderByAge            | .NET 9.0             | 11.26 ms | 0.710 ms |   1.53 MB
+| OrderByAgeOrderByAge  | .NET 9.0             | 13.23 ms | 0.178 ms |   3.06 MB
```
sebastien-marichal, line 36: Should this be `OrderByAgeTwice` to match the benchmark snippet?
```suggestion
| OrderByAgeTwice       | .NET 9.0             | 13.23 ms | 0.178 ms |   3.06 MB
```

rules/S3169/csharp/rule.adoc
```diff
@@ -27,6 +27,83 @@ var x = personList
   .ToList();
 ----
 
+=== Benchmarks
+
+[options="header"]
+|===
+| Method                | Runtime              | Mean     | StdDev   | Allocated
+| OrderByAge            | .NET 9.0             | 11.26 ms | 0.710 ms |   1.53 MB
```
sebastien-marichal, line 35: Any reason to not include .NET 8?

rules/S3169/csharp/rule.adoc
```diff
@@ -1,9 +1,9 @@
 == Why is this an issue?
 
-There's no point in chaining multiple ``++OrderBy++`` calls in a LINQ; only the last one will be reflected in the result because each subsequent call  completely reorders the list. Thus, calling ``++OrderBy++`` multiple times is a performance issue as well, because all of the sorting will be executed, but only the result of the last sort will be kept.
+There's no point in chaining multiple `OrderBy` calls in a LINQ; only the last one will be reflected in the result because each subsequent call  completely reorders the list. Thus, calling `OrderBy` multiple times is a performance issue as well, because all of the sorting will be executed, but only the result of the last sort will be kept.
```
sebastien-marichal, line 3:
```suggestion
There's no point in chaining multiple `OrderBy` calls in a LINQ; only the last one will be reflected in the result because each subsequent call completely reorders the list. Thus, calling `OrderBy` multiple times is a performance issue as well, because all of the sorting will be executed, but only the result of the last sort will be kept.
```

rules/S3169/csharp/rule.adoc
```diff
@@ -27,6 +27,83 @@ var x = personList
...
+    }).ToArray();
+}
+
+[Benchmark(Baseline = true)]
+public void OrderOnceByAge() =>
+    array.OrderBy(x => x.Age).Consume(consumer);
+
+[Benchmark]
+public void OrderByAgeTwice() =>
+    array.OrderBy(x => x.Age).OrderBy(x => x.Age).Consume(consumer);
```
sebastien-marichal, line 86: I am not sure these bring value to the rule.
I would drop them to only have the other two.
It has more focus on the multiple `OrderBy` and the `ThenBy` solution.

rules/S3169/csharp/rule.adoc
sebastien-marichal: Could be it dropped, see my next comment.

sebastien-marichal (changes requested):
I think we can make the benchmark simpler.

sebastien-marichal (approved):
LGTM!

PR #4597, opened 2025-01-07
SONARIAC-1856 Modify S7019: add EXEC alternatives and exceptions

rudy-regazzoni-sonarsource:


rules/S7019/docker/rule.adoc
```diff
@@ -22,6 +31,22 @@ FROM scratch
...
+----
+
+[source,docker,diff-id=3,diff-type=noncompliant]
+----
+FROM scratch
+ENTRYPOINT echo "Long script with chaining commands" \
+  && echo "Welcome!" \
+  && echo "Goodbye"
+----
+
```
jonas-wielage-sonarsource, line 49: I would remove the diff-id=3 noncompliant example. 
Keep only the compliant one with the wrapper script, but add the relation to a non compliant one.

jonas-wielage-sonarsource (approved):
Please have a look at my suggestions before merging.

rules/S7019/docker/rule.adoc
rudy-regazzoni-sonarsource: Ah very good point! I didn't know we could link multiple compliant examples to a single non-compliant one. Way better!

jonas-wielage-sonarsource: That's not what I meant, sorry for the confusion.
I meant that diff-id and diff-type will be removed from the wrapper file solution.
Unfortunately we can only use one diff-id per unique example, so I we need to adress this in a follow up PR 😅

PR #4664, opened 2025-02-14
Fix Secrets template rule

frederic-tingaud-sonarsource:
DefaultSeverity and code impact should be in sync with each other.


daniel-teuchert-sonarsource (approved):
LGTM! Thanks for fixing this 🙏

PR #4669, opened 2025-02-18
CPP-6157 Fix discrepancies between MQR and severity for CFamily rules

frederic-tingaud-sonarsource:
We noticed multiple discrepancies between MQR and defaultSeverity. This PR fixes those on CFamily-specific rules


rules/S7032/cfamily/metadata.json
```diff
@@ -17,7 +17,7 @@
   "quickfix": "infeasible",
   "code": {
     "impacts": {
-      "MAINTAINABILITY": "HIGH"
+      "MAINTAINABILITY": "MEDIUM"
```
loic-joly-sonarsource, line 20: HIGH seems ok for me here. I really don't understand the fact that `HIGH` and `CRITICAL` are related. Or to say in another way, in non-MQR mode, the names cover a 2 dimension design space, while in MQR mode, they convey a single dimension scale...

Anyways, I think there is a high impact / low likelihood risk there, in the case where the mutex is not locked.

rules/S7127/cfamily/metadata.json
```diff
@@ -8,7 +8,7 @@
   },
   "tags": [
   ],
-  "defaultSeverity": "Minor",
+  "defaultSeverity": "Major",
```
loic-joly-sonarsource, line 11: I think I would say `critical `/ `HIGH `on this rule...

loic-joly-sonarsource (changes requested):
Very unpleasant PR to review. Those levels are quite difficult to determine, and it's hard to reach any level of confidence in the estimation process...

rules/S7032/cfamily/metadata.json
frederic-tingaud-sonarsource: I put it as MEDIUM because I disagree with the mutex not locked case: creating an unnamed lock would be exactly the same bug outside an if and we already catch it with [S5184](https://sonarsource.github.io/rspec/#/rspec/S5184)

rules/S7127/cfamily/metadata.json
frederic-tingaud-sonarsource: Works for me. Done

rules/S7119/cfamily/metadata.json
```diff
@@ -9,16 +9,16 @@
...
   "ruleSpecification": "RSPEC-7119",
   "sqKey": "S7119",
   "scope": "All",
   "defaultQualityProfiles": ["Sonar way"],
   "quickfix": "infeasible",
   "code": {
     "impacts": {
-      "MAINTAINABILITY": "HIGH",
-      "RELIABILITY": "LOW"
+      "MAINTAINABILITY": "LOW",
```
loic-joly-sonarsource, line 20:
```suggestion
      "MAINTAINABILITY": "MEDIUM",
```
The impact I see is that the code might be working by luck, and some unrelated change breaks it in a non-obvious way, which make changing code complex.

loic-joly-sonarsource (approved):

PR #4673, opened 2025-02-19
SONARKT-569 Modify rule S4830: add support for WebViews

antonioaversa:
## Review

- [ ] text style and tone


rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
...
+----
+class MyWebViewClient : WebViewClient() {
+    override fun onReceivedSslError(view: WebView, handler: SslErrorHandler, error: SslError) =
+        handler.proceed() // Noncompliant
+}
+----
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
```
antonioaversa, line 23: Here I have included the validation sections that all the languages have in common. It is composed of three parts.

The first part is fairly general and it gives the advice to avoid the "invalid certificate" scenario altogether: `Addressing the vulnerability of disabled TLS certificate validation primarily involves re-enabling the default validation.`. I think the suggestion makes the most sense, even on Android.

The second part is the `Using trusted certificates` section, which seems to perfectly apply to Android.

The third part is the `Working with self-signed certificates or non-standard CAs` section, which may not be straighforward in Android: unlike the "Java Cryptography Extension" scenario, you cannot simply add a certificate to the local truststore via `keytool`, to my knowledge.

I kept it as it seems doable in principle, but I didn't include the `include::../../common/fix/keytool.adoc[]` section, which is really specific to the "Java Cryptography Extension" scenario.

Instead, I have added a `==== Implementing a server certificate validation` section, specific to Android, to show how to go about actually validating a certificate rejected by the system.

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
+== How to fix it in Android WebView
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+The certificate validation gets disabled by overriding the `onReceivedSslError` method of the `WebViewClient` class with an implementation that calls `SslErrorHandler.proceed()` unconditionally, and that never calls `SslErrorHandler.cancel()`.
```
antonioaversa, line 7: This could be factorized between Java and Kotlin, so that, when the rule will be implemented in Java, the text will be shared.

I didn't do it here just yet. I think it's better to extract that into a `/common/fix` section once we have this rule extension implemented in sonar-java.

docs/header_names/allowed_framework_names.adoc
```diff
@@ -47,7 +48,6 @@
 * Legacy Mongo Java API
 * OkHttp
 * Realm
-* Java Cryptography Extension
```
antonioaversa, line 50: This was repeated twice (here and [few lines above](https://github.com/SonarSource/rspec/pull/4673/files#diff-58a4a0dab03d1be43a178257907cc7a3ffbb9bc95a557b7e7ab2150c1b5ca1c0L41)). I don't think there is any good reason to do that, so I have removed the duplicate.

rules/S4830/kotlin/rule.adoc
```diff
@@ -10,11 +10,14 @@ include::../impact.adoc[]
...
 
+include::how-to-fix-it/android-webview.adoc[]
+
 == Resources
 
 include::../common/resources/standards.adoc[]
 
-* https://wiki.sei.cmu.edu/confluence/display/java/MSC61-J.+Do+not+use+insecure+or+weak+cryptographic+algorithms
+* CERT - https://wiki.sei.cmu.edu/confluence/display/java/MSC61-J.+Do+not+use+insecure+or+weak+cryptographic+algorithms
+* Android Documentation - https://developer.android.com/reference/android/webkit/WebViewClient?hl=en#onReceivedSslError(android.webkit.WebView,%20android.webkit.SslErrorHandler,%20android.net.http.SslError)[WebViewClient.onReceivedSslError] method
```
egon-okerman-sonarsource, line 20: Please also link https://support.google.com/faqs/answer/7071387?hl=en which is Google's official remediation

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+This means that a certificate initially rejected by the system will be accepted by the `WebViewClient`, regardless of its origin.
+
+=== Noncompliant code example
+
+[source,kotlin]
```
egon-okerman-sonarsource, line 13: Interestingly, I can see that the Java and Kotlin documentation was never updated to fully support LaYC. You can look at Docker as an example of the recommended structure.
The source tag needs extra information to support the diff view correctly (and we need to go back and fix this in the other parts of S4830 too).
```suggestion
[source,kotlin,diff-id=101,diff-type=noncompliant]
```
(diff-id cannot collide with any of the IDs in other languages, which is why I put 101 here.)

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
...
+    override fun onReceivedSslError(view: WebView, handler: SslErrorHandler, error: SslError) =
+        handler.proceed() // Noncompliant
+}
+----
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+==== Implementing a server certificate validation
```
egon-okerman-sonarsource, line 25: Like the comment above, to follow the format correctly, the title here should be renamed

 ```suggestion
==== Compliant solution
```

and it should be moved above "How to fix it". We generally keep the structure here of:

* Noncompliant code sample + short explanation if necessary
* Compliant code sample + short explanation if necessary
* Longer explanation and possible pitfalls

The text below and code sample are fine otherwise.

File: rules/S4830/kotlin/how-to-fix-it/android-webview.adoc [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+
+Alternatively, you need to implement a validation of the server certificate received in the `SslErrorHandler` object, calling `proceed` and `cancel` appropriately.
+
+[source,kotlin]
```
egon-okerman-sonarsource, line 29:
```suggestion
[source,kotlin,diff-id=101,diff-type=compliant]
```

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
...
+}
+----
+
+=== How does this work?
+
+include::../../common/fix/validation.adoc[]
+
+==== Implementing a server certificate validation
+
+Alternatively, you need to implement a validation of the server certificate received in the `SslErrorHandler` object, calling `proceed` and `cancel` appropriately.
```
egon-okerman-sonarsource, line 27: This sentence probably needs a rewrite considering it will be above the common fix text.

egon-okerman-sonarsource (changes requested):
Needs a few structural changes to be fully "LaYC up-to-date", which we apparently forgot to do for S4830 in Java/Kotlin.

antonioaversa:
> Needs a few structural changes to be fully "LaYC up-to-date", which we apparently forgot to do for S4830 in Java/Kotlin.

As suggested, I have adapted the "How to fix section" to have both compliant and non-compliant code examples, to have a coherent experience with other languages such as docker. 

In order to match that, I have moved the `==== Implementing a server certificate validation` up, and make it become `=== Compliant solution`, so that the two code snippets are now next to each other, and the code diff between the two makes sense. I have kept the `=== How does this work?` the same as it is in docker and other languages.

Let me know if that looks better for you.

Regarding LaYC: the migration of this rule for Kotlin was done in https://github.com/SonarSource/rspec/pull/2176
In general, I don't think that only having non-compliant makes this section non-fully LaYC compliant, as the "How to fix it" section can have any free structure, if that fits better the story-telling:
> This tab could also use a freeform 'story-telling' style if that makes it clearer for the user. Feel free to use any of the titles below, or any other titles you find useful. ([ref](https://github.com/SonarSource/rspec/blob/master/docs/description.adoc#:~:text=This%20tab%20could%20also%20use%20a%20freeform%20%27story%2Dtelling%27%20style%20if%20that%20makes%20it%20clearer%20for%20the%20user.%20Feel%20free%20to%20use%20any%20of%20the%20titles%20below%2C%20or%20any%20other%20titles%20you%20find%20useful.))

The optionality also applies to "How does this work", and any other level 3 section.

Given the above, I think that the way Java was done looks good to me: the non-compliant code example doesn't have a corresponding compliant, since all the proposed solutions for Java are ways to not use a custom trust manager at all, so I don't think we should bend the structure to meet a LaYC layout which is defined as optional, and proposed as an example of how things should look.

If anything I would remove the `==== Noncompliant code example` and `=== How does this work?` titles: the first kind of implies that there is a compliant solution, and the second in my opinion should be the incipit of the entire How to fix section. But I think this is more for a later change, across all languages.

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
antonioaversa: Yes, I have removed the "Alternatively" and moved above the entire section, so that the two code examples and next to each other, and it makes more sense for them to be in diff with each other.

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+The certificate validation gets disabled by overriding the `onReceivedSslError` method of the `WebViewClient` class with an implementation that calls `SslErrorHandler.proceed()` unconditionally, and that never calls `SslErrorHandler.cancel()`.
+
+This means that a certificate initially rejected by the system will be accepted by the `WebViewClient`, regardless of its origin.
+
+=== Noncompliant code example
```
egon-okerman-sonarsource, line 11:
```suggestion
==== Noncompliant code example
```

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+[source,kotlin,diff-id=101,diff-type=noncompliant]
+----
+class MyWebViewClient : WebViewClient() {
+    override fun onReceivedSslError(view: WebView, handler: SslErrorHandler, error: SslError) =
+        handler.proceed() // Noncompliant
+}
+----
+
+=== Compliant solution
```
egon-okerman-sonarsource, line 21:
```suggestion
==== Compliant solution
```

egon-okerman-sonarsource (approved):
LGTM, I suggested two changes for consistency with other security rules.

rules/S4830/kotlin/how-to-fix-it/android-webview.adoc
antonioaversa: Good catch!

PR #4717, opened 2025-02-28
S2486 in SonarJS is only for tests

zglicz:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S2486/javascript/metadata.json
+++ b/rules/S2486/javascript/metadata.json
@@ -1,2 +1,3 @@
 {
+  "scope": "Tests"
 }


vdiez (changes requested):
Why is this only for tests?

zglicz:
https://github.com/SonarSource/SonarJS/blob/master/sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/S2486.java You tell me :)

vdiez:
Our Java check is wrong. This rule is for all code. Check rule description

zglicz:
@vdiez - then I guess we are good, as with the new generation it will be for all code

PR #4731, opened 2025-03-10
RULEAPI-836 - Add `compatibleLanguages` to javascript rule manifests

ericmorand-sonarsource:
I did set the compatible languages to their actual key exposed by SonarJS. It is fundamental: RSPEC _is_ the specification that drives the plugin implementations. Putting "JAVASCRIPT" and "TYPESCRIPT" in RSPEC would render generating source codes impossible without having to know the mapping between these arbitrary keys and the actual language ones.

For information, here is the JavaScript script that I executed to update the rules from SonarJS's `master`.

```js
import {readdirSync, existsSync, readFileSync, writeFileSync} from "node:fs";
import {basename, extname, join} from "node:path";

const sourceDirectory = '/home/eric.morand/Projects/SonarJS/sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript';
const targetDirectory = '/home/eric.morand/Projects/rspec/rules';

const entries = readdirSync(sourceDirectory, {
    withFileTypes: true
});

for (const entry of entries) {
    if (entry.isFile && entry.name.endsWith(".json")) {
        if (entry.name === 'Sonar_way_profile.json') {
            continue;
        }
        
        const manifest = JSON.parse(readFileSync(join(sourceDirectory, entry.name), "UTF-8"));
        const compatibleLanguages = manifest["compatibleLanguages"].map((language) => {
            return language === "JAVASCRIPT" ? "js" : "ts";
        });
        
        const ruleName = basename(entry.name, extname(entry.name));
        const targetFile = join(targetDirectory, ruleName, 'javascript', 'metadata.json');
        
        if (existsSync(targetFile)) {
            const manifest = JSON.parse(readFileSync(targetFile, "UTF-8"));
            
            manifest["compatibleLanguages"] = compatibleLanguages;

            writeFileSync(targetFile, JSON.stringify(manifest, null, 2));
        }
    }
}
```

As a side-effect, it streamlined every javascript rule manifest to a 2-spaces identation.

To demonstrate that using SonarQube language keys in RSPEC makes generating the code easier and more generic, I pushed to RSPEC a dedicated branch providing `compatibleLanguages` to both S100 (`js` and `ts`) and S101 (`js` only), and updated the RSPEC Maven plugin, to very easily get the following correct results:

image

image

## Review

- [ ] text style and tone


rules/S1090/javascript/metadata.json
```diff
@@ -3,7 +3,7 @@
   "type": "CODE_SMELL",
   "status": "ready",
   "remediation": {
-    "func": "Constant\/Issue",
+    "func": "Constant/Issue",
```
vdiez, line 6: why are we missing the backslash in many jsons?

vdiez (approved):
some diffs are very different instead of just compatible languages? I guess it's due to some JSON strigify method you are using? pls double check the diff in SonarJS in jsons and html after running rule-api against this branch

rules/S1090/javascript/metadata.json
ericmorand-sonarsource: Because they were not needed to begin with, and `JSON.stringify` is actually correctly serializing the `/` character.

Also note this amusing part of Analyzer Commons:

https://github.com/SonarSource/sonar-analyzer-commons/blob/3c44ef68806f8e7c9b72c666da58e06ac6f7b688/commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java#L305

The part _after_ the `/` is not even considered. This is likely a legacy value.

ericmorand-sonarsource: > I guess it's due to some JSON strigify method you are using

Yes, the JSON are now formatted consistently - which was not the case before.

> pls double check the diff in SonarJS in jsons and html after running rule-api against this branch

Good point, give me a few minutes and I'll share the diff - ignoring the whitespaces.

ericmorand-sonarsource: @vdiez you can check the diff between current JSON files and their counterparts generated from this PR changes. Note the `extra` property that is now present, because our Maven plugin doesn't mutate the metadata. This property is not used and has no impact on SonarJS.

https://github.com/SonarSource/SonarJS/compare/rspec-compatible-languages

PR #4777, opened 2025-03-13
Modify rule S7180: improve description

leonardo-pilastri-sonarsource:
## Review

- [ ] text style and tone


rules/S7180/java/rule.adoc
```diff
@@ -1,8 +1,16 @@
 == Why is this an issue?
 
-Annotating interfaces or interface methods with ``++@Cache*++`` annotations is not recommended by the official Spring documentation.
+Annotating interfaces or interface methods with ``++@Cache*++`` annotations is not recommended by the official Spring documentation:
+```text
+Spring recommends that you only annotate concrete classes (and methods of concrete classes) with the @Cache* annotations, as opposed to annotating interfaces.
+You certainly can place an @Cache* annotation on an interface (or an interface method), but this works only if you use the proxy mode (mode="proxy").
+If you use the weaving-based aspect (mode="aspectj"), the caching settings are not recognized on interface-level declarations by the weaving infrastructure.
+```
 If you use the weaving-based aspect (mode="aspectj"), the ``++@Cache*++`` annotations will be ignored, and no caching proxy will be created.
```
erwan-serandour, line 9: we can remove this sentence the text above is explicit enough and we don't need to repeat ourselves

rules/S7180/java/rule.adoc
```diff
@@ -1,8 +1,16 @@
...
 
+Also, when a method is annotated as cacheable inside an interface, if two different implementations of that method exist, the first one to be invoked will populate the cache.
+Subsequent calls will always return the cached value, even if it's the other implementation being called.
+
 === What is the potential impact?
 
 * *Confusing Code*: Developers may mistakenly believe that caching is in effect, leading to confusion and incorrect assumptions about application performance.
```
erwan-serandour: let add also the effect, it is difficult to know what method populated the cache

erwan-serandour (changes requested):
quite better than the previous version

erwan-serandour (approved):

PR #4787, opened 2025-03-14
Modify rule S7186: update description to focus more explicitly on the Data Repository

erwan-serandour:


leonardo-pilastri-sonarsource (approved):

jonathanvila:
It would be worth it to revisit it, maybe in the future expanding the check to concrete classes. 
Here is a sample 
```
@Transactional
public class CustomerRepository {

    @PersistenceContext
    private EntityManager entityManager;

    public Customer save(Customer customer) {
        if (customer.getId() == null) {
            entityManager.persist(customer);
            return customer;
        } else {
            return entityManager.merge(customer);
        }
    }

    public Customer findById(Long id) {
        return entityManager.find(Customer.class, id);
    }

    public List<Customer> findAll() {
        return entityManager.createQuery("FROM Customer", Customer.class).getResultList();
    }

    public void deleteById(Long id) {
        Customer customer = findById(id);
        if (customer != null) {
            entityManager.remove(customer);
        }
    }

    public Page<Customer> findAll(Pageable pageable) {
        List<Customer> customers = entityManager.createQuery("FROM Customer", Customer.class)
            .setFirstResult((int) pageable.getOffset())
            .setMaxResults(pageable.getPageSize())
            .getResultList();

        long totalRecords = entityManager.createQuery("SELECT COUNT(c) FROM Customer c", Long.class)
            .getSingleResult();

        return new PageImpl<>(customers, pageable, totalRecords);
    }
}
```

And for the moment change the description of the rule specifying that it only considers interfaces (it's the most common case though)

PR #4882, opened 2025-04-04
JS-676 Update mapping for js/ts files

zglicz:
[JS-676](https://sonarsource.atlassian.net/browse/JS-676)

## Review

In https://github.com/SonarSource/rspec/pull/4731, the compatibleLanguages were introduced for js/ts rules. For backward compatibility for the coverage.py script from rspec, these need to mapped to the same key. This leads to a split brain, as the older rules would map to 'JAVASCRIPT' and 'TYPESCRIPT' and in master we have 'js' and 'ts'.

Let us update the mapping so that we are consistent. I really, really hope, this will be the last time we change these mappings. Twice should be enough 😄 

## test

Ran locally the coverage.py script and checked the contents of the generated `covered_rules.json`

`pipenv run rspec-tools update-coverage --repository SonarJS --rulesdir ../rules`

The resulting file has only 2 keys for languages: 'js' and 'ts' and majority of rules are covered since a long time. See excerpt:
`
    "S6859": "SonarJS 10.11.0.25043",
    "S6861": "SonarJS 10.11.0.25043",
    "S6957": "SonarJS 10.13.0.25911",
    "S6958": "SonarJS 10.13.0.25911",
    "S6959": "SonarJS 10.13.0.25911",
    "S7059": "SonarJS 10.15.0.27423",
    "S7060": "SonarJS 10.15.0.27423",
    "S878": "SonarJS 6.0.0.9595",
    "S881": "SonarJS 6.1.0.11503",
    "S888": "SonarJS 6.2.0.12043",
    "S905": "SonarJS 6.0.0.9595",
    "S909": "SonarJS 6.5.0.13383"
`

[JS-676]: https://sonarsource.atlassian.net/browse/JS-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rspec-tools/rspec_tools/coverage.py
+++ b/rspec-tools/rspec_tools/coverage.py
@@ -41,8 +41,10 @@ REPOS = [
 
 CANONICAL_NAMES = {
   'CLOUD_FORMATION': 'CLOUDFORMATION',
-  'JS': 'JAVASCRIPT',
-  'TS': 'TYPESCRIPT',
+  'JS': 'js',
+  'TS': 'ts',
+  'JAVASCRIPT': 'js',
+  'TYPESCRIPT': 'ts',
   'WEB': 'HTML'
 }
 
--- a/rspec-tools/tests/test_coverage.py
+++ b/rspec-tools/tests/test_coverage.py
@@ -129,11 +129,11 @@ def test_update_coverage_for_repo_version(tmpdir, rules_dir: Path, mock_git_anal
     coverage = tmpdir.join('covered_rules.json')
     assert coverage.exists()
     cov = load_json(coverage)
-    assert 'JAVASCRIPT' in cov
-    assert 'S100' in cov['JAVASCRIPT']
-    assert 'MethodName' not in cov['JAVASCRIPT'] # MethodName is a legacy key for S100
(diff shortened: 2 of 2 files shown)


necto (changes requested):
Did you check the effect on the frontend/public/covered_rules.json?
Note that the coverage script scans all released versions of the analyzers and gathers rule coverage for all of them.
In particular, I am concerned about the previous versions of SonarJS: are they still recognized to cover the same rules?

necto (commented):
Disregard my previous comment, you've addressed it in the PR summary

necto:
> Disregard my previous comment, you've addressed it in the PR summary

So it appears good to me. @frederic-tingaud-sonarsource , what do you think?

frederic-tingaud-sonarsource (approved):
Looking good to me!

necto (approved):

zglicz:
@frederic-tingaud-sonarsource @necto - When do you think the update rule coverage action/script will update the json file? Its been a week now almost.

PR #4890, opened 2025-04-05
Batch-reformat rspec-tools to follow the python conventions and enforce code style in CI

necto:
Add the default python codestyle tools: Black and usort and reformat code to enforce the consistent coding style.
The PR features massive changes due to batch reformatting. You can validate the changes by running the tools yourself:
```
cd rspec-tools
pipenv run black rspec_tools tests/
pipenv run usort format rspec_tools tests
```
Once merged, we should add .git-blame-ignore-revs file to exclude it from git blame.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -43,6 +43,8 @@ tooling_tests_task:
     - bash ci/fetch_branches.sh
     - cd rspec-tools
     - pipenv run pytest --cov=rspec_tools --cov-report=xml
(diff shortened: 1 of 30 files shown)


necto:
This PR relies on #4889

necto:
The force-pushed branch is the same as the original one. It was easier to recreate the branch then to resolve the numerous merge conflicts with `master`. Here is the full diff: https://gist.github.com/necto/9434723a061c9ffb0d010373a8611c07

PR #4903, opened 2025-04-06
Refactor and expand tests for check-links

necto:
Collect the file tree used to test different scenarios for check-links. Instead, create all the dummy files during the test setup in a temporary directory. This makes the test easier to understand for the LLM that does not need to crawl around the filesystem anymore.

Additionally, fix a test name that resulted in the test shadowing.

Additionally, avoid polutting the current directory with link_probes.history file that was generated by the tests

Also, add an array of new tests exercising the temporal conditions (link probed before, link inaccessible for a short period), and link aggregation (many links in a file, same link in multiple files). Incidentally, this increased the coverage of checklinks.py from 80 to 89% percent.

### DIFF
--- a/rspec-tools/rspec_tools/checklinks.py
+++ b/rspec-tools/rspec_tools/checklinks.py
@@ -239,7 +239,7 @@ def report_errors(errors, urls):
         show_files(urls[key])
 
 
-def check_html_links(dir, history_file="./link_probes.history"):
+def check_html_links(dir, history_file):
     load_url_probing_history(history_file)
     urls = get_all_links_from_htmls(dir)
     success = probe_links(urls)
--- a/rspec-tools/tests/test_links.py
+++ b/rspec-tools/tests/test_links.py
@@ -315,7 +315,7 @@ def test_url(setup_temp_dir):
 
 def test_ok(setup_temp_dir):
     temp_path = setup_temp_dir
-    
+
     # Create test directories and files
     test_dirs = {
         "OK": {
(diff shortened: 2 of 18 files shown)


necto:
This PR builds on #4890 which was necessary to productively employ LLM for collaboration

rspec-tools/tests/test_links.py
```diff
@@ -0,0 +1,741 @@
...
+
+@pytest.fixture
+def setup_test_files():
+    """Create a temporary directory with test files for each test case."""
+    # Create a temp directory
+    temp_dir = tempfile.mkdtemp()
+    temp_path = Path(temp_dir)
+
+    # Create test directories
+    test_dirs = {
```
frederic-tingaud-sonarsource, line 108: Perhaps the test would be more convenient to read if each test created its own files like in test_dead_link_in_multiple_files. That would put the content and the test on the same screen.

necto:
Force push: rebased on master, no conflicts

rspec-tools/tests/test_links.py
necto: Make sense. I refactored that, so now each test creates its own files

rspec-tools/rspec_tools/checklinks.py
```diff
@@ -241,11 +239,11 @@ def report_errors(errors, urls):
         show_files(urls[key])
 
 
-def check_html_links(dir):
-    load_url_probing_history()
+def check_html_links(dir, history_file="./link_probes.history"):
```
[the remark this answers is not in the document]
necto: no, removed
53c53c06a8 origin/az/linkcheck-testing Remove default value for the check_html_links parameter

PR #4952, opened 2025-04-11
JS-605 Update rules from 'All' -> 'Main'

zglicz:
[JS-605](https://sonarsource.atlassian.net/browse/JS-605)

Currently, we only support rules being either 'Main' or 'Test'. If a rule is set to be 'All', we default it to 'Main'. This is probably not what we want to do in the long run, but for now, this is the truth about what is happening behind the scenes.
Later, we might want to check the scope of the parent rule, if it is 'All', perhaps we might want to re-enable it for both 'Main' and 'Tests' ('All') in our rule as well.

Script used to make this update:
```import {readdir, stat, readFileSync, writeFileSync} from 'node:fs';
import * as path from 'path';

const directoryPath = './rules'; // Replace with your directory path

function readDirectory(directory: string): void {
    readdir(directory, (err, files) => {
        if (err) {
            console.error('Error reading directory:', err);
            return;
        }

        files.forEach(file => {
            const filePath = path.join(directory, file);
            stat(filePath, (err, stats) => {
                if (err) {
                    console.error('Error getting file stats:', err);
                    return;
                }

                if (stats.isDirectory()) {
                    readRuleFolder(filePath);
                }
            });
        });
    });
}

function readRuleFolder(folderPath: string): void {
    readdir(folderPath, (err, files) => {
        if (err) {
            console.error('Error reading javascript folder:', err);
            return;
        }

        // console.log(`Contents of rule folder:`, files);
        files.forEach(folder => {
            if (folder === 'javascript') {
                // read javascript json
                const jsMetadataPath = path.join(folderPath, folder, 'metadata.json');
                let jsContents = JSON.parse(readFileSync(jsMetadataPath, 'utf-8'));
                let ruleContents = JSON.parse(readFileSync(path.join(folderPath, 'metadata.json'), 'utf-8'));

                const derivedScope = (jsContents.scope ?? ruleContents.scope);
                if (derivedScope === 'All') {
                    console.log(folderPath, derivedScope);
                    jsContents.scope = 'Main';
                    writeFileSync(jsMetadataPath, JSON.stringify(jsContents, null, 2));
                }
            }
        });
    });
}

readDirectory(directoryPath);
console.log('running');```

[JS-605]: https://sonarsource.atlassian.net/browse/JS-605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdiez (approved):
LGTM

PR #4964, opened 2025-04-16
Modify rule S3236: Add rule exception

mary-georgiou-sonarsource:


rules/S3236/rule.adoc
```diff
@@ -41,5 +41,6 @@ void MyMethod()
 
 === Exceptions
 
-``++CallerMemberName++`` is not checked to avoid False-Positives with WPF/UWP applications.
-
+`CallerMemberName` is not checked to avoid False-Positives with WPF/UWP applications.
+`System.Diagnostics.Debug.Assert` is excluded. The signature of this method has been updated from .NET9 onwards from `public static void Assert(bool condition, string message)` to `public static void Assert([DoesNotReturnIf(false)] bool condition, [CallerArgumentExpression(nameof(condition))] string? message = null)`.
+The default message provided by this method is the failed assertion and its position in the code, however a custom message from the developer is many times preferred as an explanation of the failed assertion.
```
sebastien-marichal, line 46:
```suggestion
* `CallerMemberName` is not checked to avoid False-Positives with WPF/UWP applications.
* `System.Diagnostics.Debug.Assert` is excluded. The signature of this method has been updated from .NET9 onwards from `public static void Assert(bool condition, string message)` to `public static void Assert([DoesNotReturnIf(false)] bool condition, [CallerArgumentExpression(nameof(condition))] string? message = null)`.
The default message provided by this method is the failed assertion and its position in the code, however a custom message from the developer is many times preferred as an explanation of the failed assertion.
```

sebastien-marichal, line 46: I feel there is too much historical information.
Could we simplify to something like?:
```suggestion
`System.Diagnostics.Debug.Assert` is excluded as a custom message from the developer is sometimes preferred as an explanation of the failed assertion.
```

If we want to keep the decision history maybe we could add a comment in the implementation class?

sebastien-marichal (changes requested):

rules/S3236/rule.adoc
mary-georgiou-sonarsource: The repro has the ticket number which has all the info.
So I guess it's good :)

sebastien-marichal (approved):
LGTM!

PR #4967, opened 2025-04-22
Modify rule S6649: Add exceptions for x / 0

GabrielFleischer:
If the user writes 1 / 0, this is not a mistake

## Review

- [ ] text style and tone


rules/S6649/java/rule.adoc
```diff
@@ -54,6 +54,11 @@ void test_divide() {
 }
 ----
 
+
+=== Exceptions
+
+No issue is raised for `1 / 0` statements, as they are considered voluntary failures.
```
Swalkyn, line 60: I think we also don't raise for `x / 0`, so I would suggest:

```suggestion
No issue is raised when the divisor is the constant zero, as it is considered an intentional failure.
```

GabrielFleischer: I checked and you are correct, any constant denominator is excluded. So `(x + 1) / 0` is also not part of the rule.

Swalkyn (approved):
LGTM. Sorry I forgot this was assigned to me...

PR #4969, opened 2025-04-23
SONARJAVA-4881 Update S2230 to only target Spring <= 5

romainbrenguier:
[SONARJAVA-4881](https://sonarsource.atlassian.net/browse/SONARJAVA-4881)

Their is an associated PR on sonar-java to change the behaviour of the rule: [#5094](https://github.com/SonarSource/sonar-java/pull/5094)

## Review

- [ ] text style and tone

[SONARJAVA-4881]: https://sonarsource.atlassian.net/browse/SONARJAVA-4881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rules/S2230/java/rule.adoc
+++ b/rules/S2230/java/rule.adoc
@@ -1,9 +1,14 @@
 == Why is this an issue?
 
-Marking a non-public method `@Async` or `@Transactional` is misleading because Spring does not recognize non-public methods, and so makes no provision for their proper invocation.
+Marking a non-public method `@Async` or `@Transactional` is misleading because, up to version 5, Spring does not recognize non-public methods, and so makes no provision for their proper invocation.
 Nor does Spring make provision for the methods invoked by the method it called.
+Since Spring 6, protected and package-private methods can be handled, but the issue remains for private methods.
 
-Therefore marking a private method, for instance, @Transactional can only result in a runtime error or exception if the method is annotated as @Transactional.
+Therefore, marking a private method, for instance, @Transactional gives a false sense of security, and can lead to incorrect assumptions and potential bugs.
+
+== Exceptions
+
+The rule targets all non-public methods for project using Spring up to Spring 5, but only private methods for Spring 6 because Spring 6 takes into account all non-private methods.
 
 == How to fix it
 
@@ -39,6 +44,7 @@ public Future<String> asyncMethodWithReturnType() { // Compliant
 
 - https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/annotation/Async.html[Spring Framework API - Annotation Interface Async]
 - https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/annotation/Transactional.html[Spring Framework API - Annotation Interface Transactional]
+- https://docs.spring.io/spring-framework/reference/6.2.6/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility[Method visibility and @Transactional in proxy mode]
 
 === Articles & blog posts
 


rules/S2230/java/rule.adoc
```diff
@@ -1,9 +1,11 @@
 == Why is this an issue?
 
-Marking a non-public method `@Async` or `@Transactional` is misleading because Spring does not recognize non-public methods, and so makes no provision for their proper invocation.
+Marking a non-public method `@Async` or `@Transactional` is misleading because, up to version 5, Spring does not recognize non-public methods, and so makes no provision for their proper invocation.
 Nor does Spring make provision for the methods invoked by the method it called.
 
-Therefore marking a private method, for instance, @Transactional can only result in a runtime error or exception if the method is annotated as @Transactional.
+Therefore, marking a private method, for instance, @Transactional can only result in a runtime error or exception if the method is annotated as @Transactional.
```
SanderKnauff, line 6: I'm not sure if this is true actually. There just seems no way to invoke the private method through the proxy that is generated.
You can call this method just fine by calling it from a public method in the same class, but since it doesn't go through the generated proxy, it will not apply the effect of the `@Transactional` annotation.

```java
    void delegate() {
        privateMethod();
    }

    @Transactional
    private void privateMethod() {
        jdbcClient.sql("INSERT INTO foo(id) VALUES (1)").update();
        if (true) {
            throw new RuntimeException("Always crashes");
        }
        jdbcClient.sql("INSERT INTO foo(id) VALUES (2)").update();
    }

///
            try {
                testRepo.delegate();
            } catch (Exception e) {
                e.printStackTrace();
            }

            System.out.println("delegate " + testRepo.findAll());
            System.out.println(testRepo.findAll());
```
Will print `delegate [1]` even though the private transaction annotated method results in an exception and should have rolled back the insert.

romainbrenguier: You are right. I've rephrased that part.

SanderKnauff: Thank you, the new description aligns much better with the actual behavior

rules/S2230/java/rule.adoc
```diff
@@ -1,9 +1,11 @@
 == Why is this an issue?
 
-Marking a non-public method `@Async` or `@Transactional` is misleading because Spring does not recognize non-public methods, and so makes no provision for their proper invocation.
+Marking a non-public method `@Async` or `@Transactional` is misleading because, up to version 5, Spring does not recognize non-public methods, and so makes no provision for their proper invocation.
 Nor does Spring make provision for the methods invoked by the method it called.
 
-Therefore marking a private method, for instance, @Transactional can only result in a runtime error or exception if the method is annotated as @Transactional.
+Therefore, marking a private method, for instance, @Transactional gives a false sense of security, and can lead to incorrect assumptions and potential bugs.
+
+Note that the rule only targets Spring up to Spring 5, because Spring 6 takes into account all non-private methods.
```
dorian-burihabwa-sonarsource, line 8: Should we move this sentence into a new `Exceptions` section?

rules/S2230/java/rule.adoc
```diff
@@ -39,6 +41,7 @@ public Future<String> asyncMethodWithReturnType() { // Compliant
 
 - https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/annotation/Async.html[Spring Framework API - Annotation Interface Async]
 - https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/annotation/Transactional.html[Spring Framework API - Annotation Interface Transactional]
+- https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility[Method visibility and @Transactional in proxy mode]
```
dorian-burihabwa-sonarsource, line 44: Should we use the versioned URL here?
```suggestion
- https://docs.spring.io/spring-framework/reference/6.2.6/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility[Method visibility and @Transactional in proxy mode]
```

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):
Let's make sure to merge this update to the documentation in the same sonar-java PR as the implementation change.

PR #4978, opened 2025-04-29
DART-247 Modify rule S4830: Add Dart language

antonioaversa:
[DART-247](https://sonarsource.atlassian.net/browse/DART-247)

## Review

- [ ] text style and tone

[DART-247]: https://sonarsource.atlassian.net/browse/DART-247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rules/S4830/dart/how-to-fix-it/dart-io-library.adoc
+++ b/rules/S4830/dart/how-to-fix-it/dart-io-library.adoc
@@ -4,9 +4,7 @@
 
 include::../../common/fix/code-rationale.adoc[]
 
-:cert_method_name: badCertificateCallback
-
-include::../../common/fix/code-rationale-override.adoc[]
+The certificate validation gets disabled by implementing a `badCertificateCallback` that always return `true`. It is highly recommended to use the original implementation and to avoid any bypass.
 
 ==== Noncompliant code example
 
--- a/rules/S4830/dart/how-to-fix-it/dart-io-library.adoc
+++ b/rules/S4830/dart/how-to-fix-it/dart-io-library.adoc
@@ -1,0 +1,26 @@
+== How to fix it in dart:io library
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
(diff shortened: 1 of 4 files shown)


rules/S4830/dart/metadata.json
antonioaversa, line 1: Here I have copied the standards from the Kotlin version of the rule, as I they seem to equally apply to Dart.

rules/S4830/dart/how-to-fix-it/dart-io-library.adoc
```diff
@@ -0,0 +1,26 @@
+== How to fix it in dart:io library
+
+=== Code examples
+
+include::../../common/fix/code-rationale.adoc[]
+
+:cert_method_name: badCertificateCallback
+
+include::../../common/fix/code-rationale-override.adoc[]
```
pierre-loup-tristant-sonarsource, line 9: This part could be improved:

```suggestion
The certificate validation gets disabled by implementing a `badCertificateCallback` that always return `true`. It is highly recommended to use the original implementation and to avoid any bypass.
```

pierre-loup-tristant-sonarsource (approved):
LGTM. Have a look at the improvement suggestion I made before merging.

rules/S4830/dart/how-to-fix-it/dart-io-library.adoc
antonioaversa: Indeed, "empty" is not exact as the callback actually has a `return true`, at very least, in the case of Dart.

Unlike the suggestion above, I have kept the `include::../../common/fix/code-rationale.adoc[]` (`The following code contains examples of disabled certificate validation.`), since it still applies to Dart as to any other language, and also to keep the languages as aligned as possible.

pierre-loup-tristant-sonarsource (approved):
LGTM

PR #5049, opened 2025-05-13
Modify rule S6691: Use a truly non-compliant example

GabinL21:
## Context

The non-compliant example of S6691 should truly be non-compliant

## Review

- [ ] text style and tone


daniel-teuchert-sonarsource (approved):
LGTM!
I am not sure if we want to do this for every rule where some context that we expect is missing.
I think the code examples for the secret rules are a little bit different than for other languages where I think it is very important that the noncompliant code raises an issue.

In this case, I think adding the client id as context does not hurt and it is a nice improvement to the template.

PR #5051, opened 2025-05-14
SONARJAVA-5537 Modify rule S7476: update the documentation to corresponds to rule implementation

erwan-serandour:
[SONARJAVA-5537](https://sonarsource.atlassian.net/browse/SONARJAVA-5537)

[SONARJAVA-5537]: https://sonarsource.atlassian.net/browse/SONARJAVA-5537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rules/S7476/java/rule.adoc
```diff
@@ -7,9 +7,9 @@ Accidentally writing comments with three or more slashes can lead to unintended
 
 === What is the potential impact?
 
-* Accidental JavaDoc generation where normal comments were intended.
-* Misleading or broken documentation output.
-* Increased maintenance burden when upgrading to Java 23 or later.
+* accidental JavaDoc generation where normal comments were intended.
+* misleading or broken documentation output.
+* increased maintenance burden when upgrading to Java 23 or later.
```
romainbrenguier, line 12: I'm not sure about this change. Could we make full sentences instead?

erwan-serandour: :+1:

rules/S7476/java/rule.adoc
```diff
@@ -1,25 +1,27 @@
 == Why is this an issue?
 
-Starting in Java 23, comments beginning with three slashes ``++///++`` are interpreted as JavaDoc comments using Markdown syntax.
+Starting in Java 23, comments beginning with three slashes ``++///++`` are interpreted as Javadoc comments using Markdown syntax.
```
tomasz-tylenda-sonarsource, line 3: Why was this changed? It's called "JavaDoc" in CamelCase, for example, in https://openjdk.org/jeps/467

rules/S7476/java/rule.adoc
```diff
@@ -1,25 +1,27 @@
...
 In Java 22 and earlier, comments starting with more than 2 slashes were treated as normal comments.
-Accidentally writing comments with three or more slashes can lead to unintended JavaDoc being generated, when migrating to Java 23.
+Accidentally writing comments with three or more slashes can lead to unintended Javadoc being generated, when migrating to Java 23.
 
 === What is the potential impact?
 
-* Accidental JavaDoc generation where normal comments were intended.
-* Misleading or broken documentation output.
-* Increased maintenance burden when upgrading to Java 23 or later.
+Starting single-line comments with three or more slashes will increase the migration cost when upgrading to Java 23 or later.
```
tomasz-tylenda-sonarsource, line 10: Also, ide and other tools may potentially treat such comments and Markdown JavaDoc, if they are not aware of the Java version being used.

rules/S7476/java/rule.adoc
```diff
@@ -1,25 +1,27 @@
...
+
+=== Exceptions
+
+The rule does not raise issues on license headers at the start of files because they will not be considered as Javadoc comments.
```
tomasz-tylenda-sonarsource, line 14: I'm not sure about the latter part (after "because"). The first comment in the file could be a JavaDoc if there is no copyright header.

tomasz-tylenda-sonarsource: Oh, it's because it's before the package declaration.

rules/S7476/java/rule.adoc
erwan-serandour: it seems that both are used, indeed I prefer the former

rules/S7476/java/rule.adoc
erwan-serandour: :+1:

rules/S7476/java/rule.adoc
erwan-serandour: :+1:

rules/S7476/java/rule.adoc
```diff
@@ -1,17 +1,19 @@
...
 === What is the potential impact?
 
 Starting single-line comments with three or more slashes will increase the migration cost when upgrading to Java 23 or later.
+IDE or other tools may handle such comments as JavaDoc comments if they are not aware of the Java version.
 
 === Exceptions
 
-The rule does not raise issues on license headers at the start of files because they will not be considered as Javadoc comments.
+- The rule does not raise issues on license headers at the start of files because they will not be considered as JavaDoc comments.
+- The rule does not raise issues in projects using Java 16 or below.
```
tomasz-tylenda-sonarsource, line 16: Is it only because it would be a big jump?

rules/S7476/java/rule.adoc
```diff
@@ -7,19 +7,23 @@ Accidentally writing comments with three or more slashes can lead to unintended
 
 === What is the potential impact?
 
-* Accidental JavaDoc generation where normal comments were intended.
-* Misleading or broken documentation output.
-* Increased maintenance burden when upgrading to Java 23 or later.
+Starting single-line comments with three or more slashes will increase the migration cost when upgrading to Java 23 or later.
+IDE or other tools may handle such comments as JavaDoc comments if they are not aware of the Java version.
```
tomasz-tylenda-sonarsource, line 11: Let's add some connector, so it reads more natural. For example "Moreover, IDEs or other tools.."

rules/S7476/java/rule.adoc
erwan-serandour: :+1:

rules/S7476/java/rule.adoc
erwan-serandour: :+1:

rules/S7476/java/metadata.json
```diff
@@ -10,16 +10,15 @@
...
   "ruleSpecification": "RSPEC-7476",
   "sqKey": "S7476",
   "scope": "All",
   "defaultQualityProfiles": ["Sonar way"],
   "quickfix": "unknown",
   "code": {
     "impacts": {
-      "MAINTAINABILITY": "MEDIUM",
-      "RELIABILITY": "MEDIUM"
+      "MAINTAINABILITY": "LOW"
```
tomasz-tylenda-sonarsource, line 21: Do you think this should be LOW or could we go even lower and change it to INFO?

https://github.com/SonarSource/rspec/blob/master/docs/metadata.adoc#code-field is the documentation. I looked at Java checks with "MAINTAINABILITY": "INFO".

rules/S7476/java/rule.adoc
```diff
@@ -8,16 +8,16 @@ Accidentally writing comments with three or more slashes can lead to unintended
...
 === Exceptions
 
 - The rule does not raise issues on license headers at the start of files because they will not be considered as JavaDoc comments.
-- The rule does not raise issues in projects using Java 16 or below.
+- The rule only raises issues in projects using Java 17 or above, because those projects will likely be migrated to Java 23 or later.
 
 == How to fix it
 
-In versions of Java prior to 23, all single-line comments should always start with exactly 2 slashes.
+In versions of Java prior to 23, all comments should always start with exactly 2 slashes, and from Java 23 forward they should not start with more than 3.
```
tomasz-tylenda-sonarsource, line 20: Could this be confusing? "All comments" could sound like we disallow `/* */`. Should we change it to end-of-line comments (I took this term from https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.7)

rules/S7476/java/rule.adoc
```diff
@@ -8,16 +8,16 @@ Accidentally writing comments with three or more slashes can lead to unintended
 === What is the potential impact?
 
 Starting single-line comments with three or more slashes will increase the migration cost when upgrading to Java 23 or later.
```
tomasz-tylenda-sonarsource, line 10: Is `single-line comment` a common term? I think it could be confused with `/* */` on a single line. JLS calls is end-of-line comment.

erwan-serandour: We can remove the mention to single-line comment, I think no body use end-of-line comment. Single-line comments is more widely used, but may confuse some people.

rules/S7476/java/metadata.json
erwan-serandour: I like LOW. I feel the issue is a bit stronger than INFO.

rules/S7476/java/rule.adoc
erwan-serandour: let's not talk about single line or end of line comments, most people do not use the term of the jls. I will use at most so that it applies to everything.

tomasz-tylenda-sonarsource (approved):

PR #5073, opened 2025-05-23
Update architecture rule status to "beta"

irina-batinic-sonarsource:
## Review

A dedicated reviewer checked the rule description successfully for:

- [ ] logical errors and incorrect information
- [ ] information gaps and missing content
- [ ] text style and tone
- [ ] PR summary and labels follow [the guidelines](https://github.com/SonarSource/rspec/#to-modify-an-existing-rule)

PR #5074, opened 2025-05-23
Revert "Update architecture rule status to "beta""

irina-batinic-sonarsource:
Reverts SonarSource/rspec#5073

It is decided that in the end we don't want to change status.


mike-swainston-rainford-sonar (approved):
Lgtm

PR #5086, opened 2025-05-27
SONARJAVA-5581 modify S3033: add performance benchmark table

erwan-serandour:
[SONARJAVA-5581](https://sonarsource.atlassian.net/browse/SONARJAVA-5581)

[SONARJAVA-5581]: https://sonarsource.atlassian.net/browse/SONARJAVA-5581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rules/S3033/java/rule.adoc
```diff
@@ -1,31 +1,96 @@
 == Why is this an issue?
 
-Assembling a ``++StringBuilder++`` or ``++StringBuffer++`` into a ``++String++`` merely to see if it's empty is a waste of cycles. Instead, jump right to the heart of the matter and get its ``++.length()++`` instead.
+Assembling a ``++StringBuilder++`` or ``++StringBuffer++`` into a ``++String++`` merely to see if it's empty is a waste of cycles. Instead, jump right to the heart of the matter and use ``++.isEmpty++`` instead.
```
romainbrenguier, line 3: This sounds very informal and not precise or succinct. (also instead is repeated twice in the same sentence).
I'd suggest for instance: "It is inefficient to build a String from a StringBuilder or StringBuffer just to check if it's empty. Instead, directly use the .isEmpty() method.

erwan-serandour: :+1:

PR #5092, opened 2025-05-28
SONARJAVA-5589 modify S1643: add performance benchmark table

erwan-serandour:


rules/S1643/java/rule.adoc
```diff
@@ -26,6 +26,45 @@ StringBuilder bld = new StringBuilder();
...
+The results were generated by running the following snippet with https://github.com/openjdk/jmh[JMH]:
+
+[source,java]
+----
+@Param({"100", "1000", "25000"})
+int size;
+private String word = "append";
+
+@Benchmark
+public String concat() {
```
tomasz-tylenda-sonarsource, line 49: Let's change it to some other name, for example, `plus`. The reason is that String has a method `concat` (which, btw, we could benchmark as well), so the current name could be confusing.

rules/S1643/java/rule.adoc
```diff
@@ -26,6 +26,45 @@ StringBuilder bld = new StringBuilder();
...
+| stringBuilder| 100| Temurin 21| 0.93 us/op| ±0.03 us/op
+| stringBuilder| 1000| Temurin 21| 9.27 us/op| ±0.49 us/op
+| stringBuilder| 25000| Temurin 21| 242 us/op| ±4.86 us/op
+|===
+
+The results were generated by running the following snippet with https://github.com/openjdk/jmh[JMH]:
+
+[source,java]
+----
+@Param({"100", "1000", "25000"})
```
tomasz-tylenda-sonarsource, line 44: Could we change it to 100, 100, 10000, 100000? It's easier to make sense of this data if the rows in the benchmark table differ by a constant factor.

tomasz-tylenda-sonarsource (changes requested):

rules/S1643/java/rule.adoc
```diff
@@ -29,24 +29,29 @@ StringBuilder bld = new StringBuilder();
...
-| concat| 100| Temurin 21| 4.16 us/op| ±0.15 us/op
-| concat| 1000| Temurin 21| 3.66 us/op| ±5.64 us/op
-| concat| 25000| Temurin 21| 296000 us/op| ±4160 us/op
-| stringBuilder| 100| Temurin 21| 0.93 us/op| ±0.03 us/op
-| stringBuilder| 1000| Temurin 21| 9.27 us/op| ±0.49 us/op
-| stringBuilder| 25000| Temurin 21| 242 us/op| ±4.86 us/op
+| plus| 100| Temurin 21| 4.19 us/op| ±0.34 us/op
+| plus| 1000| Temurin 21| 377.08 us/op| ±17.36 us/op
+| plus| 10000| Temurin 21| 40221.49 us/op| ±1342.76 us/op
+| plus| 100000| Temurin 21| 5286.84 ms/op| ±185.79 ms/op
```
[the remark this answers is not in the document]
romainbrenguier: Also I'm not sure `us` is clear to everyone to mean micro seconds, could we use &#956;s instead? (we can use `&#956;` for &#956;)

tomasz-tylenda-sonarsource (approved):

PR #5137, opened 2025-06-18
Move java S6555 description to java S2259

andreaguarino:
## Review

- [ ] text style and tone


rules/S2259/java/rule.adoc
```diff
@@ -1,33 +1,69 @@
...
 ==== Noncompliant code example
 
-include::noncompliant-code.adoc[]
+[source,java,diff-id=1,diff-type=noncompliant]
+----
+public boolean isNameEmpty(String givenName, String familyName) {
+  if (givenName != null && familyName != null) {
+    return givenName.isEmpty() && familyName.isEmpty();
+  }
+  return familyName.isEmpty(); // Noncompliant; familyName may be null
```
Swalkyn, line 17: Because we translate the left-hand side of the `&&` expression as an implicit condition, we don't actually raise here. We should update this example. Maybe we can replace it by some of the example from the previous S2259 description?

andreaguarino: good catch, thanks!

Swalkyn (approved):

PR #5177, opened 2025-07-04
Modify rule S3267: Provide better examples

alex-meseldzija-sonarsource:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S3267/csharp/rule.adoc
+++ b/rules/S3267/csharp/rule.adoc
@@ -81,17 +81,26 @@ List<string> Method(IEnumerable<MyDto> collection)
 
 [source,csharp,diff-id=3,diff-type=noncompliant]
 ----
-async Task<List<string>> Method(IAsyncEnumerable<string> collection)
+public void Method(List<string> list)
 {
-  var result = new List<string>();
-  await foreach (var element in collection) // Noncompliant
+  foreach (var element in list)
   {
-    if (element != null)
+    var someValue = element.Length;
+  }
+}
+----
+
+[source,csharp,diff-id=4,diff-type=noncompliant]
+----
+void Method(List<int> list)
+{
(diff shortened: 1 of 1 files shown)


rules/S3267/csharp/rule.adoc
```diff
@@ -81,17 +81,26 @@ List<string> Method(IEnumerable<MyDto> collection)
 
 [source,csharp,diff-id=3,diff-type=noncompliant]
 ----
-async Task<List<string>> Method(IAsyncEnumerable<string> collection)
+public void Method(List<string> list)
```
sebastien-marichal, line 84: I think we should keep am example with IAsyncEnumerable as we explicitly mention it in the How to fix it section.

It was added after this community post: https://community.sonarsource.com/t/rspec-3267-wrongly-triggering-on-iasyncenumerable/139018

sebastien-marichal (changes requested):
We should keep IAsyncEnumerable example

rules/S3267/csharp/rule.adoc
alex-meseldzija-sonarsource: Good point!

rules/S3267/csharp/rule.adoc
```diff
@@ -81,17 +81,26 @@ List<string> Method(IEnumerable<MyDto> collection)
...
   {
-    if (element != null)
+    var someValue = element.Length;
+  }
+}
+----
+
+[source,csharp,diff-id=4,diff-type=noncompliant]
+----
+async void Method(List<Task<int>> tasks)
```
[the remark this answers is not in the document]
alex-meseldzija-sonarsource: Task.WhenEach() is an async enumerable.
But you're right, this should be as explicit as possible.

PR #5200, opened 2025-07-15
S5144 S6173 S6350 S6399 S6641 S7044

roberto-orlandi-sonarsource:


rules/S6173/vbnet/how-to-fix-it/dotnet.adoc
```diff
@@ -35,30 +35,30 @@ End Interface
...
-        If Not EFFECT_ALLOW_LIST.Contains(EffectName) Then
+        If Not EffectAllowList.Contains(EffectName) Then
             Return BadRequest("Invalid effect name. The effect is not allowed.")
         End If
 
         Dim EffectInstance = Activator.CreateInstance(Nothing, EffectName)
         Dim EffectPlugin As Object = EffectInstance.Unwrap()
 
-        If CType(EffectPlugin, IEffect).ApplyFilter() Then
+        If CType(EffectPlugin, IEffectCompliant).ApplyFilter() Then
```
renaud-tognelli-sonarsource, line 53: I don't think the interface is suppose to change name between the two implementations. 
I think this could be confusing for reader, wdyt?

```suggestion
        If CType(EffectPlugin, IEffect).ApplyFilter() Then
```

rules/S6641/vbnet/how-to-fix-it/dotnet-core.adoc
```diff
@@ -15,37 +15,45 @@ malicious, or must be properly quoted so that they cannot interfere with other c
...
-        Me.ConnectionString,
-        request.Form("username"),
-        request.Form("password"))
-
-    Dim connection As New SqlConnection()
-    connection.ConnectionString = connectionString ' Noncompliant
-    connection.Open()
-    Return connection
-End Function
+Public Class ExampleController
```
renaud-tognelli-sonarsource, line 18: Was it intentional to add the class?

```suggestion
Public Class ExampleController
```

renaud-tognelli-sonarsource (approved):
LGTM,
made minor comments.

rules/S6641/vbnet/how-to-fix-it/dotnet-core.adoc
roberto-orlandi-sonarsource: I added it as I noticed that almost all the examples contain the class. Maybe I will remove it for the moment to keep it consistent with the csharp one, and in case I will introduce the class when we merge the feature branch so that we could also apply a consistency check between rules if we see it needed

PR #5233, opened 2025-07-22
Modify S2083(all): Add example user input for readability

loris-s-sonarsource:
A user thought that the point of the rule was partial path traversal, most probably by misreading the "pitfall" section. This tiny update aims to prevent this for future users.

### DIFF
--- a/rules/S2083/common/fix/code-rationale.adoc
+++ b/rules/S2083/common/fix/code-rationale.adoc
@@ -1,5 +1,10 @@
 The following code is vulnerable to path injection as it creates a path using
 untrusted data without validation.
 
-An attacker can exploit the vulnerability in this code to {code_impact}
-arbitrary files.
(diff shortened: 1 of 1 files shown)


rules/S2083/common/fix/code-rationale.adoc
```diff
@@ -1,5 +1,10 @@
 The following code is vulnerable to path injection as it creates a path using
 untrusted data without validation.
 
-An attacker can exploit the vulnerability in this code to {code_impact}
-arbitrary files.
+In this particular case, the code can be exploited with the following string, to
+{code_impact} arbitrary files outside of the intended directory:
```
teemu-rytilahti-sonarsource, line 5:
```suggestion
In this particular case, the code can be exploited with the following input to
{code_impact} arbitrary files outside of the intended directory:
```

PR #5237, opened 2025-07-23
S2301 Fix inappropriate code example

vdiez:


zglicz (approved):
Do you want to add an example with the usage of an object property?

PR #5248, opened 2025-07-28
Modify rule S4426: remove references to AES

egon-okerman-sonarsource:
This PR removes all references to AES in S4426. It is removed as it makes no sense: our recommended key size for AES is 128bit, which is also the minimum.

The noncompliant example throws an exception when ran (when the import error is fixed, as `java.security.KeyGenerator` does not exist. Java and Kotlin are the only languages to consider AES. For other languages, there is nothing to fix.

## Review

- [ ] text style and tone


teemu-rytilahti-sonarsource (approved):
This looks good to me, maybe the rationale.adoc should also be updated to remove the references to AES?

PR #5278, opened 2025-08-13
Modify S6350(python): Fix typo

loris-s-sonarsource:
## Review

- [ ] text style and tone


rules/S6350/python/rule.adoc
```diff
@@ -7,21 +7,32 @@ include::../recommended.adoc[]
...
-if input in allowed:
-    subprocess.run(["/usr/bin/find", input])
+
+ALLOWED = ["file1", "file2", "file3"]
+
+@app.route('/example')
+def example():
+    input = request.args.get("data")
+    if input in ALLOWED:
+        subprocess.run(["/usr/bin/find", input])
```
nicolas-gauthier-sonarsource, line 35: Yeah a bug, if the ALLOWED is defined outside, we will raise, probably some type/field inference issue, do you now if we have a ticket for that?
```suggestion
@app.route('/example')
def example():
    ALLOWED = ["file1", "file2", "file3"]
    input = request.args.get("data")
    if input in ALLOWED:
        subprocess.run(["/usr/bin/find", input])
```

nicolas-gauthier-sonarsource (approved):
LGTM, just be careful to rename the PR, I don't think we have rule S63050 yet ^^
And the compliant example might deserve a modification to not raise

PR #5358, opened 2025-09-02
Modify S6967: Clarify a bit better the security vulnerability point.

mary-georgiou-sonarsource:
After this [discuss post](https://discuss.sonarsource.com/t/should-s6967-csharp-be-a-vulnerability/23159) I felt that it was needed to be a bit more specific in our vulnerability point.


rules/S6967/csharp/rule.adoc
```diff
@@ -12,7 +12,7 @@ Skipping model validation can lead to:
 
 * Data Integrity Issues: Without validation, incorrect or inconsistent data can be saved to your database, leading to potential data corruption or loss.
 
-* Security Vulnerabilities: Skipping validation can expose your application to security risks.
+* Security Vulnerabilities: Skipping validation can expose your application to security risks because incoming user data isn't properly checked. For example, any https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-9.0#built-in-attributes[validation attributes]] applied to your model properties, will be ignored. This lack of proper input validation could lead to various attacks, including, for example, injection attacks in case of not validated string inputs.
```
pavel-mikula-sonarsource, line 15: I find the part about validating input strings redundant to the beginning of the paragraph
```suggestion
* Security Vulnerabilities: Skipping validation can expose your application to security risks because incoming user data isn't properly checked. For example, any https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-9.0#built-in-attributes[validation attributes]] applied to your model properties, will be ignored. This lack of proper input validation could lead to various injection attacks, like SQL injection, cross-site scripting and others.
```

pavel-mikula-sonarsource (approved):
What about this?

PR #5386, opened 2025-09-09
Modify rule S6437: Usernames are not sensitive

teemu-rytilahti-sonarsource:
Make the documentation to follow the implementation, where we do not raise for usernames.

Part of APPSEC-2642

## Review

- [ ] text style and tone


daniel-teuchert-sonarsource (changes requested):
I feel like the compliant example should still not contain a hard-coded username (even if we do not raise for it).
Can we either have examples that do not contain hard-coded usernames or also not have hard-coded usernames in the non-compliant example?

PR #5392, opened 2025-09-10
Modify rule S6383: update key vault argument

GabinL21:
## Review

- [ ] text style and tone

### DIFF
--- a/rules/S6383/terraform/rule.adoc
+++ b/rules/S6383/terraform/rule.adoc
@@ -28,7 +28,7 @@ resource "azurerm_kubernetes_cluster" "example2" {
 }
 ----
 
-For https://azure.microsoft.com/fr-fr/services/key-vault/[Key Vaults] (before 4.42):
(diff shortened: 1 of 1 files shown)


rules/S6383/terraform/rule.adoc
```diff
@@ -33,7 +33,7 @@ For https://azure.microsoft.com/fr-fr/services/key-vault/[Key Vaults]:
 [source,terraform,diff-id=2,diff-type=noncompliant]
 ----
 resource "azurerm_key_vault" "example" {
-  enable_rbac_authorization = false # Sensitive
+  rbac_authorization_enabled = false # Sensitive
```
GabinL21, line 36: I'm not sure if we should keep the previous argument name (an example with "before 4.42"), because it can help especially as the change is pretty recent, but it can be more confusing and it will be less and less relevant in the long run. WDYT? Did you already handle such a case?

petertrr (approved):

